PR #1, opened 2015-03-18
add first version of launch api

dirk-thomas:
And an example launch file.

Connects to ros2/ros2#4

@esteve @tfoote @wjwwood Please review.


launch/launch/launcher.py
```diff
@@ -0,0 +1,154 @@
...
+        # terminate all remaining processes
+        if all_futures:
+
+            # sending SIGINT to remaining processes
+            for index in all_futures.values():
+                p = self.process_descriptors[index]
+                self._process_message(p, 'signal SIGINT')
+                p.transport.send_signal(signal.SIGINT)
+
+            yield from asyncio.wait(all_futures.keys(), timeout=3)
```
wjwwood: Should this timeout be parameterized?

launch/launch/launcher.py
```diff
@@ -0,0 +1,154 @@
...
+from threading import Lock
+
+from launch import ProcessDescriptor
+from launch.output_handler import CompositeOutputHandler
+from launch.output_handler import ConsoleOutput
+from launch.exit_handler import DefaultExitHandler
+from launch.protocol import SubprocessProtocol
+
+
+class DefaultLauncher(object):
```
wjwwood: The `(object)` isn't required in Python3 for "new" style classes. Leaving it doesn't hurt though.

launch/launch/output_handler.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def get_description(self):
+        return ', '.join([h.get_description() for h in self.output_handlers])
+
+    def support_stderr2stdout(self):
+        return all([h.support_stderr2stdout() for h in self.output_handlers])
+
+    def __getattr__(self, name):
+        def wrapper(*args, **kwargs):
+            [getattr(h, name)(*args, **kwargs) for h in self.output_handlers]
```
wjwwood: Does this need to be returned?

launch/launch/output_handler.py
```diff
@@ -0,0 +1,196 @@
...
+        super(LineOutput, self).__init__()
+        self.left_over_stdout = b''
+        self.left_over_stderr = b''
+
+    def on_stdout_received(self, data):
+        lines, self.left_over_stdout = self._process_incoming_lines(data, self.left_over_stdout)
+        self.on_stdout_lines(lines)
+
+    def on_stdout_lines(self, lines):
+        pass
```
wjwwood: Should these do something else by default? It seems bad that if someone makes a typo in the subclass, then nothing happens (silently passes).

Maybe it should print with a prefix or even raise NotImplementedError?

launch/launch/output_handler.py
dirk-thomas: What should it return? A list of all return values?

launch/launch/launcher.py
dirk-thomas: We currently use it in 33 other places with the `object` and only in one location without. I would keep it for now but we can certainly remove it in all locations if everyone agrees.

launch/launch/output_handler.py
wjwwood: I don't know, I was just curious. Depends on what the output handlers might return.

dirk-thomas: These are all callback functions. None of them currently returns anything. I would rather not return anything then a list of `None`s.

launch/package.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>launch</name>
+  <version>0.0.0</version>
+  <description>The ROS launch tool.</description>
+  <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
+  <license>Apache License 2.0</license>
+
```
wjwwood: Does this need dependencies?

launch/setup.py
```diff
@@ -0,0 +1,32 @@
...
+    keywords=['ROS'],
+    classifiers=[
+        'Intended Audience :: Developers',
+        'License :: OSI Approved :: Apache Software License',
+        'Programming Language :: Python',
+        'Topic :: Software Development',
+    ],
+    description='Launch processes specified in launch files.',
+    long_description='''\
+The ability to run multiple processes and react on individual processes exiting.''',
```
wjwwood: This is a sort of strange description. Maybe expand on this and make proper sentences like "This package provides the ability..." rather than just "The ability...".

launch/package.xml
dirk-thomas: Not that I am aware of. It is a pure Python package and only uses Python internal modules.

launch/test/counter.py
```diff
@@ -0,0 +1,50 @@
...
+import signal
+import sys
+import time
+
+
+def main(argv=sys.argv[1:]):
+    parser = argparse.ArgumentParser(
+        description=(
+            'Simple program outputting a counter. '
+            'Even values go to STDERR, odd values go to STDOUT.'))
```
wjwwood: The inner parenthesis are not needed.

esteve: And add a backslash after counter.' Multiline strings without \ tend to be prone to breakage.

dirk-thomas: Without them the indentation does not follow PEP8. Or the second line would need to start aligned to the equal sign.

launch/test/launch_multiple_counters.py
```diff
@@ -0,0 +1,35 @@
...
+import sys
+
+from launch.exit_handler import IgnoreExitHandler
+from launch.exit_handler import RestartExitHandler
+from launch.output_handler import FileOutput
+from launch.output_handler import ConsoleOutput
+
+
+def launch(launch_descriptor):
+    counter_file = os.path.join(os.path.dirname(__file__), 'counter.py')
```
wjwwood: This might be fine as is, but `__file__` can be relative.

launch/test/launch_multiple_counters.py
```diff
@@ -0,0 +1,35 @@
...
+
+    ld = launch_descriptor
+    ld.add_process(
+        cmd=[sys.executable, '-u', counter_file, '--limit', '9', '--sleep', '0.5'],
+        name='foo',
+        output_handlers=[FileOutput(filename='/tmp/foo.log')],
```
wjwwood: You should probably use the `tempfile` module, `/tmp` may not exist or that file may already exist or be in use.

launch/test/launch_multiple_counters.py
```diff
@@ -0,0 +1,35 @@
...
+    ld.add_process(
+        cmd=[sys.executable, '-u', counter_file, '--limit', '10', '--sleep', '0.25'],
+        name='bar',
+        env=None,
+        exit_handler=IgnoreExitHandler(),
+    )
+    ld.add_process(
+        cmd=[sys.executable, '-u', counter_file, '--limit', '5', '--sleep', '1'],
+        name='baz',
+        output_handlers=[FileOutput(filename_stderr='/tmp/baz-err.log')],
```
wjwwood: Same comment about using `tempfile`.

launch/launch/output_handler.py
wjwwood: Fine by me.

launch/package.xml
wjwwood: Cool, I guess `setuptools` and `argparse` are implied. That should be fine.

launch/test/counter.py
wjwwood: @dirk-thomas if it is for style reasons, that I suppose that is ok, but it's not technically needed.

@esteve I do not believe you need the `\`. As long as it is in the containing `()` it is safe, and if you remove the `()` it would become a syntax error.

esteve: Not needed, but I've seen too many times that a multiline like this causes some breakage when moving variables around or reformatting. In this case it's fine because of the parenthesis, though.

dirk-thomas:
@wjwwood Thanks, I have address all your comments.

wjwwood:
One more comment is that there are no headers in the files with copyright and license snippet.

Other than that, lgtm.

wjwwood:
Re-reviewed, +1

tfoote:
lgtm

esteve:
+1

PR #11, opened 2015-11-11
Added send_signal to ProcessDescriptor

esteve:
This PR adds a `send_signal` method to `ProcessDescriptor`, so that a signal can be sent to a process (e.g. `SIGTERM`)

### DIFF
--- a/launch/launch/__init__.py
+++ b/launch/launch/__init__.py
@@ -27,8 +27,10 @@ class LaunchDescriptor(object):
             raise RuntimeError("Task name '%s' already used" % name)
         if exit_handler is None:
             exit_handler = default_exit_handler
-        self.task_descriptors.append(CoroutineDescriptor(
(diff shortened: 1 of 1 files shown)


esteve:
@gerkey could you test it with the examples?

gerkey:
Working on it now.

launch/launch/__init__.py
```diff
@@ -47,6 +51,9 @@ class TaskDescriptor(object):
     def __init__(self):
         self.task_state = None
 
+    def send_signal(self, signal):
+        pass
+
```
dirk-thomas: Since the other subclass `CoroutineDescriptor` can't implement this the base class should probably not contain this method.

esteve: Removed.

dirk-thomas:
+1

PR #12, opened 2015-11-13
Added terminate

esteve:
This adds a `terminate` method to terminate a process.


gerkey:
+1, good to get rid of the Windows-specific SIGINT vs SIGTERM logic in other places.

dirk-thomas:
How is this being used? Eventually invoking close would be better since it also ensures that pipes etc. are closed.

esteve:
This is being called in https://github.com/ros2/examples/blob/test_talker_listener_rebased/rclcpp_examples/test/test_executables_example.py.in#L56 but if it's better to just call close instead, I can change the code there.

PR #13, opened 2015-11-16
Added launch_testing package

esteve:
This PR adds a handler that can check that a process has produced output that matches a predefined string or a regular expression.


launch_testing/launch_testing/__init__.py
```diff
@@ -0,0 +1,61 @@
+import io
+import re
+from launch.output_handler import LineOutput
+
+
+class InMemoryHandler(LineOutput):
+
+    PREFIXES_TO_FILTER = [
```
tfoote: These prefixes should be able to be adjusted or at least disabled. I'd suggest a constructor argument with these default values. And call these default values something like DDS_PREFIXES or some other semantically correct name.

tfoote:
This looks great. It would be useful to be able to enable this on stderr too. Often an error is something you would want to test/assert is thrown.

launch_testing/launch_testing/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+                continue
+            self.stdout_data.write(line + b'\n')
+
+            # Are we ready to quit?
+            if self.regex_match:
+                self.matched = re.match(self.expected_lines[0], line)
+            else:
+                self.matched = (
+                    len(self.stdout_data.getvalue().splitlines()) >=
+                    len(self.expected_lines))
```
dirk-thomas: Doesn't the line content have to match in order to abort? Currently if the test prints a few unrelated lines the test will abort but the later `check` function will return `False`.

launch_testing/launch_testing/__init__.py
esteve: Done.

launch_testing/launch_testing/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class InMemoryHandler(LineOutput):
+
+    def __init__(
+        self, name, ld, expected_lines, regex_match=False,
+        filtered_prefixes=[
+            b'pid', b'rc',
+            b'RTI Data Distribution Service Evaluation License issued to',
+            b'Expires on']):
```
dirk-thomas: This requires internal knowledge about the rmw implementations available. For any other rmw implementation it is not possible to extend this. So it's not *dogfooding*. (I don't have a proposal how to handle this atm.)

codebot: in looking around for a dogfood emoji, the best I could quickly find was [colon]meat_on_bone[colon] which renders as  :meat_on_bone:

codebot: a key improvement suggested by tully is to prefix the food with a dog symbol:  :dog: :meat_on_bone:

launch_testing/launch_testing/__init__.py
esteve: Fixed.

launch_testing/launch_testing/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+            if any([line.startswith(prefix) for prefix in self.filtered_prefixes]):
+                continue
+            self.stdout_data.write(line + b'\n')
+
+            # Are we ready to quit?
+            if self.regex_match:
+                self.matched = re.match(self.expected_lines[0], line)
+            else:
+                output_lines = self.stdout_data.getvalue().splitlines()
+                self.matched = output_lines == self.expected_lines
```
dirk-thomas: Does it have to be the exact list? Shouldn't it be possible to only look for one specific line out of many?

launch_testing/launch_testing/__init__.py
esteve: I can't come up with a better idea either, so I've added a `TODO` that we can address in the future.

launch_testing/launch_testing/__init__.py
esteve: The user can pass a regular expression and it will only match one line, but if passing multiple literal lines, it'll match all of them. I think giving the flexibility to users to choose either behavior is useful.

dirk-thomas: I think we have two different dimensions here:

* How to express expected output:

  * **A** plain text
  * **B** reg ex
  * **C** Python script

* What is considered "enough":

  * **1** the expected output must be identical to the output
  * **2** the expected output is in the output (but the output might have more)
  * **3** one of the expected output lines is enough to be in the output

This patch currently prescribed **B3** and **A1** and doesn't give the user any choose of the other valid combinations.

esteve: The thing is that what is currently in this PR is already "enough" right now. The tests I'm writing pass and exercise the code in this PR. If I need to revisit this code (which I'm sure I'll have to), I'm ok with that. However I prefer to not preemptively write code that may or may not be needed, especially in a case like this which is just a helper class for a very limited usecase.

dirk-thomas: Then this should at least be documented since the current combination of "rules" is only obvious to the person who has implemented the code. Just from the PR using this I would have expected something very different.

esteve: Agreed. I added a docstring to document this behavior.

launch_testing/launch_testing/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+        that will only need one line to match, instead of the entire output.
+    @param regex_match: If true, treat the expected_lines as a regular expression in match
+        accordingly.
+    @param filtered_prefixes: A list of prefixes that will be ignored. By default the output of
+        RTI Connext will be ignored.
+    '''
+    # TODO(esteve): This requires internal knowledge about the rmw implementations available.
+    def __init__(
+        self, name, ld, expected_lines, regex_match=False,
+        filtered_prefixes=[
```
dirk-thomas: This cannot use a list as a default value. Instead use `None` and set the default value in the constructor.

launch_testing/launch_testing/__init__.py
```diff
@@ -0,0 +1,74 @@
+import io
+import re
+from launch.output_handler import LineOutput
```
dirk-thomas: The package lacks a run dependency on `lauch`: http://ci.ros2.org/job/ros2_batch_ci_linux/597/testReport/nose.failure/Failure/runTest/

That is why I suggested to run a CI job before merging. Did you run it locally? In isolation?

esteve: Yes, I ran it locally. Not in isolation. And thanks for keeping an eye on this, but I was already working on a fix.

esteve: Also, we agreed by consensus to merge this and fix any issues we may find.

dirk-thomas:
This new package should use the linters (and pass them).

esteve:
@dirk-thomas how do I enable them? I based this package off `launch` and can't find a reference to linters in it.

dirk-thomas:
Since it is a Python package you need to add tests for each linter. Same as here: https://github.com/ament/ament_index/tree/master/ament_index_python/test

launch_testing/launch_testing/__init__.py
jacquelinekay: For the changes I'm making to the test in ros2/demos#50, I need to support case B2: the expected output multi-line regex which is a subset of the output of the program.

From what I can tell right now, the current behavior for regex expected outputs is that only the first line of the file is considered for matching, and the rest is ignored. I would actually expect case "3" to mean that all of the lines in the regex are considered but only one of them has to match (which is not happening right now).

I'm going to change the behavior to 2, since that should work for my case and not break existing tests. Implementing the different behaviors Dirk described does not sound difficult, but design how to expose and select between the different options sounds less trivial.

esteve: Thanks for working on it!

PR #18, opened 2015-12-12
Minor fixes

esteve:
This enables matching programs that don't display any output and fix an error where stdout lines are evaluated multiple times.

### DIFF
--- a/launch_testing/launch_testing/__init__.py
+++ b/launch_testing/launch_testing/__init__.py
@@ -65,7 +65,7 @@ class InMemoryHandler(LineOutput):
                     if td.name == self.name:
                         td.terminate()
                         self.killed = True
-                        break
+                        return
 
     def on_stderr_lines(self, lines):
         self.stderr_data.write(lines)


launch_testing/launch_testing/__init__.py
```diff
@@ -41,7 +41,7 @@ def __init__(
         self.stdout_data = io.BytesIO()
         self.stderr_data = io.BytesIO()
         self.regex_match = regex_match
-        self.matched = False
+        self.matched = len(self.expected_lines) == 0
```
dirk-thomas: Why should a test use this fancy `InMemoryHandler` if it doesn't specify any expected lines? I would think the launch file should not specify the handler in this case instead.

esteve: Makes, that was an oversight. Fixed.

dirk-thomas:
lgtm

dirk-thomas:
(Hit the wrong button and deleted my previous comment by accident): lgtm

PR #19, opened 2015-12-15
Complete matching

jacquelinekay:
Ready for review.

This pull request implements partial multi-line matching in regexes. e.g. the InMemoryHandler will now accept a test if the output contains a string that matches a multi-line regex.

### DIFF
--- a/launch_testing/launch_testing/__init__.py
+++ b/launch_testing/launch_testing/__init__.py
@@ -35,6 +35,7 @@ class InMemoryHandler(LineOutput):
         self.name = name
         self.launch_descriptor = launch_descriptor
         self.expected_lines = expected_lines
+        self.expected_output = b'\n'.join(self.expected_lines)
         self.killed = False
         self.left_over_stdout = b''
         self.left_over_stderr = b''
@@ -52,20 +53,21 @@ class InMemoryHandler(LineOutput):
             if any([line.startswith(prefix) for prefix in self.filtered_prefixes]):
                 continue
             self.stdout_data.write(line + b'\n')
-
-            # Are we ready to quit?
-            if self.regex_match:
-                self.matched = re.match(self.expected_lines[0], line)
-            else:
+            if not self.regex_match and not self.matched:
                 output_lines = self.stdout_data.getvalue().splitlines()
                 self.matched = output_lines == self.expected_lines
-            if self.matched:
-                # We have enough output to compare; shut down my child
-                for td in self.launch_descriptor.task_descriptors:
-                    if td.name == self.name:
-                        td.terminate()
-                        self.killed = True
-                        return
+
+        # Are we ready to quit?
+        if self.regex_match and not self.matched:
+            self.matched = re.search(self.expected_output, self.stdout_data.getvalue())
+
+        if self.matched:
+            # We have enough output to compare; shut down my child
+            for td in self.launch_descriptor.task_descriptors:
+                if td.name == self.name:
+                    td.terminate()
+                    self.killed = True
+                    return
 
     def on_stderr_lines(self, lines):
         self.stderr_data.write(lines)
--- a/launch_testing/setup.py
+++ b/launch_testing/setup.py
@@ -23,4 +23,5 @@ setup(
     long_description='''\
 This package provides helper scripts for writing tests that use the ROS launch tool.''',
     license='Apache License, Version 2.0',
+    test_suite='test',
 )
--- a/launch_testing/test/matching.py
+++ b/launch_testing/test/matching.py
@@ -1,0 +1,26 @@
+#!/usr/bin/env python3
+
+# Copyright 2015 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+
+
+def main():
+    print("this is line 1", file=sys.stdout)
+    print("this is line b", file=sys.stdout)
+    return 0
+
+if __name__ == '__main__':
+    sys.exit(main())
--- a/launch_testing/test/test_matching.py
+++ b/launch_testing/test/test_matching.py
@@ -1,0 +1,66 @@
+# Copyright 2015 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import sys
+import tempfile
+from launch import LaunchDescriptor
+from launch.exit_handler import ignore_exit_handler
+from launch.launcher import DefaultLauncher
+from launch_testing import create_handler
+
+
+def test_matching():
+    output_handlers = []
+
+    launch_descriptor = LaunchDescriptor()
+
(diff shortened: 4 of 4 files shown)


jacquelinekay:
http://ci.ros2.org/job/ci_linux/728/
http://ci.ros2.org/job/ci_osx/604/
http://ci.ros2.org/job/ci_windows_connext_dynamic/51
http://ci.ros2.org/job/ci_windows_connext_static/61
http://ci.ros2.org/job/ci_windows_opensplice/792

launch_testing/launch_testing/__init__.py
```diff
@@ -35,6 +35,7 @@ def __init__(
         self.name = name
         self.launch_descriptor = launch_descriptor
         self.expected_lines = expected_lines
+        self.expected_output = b'\n'.join(self.expected_lines)
```
dirk-thomas: This comment still applies: https://github.com/ros2/launch/pull/18#discussion_r47428466

jacquelinekay: If launch was documented better, I would know how to use the existing handlers in order to implement what I need for case of matching a multi-line output expectation.

jacquelinekay:
the test fails on Windows:

http://ci.ros2.org/job/ci_windows_opensplice/793/testReport/%28root%29/test_matching/test_matching/

I have no clue why, I'm not familiar at all with Python development on Windows. Do you have any ideas, @dirk-thomas or @esteve?

jacquelinekay:
Possibly fixed:

http://ci.ros2.org/job/ci_windows_opensplice/805

esteve:
+1

launch_testing/launch_testing/__init__.py
```diff
@@ -53,19 +54,19 @@ def on_stdout_lines(self, lines):
...
-                output_lines = self.stdout_data.getvalue().splitlines()
-                self.matched = output_lines == self.expected_lines
-            if self.matched:
-                # We have enough output to compare; shut down my child
-                for td in self.launch_descriptor.task_descriptors:
-                    if td.name == self.name:
-                        td.terminate()
-                        self.killed = True
-                        return
+        # Are we ready to quit?
```
gerkey: I think that there's a risk here of not correctly matching for our existing multi-line non-regex tests (e.g., https://github.com/ros2/examples/blob/master/rclcpp_examples/test/talker.txt#L1-L2). If you get unlucky with how the (potentially buffered) lines are handed to `on_stdout_lines()`, then you might have an extra line in `self.stdout_data` when it comes time to match.

That's why the check for being ready to quit was in the line-by-line loop: the non-regex semantics were that the expected output should match the actual output up to the length (measured in lines) of the expected output, and that any trailing lines in the actual output would be ignored.

dirk-thomas:
This change (https://github.com/ros2/launch/pull/19/files#r47718938) should not be merged.

launch_testing/launch_testing/__init__.py
jacquelinekay: I see.

Perhaps if this repository had been documented better, I would understand the nature of `on_stdout_lines`. I could not find any information about how `on_stdout_lines` behaves in my perusal of the source code of `launch`.

jacquelinekay:
@dirk-thomas, can you please clarify your comment? I don't understand how it applies here.

From my understanding, you're saying that these change renders the InMemoryHandler useless for the case that it was designed for. Based on the line you commented on, you think that the expectation lines should not be concatenated together because they need to be separated as a list, for reasons I don't totally understand. 

Since you think this is wrong, can you clarify what the requirements of the InMemoryHandler are and suggest an alternative strategy for the case that I am trying to design a launch handler for?

launch_testing/launch_testing/__init__.py
gerkey: Yep, no way for you to know that. My comment isn't from first principles, but rather the memory of working with @esteve to write the test output-matching support, and finding that the ready-to-quit check needed to be done on a line-by-line basis.

How about putting the non-regex check back inside the loop, conditioned on `self.regex == False`, and leave the regex check outside, conditioned on `self.regex == True`?

We might want to further adopt the pattern of adding to `.regex` files a last line containing `.*`, to allow trailing lines to be ignored (at least in test cases where we don't know exactly how many lines will be output).

launch_testing/launch_testing/__init__.py
dirk-thomas: Forget my original comment on this. It doesn't apply to this line anymore.

Anyway I don't think we need `expected_output` if we already store `expected_lines`. Since we don't support multi-line regex matching.

dirk-thomas:
I think the code would benefit significantly from clearly separating the two different aspects (how is output matched, what is consider "enough": https://github.com/ros2/launch/pull/13#discussion_r44998991). It would not only reduce complexity but also improve readability.

launch_testing/launch_testing/__init__.py
jacquelinekay: The whole point of this pull request is that I'm trying to support multi-line regex matching. Sorry if that wasn't clear.

dirk-thomas: I mean a single regex which can match across newlines. I thought this PR only aims to support multiple single-line regexps?

jacquelinekay:
Do you consider that a requirement for merging this pull request? I expect that would take some consideration and reworking of this code. There is already a pull request blocked on merging this one (ros2/demos#50).

launch_testing/launch_testing/__init__.py
jacquelinekay: The functionality that I am trying to achieve with this pull request is supporting a single regex that matches multiple lines.

e.g. Suppose the "pass" condition of my test is see the following lines printed in order

"foo 1
bar 2"

but 1 and 2 can be any number.

Then naturally I would want to match them with a single regex that spans 2 lines, something like "foo \d+\nbar\d+"

jacquelinekay: The use case is in the pendulum demo, in which I expect to see a particular block of output printed at the end of the demo, e.g. 

```
rttest statistics:
  - Minor pagefaults: 48
  - Major pagefaults: 0
  Latency (time after deadline was missed):
    - Min: 4218 ns
    - Max: 23552 ns
    - Mean: 5735.88 ns
    - Standard deviation: 614.286


PendulumMotor received 261 messages
PendulumController received 262 messages
```

Accepting multiple single-line regexes is not adequate in this situation, since if the program exits early before getting to the last line due to an error, the test will still pass.

We should eventually work towards implementing multiple single-line regexes as well and selecting between modes. I do not currently have an need for it, however, and the functionality proposed is a superset of the existing functionality.

dirk-thomas:
I think the current code - especially this line:

    re.search(self.expected_output, self.stdout_data.getvalue())

is fragile at best. I think constructing an example which won't work as expected should be easy. My first idea would be a single regex like:

    ^foo.*

If the line starting with "foo" is not the first one it won't match.

I won't insist on doing it now (before the merge). I just wouldn't rely on the test results until then :wink:

launch_testing/launch_testing/__init__.py
jacquelinekay: How does 4417cf1 look to you?

launch_testing/launch_testing/__init__.py
dirk-thomas: How would you write your regex for the example:

    foo 1
    bar 2

if another line might be printed before?

jacquelinekay: It succeeds because I am using `re.search` instead of `re.match`.

If you run this in an interpreter:

```
import re
input = "baz 0\nfoo 1\nbar 2"
pattern = "foo \d\nbar \d"
re.search(pattern, input) is not None
```

you'll get that the last statement is true (meaning that the regex matched)

But if you use "match" instead of "search", the statement is false.

dirk-thomas: But if you anchor your regex with a leading `^` it won't work anymore. And without that you might match other lines.

I am just saying that the current code works only fine in a very narrow set of use cases.

jacquelinekay: Yes, as I said below, I'm not claiming to cover every situation in this pull request.

I still think this is adequate for existing use cases, since the previous functionality of launch_testing only matched a single-line regex to the output.

It creates room for error in that miscellaneous output could be printed to the screen and the test would still pass in tests that expect a regex.

However, in the places where this package is used (examples and demos) the expectation is generally a line that is printed when some set of conditions is met. e.g. a message has been received by a subscriber.

So if the miscellaneous output printed to the screen is due to a fatal error that causes the tested process to crash before meeting the required condition, then the test will fail.

launch_testing/launch_testing/__init__.py
gerkey: lgtm

gerkey:
+1 to merging this now and continuing to improve launch test support over time as needed.

PR #21, opened 2015-12-18
Fix talker listener test

jacquelinekay:
Connects to ros2/examples#83

### DIFF
--- a/launch/launch/exit_handler.py
+++ b/launch/launch/exit_handler.py
@@ -24,11 +24,13 @@ class ExitHandlerContext(object):
         self.task_state = task_state
 
 
-def default_exit_handler(context):
+def default_exit_handler(context, ignore_returncode=False):
     """
     Trigger teardown of launch.
 
     Use the returncode of the task for the launch if the launch was not already tearing down.
+
+    @param ignore_returncode: If True, then ignore process return code.
     """
     # trigger tear down if not already tearing down
     if not context.launch_state.teardown:
@@ -39,7 +41,9 @@ def default_exit_handler(context):
         not context.launch_state.returncode and
         # since Python < 3.5 on Windows does not support signaling SIGINT to the subprocesses
         # we can't expect them to shutdown cleanly, therefore we ignore their return code
-        (os.name != 'nt' or not context.launch_state.teardown)
+        (os.name != 'nt' or not context.launch_state.teardown) and
+        # allow force ignoring
+        not ignore_returncode
     ):
         try:
             rc = int(context.task_state.returncode)
@@ -70,3 +74,16 @@ def primary_exit_handler(context):
             context.launch_state.returncode = 1
 
     default_exit_handler(context)
+
+
+def primary_ignore_returncode_exit_handler(context):
+    """
+    Trigger teardown of launch and ignore return codes.
+
+    Same as primary exit handler but ignore return codes from teardown.
+    """
+    if context.launch_state.teardown:
+        if not context.launch_state.returncode:
+            context.launch_state.returncode = 1
+
+    default_exit_handler(context, ignore_returncode=True)
--- a/launch_testing/launch_testing/__init__.py
+++ b/launch_testing/launch_testing/__init__.py
@@ -17,12 +17,14 @@ class InMemoryHandler(LineOutput):
         accordingly.
     @param filtered_prefixes: A list of prefixes that will be ignored. By default the output of
         RTI Connext will be ignored.
+    @param exit_on_match: If True, then when its output is matched, this handler
+        will terminate; otherwise it will simply keep track of the match.
     """
 
     # TODO(esteve): This requires internal knowledge about the rmw implementations available.
     def __init__(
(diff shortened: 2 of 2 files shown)


launch_testing/launch_testing/__init__.py
```diff
@@ -17,12 +17,14 @@ class InMemoryHandler(LineOutput):
...
         RTI Connext will be ignored.
+    @param test_process: If True, then when its output is matched, this handler
+        will terminate all processes; otherwise it will simply keep track of the match.
     """
 
     # TODO(esteve): This requires internal knowledge about the rmw implementations available.
     def __init__(
         self, name, launch_descriptor, expected_lines, regex_match=False,
-        filtered_prefixes=None
+        filtered_prefixes=None, test_process=True
```
dirk-thomas: Would a name like `exit_on_match` be more specific? It might or might not be related to testing.

jacquelinekay: What about `primary_process` or `primary_match_handler` to mirror the `primary_exit_handler`? Is that accurate?

launch_testing/launch_testing/__init__.py
```diff
@@ -61,13 +63,10 @@ def on_stdout_lines(self, lines):
...
-        if self.matched:
-            # We have enough output to compare; shut down my child
+        if self.matched and self.test_process:
+            # We matched and we're in charge; shut it all down
             for td in self.launch_descriptor.task_descriptors:
-                if td.name == self.name:
-                    td.terminate()
-                    self.killed = True
-                    return
+                td.terminate()
```
dirk-thomas: I don't think this should terminate all but only itself. The strategy if that also terminates the others is defined in the launch file already.

launch_testing/launch_testing/__init__.py
dirk-thomas: Since this class only deals with a single process I think `primary` doesn't have enough context here.

launch_testing/launch_testing/__init__.py
gerkey: Yeah, you could restore the just-kill-myself behavior here and then condition the assignment of the exit_handler (between `primary_exit_handler` and `ignore_exit_handler`) on whether this is the test process: https://github.com/ros2/examples/pull/84/files#diff-1988d847d73fdc4f80e5573bb07a0466R33 .

jacquelinekay: yeah, I have a branch that does something like that:

https://github.com/ros2/examples/compare/talker_test

but one of the processes exits with exit code -15. If you of you want to investigate that alternative approach, feel free.

gerkey: Oh, right, they exit with -15 when we call `terminate()`, which means that we can't use the `primary_exit_handler`, which demands a return code of 0. Let's leave it like this.

dirk-thomas: My conclusion would be that we need a different exit handler then which doesn't care about the exit code.

gerkey: @jacquelinekay, want me to look into the new don't-care-about-return-code exit handler?

jacquelinekay: sure. I'm focusing on `waitset_handles` and the services test right now.

gerkey: OK, I'm on it

gerkey: New `primary_ignore_returncode_exit_handler` added in 0504481 and used in https://github.com/ros2/examples/commit/b45e7954c1c5e56c12eff305f2e601c6440877fe

launch_testing/launch_testing/__init__.py
```diff
@@ -17,12 +17,14 @@ class InMemoryHandler(LineOutput):
         accordingly.
     @param filtered_prefixes: A list of prefixes that will be ignored. By default the output of
         RTI Connext will be ignored.
+    @param exit_on_match: If True, then when its output is matched, this handler
+        will terminate all processes; otherwise it will simply keep track of the match.
```
dirk-thomas: Doc string doesn't describe current behavior.

dirk-thomas:
Beside the comment lgtm

launch/launch/exit_handler.py
```diff
@@ -24,11 +24,13 @@ def __init__(self, launch_state, task_state):
...
 
 
-def default_exit_handler(context):
+def default_exit_handler(context, ignore_returncode=False):
     """
     Trigger teardown of launch.
 
     Use the returncode of the task for the launch if the launch was not already tearing down.
+
+    @param ignore_returncode: If True, then ignore all process return codes.
```
dirk-thomas: The exit handler is specific to a single process. `all` doesn't make sense here.

launch_testing/launch_testing/__init__.py
gerkey: db9ed36

PR #22, opened 2016-02-05
print command to stderr if running it fails

dirk-thomas:
This help in debugging when the command in launch file is not invokable.

Since the `_TaskException` is only used internally and not exposed anywhere I prefixed the name with underscore. Hope that makes sense.

### DIFF
--- a/launch/launch/launcher.py
+++ b/launch/launch/launcher.py
@@ -25,6 +25,13 @@ from launch.protocol import SubprocessProtocol
 from launch.task import TaskState
 
 
+class _TaskException(Exception):
+
+    def __init__(self, task_descriptor_index, exception):
+        self.task_descriptor_index = task_descriptor_index
+        self.exception = exception
+
+
 class DefaultLauncher(object):
 
     def __init__(self, name_prefix='', sigint_timeout=10):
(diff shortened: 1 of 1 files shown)


tfoote:
The _ seems ok to me. Though a doc block comment saying that this is only internal would be even less ambiguous.

dirk-thomas:
PEP8 clearly define the semantic of a leading underscore (https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces). The docblock would be redundant - therefore I will leave it without.

tfoote:
The leading underscore is typically used to indicate that it's for internal use. But according to the pep: "_single_leading_underscore : weak "internal use" indicator. E.g. from M import * does not import objects whose name starts with an underscore." Adding a doc block to any piece of code is rarely redundant. If it's worth a comment in the PR, it's probably worth a comment in the code, since someone reading the code at a later date will not likely find the comment in the PR.

wjwwood:
@tfoote does this look ok otherwise (it seems you guys have agreed to disagree on the docstring note)?

tfoote:
Yes, otherwise it looks good.

PR #23, opened 2016-02-17
Allow for variation on RTI licence in default filtered_prefixes

dhood:
The console output of my RTI free-trial licence is a bit different to what's being filtered in the default filtered_prefixes. Mine says "RTI Data Distribution Service EVAL License issued to ..."

I understand from this [TODO](https://github.com/ros2/launch/blob/master/launch_testing/launch_testing/__init__.py#L24) that this part of the code might be removed one day, but in the meantime this allows my tests on nodes using connext to pass and might help someone else.


dirk-thomas:
Thanks for the patch. While we want to clean this part up it makes sense to check for a shorter hard coded prefix until then.

PR #24, opened 2016-03-29
Allow specification of filtered_prefixes when creating handler

dhood:
Working towards moving the [RTI licence filtering](https://github.com/dhood/launch/blob/be02ffa5fa7adbe99862c60f588d8b75af588f67/launch_testing/launch_testing/__init__.py#L33) outside of launch

connects to ros2/rmw_connext#156

### DIFF
--- a/launch_testing/launch_testing/__init__.py
+++ b/launch_testing/launch_testing/__init__.py
@@ -82,21 +82,23 @@ class InMemoryHandler(LineOutput):
             (output_lines, self.expected_lines)
 
 
-def create_handler(name, launch_descriptor, output_file, exit_on_match=True):
+def create_handler(
+    name, launch_descriptor, output_file, exit_on_match=True, filtered_prefixes=None
+):
     literal_file = output_file + '.txt'
     if os.path.isfile(literal_file):
         with open(literal_file, 'rb') as f:
             expected_output = f.read().splitlines()
         return InMemoryHandler(
             name, launch_descriptor, expected_output, regex_match=False,
-            exit_on_match=exit_on_match)
+            exit_on_match=exit_on_match, filtered_prefixes=filtered_prefixes)
     regex_file = output_file + '.regex'
     if os.path.isfile(regex_file):
         with open(regex_file, 'rb') as f:
             expected_output = f.read().splitlines()
         return InMemoryHandler(
             name, launch_descriptor, expected_output, regex_match=True,
-            exit_on_match=exit_on_match)
+            exit_on_match=exit_on_match, filtered_prefixes=filtered_prefixes)
     py_file = output_file + '.py'
     if os.path.isfile(py_file):
         checker_module = SourceFileLoader(
--- a/launch_testing/launch_testing/__init__.py
+++ b/launch_testing/launch_testing/__init__.py
@@ -2,37 +2,54 @@ from importlib.machinery import SourceFileLoader
 import io
 import os
 import re
+
+import ament_index_python
 from launch.output_handler import LineOutput
+
+
+class UnmatchedOutputError(BaseException):
+    pass
 
 
 class InMemoryHandler(LineOutput):
     """Aggregate data from standard output.
 
-    @param name: Name of the process being tested.
-    @param launch_descriptor: L{LaunchDescriptor} object that contains the processes in the test.
-    @param expected_lines: A list of lines to match the output literally or a regular expression
+    :param name: Name of the process being tested.
+    :param launch_descriptor: :py:obj:`LaunchDescriptor` object that contains the processes in the
+        test.
+    :param expected_lines: A list of lines to match the output literally or a regular expression
         that will only need one line to match, instead of the entire output.
-    @param regex_match: If true, treat the expected_lines as a regular expression in match
+    :param regex_match: If true, treat the expected_lines as a regular expression in match
         accordingly.
-    @param filtered_prefixes: A list of prefixes that will be ignored. By default the output of
-        RTI Connext will be ignored.
-    @param exit_on_match: If True, then when its output is matched, this handler
+    :param filtered_prefixes: A list of byte strings representing prefixes that will cause output
+        lines to be ignored if they start with one of the prefixes. By default lines starting with
+        the process ID (`b'pid'`) and return code (`b'rc'`) will be ignored.
+    :param filtered_rmw_implementation: RMW implementation for which the output will be ignored
+        in addition to the default/`filtered_prefixes`.
+    :param exit_on_match: If True, then when its output is matched, this handler
         will terminate; otherwise it will simply keep track of the match.
+    :raises: :py:class:`UnmatchedOutputError` if :py:meth:`check` does not find that the output
+        matches as expected.
+    :raises: :exc:`LookupError` if the `rmw_output_filter` of the `filtered_rmw_implementation`
+        cannot be found.
+    :raises: :exc:`IOError` if the `rmw_output_filter` of the `filtered_rmw_implementation`
+        cannot be opened.
     """
 
-    # TODO(esteve): This requires internal knowledge about the rmw implementations available.
     def __init__(
         self, name, launch_descriptor, expected_lines, regex_match=False,
-        filtered_prefixes=None, exit_on_match=True
+        filtered_prefixes=None, filtered_rmw_implementation=None, exit_on_match=True
     ):
         super(LineOutput, self).__init__()
         if filtered_prefixes is None:
-            self.filtered_prefixes = [
-                b'pid', b'rc',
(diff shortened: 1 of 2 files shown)


dirk-thomas:
lgtm

dhood:
thinking about your comment here https://github.com/ros2/examples/pull/99/files/8dd30a1151c8b5f5e8c8003658652709f7e0f0d5#r57930461, perhaps it makes more sense to pass 'additional filtered prefixes' instead?

dhood:
or:
add a get_default_filtered_prefixes method so that https://github.com/ros2/examples/pull/99/files#diff-1988d847d73fdc4f80e5573bb07a0466R39 would instead be
```
filtered_prefixes = get_default_filtered_prefixes()
filtered_prefixes.extend([...
```

dhood:
I implemented the other things I suggested (and pep8 checked :wink:) - what do you think?

once I've checked that all tests in demo repositories are explicitly filtering RTI licence output themselves, I'll remove the licence filtering from the defaults, tackling [this TODO](https://github.com/dhood/launch/blob/master/launch_testing/launch_testing/__init__.py#L26)

dirk-thomas:
Taking a step back: the goal is that other packages don't need to know about specific information about output coming from rmw implementations. To actually achieve this we could do the following:

* Each rmw implemenation puts a resource in the ament index e.g. `rmw_output_filter` which can contain multiple lines
  * Each line is either a prefix, substring or regex (?) whatever we think is sufficient for now (it looks like *prefix* covers all the cases atm
* Launch can then optionally filter rmw specific output (by looking up the patterns in the ament index)
* The tests can optionally also filter the output from `launch` like `pid` and `rc` (or launch gets an option to  suppress them in the first place?)

That would solve the issue we are having as well as cleanly separates the information to extend it to arbitrary rmw implementations. How does that sound?

dhood:
that sounds heaps better than adding the filtering in each of the tests. I'll give it a shot. I'll implement it as prefix filtering to start with but I could see regex being useful at some point (if the RTI licence output moves around the names and dates in it, for example)

dhood:
ok, I think that now this PR, when combined with https://github.com/dhood/rmw_connext/commit/4f797f1ae9810f75f2a09cece861cd6e8ae5e0b2 and https://github.com/ros2/examples/pull/99, should implement it

This is the build job with the three branches: http://ci.ros2.org/job/ci_linux/1209/
Note that the ros2/demos tests do not fail even though I haven't modified them to filter for license outputs because of https://github.com/ros2/launch/issues/26

launch_testing/launch_testing/__init__.py
```diff
@@ -15,25 +17,31 @@ class InMemoryHandler(LineOutput):
...
-    @param filtered_prefixes: A list of prefixes that will be ignored. By default the output of
-        RTI Connext will be ignored.
+    @param filtered_prefixes: A list of prefixes that will cause output lines to be ignored if 
+        they start with one of the prefixes. By default lines starting with 
+        the process ID (`pid`) and return code (`rc`) will be ignored.
+    @param rmw_implementations_to_filter: A list of RMW implementations for which the output will
+        be ignored in addition to the default/``filtered_prefixes``.
     @param exit_on_match: If True, then when its output is matched, this handler
         will terminate; otherwise it will simply keep track of the match.
+    :raises: :exc:`LookupError`
```
dhood: is it ok to mix the @ and : styles here? should I change all lines to use : as is done [here](https://github.com/ament/ament_index/blob/master/ament_index_python/ament_index_python/__init__.py#L46)?

dirk-thomas: Consistency would be good. You can update all to `:` since that is what we use almost everywhere now.

dhood: ok I've done that

launch_testing/launch_testing/__init__.py
```diff
@@ -2,38 +2,52 @@
...
         will terminate; otherwise it will simply keep track of the match.
+    :raises: :py:class:`UnmatchedOutputError` if :py:meth:`check` does not find that the output
+        matches as expected.
     """
 
-    # TODO(esteve): This requires internal knowledge about the rmw implementations available.
     def __init__(
         self, name, launch_descriptor, expected_lines, regex_match=False,
-        filtered_prefixes=None, exit_on_match=True
+        filtered_prefixes=None, filtered_rmw_implementations=None, exit_on_match=True
```
dirk-thomas: In which case is more than one rmw implementation necessary for a single executable?

launch_testing/launch_testing/__init__.py
```diff
@@ -2,38 +2,52 @@
...
         that will only need one line to match, instead of the entire output.
-    @param regex_match: If true, treat the expected_lines as a regular expression in match
+    :param regex_match: If true, treat the expected_lines as a regular expression in match
         accordingly.
-    @param filtered_prefixes: A list of prefixes that will be ignored. By default the output of
-        RTI Connext will be ignored.
-    @param exit_on_match: If True, then when its output is matched, this handler
+    :param filtered_prefixes: A list of prefixes that will cause output lines to be ignored if
+        they start with one of the prefixes. By default lines starting with
+        the process ID (`pid`) and return code (`rc`) will be ignored.
```
dirk-thomas: Does this need to be a list of `bytes`?

launch_testing/launch_testing/__init__.py
```diff
@@ -78,26 +92,63 @@ def get_description(self):
...
+                additional_filtered_prefixes = [
+                    str.encode(l) for l in rmw_filter_output_lines]
+                rmw_filter_output_found = True
+        if not rmw_filter_output_found:
+            raise LookupError(
+                "Could not read RMW output filter for RMW implementation "
+                "'%s' at expected location %s" % (rmw_implementation, rmw_filter_output_path))
+    else:
+        additional_filtered_prefixes = []
+    return additional_filtered_prefixes
```
dirk-thomas: The branching complexity of this function is pretty high. I think it would be easier to read the function if it would reduce the nested indentations by e.g. returning early:

```
def get_rmw_output_filter(rmw_implementation):
    prefix_with_resource = ament_index_python.has_resource(
        'rmw_output_filter', rmw_implementation)
    if not prefix_with_resource:
        return []
    
    relative_rmw_filter_output_path = ament_index_python.get_resource(
        'rmw_output_filter', rmw_implementation)
    rmw_filter_output_path = os.path.join(
        prefix_with_resource, 'share', rmw_implementation, relative_rmw_filter_output_path)
    if not os.path.isfile(rmw_filter_output_path):
        raise LookupError(
            "Could not read RMW output filter for RMW implementation "
            "'%s' at expected location %s" % (rmw_implementation, rmw_filter_output_path))

    with open(rmw_filter_output_path, 'r') as h:
        rmw_filter_output = h.read()
        rmw_filter_output_lines = rmw_filter_output.strip().split('\n')
        additional_filtered_prefixes = [
            str.encode(l) for l in rmw_filter_output_lines]
    return additional_filtered_prefixes
```

launch_testing/launch_testing/__init__.py
```diff
@@ -78,26 +92,63 @@ def get_description(self):
...
+def get_rmw_output_filter(rmw_implementation):
+    prefix_with_resource = ament_index_python.has_resource(
+        'rmw_output_filter', rmw_implementation)
+    if prefix_with_resource:
+        rmw_filter_output_found = False
+        relative_rmw_filter_output_path = ament_index_python.get_resource(
+            'rmw_output_filter', rmw_implementation)
+        rmw_filter_output_path = os.path.join(
+            prefix_with_resource, 'share', rmw_implementation, relative_rmw_filter_output_path)
+        if os.path.isfile(rmw_filter_output_path):
```
dirk-thomas: This should also check if the file is readable.

launch_testing/launch_testing/__init__.py
```diff
@@ -78,26 +92,63 @@ def get_description(self):
...
+            with open(rmw_filter_output_path, 'r') as h:
+                rmw_filter_output = h.read()
+                rmw_filter_output_lines = rmw_filter_output.strip().split('\n')
```
dirk-thomas: Instead of `read` and `split` use `str.splitlines()` since it will take care of platform specific new lines as well as empty lines.

launch_testing/launch_testing/__init__.py
```diff
@@ -78,26 +92,63 @@ def get_description(self):
...
+                additional_filtered_prefixes = [
+                    str.encode(l) for l in rmw_filter_output_lines]
```
dirk-thomas: Instead of reencoding the previously decoded content open the file in binary mode.

PR #25, opened 2016-03-30
Multiple vendors

wjwwood:
These changes are supporting my refactor of the `rclpy` tests in https://github.com/ros2/rclpy/pull/5.

Connects to ros2/rclpy#5

### DIFF
--- a/launch/launch/launcher.py
+++ b/launch/launch/launcher.py
@@ -39,6 +39,10 @@ class DefaultLauncher(object):
         self.sigint_timeout = sigint_timeout
         self.task_descriptors = []
         self.print_mutex = threading.Lock()
+        self.loop = None
+        self.loop_lock = threading.Lock()
+        self.interrupt_future = asyncio.Future()
+        self.launch_complete = threading.Event()
 
     def add_launch_descriptor(self, launch_descriptor):
         for task_descriptor in launch_descriptor.task_descriptors:
@@ -51,13 +55,29 @@ class DefaultLauncher(object):
 
             self.task_descriptors.append(task_descriptor)
 
+    def interrupt_launch(self):
+        with self.loop_lock:
+            if self.loop is not None:
+                self.loop.call_soon_threadsafe(self.interrupt_launch_non_threadsafe)
+
+    def interrupt_launch_non_threadsafe(self):
+        self.interrupt_future.set_result(True)
+
+    def wait_on_launch_to_finish(self, timeout):
+        return self.launch_complete.wait(timeout)
+
+    def launch_is_running(self):
+        return not self.launch_complete.is_set()
+
     def launch(self):
-        if os.name == 'nt':
-            # Windows needs a custom event loop to use subprocess transport
-            loop = asyncio.ProactorEventLoop()
-        else:
-            loop = asyncio.new_event_loop()
-        asyncio.set_event_loop(loop)
+        with self.loop_lock:
+            if os.name == 'nt':
+                # Windows needs a custom event loop to use subprocess transport
+                self.loop = asyncio.ProactorEventLoop()
+            else:
+                self.loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(self.loop)
+        loop = self.loop
         try:
             returncode = loop.run_until_complete(self._run())
         except _TaskException as e:
@@ -66,6 +86,8 @@ class DefaultLauncher(object):
                 ' '.join(self.task_descriptors[e.task_descriptor_index].cmd), file=sys.stderr)
             raise e.exception
         loop.close()
+        with self.loop_lock:
+            self.loop = None
         if os.name != 'nt':
             # the watcher must be reset otherwise a repeated invocation fails inside asyncio
             asyncio.get_event_loop_policy().set_child_watcher(None)
@@ -74,6 +96,8 @@ class DefaultLauncher(object):
 
     @asyncio.coroutine
     def _run(self):
+        self.interrupt_future = asyncio.Future()
+        self.launch_complete.clear()
         launch_state = LaunchState()
         for p in self.task_descriptors:
             p.task_state = TaskState()
@@ -108,7 +132,11 @@ class DefaultLauncher(object):
             # wake up frequently and check if any subprocess has exited
             if not isinstance(threading.current_thread(), threading._MainThread):
                 kwargs['timeout'] = 0.5
-            yield from asyncio.wait(all_futures.keys(), **kwargs)
+            yield from asyncio.wait(list(all_futures.keys()) + [self.interrupt_future], **kwargs)
+
+            # if asynchronously interrupted, stop looping and shutdown
+            if self.interrupt_future.done():
+                break
 
             # when the event loop run does not run in the main thread
             # use custom logic to detect that subprocesses have exited
@@ -188,6 +216,7 @@ class DefaultLauncher(object):
                     p.task_states[index].restart_count += 1
                     yield from self._spawn_process(index)
                     all_futures[p.protocol.exit_future] = index
+        # end while True
 
         # terminate all remaining processes
         if all_futures:
@@ -257,6 +286,7 @@ class DefaultLauncher(object):
 
         if launch_state.returncode is None:
             launch_state.returncode = 0
+        self.launch_complete.set()
         return launch_state.returncode
 
     def _spawn_process(self, index):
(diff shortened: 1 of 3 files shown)


launch/launch/launcher.py
```diff
@@ -39,6 +40,11 @@ def __init__(self, name_prefix='', sigint_timeout=10):
         self.sigint_timeout = sigint_timeout
         self.task_descriptors = []
         self.print_mutex = threading.Lock()
+        self.loop = None
+        self.loop_lock = threading.Lock()
+        self.interrupt_future = asyncio.Future()
+        self.launch_complete = threading.Event()
+        self.processes_launched = threading.Event()
```
dirk-thomas: Maybe better ``processes_spawned` since *to spawn* is the commonly used verb for processes?

launch/launch/launcher.py
```diff
@@ -51,13 +57,35 @@ def add_launch_descriptor(self, launch_descriptor):
...
+            if self.loop is not None:
+                self.loop.call_soon_threadsafe(self.interrupt_launch_non_threadsafe)
+
+    def interrupt_launch_non_threadsafe(self):
+        self.interrupt_future.set_result(True)
+
+    def wait_on_launch_to_finish(self, timeout):
+        return self.launch_complete.wait(timeout)
+
+    def launch_is_running(self):
```
dirk-thomas: What is the semantic difference to `Thread.isAlive()`?

If the method has a different semantic and should stay it should start with the verb.

launch/launch/launcher.py
```diff
@@ -51,13 +57,35 @@ def add_launch_descriptor(self, launch_descriptor):
...
+        return not self.launch_complete.is_set()
+
+    def wait_on_processes_to_launch(self, timeout):
+        return self.processes_launched.wait(timeout)
+
+    def processes_are_launched(self):
```
dirk-thomas: This should start with the verb: `are_processes_launched` / `are_processes_spawned`.

launch/launch/launcher.py
```diff
@@ -51,13 +57,35 @@ def add_launch_descriptor(self, launch_descriptor):
...
-            loop = asyncio.new_event_loop()
-        asyncio.set_event_loop(loop)
+        with self.loop_lock:
+            if os.name == 'nt':
+                # Windows needs a custom event loop to use subprocess transport
+                self.loop = asyncio.ProactorEventLoop()
+            else:
+                self.loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(self.loop)
+        loop = self.loop
```
dirk-thomas: Why does this block need a lock? I don't think calling `launch` multiple times should be a supported use case. With the `self.loop` variable being accessed outside the lock this method is not thread safe anyway.

launch/launch/launcher.py
```diff
@@ -225,7 +254,18 @@ def _run(self):
                         except ProcessLookupError:
                             pass
 
-            yield from asyncio.wait(all_futures.keys())
+            kwargs = {}
+            if not isinstance(threading.current_thread(), threading._MainThread):
+                # wake up periodically if we are not in the main thread
```
dirk-thomas: Couldn't it be that launch itself is being using from a non `_MainThread`? Then this check wouldn't work.

launch/launch/launcher.py
```diff
@@ -320,6 +391,9 @@ def __init__(self, launcher):
         super(AsynchronousLauncher, self).__init__()
         self.launcher = launcher
 
+    def terminate(self):
+        self.launcher.interrupt_launch()
```
dirk-thomas: From terminate I would expect that it also waits until the thread has actually finished (`Thread.join()`?).

launch/test/test_interrupt_asynchronous_launcher.py
```diff
@@ -0,0 +1,46 @@
...
+    async_launcher = AsynchronousLauncher(launcher)
+    async_launcher.start()
+
+    # wait up to 10 seconds to get to the point where at least all of the
+    # asyncio-subprocess coroutines have been run (the processes are still
+    # not guaranteed to be running yet)
+    launcher.wait_on_processes_to_launch(10)
+    if not launcher.processes_are_launched():
+        # if the processes didn't launch after 10 seconds, fail the test
+        assert False, "launcher never reported processes launched"
```
dirk-thomas: Nitpick: single quotes wherever possible.

Same below.

launch/test/test_interrupt_asynchronous_launcher.py
```diff
@@ -0,0 +1,46 @@
...
+    async_launcher.terminate()
+    # now wait for the launcher to finish and error if if doesn't
+    async_launcher.join(60)
```
dirk-thomas: These timeouts (60s, 30s, 10s) are very high and make the test unnecessary long. Can they be reduced savely without resulting in getting flaky?

dirk-thomas:
Are all new members and methods intentionally public?

launch/launch/launcher.py
wjwwood: By similar logic, you could argue the launch function should be spawn, but I think this makes sense to change: 32a840c

launch/launch/launcher.py
wjwwood: This is on the `DefaultLauncher` class which is not a Thread. Presumably you can call launch multiple times on this object, so I think it isn't even comparable to a thread.

The `AsynchronousLauncher` class, which inherits from `threading.Thread`, only has the custom `terminate` function and otherwise the `Thread.wait` and `Thread.is_alive` methods are used.

launch/launch/launcher.py
wjwwood: Is that a written rule? It was easier to read as `if [not] launcher.processes_are_launched:` rather than `if [not] launcher.are_processes_launched`, but I guess I'll change it: 1325b3d

launch/launch/launcher.py
wjwwood: This is and at the end of this function are the only places `self.lock` is being set, so `loop = self.loop` is safe so long as `launch()` is not called concurrently. The intention is not to make `launch()` thread-safe but `interrupt_launch()` thread-safe, which this does accomplish.

launch/launch/launcher.py
wjwwood: What?

launch/launch/launcher.py
wjwwood: `terminate` does not wait on something to be done in my experience. For example, subprocess's terminate just sends the signal: https://docs.python.org/3.5/library/subprocess.html#subprocess.Popen.terminate

The expectation is that you need to then wait for the thing to finish, which is what `join` is for and I think it makes sense to keep them as separate steps. I use `join` in conjunction with this in the tests:

https://github.com/ros2/launch/pull/25/files#diff-b620208b20ed521101ca3dbcc3cdc2e1R43

launch/test/test_interrupt_asynchronous_launcher.py
wjwwood: They *should* return quickly, but if the machine is slow then it could fail if the timeout is not large enough. I don't know how short I could make them without making them flakey. So I don't see any advantage to shortening them especially since they normally won't take anytime at all and we risk making them flakey by decreasing the timeout. That being said the times are completely arbitrary.

I can also make them blocking, but that will result in the entire test getting timed-out and potentially preventing other tests from running when they otherwise should.

wjwwood:
> Are all new members and methods intentionally public?

All of the methods are intentionally public, but the members I was just following what the class was doing before (all members are "public").

launch/launch/launcher.py
wjwwood: I looked over this again and I don't understand the question. This same check was being used in this function already. Maybe you could describe a scenario where this wouldn't work?

dirk-thomas: When a launch file is being executed directly the semantic of `_MainThread` is clear and this works. But what if a Python script starts a new thread and calls `launch` in there? I would expect the check for `_MainThread` to not work in this case.

wjwwood: I think that's the whole point of this check, if you're not calling `launch()` in the main thread then it needs to do something special. I don't know what you mean by "I would expect the check for _MainThread to not work in this case.", do you mean it will give a false positive or a false negative?

dirk-thomas: False negative because the function is *not* invoked in the main thread.

wjwwood: If the function is *not* invoked in the main thread then this condition should be satisfied and the timeout argument added, as far as I can see. I don't think there is any semantic ambiguity on what the "_MainThread" is, so I don't see how there can be a false negative?

wjwwood: Maybe you could describe a test with pseudo code that would expose this issue? I'm happy to try and write a test to cover this case, but I still don't see what could happen here that's unintended or broken.

wjwwood:
If there's any more feedback on this pr, please give it as this is blocking me from squashing, running CI on, and then merging the related pr's.

dirk-thomas:
LGTM.

PR #28, opened 2016-04-21
fix PEP257 warning D211

dirk-thomas:
Since we don't want to follow D203 (ament/ament_lint#47) I assume we will follow D211 (which is checked by newer versions of `pydocstyle`).

### DIFF
--- a/launch/launch/exit_handler.py
+++ b/launch/launch/exit_handler.py
@@ -16,7 +16,6 @@ import os
 
 
 class ExitHandlerContext(object):
-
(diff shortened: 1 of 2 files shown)


dhood:
think there's one here too https://github.com/ros2/launch/blob/master/launch_testing/launch_testing/__init__.py#L9 which I will fix in #24 if it's not appropriate here

dirk-thomas:
The referenced line seems to already be part of this patch?

dhood:
yes, my mistake

PR #30, opened 2016-05-03
fix unused test script

dirk-thomas:
Fixes ros2/ros2#219.

http://ci.ros2.org/job/ci_linux/1250/


jacquelinekay:
Is this test script important and should it be added to CI, i.e. in a nosetest?

tfoote:
Reminder for @dhood to think about making a test for this while she is doing other tests for launch.

jacquelinekay:
ok, +1 for now and I'll wait on tests in a separate PR.

Issue #26, opened 2016-04-19
launch_testing always ignores license output when checking regexes


dhood:
`launch_testing` has support for filtering the RTI license output when checking that the received output matches the expected output. If the license output appears and is not being filtered, then the check fails. I think that this is the desired behaviour.

However, this concept seems to be implemented differently for the case when `regex_match` is `True`. The license output is always ignored, whether or not it was supposed to be. (And any other output that appears that doesn't match the regex)

These non-matching, un-filtered lines are written [here](https://github.com/ros2/launch/blob/master/launch_testing/launch_testing/__init__.py#L57) and read [here](https://github.com/ros2/launch/blob/master/launch_testing/launch_testing/__init__.py#L80), but since `self.matched` is `True`, the assertion does not fail.

Am I right in interpreting this as a bug that I should fix?

dhood:
this could also be addressed by adding `^` to the start of regexes (https://github.com/ros2/demos/blob/master/intra_process_demo/test/two_node_pipeline.regex for example), but that leads to test timeouts instead of failures

dirk-thomas:
The matrix of what to test for and how to test for it (https://github.com/ros2/launch/pull/13#discussion_r44998991) was implemented incompletely in that PR. I agree with your conclusion that this should be fixed by making the different tests do the same thing.

dhood:
thanks for that link! OK good, not a bug, just not yet implemented. 
I'll work on getting it to fail if lines which aren't covered by the regex are detected. (**B1**, I believe)

dhood:
I've started implementing this. At the moment, any lines that are not filtered and are not matched by the regex cause the test to fail.

This affects `ros2/demos` (pinging @jacquelinekay) because there are currently some test regexes which do not include all output lines, e.g.:
http://ci.ros2.org/job/ci_linux/1216/testReport/%28root%29/test_pendulum__rmw_connext_cpp/test_executable/

In this example case, do you think it's best to:
1. change the test regexes to expect the `'Logger node initialized.'` line
2. leave the regex as it is, but change the executable output to be something like `[debug] 'Logger node initialized.'` and then add an `[debug]` prefix filter to the test (or somehow use a different logging level)
3. add an option in `launch_testing` to ignore any lines that don't match the regex, as long as at some point the regex is matched

thanks

sloretz:
> At the moment, any lines that are not filtered and are not matched by the regex cause the test to fail.

[...]

PR #34, opened 2016-05-31
Make launch_testing fail on unexpected license output when regex matching

dhood:
Connects to ros2/launch#26

I have made the behaviour of `launch_testing` when matching with regexes vs literal text consistent.

Have added test cases for both regex and text matching which describe how I understand the system should function. Namely, tests will fail if unexpected lines are received before, during or after the expected output, unless those lines are filtered with the `filtered_prefixes` option I added to `launch_testing` recently.

### DIFF
--- a/launch_testing/launch_testing/__init__.py
+++ b/launch_testing/launch_testing/__init__.py
@@ -74,10 +74,10 @@ class InMemoryHandler(LineOutput):
             if any([line.startswith(prefix) for prefix in self.filtered_prefixes]):
                 continue
             self.stdout_data.write(line + b'\n')
-            if not self.regex_match and not self.matched:
+            if not self.regex_match:
                 output_lines = self.stdout_data.getvalue().splitlines()
-                self.matched = output_lines == self.expected_lines
-                self.complete_match = self.matched
+                self.complete_match = output_lines == self.expected_lines
+                self.matched |= self.complete_match
 
         # Are we ready to quit?
         if self.regex_match and not self.matched:
--- a/launch_testing/launch_testing/__init__.py
+++ b/launch_testing/launch_testing/__init__.py
@@ -21,7 +21,7 @@ class InMemoryHandler(LineOutput):
         test.
     :param expected_lines: A list of lines to match the output literally or a regular expression
         that will only need one line to match, instead of the entire output.
-    :param regex_match: If true, treat the expected_lines as a regular expression in match
+    :param regex_match: If True, treat the expected_lines as a regular expression in match
         accordingly.
     :param filtered_prefixes: A list of byte strings representing prefixes that will cause output
         lines to be ignored if they start with one of the prefixes. By default lines starting with
@@ -30,6 +30,9 @@ class InMemoryHandler(LineOutput):
         in addition to the default/`filtered_prefixes`.
     :param exit_on_match: If True, then when its output is matched, this handler
         will terminate; otherwise it will simply keep track of the match.
+    :param exact_match: If True, the output received must exactly match the expected output, with
+        no unfiltered lines appearing. If False, the output must simply contain the expected
+        output, but unfiltered lines will be tolerated. Default value is True.
     :raises: :py:class:`UnmatchedOutputError` if :py:meth:`check` does not find that the output
         matches as expected.
     :raises: :exc:`LookupError` if the `rmw_output_filter` of the `filtered_rmw_implementation`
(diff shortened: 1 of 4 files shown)


wjwwood:
Changes lgtm, can you run CI on this (including Connext) and we can look to merge it after the alpha release is out of the door.

launch_testing/launch_testing/__init__.py
```diff
@@ -53,14 +53,15 @@ def __init__(
...
-        self.expected_output = b'\n'.join(self.expected_lines)
+        self.expected_output = b'\n'.join(self.expected_lines) + b'\n'
         self.left_over_stdout = b''
         self.left_over_stderr = b''
         self.stdout_data = io.BytesIO()
         self.stderr_data = io.BytesIO()
         self.regex_match = regex_match
         self.exit_on_match = exit_on_match
         self.matched = False
+        self.matched_exactly = False
```
dirk-thomas: Just from reading the code I would expect `matched_exactly` to have a wrong value if `matched` becomes try early while processing the lines. As soon as that happens `on_stdout_lines` stops processing further lines and therefore never computes `matched_exactly` again.

dhood: what I had in mind while writing this code was that `matched_exactly` is true if no unexpected lines were received, whereas `matched` can be true even if non-matching lines were found. so I think by the time `matched` is able to go true, `matched_exactly` will have had its chance to go true it it's supposed to. if `matched_exactly` isn't true by that time, it will never be true

e.g. if you are looking for 'a; b; c' and you get 'x; a; b; y; c; z' then I expect `matched_exactly` to be false and `matched` to be true. then it will stop processing. if 'a;b;c' is received again at a later stage, I wouldn't have expected that to be allowed to set `matched_exactly` true, because extra lines were still received in the meantime

does that match with what you would expect it to do?

dirk-thomas: If you expect `a` but get `a; b` the flag `matched` might turn `true` after reading `a` and then no further content is being considered and `matched_exactly` has no chance of turning back to `false`.

dhood: thanks - 
the tests written with this in mind didn't catch this because all of the output lines were being received in the one callback. Is https://github.com/ros2/launch/pull/34/commits/4167c2eb831d05fe82a33a1683255a4c2e2567fb an appropriate way to address this? (it works but you might have better ideas)

I have removed [these lines](https://github.com/ros2/launch/pull/34/files#diff-95fa932db2912f5a4060524ea72cd0cfL66). There are some tests that start failing in other places e.g. talker-listener in `rclcpp_examples` as a result, but I'll adapt them (assuming you agree that requiring `matched_exactly` should be the default behaviour)

dirk-thomas: I would assume that you also need to flush the output. Maybe that is even sufficient without the sleep. Do the tests then fail as expected for this case (without removing the return)?

dhood: yes they fail without removing the return, but only if I both sleep and (either flush or invoke python with -u). (I also was surprised that the sleep was necessary which prompted me to ask you)

dirk-thomas: Sounds good.

dhood: thank you

launch_testing/launch_testing/__init__.py
```diff
@@ -53,38 +57,57 @@ def __init__(
...
+        if regex_match:
+            self.expected_output = self.expected_lines
+            # Add a surrounding capture group
+            self.expected_output_captured = \
+                b'(?P<launch_testing_capture>' + self.expected_output + b')'
+            # Increment any group IDs in the original regex to compensate for the surrounding group
+            self.expected_output_captured = re.sub(
+                b'\\\(\d)',  # TODO(dhood): only match single backslashes
+                lambda matchobj: b'\\' + str(int(matchobj.group(1)) + 1).encode(),
+                self.expected_output_captured)
```
dhood: open to suggestions of less intrusive approaches to this. 
the situation is that I want to check if the lines received are _completely_ covered by the regex. I wrap the regex in a capture group and check that the captured group contains the entirety of the text received. because of that, I then have to increment any references to groups inside the origin regex because `\1` then refers to the outer group, not whatever group it originally referred to. (which is a problem for something like [this](https://github.com/ros2/demos/blob/master/intra_process_demo/test/cyclic_pipeline.regex#L2), for example)

@wjwwood and I also considered using regex subroutines or having capture groups both before and after the regex and checking that they're empty, but subroutines don't seem to be supported in python for the former approach, and for the latter the same situation is introduced with the group numbering being changed.

PR #35, opened 2016-06-06
Extend the available exit handlers to support a no teardown option

tfoote:
This should be used for test fixture processes that may terminate cleanly before the end of the test. However it will not block teardown if the test process finishes.

### DIFF
--- a/launch/launch/exit_handler.py
+++ b/launch/launch/exit_handler.py
@@ -23,7 +23,7 @@ class ExitHandlerContext(object):
         self.task_state = task_state
 
 
-def default_exit_handler(context, ignore_returncode=False):
+def default_exit_handler(context, ignore_returncode=False, trigger_teardown=True):
     """
     Trigger teardown of launch.
 
@@ -32,7 +32,7 @@ def default_exit_handler(context, ignore_returncode=False):
     @param ignore_returncode: If True, then ignore process return code.
     """
     # trigger tear down if not already tearing down
-    if not context.launch_state.teardown:
+    if not context.launch_state.teardown and trigger_teardown:
         context.launch_state.teardown = True
 
     # set launch return code if not already set
@@ -66,6 +66,7 @@ def primary_exit_handler(context):
     Trigger teardown of launch and if teardown already in place set non-zero return code.
 
     Same as default exit handler but if teardown was triggered by another task
+
     ensure that the returncode is non-zero.
     """
     if context.launch_state.teardown:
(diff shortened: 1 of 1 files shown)


launch/launch/exit_handler.py
```diff
@@ -23,17 +23,34 @@ def __init__(self, launch_state, task_state):
...
     Trigger teardown of launch.
 
     Use the returncode of the task for the launch if the launch was not already tearing down.
 
     @param ignore_returncode: If True, then ignore process return code.
+    @param teardown_on_error_only: If True, trigger launch context teardown
+    on exit if the return code indicates an error. This is mutually not
+    relevant if ignore_returncode is True or trigger_teardown is False.
+    @param trigger_teardown: If True, trigger launch context teardown
+    on exit if not already triggered.
```
dirk-thomas: The entanglement of the different keyword arguments seems pretty complicated to me. At least to me even after reading the doc it is not straight forward what to pass. Maybe the arguments can be changed somehow or it would be easier to have a separate exit handler if the behavior is so different from the default.

mikaelarguedas:
@tfoote and @dirk-thomas to clarify what should be done here

launch/launch/exit_handler.py
tfoote: I'd be happy to change the keywords. These were the ones that made sense to me. They did accumulate over time.

My vision of this is that there's a combinatoric set of possible exit handlers. And if each combo can be combined in a single implementation then a thin wrapper added on top which calls it with the right argument that will be easier to maintain. (Like ignore_signal_exit_handler vs ignore_signal_exit_handler_no_teardown)

Similarly the default signal handler provides arguments such that the behavior can be modified from higher level handlers, but we keep a single implementation for each behavior based on the arguments passed through. This makes it relatively clear, how the different behaviors compare to each other.

dirk-thomas:
I would suggest to close the current PR. First it has conflicts and second the added keyword arguments are very complicated (at least for me) which makes it pretty hard to use. Until we have a specific use case for the cases and someone will spend time to actually work on this I think it is better to just close.

---

Some comments regarding the design. Some behaviors implemented by exit handlers are orthogonal. And imo those should not be implemented in a single exit handler. The launch API allows to pass multi exit handlers and I think passing multiple orthogonal ones is prefer over passing a single monolithic and complicated one.

For the user a nice-to-use single exit handler can be offered (as syntactic sugar) which under the hood should delegate the work to the separately implemented exit handlers.

mikaelarguedas:
@tfoote to respond to comments

tfoote:
With the new version of launch this is now deprecated and not worth following up.

PR #36, opened 2016-06-24
Send sigint when regex gets matched

jacquelinekay:
Fixes ros2/ros2#239

When a process that was instrumented by gcov is interrupted by SIGKILL, SIGHUP or SIGTERM, it does not write a gcda file. The reason no gcda files were written for intra_process_demo is that all of the nosetests for the demos use regex matching, which through the codepath of `TaskDescriptor.terminate` sends SIGHUP to the child processes.

This affected files in other packages as well, for example the `talker/listener` in rclcpp_examples only has coverage information for the talker, because the listener uses regex matching and "exit_on_match":

http://ci.ros2.org/view/nightly/job/nightly_linux_coverage/5/cobertura/_home_rosbuild_ci_scripts_ws_src_ros2_examples_rclcpp_examples_src_topics/

We should expect this job to have results for the intra_process_demo package:
http://ci.ros2.org/job/ci_linux/1539/

### DIFF
--- a/launch/launch/__init__.py
+++ b/launch/launch/__init__.py
@@ -16,6 +16,9 @@ from launch.output_handler import CompositeOutputHandler
 from launch.output_handler import ConsoleOutput
 from launch.exit_handler import default_exit_handler
 
+import os
+import signal
+
 
 class LaunchDescriptor(object):
 
@@ -76,7 +79,10 @@ class ProcessDescriptor(TaskDescriptor):
     def send_signal(self, signal):
         if self.transport:
             self.transport.send_signal(signal)
+            self.task_state.signals_received.append(signal)
 
     def terminate(self):
-        if self.transport:
+        if os.name != 'nt':
+            self.send_signal(signal.SIGINT)
+        elif self.transport:
             self.transport.terminate()
--- a/launch/launch/__init__.py
+++ b/launch/launch/__init__.py
@@ -78,8 +78,8 @@ class ProcessDescriptor(TaskDescriptor):
 
     def send_signal(self, signal):
         if self.transport:
-            self.transport.send_signal(signal)
             self.task_state.signals_received.append(signal)
+            self.transport.send_signal(signal)
 
     def terminate(self):
         if os.name != 'nt':
--- a/launch_testing/launch_testing/__init__.py
+++ b/launch_testing/launch_testing/__init__.py
@@ -2,6 +2,7 @@ from importlib.machinery import SourceFileLoader
 import io
 import os
 import re
+import signal
 
 import ament_index_python
 from launch.output_handler import LineOutput
@@ -83,7 +84,10 @@ class InMemoryHandler(LineOutput):
             # We matched and we're in charge; shut myself down
             for td in self.launch_descriptor.task_descriptors:
                 if td.name == self.name:
-                    td.terminate()
+                    if os.name != 'nt':
+                        td.send_signal(signal.SIGINT)
+                    else:
+                        td.terminate()
                     return
 
     def on_stderr_lines(self, lines):
--- a/launch_testing/launch_testing/__init__.py
+++ b/launch_testing/launch_testing/__init__.py
@@ -2,7 +2,6 @@ from importlib.machinery import SourceFileLoader
 import io
 import os
 import re
-import signal
 
 import ament_index_python
 from launch.output_handler import LineOutput
--- a/launch_testing/launch_testing/__init__.py
+++ b/launch_testing/launch_testing/__init__.py
@@ -84,10 +84,8 @@ class InMemoryHandler(LineOutput):
             # We matched and we're in charge; shut myself down
             for td in self.launch_descriptor.task_descriptors:
                 if td.name == self.name:
-                    if os.name != 'nt':
-                        td.send_signal(signal.SIGINT)
-                    else:
-                        td.terminate()
+                    td.terminate()
+                    td.protocol.exit_future.set_result(True)
                     return
 
     def on_stderr_lines(self, lines):
--- a/launch_testing/test/hang_after_match.py
+++ b/launch_testing/test/hang_after_match.py
@@ -19,30 +19,15 @@ import sys
 import time
 
 
-# from http://stackoverflow.com/questions/842557
-class IgnoreKeyboardInterrupt(object):
-    def __enter__(self):
-        self.signal_received = False
-        self.old_handler = signal.getsignal(signal.SIGINT)
-        signal.signal(signal.SIGINT, self.handler)
-
-    def handler(self, sig, frame):
-        self.signal_received = (sig, frame)
-
-    def __exit__(self, type, value, traceback):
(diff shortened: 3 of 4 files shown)


launch_testing/launch_testing/__init__.py
```diff
@@ -83,7 +84,10 @@ def on_stdout_lines(self, lines):
             # We matched and we're in charge; shut myself down
             for td in self.launch_descriptor.task_descriptors:
                 if td.name == self.name:
-                    td.terminate()
```
dhood: We could consider changing the behaviour of `terminate` itself, instead - I think `exit_on_match` might be the only thing that uses it at the moment

dhood:
I think you'll need to add the signal to `signals_received` in order to use the `ignore_signal_exit_handler` with `exit_on_match` in https://github.com/ros2/demos/pull/73

jacquelinekay:
True... Another reason why it might be more efficient to wrap that logic in
"terminate" instead so it can be reused elsewhere...

dhood:
yeah, the `teardown` flag is another thing that needs to be set I think. I have noticed these points before but hadn't decided yet at which 'layer' to make the changes.. so I'll be watching this PR!

tfoote:
This also could be helped by #35 right now we have to guarentee that all nodes under test are the first to shutdown in the fixture.

jacquelinekay:
OK, in 88cc46f I think I solved the problem @dirk-thomas brought up when I asked him about this pull request, which is: what happens when the launched process blocks after receiving the regex? Calling `ProcessDescriptor.protocol.set_value(true)` causes the future for the process to be fulfilled, which is what we want to communicate to the main _run loop, which takes care of cleanup for us after waking up from the `yield`.

jacquelinekay:
I can write a standalone test for this if desired, I was testing locally with the `intra_process_demo` and adding a `while(true){}` at the end of the program.

gerkey:
That's a great explanation for why we weren't getting coverage for those tests. Nice!

launch/launch/__init__.py
```diff
@@ -75,8 +78,11 @@ def __init__(self, cmd, name, output_handler, exit_handler, env=None):
 
     def send_signal(self, signal):
         if self.transport:
+            self.task_state.signals_received.append(signal)
             self.transport.send_signal(signal)
 
     def terminate(self):
-        if self.transport:
+        if os.name != 'nt':
+            self.send_signal(signal.SIGINT)
```
dirk-thomas: As mentioned on the other ticket before: only sending a `SIGINT` is not sufficient since the process might not exit based on that.

launch_testing/launch_testing/__init__.py
```diff
@@ -2,6 +2,7 @@
 import io
 import os
 import re
+import signal
```
dirk-thomas: Why is this being added?

launch/launch/__init__.py
jacquelinekay: I think that line 88 covers the case where the process doesn't exit when it receives SIGINT.
https://github.com/ros2/launch/pull/36/files#diff-95fa932db2912f5a4060524ea72cd0cfR88

The task descriptor's future will be considered finished, so the thread waiting in the `yield` statement in the `while True` loop in `launcher.py` will wake up and initialize exit.

https://github.com/ros2/launch/blob/master/launch/launch/launcher.py#L149

I was testing this manually but I will add a unit test for this case.

dirk-thomas: Are you referring to two lines below this? That code is in the else block and not being considered if the if block is used.

jacquelinekay: no, I must have linked incorrectly, I'm referring to line 87 of `launch_testing/launch_testing/__init__.py`

```
td.protocol.exit_future.set_result(True)
```

dirk-thomas: How does manually overriding the `exit_future` ensure that the process is terminated? It will "signal" like it has been terminated but the process might still be running, no?

jacquelinekay: In `launcher.py` the launcher calls `close_process` on all the exited future, which calls `Transport.close()` on the process, which send SIGTERM to the process if it hasn't exited yet.

https://github.com/ros2/launch/blob/master/launch/launch/launcher.py#L183
https://docs.python.org/3/library/asyncio-protocol.html#asyncio.BaseSubprocessTransport.close

dirk-thomas: Currently `terminate` is being called on the task descriptor when the expected output is matched. If we assume that the process is not handling SIGINT perfectly then this change renders that `terminate` call useless. Instead it relies on the launcher to cleanup all process afterwards.

From my point of view the current patch is not consistent. Either the `terminate` method should be removed from the task descriptor (since its not doing anymore what it is supposed to) or it needs to keep actually terminating the subprocess. If we keep the responsibility to terminate processes to the launch only then the task descriptor should not have a `terminate` method.

Also the launcher enforces `SIGINT` and `SIGTERM` on all processes after exiting the main loop. So a process might hang around for a long time if the lanuch file is supposed to be running longer than one individual process.

jacquelinekay: OK, that's fine. I can change this pull request so that the `terminate` method remains as is and add a new function that tries to call `SIGINT` first, e.g. `request_exit`, so that the names match the behavior.

Because of the problem described in the description of this PR, I need a way to call SIGINT after the match condition is met, because gcov doesn't produce output for SIGTERM'd process.

So what do you think should happen to processes that receive SIGINT after the output was matched? Is the basic idea in this PR acceptable? Or do you have an alternative suggestion?

dirk-thomas: It needs to be made sure that the process has "enough" time to process the `SIGINT`. Otherwise it might be `SIGTERM`ed shortly after.

I think it should be ensured that this process actually terminates in a finite time. Independent of the fact that the launch file will ensure that when its shutting down. Not sure what a good way to achieve that is.

jacquelinekay: Yeah, I don't think there's a good way of determining how much time is enough to wait until sending SIGTERM after SIGINT besides a heuristic provided the user.

jacquelinekay: I still think that setting the result of the future from `on_stdout_lines` in the matching handler so that the run loop wakes up from yield is necessary. Then I think I need to add some logic so that launch waits a short time before sending SIGTERM to processes that have already received SIGINT (the signals are recorded in the `signals_received` list in TaskState).

jacquelinekay: Does an `exit_handler` that sleeps for a finite amount of time if SIGINT is in signals_received make sense?

dirk-thomas: That will prevent the launcher to handle any further event on the concurrent futures: https://github.com/ros2/launch/blob/e3ec7bbfb180f8ede93a7ed85737d0def9c9c5bf/launch/launch/launcher.py#L190

sloretz:
It looks like ros2/ros2#239 referenced in the description was reportedly fixed by #44. I'll close this under the assumption that this PR is no longer required.

Issue #37, opened 2016-07-05
Environment passing to launch has no effect


mikaelarguedas:
Spawned processes always get the parent's environment even though a specific env parameter is given.

Reproducible example [here](https://gist.github.com/mikaelarguedas/85a7eebb1dafe1a73c3b26d95ca075e3)

dirk-thomas:
The `env` parameter of `add_process` is not being used anywhere at the moment. If you need to pass custom environments you need to make it pass this along to the subprocess.

mikaelarguedas:
I see. I discovered this while testing https://github.com/ros2/system_tests/blob/master/test_communication/test/test_publisher_subscriber.py.in which I think is the only place that uses it for now.
Just to confirm my understanding, the purpose of the `env` parameter is indeed to pass custom environment but doesn't pass them to the spawned subprocesses yet, is that what you mean ?

dirk-thomas:
Yes, the function has such a parameter but it is not applied anywhere.

PR #38, opened 2016-07-05
Pass environment to launched processes

mikaelarguedas:
Connects to #37

### DIFF
--- a/launch/launch/launcher.py
+++ b/launch/launch/launcher.py
@@ -342,6 +342,7 @@ class DefaultLauncher(object):
         transport, protocol = yield from loop.subprocess_exec(
             lambda: SubprocessProtocol(p.output_handler),
             *p.cmd,
+            env=p.env,
             **kwargs)
         p.transport = transport
         p.protocol = protocol
--- a/launch/launch/launcher.py
+++ b/launch/launch/launcher.py
@@ -338,11 +338,12 @@ class DefaultLauncher(object):
         kwargs = {}
         if p.output_handler.support_stderr2stdout():
             kwargs['stderr'] = asyncio.subprocess.STDOUT
+        if p.env is not None:
+            kwargs['env'] = p.env
         loop = asyncio.get_event_loop()
         transport, protocol = yield from loop.subprocess_exec(
             lambda: SubprocessProtocol(p.output_handler),
             *p.cmd,
-            env=p.env,
             **kwargs)
         p.transport = transport
         p.protocol = protocol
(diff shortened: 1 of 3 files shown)


launch_testing/test/check_env.py
```diff
@@ -0,0 +1,4 @@
+if __name__ == '__main__':
+    import os
```
dirk-thomas: Please move the import to the module level.

launch/launch/launcher.py
```diff
@@ -342,6 +342,7 @@ def _spawn_process(self, index):
         transport, protocol = yield from loop.subprocess_exec(
             lambda: SubprocessProtocol(p.output_handler),
             *p.cmd,
+            env=p.env,
```
dirk-thomas: Can you just add `env` to `kwargs`?

launch_testing/test/check_env.py
```diff
@@ -0,0 +1,4 @@
...
+    assert(os.environ.get('testenv2', 'Not Set') == 'Not Set')
```
dirk-thomas: Assert is not a function but a keyword. This will always be true since the arguments are a tuple.

dirk-thomas: This line should better test that the environment variable is not set: `'testenv2' not in os.environ`

dirk-thomas:
The new files need copyright headers.

launch/launch/launcher.py
mikaelarguedas: Done https://github.com/ros2/launch/pull/38/commits/6368d9c8610db0497b2ce0845fdac263e39a8b82

launch_testing/test/check_env.py
mikaelarguedas: Surprisingly it raised errors as expected with the extra brackets.
Removed in https://github.com/ros2/launch/pull/38/commits/f49d6b681fa3c475976730e42f04b1d1a34de890

launch_testing/test/check_env.py
mikaelarguedas: Done https://github.com/ros2/launch/pull/38/commits/f49d6b681fa3c475976730e42f04b1d1a34de890

launch_testing/test/check_env.py
mikaelarguedas: Makes sense. Changed in https://github.com/ros2/launch/pull/38/commits/ede0634c469bb3a12e3f89a6b2faec93a2be4dfd

wjwwood:
These changes lgtm for merge after the alpha release and after CI.

mikaelarguedas:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/1670/)
* OSX [Build Status](http://ci.ros2.org/job/ci_osx/1245/)

PR #40, opened 2016-07-28
Use Python 3.5 async

rohbotics:
Python 3.5 adds a lot of new asyncio functionality, and launch was still using the 3.4 asyncio stuff.
As python 3.5 is our minimum version as of alpha7, this PR pulls forward to the 3.5 asyncio.

### DIFF
--- a/launch/launch/launcher.py
+++ b/launch/launch/launcher.py
@@ -27,12 +27,14 @@ from launch.task import TaskState
 
 
 class _TaskException(Exception):
+
     def __init__(self, task_descriptor_index, exception):
         self.task_descriptor_index = task_descriptor_index
         self.exception = exception
 
 
 class DefaultLauncher(object):
+
     def __init__(self, name_prefix='', sigint_timeout=10):
         self.name_prefix = name_prefix
         self.sigint_timeout = sigint_timeout
@@ -387,6 +389,7 @@ class DefaultLauncher(object):
 
 
 class AsynchronousLauncher(threading.Thread):
+
     def __init__(self, launcher):
         super(AsynchronousLauncher, self).__init__()
         self.launcher = launcher
--- a/launch/launch/launcher.py
+++ b/launch/launch/launcher.py
@@ -27,14 +27,12 @@ from launch.task import TaskState
 
 
 class _TaskException(Exception):
-
     def __init__(self, task_descriptor_index, exception):
         self.task_descriptor_index = task_descriptor_index
         self.exception = exception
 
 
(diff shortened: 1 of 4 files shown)


rohbotics:
bump

launch/launch/launcher.py
```diff
@@ -27,14 +27,12 @@
...
 
 class _TaskException(Exception):
-
     def __init__(self, task_descriptor_index, exception):
         self.task_descriptor_index = task_descriptor_index
         self.exception = exception
 
 
 class DefaultLauncher(object):
-
```
dirk-thomas: Why should those two empty lines be removed?

dirk-thomas:
The patch lgtm beside the minor comment. But please run CI jobs for the change and add the links here.

dirk-thomas:
Also if you would like the PRs to be merged please add the "in review" label.

launch/launch/launcher.py
rohbotics: PyCharm was complaining about PEP8 violations,
Looks like it isn't actually specified whether they should be there or not

I will add them back.

launch/launch/launcher.py
```diff
@@ -390,7 +389,6 @@ def _print_process_stacktrace(self, name, future, exception):
 
 
 class AsynchronousLauncher(threading.Thread):
-
```
dirk-thomas: Same here.

rohbotics:
I had to disable fastrtps because of this https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/54

Linux: [Build Status](http://ci.ros2.org/job/ci_linux/1695/)

OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1257/)

launch/launch/launcher.py
wjwwood: IIRC, it was a change in PEP257. They used to say to put this blank line in so it matched classes with docstrings which have a blank line between the class definition and the dosctring. But since then they changed that so that you don't put a blank line between the class definition and the docstring so this blank line is no longer required. I don't think it is explicitly forbidden, but personally I wouldn't have added it if writing the class from scratch:

https://hg.python.org/peps/rev/9b715d8246db

Maybe related: https://github.com/ament/ament_lint/pull/47

I don't think it matters if they're removed or not, so don't spend anymore time changing these.

rohbotics:
CI Looks good.

wjwwood:
Thanks!

Issue #41, opened 2016-09-26
No exit_handler to check the error code of a program that closes 


geoffviola:
Inside the https://github.com/ros2/launch/blob/master/launch/launch/exit_handler.py file, all the exit handlers ignore output, restart the program, or run the default_exit_handler, which shuts all the programs down. I want to launch a program and check its return code for failure. If the error code is non-zero, the launch file should tear itself down. Is that in line with the current design?

dirk-thomas:
I am not sure I understand your exact use case and how it diverges from what is possible with the existing  exit handlers but the design aims to enable implementing any logic using custom handlers.

geoffviola:
Here's the scenario. I have a launch file that starts a node and a custom parameter client. The parameter client will set the parameters and return with a zero exit code, if it ran properly. I do want the launch file to tear down if the error code is non-zero, because I don't want to run the node without the correct parameters. Currently, no exit_handler supports this scenario.

dirk-thomas:
Likely we haven't had this use case in any of the exiting demos and examples yet. A custom written exit handler should be able to fulfill this use case. You can either keep it in your package or create a pull request to contribute it to the `launch` package.

clalancette:
I'm going to close this out since I closed out the PR that implemented this.  Feel free to re-open if we want to keep this feature request.

PR #43, opened 2016-10-04
added  check_error_code_on_close

geoffviola:
https://github.com/ros2/launch/issues/41

### DIFF
--- a/launch/launch/exit_handler.py
+++ b/launch/launch/exit_handler.py
@@ -108,7 +108,7 @@ def ignore_signal_exit_handler(context):
 
 def check_error_code_on_close(context):
     """
-    Succeed if the process shuts down only if the error code is 0.
+    Trigger teardown of launch, if the process returns a non-zero error code.
     """
 
     if int(context.task_state.returncode):
--- a/launch/launch/exit_handler.py
+++ b/launch/launch/exit_handler.py
@@ -35,6 +35,10 @@ def default_exit_handler(context, ignore_returncode=False):
     if not context.launch_state.teardown:
         context.launch_state.teardown = True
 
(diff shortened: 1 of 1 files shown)


launch/launch/exit_handler.py
```diff
@@ -35,6 +35,10 @@ def default_exit_handler(context, ignore_returncode=False):
     if not context.launch_state.teardown:
         context.launch_state.teardown = True
 
+    __default_exit_handler_helper(context)
```
dirk-thomas: The keyword argument `ignore_returncode` must be passed to the new internal function:

```
__default_exit_handler_helper(context, ignore_returncode=ignore_returncode)
```

launch/launch/exit_handler.py
```diff
@@ -100,3 +104,14 @@ def ignore_signal_exit_handler(context):
             context.task_state.returncode = 0
 
     default_exit_handler(context)
+
+
+def check_error_code_on_close(context):
+    """
+    Succeed if the process shuts down only if the error code is 0.
```
dirk-thomas: This docblock is unclear to me. It should describe what it is actually doing. E.g. trigger teardown of launch if the task has a non-zero return code

mikaelarguedas:
@geoffviola What is the status on this ? can you rebase this PR and address comments please ? Thanks!

geoffviola:
I probably won't be able to finish this. The use case is less relevant now. The team moved towards bash scripts instead of launch files, which ignores failures. Beta 2 is supposed to have command line parameters.

clalancette:
The original author isn't that interested in this right now, so I'll close this out.  Feel free to reopen if you still want this.

PR #44, opened 2016-11-01
send SIGINT instead of terminate() if possible

dirk-thomas:
Currently calling `terminate()` will make the process fail with a potentially non-zero return code. If a `SIGINT` is sent instead the process might not respond to it. Imo that is an error case (the process not honoring the `SIGINT` signal) so if the process keeps running it will be catched by the timeout which will then fail the test.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/1863/)
* OS X [Build Status](http://ci.ros2.org/job/ci_osx/1398/)
* [Build Status](http://ci.ros2.org/job/ci_windows/1781/)

### DIFF
--- a/launch_testing/launch_testing/__init__.py
+++ b/launch_testing/launch_testing/__init__.py
@@ -2,6 +2,7 @@ from importlib.machinery import SourceFileLoader
 import io
 import os
 import re
+import signal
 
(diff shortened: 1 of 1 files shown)


tfoote:
+1

PR #46, opened 2016-11-09
yet another exit handler

dirk-thomas:
If the process exits cleanly with a return code of 0 it just continue with the launch file.

If the process returns with an error code it tears down the launch file.

### DIFF
--- a/launch/launch/exit_handler.py
+++ b/launch/launch/exit_handler.py
@@ -100,3 +100,22 @@ def ignore_signal_exit_handler(context):
             context.task_state.returncode = 0
 
     default_exit_handler(context)
+
+
+def exit_on_error_exit_handler(context):
+    """
+    If the return code indicates an error trigger a tear down.
+
+    If the return code is zero continue the other tasks.
+    """
+    try:
+        rc = int(context.task_state.returncode)
+    except (TypeError, ValueError):
+        rc = 1 if bool(context.task_state.returncode) else 0
+    if rc:
+        # trigger tear down if not already tearing down
+        if not context.launch_state.teardown:
+            context.launch_state.teardown = True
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
Note @tfoote : this will replace the parameter `teardown_on_error_only` proposed in https://github.com/ros2/launch/pull/35

tfoote:
I would advocate using the approach in #35 as it has one core implementation and having the different entry points instead of duplicating the logic across each of the exit handlers. #35 was deprioritized so I didn't follow up on the comments, but each of the extensions are valuable and as we approach the combinitoric number of handlers for each option we will definitely want to be avoiding code duplication.

dirk-thomas:
While #35 might address this that PR has been stale for several month. The exit handler added in this PR is being needed for the composition PR and the patch is a easy to review addition. Therefore I suggest merging this as-is. #35 can be revisited afterwards when someone finds the time to do so.

mikaelarguedas:
The code looks good to me, it can be merged if the refactor of the exit handlers is ticketed somewhere.

Question: In what cases do you see the try/except necessary ? In other words when do we expect to have a return code that cannot be converted to `int` but convertible to `bool` ? (I can imagine an empty return code or a garbage string or list/tuple) but have trouble to see in what scenario this may happen.
It doesn't cost anything to keep the check, just asking out of curiosity.

dirk-thomas:
The task can not only be a subprocess (which always returns an int) but also a coroutine. Such a Python function can return anything. Often functions return a string as the error which when being used as a return code ends up to be a non-zero value. Therefore the exit handler needs to try that conversion / exception handling.

Imo the idea of the refactoring is covered by the existing PR #35.

PR #47, opened 2016-11-10
handle Windows case

dirk-thomas:
This is a follow up of #46. I certainly don't think these exit handlers should stay that way. But this will get composition and Windows to be happy together... [Build Status](http://ci.ros2.org/job/ci_windows/1855/)

E.g. this special case should be addressed within launch itself since it is a platform flaw which the exit handlers shouldn't need to work around.


mikaelarguedas:
Should this be ticketed in launch then so that we don't forget ?

dirk-thomas:
I still think #35 is sufficient to revisit the exit handlers.

PR #53, opened 2017-06-09
Add python3 buildtool and exec depends.

nuclearsandwich:
In order to be packaged these python3 dependencies are needed.


dirk-thomas:
Same question as on similar tickets about the need for a `python3` dependency.

nuclearsandwich:
> Same question as on similar tickets about the need for a python3 dependency.

https://github.com/ament/ament_lint/pull/76#discussion_r122558187

nuclearsandwich:
Closing for reasons outlined in https://github.com/ament/ament_lint/pull/76#issuecomment-310195822

PR #57, opened 2017-07-06
Added support for command line launch arguments

bponsler:
Updated to support passing in command line arguments in the same format used by roslaunch (i.e., arg_name:=value). This supports zero or more arguments added after one or more launch files passed on the command line.

Updated usage:

    usage: launch [-h] launch_file [launch_file ...] [arg [arg ...]]
    
    Launch the processes specified in a launch file.
    
    positional arguments:
      launch_file  The launch file.
      arg          An argument to the launch file (e.g., arg_name:=value).
    
    optional arguments:
      -h, --help   show this help message and exit

### DIFF
--- a/launch/launch/arguments.py
+++ b/launch/launch/arguments.py
@@ -1,0 +1,55 @@
+# Copyright 2015 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+def get_launch_args(argv, separator=":="):
+    """Return a dictionay of key value pairs for launch arguments passed
+    on the command line using the format "key:=value". This will process
+    every string in the argv list.
+
+    NOTE: all key value pairs will be returned as strings and no type
+    checking is applied to the values.
+
+    For example:
+
+        argv = ["arg1:=123", "hello", "arg2:=true", "arg3:=true", "/path/to/file"]
+        args = get_launch_args(argv)
+        # Results in args = {'arg1': '123', 'arg2': 'true', 'arg3': 'true'}
+
+    :param argv: is the list of string command line arguments
+    :param separator: is the string separator for each key:=value pair (e.g., ':=')
+    :returns: a dictionary of string key value pairs
+    :rtype: dictionary
+
+    """
+    launch_args = {}
+
+    # Separate the command line arguments into a dictionary of
+    # launch key value pairs of the format key:=value. Process the
+    # first element even though it will likely be the executable name
+    # to support other lists.
+    for arg in argv:
+        # Split the arg into a key value pair
+        arg_pair = arg.split(separator)
+        if len(arg_pair) != 2:
+            continue  # Skip non launch args
+
+        key, value = arg_pair
+
+        # Ignore pairs with an empty key, which isn't a useful argument pair
+        if len(key) == 0:
+            continue
+
+        launch_args[key] = value
+
+    return launch_args
--- a/launch/launch/arguments.py
+++ b/launch/launch/arguments.py
@@ -18,7 +18,7 @@ def get_launch_args(argv, separator=':='):
     Get the list of launch arguments passed on the command line.
 
     Return a dictionary of key value pairs for launch arguments passed
-    on the command line using the format "key{separator}value". This will
+    on the command line using the format 'key{separator}value'. This will
     process every string in the argv list.
 
     NOTE: all key value pairs will be returned as strings and no type
@@ -26,7 +26,7 @@ def get_launch_args(argv, separator=':='):
 
     For example:
 
-        argv = ["arg1:=123", "hello", "arg2:=true", "arg3:=true", "/path/to/file"]
+        argv = ['arg1:=123', 'hello', 'arg2:=true', 'arg3:=true', '/path/to/file']
         args = get_launch_args(argv)
         # Results in args = {'arg1': '123', 'arg2': 'true', 'arg3': 'true'}
 
@@ -44,8 +44,9 @@ def get_launch_args(argv, separator=':='):
     # first element even though it will likely be the executable name
     # to support other lists.
     for arg in argv:
-        # Split the arg into a key value pair
-        arg_pair = arg.split(separator)
+        # Split the arg into a key value pair (allow values to contain
+        # the separator)
+        arg_pair = arg.split(separator, 1)
         if len(arg_pair) != 2:
             continue  # Skip non launch args
 
--- a/launch/launch/arguments.py
+++ b/launch/launch/arguments.py
@@ -14,7 +14,8 @@
 
 
 def get_launch_args(argv, separator=":="):
-    """Get the list of launch arguments passed on the command line.
+    """
+    Get the list of launch arguments passed on the command line.
 
     Return a dictionary of key value pairs for launch arguments passed
     on the command line using the format "key{separator}value". This will
--- a/launch/launch/arguments.py
+++ b/launch/launch/arguments.py
@@ -1,4 +1,4 @@
-# Copyright 2015 Open Source Robotics Foundation, Inc.
+# Copyright 2017 Open Source Robotics Foundation, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -12,10 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
 def get_launch_args(argv, separator=":="):
-    """Return a dictionay of key value pairs for launch arguments passed
-    on the command line using the format "key:=value". This will process
-    every string in the argv list.
+    """Get the list of launch arguments passed on the command line.
+
+    Return a dictionary of key value pairs for launch arguments passed
+    on the command line using the format "key{separator}value". This will
+    process every string in the argv list.
 
     NOTE: all key value pairs will be returned as strings and no type
     checking is applied to the values.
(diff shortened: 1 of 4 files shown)


launch/launch/main.py
```diff
@@ -22,25 +22,86 @@
...
-    if not os.path.exists(filename) or not os.path.isfile(filename):
-        raise argparse.ArgumentError("'%s' does not exist" % filename)
-    return filename
+    return os.path.exists(filename) and os.path.isfile(filename)
+
+
+def get_launch_arg(arg, separator=":="):
+    # Make sure the separator exists
+    if separator not in arg:
+        return None
```
wjwwood: This check is redundant with the one below. If `:=` is not in the string then a list of length one with be returned. I don't think checking this before hand will really save much time especially since this is a one time cost at startup of a launch file.

bponsler: These changes actually parse the `key:=value` pairs and creates the `arguments` dictionary that contains all of the mappings from key to value. Then the `launch()` functions just need to deal with a regular dictionary for argument checking.

I figured that the launch tool might be drastically different in the future. No worries if these changes don't get pulled in, or entirely removed at a later date.

Good catch on the redundant check, I will remove it.

dhood:
An observation I have (from trying (not very hard) to get arguments through to a launch file in https://github.com/ros2/turtlebot2_demo/pull/44): This passes a dictionary of parsed arguments through to the launch file, but existing launch files [are expecting `argv`](https://github.com/ros2/launch/blob/c7a34bf1ce9e22b55004e06b36854f7e954bdc5b/launch/launch/loader.py#L18). Granted, before this change, the arguments were actually being passed through to the launch file as an empty dictionary, but I assumed the intention was to pass argv through and let launch files do what they want with it which seems like more flexible design.

if we were to instead pass the parser through to the launch file (and use `parse_known_args` here), @bponsler can access their arguments in their launch file [like this](https://github.com/ros2/turtlebot2_demo/commit/74f59f93d1d49d57a342562c359d786d0609934f) and can leverage argparse's options instead of re-implementing them here (e.g. support for `--my-flag` booleans).

There are other challenges that would come with the argument parser extension approach (e.g. multiple definition of arguments, which perhaps could be resolved with subparsers), so it may not actually be feasible, but if nothing else we could consider passing argv through in addition to the dictionary

bponsler:
@dhood I was a bit confused with the launch files expected `argv` but passing down a dictionary, so I stuck with the dictionary. Your proposal seems like a perfectly reasonable, and flexible, way of passing arguments to launch files. Additionally, it has an added benefit of supporting syntax like: `launch my_launch.py -- --help` to view the help of the launch file which would automatically include arguments (where my implementation could not support this). I agree that passing down the parser is preferred to my suggested implementation (assuming it is feasible).

CodingPeon:
I've tried this on my project and passing in command line arguments works. What does it need to get the PR merged? Is there anything I could contribute to get this going?

dhood:
First up, sorry for the delay in response. We've discussed this "offline" and have decided it's best to not have launch files tied to anything python-specific for now (dictionaries, argparsers). So, the following would be the appropriate direction for this PR:

1. Pass `argv` through to launch files, instead of a pre-parsed dictionary.
1. Move the parsing of `arg_name:=value` parameters to a function that can be called from within python launch files that want to have this functionality.

@bponsler are you interested in re-factoring this (and ideally adding some tests)? Otherwise sounds like there are others willing to pitch in

bponsler:
@dhood that sounds like a good idea and thanks for the update. I should have some time in the next few days to throw together the desired changes.

mikaelarguedas:
@clalancette is this back in prograss ? it looks like the comments have been addressed in https://github.com/ros2/launch/pull/57/commits/4247584eeb9b9f17f8b5ff54f91748db7d47e6e9

clalancette:
Oh, I may have misread the ordering of events.  @bponsler , is this ready for another review?

bponsler:
@clalancette Yep, I made new changes to address the comments raised by @dhood

clalancette:
Apologies for erroneously changing the status, switching back.

bponsler:
Thanks @dhood for catching those mistakes! 0584f2e should address all of your concerns, but let me know if I missed something.

dhood:
I pushed https://github.com/ros2/launch/pull/57/commits/1eb006fbe02cd7c0cff7c7a45fd76625054d2f9a and this CI should come out green now

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3007/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/431/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2428/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3102/)

launch/launch/arguments.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # Separate the command line arguments into a dictionary of
+    # launch key value pairs of the format key${separator}value. Process the
+    # first element even though it will likely be the executable name
+    # to support other lists.
+    for arg in argv:
+        # Split the arg into a key value pair
+        arg_pair = arg.split(separator)
+        if len(arg_pair) != 2:
+            continue  # Skip non launch args
```
dirk-thomas: Why shouldn't this allow values containing `:=`? I would suggest tu call `split` with a limit of 2 and remove this restriction.

launch/launch/main.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2015 Open Source Robotics Foundation, Inc.
+# Copyright 2017 Open Source Robotics Foundation, Inc.
```
dirk-thomas: The previous year must be kept: `2015-2017`.

launch/launch/main.py
```diff
@@ -22,27 +22,49 @@
 
 
 def file_exists(filename):
-    if not os.path.exists(filename) or not os.path.isfile(filename):
-        raise argparse.ArgumentError("'%s' does not exist" % filename)
-    return filename
+    return os.path.exists(filename) and os.path.isfile(filename)
```
dirk-thomas: This check is redundant. `isfile` already implies `exists`. And in that case the function should be removed since `file_exists` maps to only `is_file`.

launch/launch/main.py
```diff
@@ -22,27 +22,49 @@
...
+        if file_exists(arg):
+            launch_files.append(arg)
+        else:
+            break
+
+    # Ensure that at least one existing launch file was given
+    if len(launch_files) == 0:
+        print('error: you must pass at least one valid launch file', file=sys.stderr)
+        parser.print_help()
+        sys.exit(2)
```
dirk-thomas: This could use `parser.error(...)` instead.

launch/test/test_launch_args.py
```diff
@@ -0,0 +1,106 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from launch.arguments import get_launch_args
+
+
+def test_launch_args():
+    # Test empty arguments
+    argv = []
+    args = get_launch_args(argv)
+    assert len(args) == 0, "Failed empty list"
```
dirk-thomas: Please use single quote throughout the file where possible (when no `'` are in the string which would require quoting).

launch/launch/arguments.py
bponsler: Great point that I had not thought about -- thanks.

launch/launch/main.py
bponsler: Doh, thanks for catching that

bponsler:
@dirk-thomas -- thanks for the review comments! 0f0f13c should resolve all the issues you found

launch/launch/main.py
```diff
@@ -21,28 +21,45 @@
...
+        # non-existing file is found
+        if os.path.isfile(arg):
+            launch_files.append(arg)
+        else:
+            break
+
+    # Ensure that at least one existing launch file was given
+    if len(launch_files) == 0:
+        parser.error('you must pass at least one valid launch file')
+        sys.exit(2)
```
dirk-thomas: This line is never executed. [parser.error](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.error) already does exactly the same.

launch/launch/main.py
```diff
@@ -21,28 +21,45 @@
...
 
     launcher = DefaultLauncher()
-    for launch_file in args.launch_file:
+    for launch_file in launch_files:
         launch_descriptor = LaunchDescriptor()
-        load_launch_file(launch_file, launch_descriptor, arguments)
+        load_launch_file(launch_file, launch_descriptor, argv)
```
dirk-thomas: Doesn't `argv` also contain the launch file names?

bponsler: Yes, it contains all command line arguments except the name of the executable

bponsler:
@dirk-thomas -- I like the --args idea, it makes it clearer and would get rid of the custom parsing code.

Should I perform any verification logic on the --args values? Like ensuring they all are of the form 'key:=value'. I think it might make more sense not to have any argument verification here and delay that logic to specific launch files as desired. What do you think? Will still verify the launch files exit as it was being done previously.

dirk-thomas:
> Should I perform any verification logic on the --args values? Like ensuring they all are of the form 'key:=value'. I think it might make more sense not to have any argument verification here and delay that logic to specific launch files as desired.

I agree, I think every launch file can and should do its own argument handling.

> Also -- should I continue passing just argv to the launch function, pass just parser.args, or pass both?

I would think that each launch file only needs `args.arg` and shouldn't get the launch files. The help should probably clarify that all args are being passed to all launch files.

bponsler:
Thanks for the clarification -- I will make these changes later today

launch/launch/main.py
```diff
@@ -35,14 +35,38 @@ def main(argv=sys.argv[1:]):
...
+    # greedy and consumes the remaining arguments. Therefore the arguments
+    # need to be validated afterward
+    launch_files = []
+    for arg in args.launch_file:
+        # Store consecutive existing files and stop once a
+        # non-existing file is found
+        if os.path.isfile(arg):
+            launch_files.append(arg)
+        else:
+            break
```
dirk-thomas: With `--args` being declared explicitly this seems to not be necessary anymore.

launch/launch/main.py
```diff
@@ -35,14 +35,38 @@ def main(argv=sys.argv[1:]):
         type=file_exists,
         nargs='+',
         help='The launch file.')
+    parser.add_argument(
+        '--args',
+        metavar="arg",
+        type=str,
+        nargs='+',
+        help='An argument to the launch file (e.g., arg_name:=value). All ' +
```
dirk-thomas: The trailing ` +` is not necessary. Python automatically concats string literals.

launch/launch/main.py
```diff
@@ -35,14 +35,38 @@ def main(argv=sys.argv[1:]):
...
+        'arguments will be passed to each launch file.')
```
dirk-thomas: Did you run the linters locally? I would expect that no indentation on the wrapped line might result in a linter warning.

bponsler: I just re-ran the linter, and fixed associated errors as well as removing the chunk of useless code. Thanks for the reviews

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3067/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/459/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2465/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3136/)

dirk-thomas:
Thank you for improving this!

PR #59, opened 2017-08-18
Filter regex patterns in addition to prefixes

dhood:
this will allow us to filter out lines based on regexes instead of just prefixes. note that even if you specify a multiline regex it will only ever be matched against one line of console output at a time (this was also true how we were matching the prefixes before)

this was motivated by wanting to filter things like the fastrtps error message in 
http://ci.ros2.org/view/nightly/job/nightly_osx_repeated/772/testReport/(root)/projectroot/test_tutorial_add_two_ints_server_add_two_ints_client_async__rmw_fastrtps_cpp/
where prefix matching wouldn't be sufficient. Even though we probably don't want to filter that error message _in general_, I found this useful while investigating flaky tests. Using this I could run tests N times ignoring this error, preventing unrelated test failures from its output.

CI including relevant changes in dependencies (for which I'll open PRs if this change looks OK conceptually)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3049/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/446/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2451/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3123/)

### DIFF
--- a/launch_testing/launch_testing/__init__.py
+++ b/launch_testing/launch_testing/__init__.py
@@ -78,7 +78,7 @@ class InMemoryHandler(LineOutput):
         for line in lines.splitlines():
             # Filter out stdout that comes from underlying DDS implementation
             # Note: we do not currently support matching filters across multiple stdout lines.
-            if any(re.fullmatch(pattern, line) for pattern in self.filtered_patterns):
+            if any(re.match(pattern, line) for pattern in self.filtered_patterns):
                 continue
             if any(line.startswith(prefix) for prefix in self.filtered_prefixes):
                 continue
--- a/launch_testing/launch_testing/__init__.py
+++ b/launch_testing/launch_testing/__init__.py
@@ -27,7 +27,7 @@ class InMemoryHandler(LineOutput):
         lines to be ignored if they start with one of the prefixes. By default lines starting with
         the process ID (`b'pid'`) and return code (`b'rc'`) will be ignored.
     :param filtered_patterns: A list of byte strings representing regexes that will cause output
-        lines to be ignored if they completely match one of the regexes.
+        lines to be ignored if they match one of the regexes.
     :param filtered_rmw_implementation: RMW implementation for which the output will be ignored
         in addition to the `filtered_prefixes`/`filtered_patterns`.
     :param exit_on_match: If True, then when its output is matched, this handler
@@ -46,14 +46,8 @@ class InMemoryHandler(LineOutput):
         exit_on_match=True,
     ):
         super(LineOutput, self).__init__()
-        if filtered_prefixes is None:
-            self.filtered_prefixes = get_default_filtered_prefixes()
-        else:
-            self.filtered_prefixes = filtered_prefixes
-        if filtered_patterns is None:
-            self.filtered_patterns = get_default_filtered_patterns()
-        else:
-            self.filtered_patterns = filtered_patterns
+        self.filtered_prefixes = filtered_prefixes or get_default_filtered_prefixes()
+        self.filtered_patterns = filtered_patterns or get_default_filtered_patterns()
 
         if filtered_rmw_implementation:
             self.filtered_patterns.extend(
@@ -80,9 +74,9 @@ class InMemoryHandler(LineOutput):
         for line in lines.splitlines():
             # Filter out stdout that comes from underlying DDS implementation
             # Note: we do not currently support matching filters across multiple stdout lines.
-            if any(re.match(pattern, line) for pattern in self.filtered_patterns):
-                continue
             if any(line.startswith(prefix) for prefix in self.filtered_prefixes):
+                continue
+            if any(re.match(pattern, line) for pattern in self.filtered_patterns):
                 continue
             self.stdout_data.write(line + b'\n')
             if not self.regex_match and not self.matched:
--- a/launch_testing/launch_testing/__init__.py
+++ b/launch_testing/launch_testing/__init__.py
@@ -23,11 +23,11 @@ class InMemoryHandler(LineOutput):
         that will only need one line to match, instead of the entire output.
     :param regex_match: If true, treat the expected_lines as a regular expression in match
         accordingly.
-    :param filtered_prefixes: A list of byte strings representing prefixes that will cause output
-        lines to be ignored if they start with one of the prefixes. By default lines starting with
-        the process ID (`b'pid'`) and return code (`b'rc'`) will be ignored.
+    :param filtered_patterns: A list of byte strings representing regexes that will cause output
+        lines to be ignored if they completely match one of the regexes. By default lines starting
+        with the process ID (`b'pid'`) and return code (`b'rc'`) will be ignored.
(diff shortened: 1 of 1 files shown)


dirk-thomas:
This breaks any code currently using this functionality. Maybe the existing API can be kept and the pattern filter introduced beside it? Updating the rmw implementations to always provide patterns seems fine since that won't affect users (except those implementing their own rmw impl. with custom filters).

dhood:
If every prefix can be expressed as a regex, I think it's better for us to drop support for the prefix option so that we don't have to maintain it and just update our usage to the new API (no one else on github uses this function)

If, however, we really like the simplicity of the prefix option, we could keep support for it, but that's a different reason

dhood:
@dirk-thomas are you in favour of keeping support for the prefix filtering, or was it just an idea you were floating? If so, I can change it, it's easy to do, I'd just like to understand the reasoning (since I think you're usually in favour of maintaining as few options as possible)

dirk-thomas:
> are you in favour of keeping support for the prefix filtering, or was it just an idea you were floating?

Yes, I think we should keep supporting prefix filtering. While every prefix can be expressed as a regex it goes imo against the design principle that a user shouldn't "pay" for a feature he isn't using.

> since I think you're usually in favour of maintaining as few options as possible

If multiple options achieve the same think I am certainly in favor reducing them to the minimal amount. In this case I see the two options as functionality different though: the prefix filtering is the "best" approach if that is sufficient for the use case, the regex allows more complicated filters but come with the "price" of additional overhead.

dhood:
ok, cool. originally it seemed because of api breakage. this makes sense to me thanks, will update

dhood:
This PR has been updated to _add_ `filtered_patterns` as an option (and update the RMW-specific filters to be regexes instead of prefixes)

CI with the attached PR updating the connext license output https://github.com/ros2/rmw_connext/pull/242:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3164/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/503/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2521/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3216/) (unrelated flaky test)

launch_testing/launch_testing/__init__.py
```diff
@@ -40,17 +42,22 @@ class InMemoryHandler(LineOutput):
...
             self.filtered_prefixes = filtered_prefixes
+        if filtered_patterns is None:
+            self.filtered_patterns = get_default_filtered_patterns()
+        else:
+            self.filtered_patterns = filtered_patterns
 
         if filtered_rmw_implementation:
             rmw_output_filter = get_rmw_output_filter(filtered_rmw_implementation)
-            self.filtered_prefixes.extend(rmw_output_filter)
+            self.filtered_patterns.extend(rmw_output_filter)
```
dirk-thomas: Why should the rmw filters need to be patterns? If a prefix filter is sufficient it would be good if an rmw impl. could use that.

launch_testing/launch_testing/__init__.py
```diff
@@ -70,6 +77,9 @@ def on_stdout_lines(self, lines):
 
         for line in lines.splitlines():
             # Filter out stdout that comes from underlying DDS implementation
+            # Note: we do not currently support matching filters across multiple stdout lines.
+            if any(re.fullmatch(pattern, line) for pattern in self.filtered_patterns):
```
dirk-thomas: Does this need to be a `fullmatch`? Could it just use `match` instead and leave it up to the regex to use `^` and `$`?

launch_testing/launch_testing/__init__.py
dhood: I understood from your comment above that, while we want users to give users the choice between patterns or prefixes, it's not necessary for "internal" rmw filters:

> Updating the rmw implementations to always provide patterns seems fine since that won't affect users (except those implementing their own rmw impl. with custom filters).

launch_testing/launch_testing/__init__.py
dhood: I think I was just trying to avoid regexes from matching accidentally, but I agree that that should be the choice of the user.

If we just use `match`, which only matches at the start of a line, then actually the prefix option could then be collapsed back into the pattern option... in which case we can just document that filters can be literal strings of prefixes for those users just wanting the simplicity of specifying prefixes. How's that sound?

launch_testing/launch_testing/__init__.py
dirk-thomas: That was 20 days ago -  a lot has happened since then :wink:

Sure, we could make all rmw impl. use regex since we kind of maintain most of them. That was the main rational in the first review. But my latest comment was about efficiency - why use a regex if a prefix does the job just fine? I haven't spend any thought on how this could be done. An rmw impl. would need to somehow declare the type of their filters - either by using a different resource key or embedding the type information in the resource. The first seems to be easier.

launch_testing/launch_testing/__init__.py
dirk-thomas: I don't think you can determine if the filter is supposed to be a string or a regular expression. E.g. if it contains a `.` you can't guess the semantic. Therefore I think you need to know the type of the filter (prefix string vs. regex) in order to invoke the corresponding logic.

Again, I think we should never use a regular expression match when a string match is sufficient  simply because of the unjustified overhead.

dhood:
OK this PR now extends both the API and the rmw-specified filters to accept both prefixes and regexes. https://github.com/ros2/rmw_connext/pull/242 has been updated to clarify it's registering a prefix to filter but otherwise no change to the codebase.

CI with this PR and https://github.com/ros2/rmw_connext/pull/242:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3278/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/557/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2612/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3319/) (Warnings from fastrtps that have been reported)

launch_testing/launch_testing/__init__.py
```diff
@@ -70,8 +73,11 @@ def on_stdout_lines(self, lines):
 
         for line in lines.splitlines():
             # Filter out stdout that comes from underlying DDS implementation
+            # Note: we do not currently support matching filters across multiple stdout lines.
             if any(line.startswith(prefix) for prefix in self.filtered_prefixes):
                 continue
+            if any(re.match(pattern, line) for pattern in self.filtered_patterns):
```
dirk-thomas: I was first worried that using the same pattern without compiling it repeatedly would be slower. But it turns out modern Python already do a great job compiling the pattern and caching it for future use. So all good here :smiley:

dhood:
thanks for the nitpick commits

Issue #62, opened 2017-09-11
Launch files broken since #57 has been merged


mikaelarguedas:
Opening this for visibility
I tried the lifecycle launchfile and the dummy_robot_bringup one and they both crash. If I revert #57 they work.

dirk-thomas:
I can confirm that the current state is broken. Are you going to work on a fix or should I look into it?

mikaelarguedas:
I'll look into it

wjwwood:
@mikaelarguedas can you please post the error you get in the crash, so others can confirm it's the same issue. I get:

```
% launch share/dummy_robot_bringup/launch/dummy_robot_bringup.py
Traceback (most recent call last):
  File "/tmp/testing_beta3/bin/ros2-osx/bin/launch", line 11, in <module>
    load_entry_point('launch==0.0.0', 'console_scripts', 'launch')()
  File "/tmp/testing_beta3/bin/ros2-osx/lib/python3.6/site-packages/launch/main.py", line 50, in main
    load_launch_file(args.launch_file, launch_descriptor, args.args)
  File "/tmp/testing_beta3/bin/ros2-osx/lib/python3.6/site-packages/launch/loader.py", line 20, in load_launch_file
    launch_module = loader.load_module()
  File "<frozen importlib._bootstrap_external>", line 399, in _check_name_wrapper
  File "<frozen importlib._bootstrap_external>", line 823, in load_module
  File "<frozen importlib._bootstrap_external>", line 682, in load_module
  File "<frozen importlib._bootstrap>", line 245, in _load_module_shim
  File "<frozen importlib._bootstrap>", line 427, in spec_from_loader
  File "<frozen importlib._bootstrap_external>", line 574, in spec_from_file_location
  File "<frozen importlib._bootstrap_external>", line 664, in is_package
  File "<frozen importlib._bootstrap_external>", line 66, in _path_split
AttributeError: 'list' object has no attribute 'rpartition'
```

dirk-thomas:
Same for me.

mikaelarguedas:
Same for me

mikaelarguedas:
fixed in #63

PR #63, opened 2017-09-11
Fix 62

mikaelarguedas:
First commit fix the error mentionned in #62 
Second commit rename the argument to make is clearer that there can be several launchfiles

### DIFF
--- a/launch/launch/main.py
+++ b/launch/launch/main.py
@@ -31,10 +31,10 @@ def main(argv=sys.argv[1:]):
     parser = argparse.ArgumentParser(
         description='Launch the processes specified in a launch file.')
     parser.add_argument(
-        'launch_files',
+        'launch_file',
         type=file_exists,
         nargs='+',
-        help='The launch file(s).')
+        help='The launch file.')
     parser.add_argument(
         '--args',
         metavar='arg',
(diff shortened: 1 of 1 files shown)


launch/launch/main.py
```diff
@@ -31,10 +31,10 @@ def main(argv=sys.argv[1:]):
     parser = argparse.ArgumentParser(
         description='Launch the processes specified in a launch file.')
     parser.add_argument(
-        'launch_file',
+        'launch_files',
```
dirk-thomas: Changing this name also changes the help text (the metavar place holder). The help text should continue to refer to each argument as a launch_file (singular).

launch/launch/main.py
```diff
@@ -45,9 +45,9 @@ def main(argv=sys.argv[1:]):
     args = parser.parse_args(argv)
 
     launcher = DefaultLauncher()
-    for launch_file in args.launch_file:
+    for launch_file in args.launch_files:
         launch_descriptor = LaunchDescriptor()
-        load_launch_file(args.launch_file, launch_descriptor, args.args)
+        load_launch_file(launch_file, launch_descriptor, args.args)
```
dirk-thomas: Only the change in this line is necessary. The rest should stay unchanged.

mikaelarguedas: ok, reverted the second commit

PR #69, opened 2018-01-21
Fix hang on keyboard interrupt

### LINKED ISSUE #64: Launcher hangs with keyboard interrupt on Windows
Most of our usage of launch files is in tests, where launch_testing causes signals to be sent. But if you run them manually and ctrl-c them, they hang (for example `demo_nodes_cpp`'s `test_set_and_get_parameters__rmw_fastrtps_cpp_Release.py`). I'm yet to find a launch file that doesn't cause this to happen. 

In https://github.com/ros2/demos/issues/175#issuecomment-329030173 @dirk-thomas noted that 
> The process hangs for me in `launcher.py` on the `loop.close()` call. Both subprocesses have been `SIGTERM`ed and the loop is not running anymore. But the `close()` call still hangs forever.

Simple ROS2-agnostic example:

```
$ cat launch_wait.py
import os

from launch import LaunchDescriptor
from launch.launcher import DefaultLauncher

def main():
    launcher = DefaultLauncher()

    if os.name == 'nt':
        cmd=['ping', '127.0.0.1', '-n', '60']
    else:
        cmd=['sleep', '60']
    launch_descriptor.add_process(
        cmd=cmd,
        name='dummy_process',
    )
    launcher.add_launch_descriptor(launch_descriptor)

    rc = launcher.launch()
    print('Return code: ' + str(rc))
    exit(rc)


if __name__ == '__main__':
    main()
```
Behaviour on linux:
```
$ python3 launch_wait.py
(dummy_process) pid 13687: ['sleep', '60'] (stderr > stdout, all > console)
^C(dummy_process) signal SIGINT
(dummy_process) rc -2
Return code: 1
```

Behaviour on windows:
```
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC>python C:\Users\osrf\Downloads\r2b3-testing\overlay\src\launch_nothing.py
(dummy_process) pid 4416: ['ping', '127.0.0.1', '-n', '60'] (stderr > stdout, all > console)
[dummy_process]
[dummy_process] Pinging 127.0.0.1 with 32 bytes of data:
[dummy_process] Reply from 127.0.0.1: bytes=32 time<1ms TTL=128
[dummy_process] Reply from 127.0.0.1: bytes=32 time<1ms TTL=128
[dummy_process] Reply from 127.0.0.1: bytes=32 time<1ms TTL=128
<manually ctrl c>
(dummy_process) signal SIGTERM
(dummy_process) rc 3221225786
<hangs>
```

rhololkeolke:
Relates to #64

When keyboard interrupt exception occurs loop.run_forever is
called, but there is no loop.stop call. This causes a hang.

I don't have a windows machine to test on.

### DIFF
--- a/launch/launch/launcher.py
+++ b/launch/launch/launcher.py
@@ -219,6 +219,7 @@ class DefaultLauncher:
         if launch_state.returncode is None:
             launch_state.returncode = 0
         self.launch_complete.set()
+        self.loop.stop()
         return launch_state.returncode
 
     async def _terminate_processes(self, launch_state, all_futures):


Karsten1987:
I can confirm that this fixes the problem on my mac machine. Will try to run CI on this asap.

Karsten1987:
Unfortunately, I believe this doesn't fix the issue on Windows.

mikaelarguedas:
Thanks @rhololkeolke for the contribution.
As this fixes the issue on some of the platforms we will take this as an incremental improvement. PR description has been modified to not close the original issue as this patch is not sufficient to fix it.
@Karsten1987 or @dhood  to run CI and merge if green

Karsten1987:
CI:

[Build Status](https://ci.ros2.org/job/ci_linux/3967/)
[Build Status](https://ci.ros2.org/job/ci_osx/3303/)
[Build Status](https://ci.ros2.org/job/ci_windows/4066/)

mikaelarguedas:
Test failures on Windows and MacOS unrelated to this PR: https://github.com/ros2/build_cop/issues/83.
Merging...

mikaelarguedas:
Thanks @rhololkeolke for the patch!

PR #70, opened 2018-01-26
update maintainer to match package.xml

mikaelarguedas:
follow-up of 3294e2a1f836867725c813e5091d0f5b1650e9b5


dirk-thomas:
We can apply the patch as-is. But in general we need to review and change the current "assignment" since it is not spread across the team evenly.

mikaelarguedas:
Arg sorry I merged before seeing your comment.
Before the release we asked people to pick up packages whose maintainers were not in the team anymore. That was done for this package, I just stumbled upon the inconsistency so rectified it. If anyone wants to take over this package and submit a PR to reflect it I'm happy to merge it :+1:

PR #73, opened 2018-06-04
move launch files into launch.legacy namespace

wjwwood:
In preparation for opening pull requests for the new launch API I've been working on, this pull request (and the connected ones) will simply move the existing launch system into a sub package (called `legacy`) and updates anything using launch to also use this new API.

That way I can introduce the new API in smaller steps without first having to have it completely replace the existing tool and work in all cases.

This is the end goal, but there are a few things in `launch_testing` and a few of the output handlers I still need to port before I can start doing that. Plus this gives us something to compare and fallback to as we start using/tryout the new tool.


wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4541/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1454/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3728/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4616/) (this is a "performance" test failure, and I think it's unrelated as nothing this test uses nor `rclpy` have been changed in these pr's)

wjwwood:
This is ready for review.

dirk-thomas:
Would the [history](https://github.com/ros2/launch/commits/refactor_launch_into_launch_legacy/launch/launch/legacy/exit_handler.py) be preserved if the move would be done in two steps: one commit just moving the file, another commit modifying the files? If yes, I would advocate for that.

wjwwood:
Btw, the feature was requested but never went anywhere, but there is a chrome extension to help with it:

https://stackoverflow.com/questions/5646174/how-to-make-github-follow-directory-history-after-renames

wjwwood:
I'll take the positive review to apply to the other pr's too since they're so automatically derivative. Thanks!

dirk-thomas:
Was there a reason why the `launch_testing` module wasn't moved into a `legacy` namespace too? I am considering doing that now to "make room" for new testing logic as well as keep the API "symmetric".

wjwwood:
No reason that’s fine for me.

PR #74, opened 2018-06-05
New launch system

wjwwood:
This is the first pull request in a series, which adds the first pass at the reference implementation described in https://github.com/ros2/design/pull/163.

I'm still working on this pull request, but some highlights to review in the mean time:

- the `architecture.rst` document: https://github.com/ros2/launch/pull/74/files#diff-e25d31144ab0a51cf4d41fa107f6c9c3
- the use of type checking with Python's `typing` module and `mypy`
  - specifically, there are some import loops which can be avoided normally, but not when doing type checking, so there are cases where "fake" imports must be done to satisfy `mypy`, and I'd like feedback on that
  - e.g. https://github.com/ros2/launch/pull/74/files#diff-84e2bbbeb5e4ca165d268d4962177343R23
- the example: https://github.com/ros2/launch/pull/74/files#diff-450e2fd128e18f66fc9c0ca9d834ca43
- console output from `ExecuteProcess` is currently modeled after `roslaunch`, but could be refactored to be like `launch` was, or something else entirely

What's known to be missing from this pr (and forth coming as soon as I can redo the commits) are tests and some example usage.

In a follow up pull request, I will add the `launch_ros` package and some demos using it, as well as the `ros2launch` package.

Things to open issues about after merging:

- Tests for non-core classes (things like specific actions, utilities, substitutions, etc...)
- Improvements to the introspection system.
- Fix introspection of event handlers
- Update introspection support for some Actions (e.g. the ExecuteProcess action could report more of what it might do in certain situations)
- Add option for hidden versus non-hidden (default) description elements (so you can see important stuff, but also visualize deep details if desired)
- Add extensions to the launch service
- Create extension to launch service for standard logging behaviors
- Conditions
- GroupAction
- Launch Arguments
- UnsetLaunchConfiguration
- Check iterable contains only types in `ensure_argument_type()`, see: https://github.com/ros2/launch/pull/74#discussion_r194928334


wjwwood:
I added an example in f353805 for anyone who wants to start reviewing.

wjwwood:
45e63bc added the tests I that I had time to clean up for this pull request. The more detailed actions still need proper automated tests, but are otherwise pretty well covered in demos I have to pull request later. I'll add expanding the test coverage to the list of new issues that need to be made as follow up work.

wjwwood:
> Does the current code come with any callables which can be passed to the OnProcessIO handler to e.g. just print to stdout? Since the example defines its own local function for that.

Not right now, the reason is that I plan to integrate the "default" one from `launch_ros` so it can hook into the ROS logging system, though I'm not 100% sure that's what we want. I might add one to launch just for convenience. I plan to still port the output handler concept from the old `launch` in some form.

> Why is the example file called launch_counter_good_bad_ugly.py? It looks pretty clean and intuitive to read or do you consider any part of it bad / ugly style?

Sorry, that's because there are three counters, one that behaves, one that ignores sigint (bad), and one that ignores sigint and sigterm (ugly). I can shorten the name. It was late when I made that file :)

launch/doc/source/architecture.rst
```diff
@@ -0,0 +1,208 @@
...
+
+- :class:`launch.actions.SetLaunchConfiguration`
+
+  - This action will set a :class:`launch.LaunchConfiguration` to a specified value, creating it if it doesn't already exist.
+  - These launch configurations can be accessed by any action via a substitution, but are scoped by default.
+
+- :class:`launch.actions.DeclareLaunchDescriptionArgument`
+
+  - This action will declare a launch description argument, which can have a name, default value, and documentation.
+  - The argument will be exposed via a command line option for a root launch description, or as action configurations to the include launch description action for the included launch description.
```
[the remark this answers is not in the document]
wjwwood: My plan was to make argument without default values required, I think this is how it works in `roslaunch`: https://wiki.ros.org/roslaunch/XML/arg#Usecase_pattern

wjwwood:
Ok, I'm going to put this one into review and build the ROS one on top.

CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4587/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1493/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3771/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4669/)

wjwwood:
Note, I will create issues for the things listed at the end of the description of this issue after merging.

launch/launch/events/process/signal_process.py
```diff
@@ -0,0 +1,63 @@
...
+
+    @property
+    def signal_name(self) -> Text:
+        """
+        Getter for signal_name.
+
+        It will be something like (e.g.) 'SIGINT', or the number if name is unknown.
+        """
+        if self.__signal_name is None:
+            self.__signal_name = get_signal_name(self.__signal)
```
dhood: Did you consider enforcing that `signal` is in the `signal.Signals` enum, so we can just do `__signal.name`? That would avoid the need for the get_signal_name_impl.py file and its hard-coded list.

launch/.flake8
```diff
@@ -0,0 +1,2 @@
+[flake8]
+max-line-length = 100
```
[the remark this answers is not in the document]
dhood: edit: later somewhere it's changed to 99. So my question is just generally if you could clarify why you chose to add a custom one instead of using the default ament_flake8 one?

dejanpan:
@sumanth-nirmal

launch/launch/utilities/ensure_argument_type_impl.py
```diff
@@ -0,0 +1,79 @@
...
+    caller: Optional[Text] = None,
+) -> None:
+    """
+    Ensure that the given argument is an instance of or subclass of one of the given types.
+
+    If the argument does not match one of the types, a TypeError is raised.
+    The caller is included in the error message if given and not None.
+    """
+    error_msg_template = "{}xpected '{}' to be one of [{}], but got '{}' of type '{}'"
+    if not isinstance(types, collections.Iterable) and not isinstance(types, type):
```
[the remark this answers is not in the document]
wjwwood: I think it's right, though it could be rewritten as `ininstance(types, type(type))`. Basically `types` either needs to be an iterable (of types) or a single type. I think it's the same signature as `isinstance()`, e.g. both `isinstance(1, int)` and `isinstance(1, (int, float))` are valid.

launch/launch/utilities/class_tools_impl.py
```diff
@@ -0,0 +1,38 @@
...
+def is_a(obj, entity_type):
+    """Return True if obj is an instance of the entity_type class."""
+    if not isclassinstance(obj):
+        raise RuntimeError("obj '{}' is not a class instance".format(obj))
+    if not inspect.isclass(entity_type):
+        raise RuntimeError("entity_type '{}' is not a class".format(obj))
+    return isinstance(obj, entity_type)
+
+
+def is_a_subclass(obj, entity_type):
```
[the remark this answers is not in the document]
wjwwood: Only if `obj` is an instance, but if it's a "class" then it won't work, e.g. `isinstance(1, int)` is ok, but `isinstance(int, int)` is not.

launch/launch/utilities/normalize_to_list_of_substitutions_impl.py
```diff
@@ -0,0 +1,42 @@
...
+from .class_tools_impl import is_a_subclass
+from ..some_substitutions_type import SomeSubstitutionsType
+from ..substitution import Substitution
+from ..substitutions import TextSubstitution
+
+
+def normalize_to_list_of_substitutions(subs: SomeSubstitutionsType) -> List[Substitution]:
+    """Return a list of Substitutions given a variety of starting inputs."""
+    def normalize(x):
+        if isinstance(x, Substitution):
```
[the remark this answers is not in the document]
wjwwood: In this case, I know `x` is not a class type, but actually an instance.

launch/launch/utilities/ensure_argument_type_impl.py
```diff
@@ -0,0 +1,79 @@
...
+            'caller',
+            'str, None',
+            caller,
+            type(caller),
+        ))
+
+    def check_argument(argument, type_var) -> bool:
+        result = False
+        result |= isinstance(argument, type_var)
+        if hasattr(argument, '__class__') and inspect.isclass(type_var):
```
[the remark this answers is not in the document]
wjwwood: Not if `argument` is a class and not an instance.

launch/launch/events/process/signal_process.py
wjwwood: TIL, that's way better, I can refactor it to use `signal.Signals` instead.

launch/launch/utilities/ensure_argument_type_impl.py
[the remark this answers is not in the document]
dhood: I didn't realise `type` could be this way because I usually use it as if it were function. Pardon the noise! We might want to check though that in the iterable case it's actually an iterable of types, but that would just be to get a nicer error message.

launch/launch/actions/log_info.py
```diff
@@ -0,0 +1,59 @@
...
+        super().__init__()
+
+        self.__msg = normalize_to_list_of_substitutions(msg)
+
+    @property
+    def msg(self) -> List[Substitution]:
+        """Getter for self.__msg."""
+        return self.__msg
+
+    def execute(self, context: LaunchContext) -> Optional[List[Action]]:
```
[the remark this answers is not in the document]
wjwwood: Looks like I can, I change this one. There are probably others in there too. I'd like to add `mypy` as a linter test, but I think doing that and fixing up all the type annotations can be done after the release.

In case it wasn't clear, the type annotations are not used at runtime.

launch/launch/utilities/ensure_argument_type_impl.py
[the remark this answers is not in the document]
wjwwood: It's fine, not like it's completely obvious :)

If you're ok with it, I'll make an issue to improve this check, because as you said, it's only a nicer error message.

launch/launch/utilities/normalize_to_list_of_substitutions_impl.py
```diff
@@ -0,0 +1,42 @@
...
+            return x
+        if isinstance(x, str):
+            return TextSubstitution(text=x)
+        raise TypeError(
+            "Failed to normalize given item of type '{}', when only "
+            "'typing.Text' or 'launch.Substitution' were expected.".format(type(x)))
```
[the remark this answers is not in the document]
wjwwood: That's true, I'll fix that.

launch/launch/actions/log_info.py
```diff
@@ -0,0 +1,59 @@
...
+        """Execute the action."""
+        _logger.info(''.join([context.perform_substitution(sub) for sub in self.msg]))
```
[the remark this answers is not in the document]
dhood: OK, the general pattern that I've come to understand (which might be useful for anyone else reading this) is that a "list of substitutions", here (for `msg`) and elsewhere (e.g. `name`/`value` [for `SetLaunchConfiguration`](https://github.com/ros2/launch/pull/74/files/165e928a4021689017683033d75bf16bbb43a3ad#diff-398c382854c9d576e95df0e25723caf4R54)), represents something like a formatted string.

launch/launch/event_handlers/event_named.py
```diff
@@ -0,0 +1,25 @@
...
+
+"""Module for event_named() function."""
+
+from typing import Callable
+from typing import Text
+
+from ..event import Event
+
+
+def event_named(name: Text) -> Callable[[Event], bool]:
```
dhood: this function doesn't seem to be used anywhere. not necessarily an issue, but wanted to point it out in case it should be!

launch/launch/event_handlers/on_process_exit.py
```diff
@@ -0,0 +1,105 @@
...
+        *,
+        target_action: 'ExecuteProcess' = None,
+        on_exit: Callable[[int], Optional[SomeActionsType]]
+    ) -> None:
+        """Overload which takes a callable to handle the exit."""
+        ...
+
+    def __init__(self, *, target_action=None, on_exit) -> None:  # noqa: F811
+        """Constructor."""
+        from ..actions import ExecuteProcess  # noqa
```
dhood: if you happen to remember why the noqa is needed here I'd be curious (it's used in the next line), but if not, don't worry about it

launch/launch/event_handlers/on_process_exit.py
```diff
@@ -0,0 +1,105 @@
...
+                    isinstance(event, ProcessExited) and (
+                        target_action is None or
+                        event.action == target_action
+                    )
+                )
+            ),
+            entities=None,
+        )
+        self.__target_action = target_action
+        # TODO(wjwwood) check that it is not only callable, but also a callable that matches
```
dhood: it'd be good to at least have the minimum type checking here, otherwise it could be a while before a developer finds out their `on_exit` was the wrong type, but maybe it's more complicated than I realise.

launch/launch/event_handlers/on_process_exit.py
```diff
@@ -0,0 +1,105 @@
...
+        # the correct signature for a handler in this case
+        self.__on_exit = on_exit
+        self.__actions_on_exit: List[LaunchDescriptionEntity] = []
+        if isinstance(on_exit, LaunchDescriptionEntity):
+            self.__on_exit = lambda event, context: on_exit
+            if isinstance(on_exit, collections.Iterable):
```
dhood: I might have missed something, but is it actually possible for `on_exit` to be both an `LaunchDescriptionEntity` and iterable? I wasn't getting that impression.

launch/launch/event_handlers/on_process_exit.py
```diff
@@ -0,0 +1,105 @@
...
+                    )
+                )
+            ),
+            entities=None,
+        )
+        self.__target_action = target_action
+        # TODO(wjwwood) check that it is not only callable, but also a callable that matches
+        # the correct signature for a handler in this case
+        self.__on_exit = on_exit
+        self.__actions_on_exit: List[LaunchDescriptionEntity] = []
```
dhood: I am surprised that the `__actions_on_exit` is only used in `describe`... I am getting the feeling that it should be used in the `handle` function as well (but tbh I feel like it maybe can't actually be a list at all). This comment is sort of a "please re-look at this" request, but there might not be anything wrong.

launch/launch/actions/timer_action.py
```diff
@@ -0,0 +1,157 @@
...
+                entities=OpaqueFunction(
+                    function=lambda context: (
+                        cast(TimerEvent, context.locals.event).timer_action.handle(context)
+                    )
+                ),
+            ))
+            _event_handler_has_been_installed = True
+
+        # Capture the current context locals so the yielded actions can make use of them too.
+        self.__context_locals = dict(context.get_locals_as_dict())
```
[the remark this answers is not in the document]
wjwwood: My intention was to force a copy.

launch/launch/event_handlers/event_named.py
wjwwood: I used it before, but swapped in all cases to using `is_a_subclass` since this is Python, but I believe (hope) it will be useful when transitioning to a markup language where you don't have handles to actions or events, and need to reference them by name.

launch/launch/event_handlers/on_process_exit.py
wjwwood: Not sure, maybe a copy-paste that can be removed (it is required when being used with quoted type annotations only).

launch/launch/event_handlers/on_process_exit.py
wjwwood: It should fail pretty obviously, something like `self.__target_action called with 2 arguments when 3 were expected`, or something near that. Python's default error is pretty decent. Adding the type checking should be possible, but I don't know how off the top of my head because I need to introspect the arguments to the callable. I'd pass on it for now, unless you think it's really important.

launch/launch/event_handlers/on_process_exit.py
wjwwood: No, this looks like a bug, I will investigate.

launch/launch/event_handlers/on_process_exit.py
wjwwood: Not sure, I'll look at it again.

launch/launch/event_handlers/on_process_exit.py
wjwwood: The reason is that flake8 marks it as a redefinition of `ExecuteProcess` even though above it's imported under `if False`.

launch/launch/event_handlers/on_process_exit.py
wjwwood: I think it's addressed in 47e498b.

launch/launch/event_handlers/on_process_exit.py
wjwwood: I think it's addressed in 47e498b.

launch/launch/events/process/process_io.py
```diff
@@ -0,0 +1,59 @@
...
+
+        Unmatched keyword arguments are passed to RunningProcessEvent, see it
+        for details on those arguments.
+
+        :param: text is the unicode data associated with the event
+        :param: fd is an integer that indicates which file descriptor the text is from
+        """
+        super().__init__(**kwargs)
+        self.__text = text
+        self.__from_stdin = fd == 0
```
dhood: is `sys.stdin.fileno()` appropriate in place of just `0` here and e.g. https://github.com/ros2/launch/pull/74/files/165e928a4021689017683033d75bf16bbb43a3ad#diff-76180495de0daffea911889f8030444fR34? I know it really just has to match between the two, so it might not make sense.

launch/launch/event_handlers/on_process_exit.py
dhood: yeah I was just imagining that a process might run for days and then you get the error only when we try to call it but it's fine to leave it as a todo

launch/launch/actions/opaque_function.py
```diff
@@ -0,0 +1,72 @@
...
+
+    The signature of the function should be:
+
+    .. code-block:: python
+
+        def function(
+            context: LaunchContext,
+            *args,
+            **kwargs
+        ) -> Optional[List[LaunchDescriptionEntity]]:
```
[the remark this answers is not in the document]
wjwwood: Actually it can be Actions or events. I believe list of entities is more accurate. All entities are visited and actions which are visited are executed. So in this way actions can return non-action entities (there are none really right now other than `LaunchDescription`).

There's no technical issue here, but maybe some inconsistency in the use of the type annotations.

launch/launch/action.py
```diff
@@ -0,0 +1,47 @@
...
+
+    The action describes the intention to do something, but also can be
+    executed given a :class:`launch.LaunchContext` at runtime.
+    """
+
+    def describe(self) -> Text:
+        """Return a description of this Action."""
+        return self.__repr__()
+
+    def visit(self, context: LaunchContext) -> Optional[List[LaunchDescriptionEntity]]:
```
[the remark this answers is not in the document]
wjwwood: Same as I said in https://github.com/ros2/launch/pull/74#discussion_r195225405.

launch/launch/actions/register_event_handler.py
```diff
@@ -0,0 +1,47 @@
...
+        self.__event_handler = event_handler
+
+    @property
+    def event_handler(self) -> EventHandler:
+        """Getter for self.__event_handler."""
+        return self.__event_handler
+
+    def execute(self, context: LaunchContext):
+        """Execute the action."""
+        context.register_event_handler(self.__event_handler)
```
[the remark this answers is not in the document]
wjwwood: Should be fine for class methods to access the variable directly.

launch/launch/events/process/process_io.py
dhood: OK 👍

dhood: OK as is I mean.

launch/launch/launch_service.py
```diff
@@ -0,0 +1,277 @@
...
+            while not run_loop_task.done():
+                try:
+                    self.__loop_from_run_thread.run_until_complete(run_loop_task)
+                except KeyboardInterrupt:
+                    base_msg = 'user interrupted with ctrl-c (SIGINT)'
+                    if not sigint_received:
+                        _logger.warn(base_msg)
+                        self._shutdown(reason='ctrl-c (SIGINT)', due_to_sigint=True)
+                        sigint_received = True
+                    else:
```
dhood: This is the opposite to what we've done in the previous `launch`: https://github.com/ros2/launch/pull/58/files
I think you've mentioned before ROS 1 does a similar thing, escalating based on the number of interrupts received, is that right?

wjwwood: The signal handling of the service needs more work, I also don't handle SIGTERM, which I probably should do in order to avoid zombie processes.

I'll spend some time on this right now, but at some point I'll need to just start making issues for things, otherwise this pr will never get merged in time.

dhood: To me it's fine if this is one of the things that gets ticketed (it will be easier to review as an incremental improvement anyway). My intent is to point out what I see while reviewing (because otherwise we might forget), then we can make the judgement on them individually.

wjwwood: For sure, didn't mean to sound like I was complaining, just saying I might fix it up some now and ticket other stuff for later, even if it ends up in the release as a follow and smaller change.

wjwwood: I pushed some enhancements in 010edcb, one of which is that repeated ctrl-c does not exit early.

launch/launch/substitutions/launch_configuration.py
```diff
@@ -0,0 +1,87 @@
...
+        super().__init__()
+
+        ensure_argument_type(variable_name, str, 'variable_name', 'LaunchConfiguration')
+
+        from ..utilities import normalize_to_list_of_substitutions
+        self.__variable_name = normalize_to_list_of_substitutions(variable_name)
+        if default is None:
+            self.__default = default
+        else:
+            # convert any items in default that are not a Substitution or str to a str
```
dhood: I want to note that it might trip users up that the default gets converted to a string underneath. It might be better to just enforce it's passed in as a string (if it's not a substitution) rather than doing it ourselves, so that a user that passes in a dictionary or something doesn't expect to be able to retrieve a dictionary later.

launch/launch/actions/execute_process.py
```diff
@@ -0,0 +1,473 @@
...
+class ExecuteProcess(Action):
+    """Action that begins executing a process and sets up event handlers for the process."""
+
+    def __init__(
+        self, *,
+        cmd: Iterable[SomeSubstitutionsType],
+        cwd: Optional[SomeSubstitutionsType] = None,
+        env: Optional[Dict[SomeSubstitutionsType, SomeSubstitutionsType]] = None,
+        shell: bool = False,
+        sigterm_timeout: SomeSubstitutionsType = LaunchConfiguration('sigterm_timeout', default=5),
```
dhood: I imagine you are setting the default via a LaunchConfiguration in order to retrieve its value when it's actually used (as opposed to just passing in a float). But is it actually possible that someone could modify it before then, externally? FWIU [it gets evaluated here](https://github.com/ros2/launch/pull/74/files#diff-61622adb9c2241797a64d96e6065c414R273) and we don't modify it internally, so maybe we can just pass it as a float?

dhood: answering myself: it is modifiable externally in the same way [the `launch-prefix` configuration is](https://github.com/ros2/launch/pull/74/files#diff-61622adb9c2241797a64d96e6065c414R163).

wjwwood: > I imagine you are setting the default via a LaunchConfiguration in order to retrieve its value when it's actually used (as opposed to just passing in a float).

Yes.

> But is it actually possible that someone could modify it before then, externally?

Yes, consider:

```python
ld = LaunchDescription([
  SetLaunchConfiguration('sigterm_timeout', 42.0),
  ExecuteProcess(cmd=['ls']),
  ExecuteProcess(
    cmd=['ls'],
    sigterm_timeout=EnvironmentVariable('CUSTOM_SIGTERM_TIMEOUT')),
  ExecuteProcess(cmd=['ls', '-la']),
  # ExecuteProcess(cmd=['ls', '-la'], sigterm_timeout=1.0),
])
```

The first and last execute processes would use the timeout set in the launch configuration, but the second would not because it was explicitly passed and the fourth (commented out) would be a nice to have interface where `sigterm_timeout` could also take a non-text/non-substitution numeric, but it's not necessary.

> FWIU it gets evaluated here and we don't modify it internally, so maybe we can just pass it as a float?

Not exactly, the substitutions that make up `sigterm_timeout` are passed to the `TimerAction` there, but they are not substituted until the `TimeAction` is visited, and the substitutions should not be evaluated until then I think. We couldn't expand the substitution there even if we wanted to, because when called via describe, it would not have a `LaunchContext` which is required to perform substitutions.

launch/launch/substitutions/launch_configuration.py
wjwwood: It's a trade-off, as it's convenient to pass literals or the result of expressions to the `SetLaunchConfiguration` action directly. I can change it.

wjwwood: I dunno, it's pretty convenient. I think I'd prefer to leave it.

If you want, I'll change it anyways, but what if I just improved the documentation a bit to be clear that the default must be a string in the end, so non-text/non-substitution things are converted to text implicitly?

launch/launch/actions/execute_process.py
dhood: yep, it doesn't get evaluated until the TimerAction is visited/executed: https://github.com/ros2/launch/pull/74/files#diff-a40b801740c81b57ba0a59a56cc0d36fR79.

I documented this behaviour in f3417d4

launch/launch/substitutions/launch_configuration.py
dhood: yeah I already clarified a bit in https://github.com/ros2/launch/pull/74/commits/760fe46c6c886f9446e7c84bd726258d1081be83#diff-55c61397a0c771113a1e78e3143e31fdR77. I can see the convenience, so let's leave it; maybe one day we will store dicts as dicts even.

launch/doc/source/architecture.rst
```diff
@@ -0,0 +1,208 @@
...
+The latter can be used to create "syntactic sugar" actions which simply yield more verbose actions.
+
+Actions may also have arguments, which can affect the behavior of the actions.
+These arguments are where :class:`launch.Substitution`'s can be used to provide more flexibility when describing reusable launch descriptions.
+
+Basic Actions
+^^^^^^^^^^^^^
+
+`launch` provides the foundational actions on which other more sophisticated actions may be built.
+This is a non-exhaustive list of actions that `launch` may provide:
```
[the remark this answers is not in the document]
dhood: I also asked myself this question.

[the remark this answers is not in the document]
wjwwood: Simply because I didn't want to have to come back here and update the list every time I added a trivial action to the list in `launch`. These are the important ones (as I see it).

launch/.flake8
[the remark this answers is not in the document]
dhood: I configured my editor to use the ament_flake8 config file as the default. I figured this was the expected workflow, so we can configure it in one place, as opposed to adding one to each repo (unless it needs to be customised for some reason). @dirk-thomas might have an opinion but there's no need to block the PR on it

launch/doc/source/architecture.rst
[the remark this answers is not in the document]
wjwwood: Simply because I didn't want to have to come back here and update the list every time I added a trivial action to the list in `launch`. These are the important ones (as I see it). This is just supposed to be an overview, not the complete API docs.

launch/.flake8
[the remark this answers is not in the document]
wjwwood: That's fine, I can remove it. I just found it useful, but I agree that it has the potential to get out of sync with `ament_flake8`'s config.

wjwwood:
Thanks for the review @dhood! Here's some CI (with ` --packages-up-to launch`):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4640/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1537/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3835/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4716/)

wjwwood:
@dhood sorry, I should have gotten you to review the new changes before merging, CI was good with them, but if you have any time to review it postmortem that'd be great:

- a9d82f9
- 54bc19e

dhood:
OK, looks good, I just have some small comments. I'm happy to open a followup PR if you agree with them.

It's fine that [there's no `else` of this check](https://github.com/ros2/launch/commit/a9d82f9a481cf6e3d389793a4d0ba043ab3abe38#diff-7133b2326a036fbf64a7548bfcc2c589R117) for the original signal handler, because it came from `getsignal`. But for custom signal handlers passed in, we're just silently ignoring signal handlers that (by mistake) are not callable. I'd add a check to the setters.

[this message](https://github.com/ros2/launch/commit/a9d82f9a481cf6e3d389793a4d0ba043ab3abe38#diff-7133b2326a036fbf64a7548bfcc2c589R146) should instead say the function needs to be _called_ from the main thread, right? Did it used to get called on module import?

[the use of "exception" in this text](https://github.com/ros2/launch/commit/a9d82f9a481cf6e3d389793a4d0ba043ab3abe38#diff-7133b2326a036fbf64a7548bfcc2c589R98) makes it sound like their handler won't be called if it raises a keyboard interrupt, where it actually will still be called.

wjwwood:
> It's fine that there's no else of this check for the original signal handler, because it came from getsignal. But for custom signal handlers passed in, we're just silently ignoring signal handlers that (by mistake) are not callable. I'd add a check to the setters.

There's no else because I believe `signal.getsignal` will return None if no handler was set. So it will be hard to know if the existing handler was intentionally not callable or in error. I'd say that's the responsibility of `signal.signal` to check.

> this message should instead say the function needs to be called from the main thread, right? Did it used to get called on module import?

That's right, it started out as something I was doing on import and decided to change it to a function.

> the use of "exception" in this text makes it sound like their handler won't be called if it raises a keyboard interrupt, where it actually will still be called.

Well there's not way of knowing if their handler will raise `KeyboardInterrupt` without running it, so I don't know how we'd do that. I didn't feel it was that confusing, but if you suggest a specific edit, I'll include it in a follow up pr.

dhood:
> There's no else because I believe signal.getsignal will return None if no handler was set. So it will be hard to know if the existing handler was intentionally not callable or in error. I'd say that's the responsibility of signal.signal to check.

Please re-read: I am referring to custom signal handlers.

> There's no else because I believe signal.getsignal will return None if no handler was set. So it will be hard to know if the existing handler was intentionally not callable or in error. I'd say that's the responsibility of signal.signal to check.

Just removing the first part of the sentence is enough to clarify.

again, I'm happy to do it, but there wasn't an open PR. now I'll build on 77

wjwwood:
Sorry, I thought you meant original handler since that's the line you linked to (117 is `if callable(__original_sigint_handler):`).

I'd check it in the `on_sigint` function before setting it then. Rather than checking it in the signal handler.

wjwwood:
But you're right, your text makes total sense. I didn't read it closely enough.

dhood:
yeah that was what I mentioned. I will do that; wasn't trying to create more work for you, just to get approval of the direction/intent :)

PR #75, opened 2018-06-12
ROS specific launch

wjwwood:
Requires #74

Things to have a look at:

- Simple pub/sub example:
  - https://github.com/ros2/launch/pull/75/files#diff-65d815fff2f1a19178fc422135bb3b66
- Lifecycle pub/sub example:
  - https://github.com/ros2/launch/pull/75/files#diff-2086d091e24f813cafb9e343fb009c64


dejanpan:
@sumanth-nirmal

sumanth-nirmal:
> Lifecycle pub/sub example:

Copy/Paste mistake, correct link is https://github.com/ros2/launch/pull/75/files#diff-2086d091e24f813cafb9e343fb009c64

wjwwood:
Rebased to one commit after https://github.com/ros2/launch/pull/74 was merged. I'll start prepping this for review now.

launch_ros/launch_ros/actions/node.py
```diff
@@ -0,0 +1,184 @@
...
+class Node(ExecuteProcess):
+    """Action that begins executing a process and sets up event handlers for the process."""
+
+    def __init__(
+        self, *,
+        package: SomeSubstitutionsType,
+        node_executable: SomeSubstitutionsType,
+        node_name: Optional[SomeSubstitutionsType] = None,
+        node_namespace: Optional[SomeSubstitutionsType] = None,
+        remappings: Optional[Dict[SomeSubstitutionsType, SomeSubstitutionsType]] = None,
```
dhood: Aren't there cases where the specific order of the remapping rules has to be preserved, because applying them in different orders causes different results? If so, another type would be more appropriate e.g. OrderedDict or list of tuples.

launch_ros/examples/lifecycle_pub_sub_launch.py
```diff
@@ -0,0 +1,94 @@
...
+            launch.actions.LogInfo(
+                msg="lifecycle node 'talker' reached the 'unconfigured' state, 'activating'."),
+            launch.actions.EmitEvent(event=launch_ros.events.lifecycle.ChangeState(
+                lifecycle_node_matcher=launch.events.process.matches_action(talker_node),
+                transition_id=lifecycle_msgs.msg.Transition.TRANSITION_ACTIVATE,
+            )),
+        ],
+    )))
+
+    # Make the talker node take the 'configure' transition.
```
dhood: for readability of this file I recommend reversing the creation of these actions so we read about them in the order that they would occur

launch_ros/launch_ros/actions/node.py
```diff
@@ -0,0 +1,184 @@
...
+
+from launch_ros.substitutions import ExecutableInPackage
+
+from rclpy.validate_namespace import validate_namespace
+from rclpy.validate_node_name import validate_node_name
+
+_logger = logging.getLogger(name='launch_ros')
+
+
+class Node(ExecuteProcess):
```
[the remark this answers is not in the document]
wjwwood: The reason I departed from `Execute` in these is because of how I planed to reuse them as arguments to the "node container", basically something like this:

```
LaunchDescription([
  ComponentNodeContainer(name='whatever', nodes=[
     Node(...),
     Node(...), 
  ]),
  Node(...),
])
```

Where the first two `Node` would never be visited, and instead would be introspected by the `ComponentNodeContainer` to figure out how to load them dynamically, and the last `Node` would actually get visited and therefore executed.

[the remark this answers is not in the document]
wjwwood: Maybe there's a better way to organize the names, with that in mind. I don't know.

I also thought about an `Execute` action which would look like:

```python
LaunchDescription([
  ComponentNodeContainer(name='whatever', nodes=[
     Node(...),
     Node(...), 
  ]),
  Execute(executable=Node(...), cwd='something'),
])
```

In which case `Node` would no longer inherit from `ExecuteProcess`.

launch_ros/launch_ros/default_launch_description.py
```diff
@@ -0,0 +1,143 @@
...
+        except KeyboardInterrupt:
+            pass
+        finally:
+            executor.remove_node(self.__rclpy_node)
+
+    def _function(self, context: launch.LaunchContext):
+        try:
+            rclpy.init(args=context.argv)
+        except RuntimeError as exc:
+            if 'rcl_init called while already initialized' in str(exc):
```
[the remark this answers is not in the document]
wjwwood: Already made one 😁: https://github.com/ros2/rclpy/issues/190

launch_ros/launch_ros/default_launch_description.py
```diff
@@ -0,0 +1,143 @@
...
+                pass
+            raise
+        self.__rclpy_node = rclpy.create_node('launch_ros')
+        context.extend_globals({
+            'ros_startup_action': self,
+            'rclpy_node': self.__rclpy_node
```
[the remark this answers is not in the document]
wjwwood: Sounds good 431853f

launch_ros/launch_ros/actions/node.py
wjwwood: That's true, let me look into that.

launch_ros/examples/lifecycle_pub_sub_launch.py
wjwwood: The event handler must be registered before sending the event. Do you have a suggestion for how to reorder it that works?

dhood: I am not sure if we are talking about the same scale. What I meant was add the action that creates the node; add the action that configures the node; add the action that activates the node; add the action that starts the listener. But I think it wasn't clear so you thought I was referring to this individual add_action call.

launch_ros/launch_ros/actions/node.py
wjwwood: So, I can't use OrderedDict as a type annotation with subscripted types for key and value, unfortunately. I don't like using a list of tuples because of how it looks/feels in the examples, but I don't see another way, and creating an ordered dict also uses a list of tuples. So I guess I'll do that, but the demos/examples need to be updated.

launch_ros/launch_ros/actions/node.py
```diff
@@ -0,0 +1,187 @@
...
+            if self.__node_name is not None:
+                self.__expanded_node_name = perform_substitutions(
+                    context, normalize_to_list_of_substitutions(self.__node_name))
+                validate_node_name(self.__node_name)
+            self.__expanded_node_name.lstrip('/')
+            if self.__node_namespace is not None:
+                self.__expanded_node_namespace = perform_substitutions(
+                    context, normalize_to_list_of_substitutions(self.__node_namespace))
+            if not self.__expanded_node_namespace.startswith('/'):
+                self.__expanded_node_namespace = '/' + self.__expanded_node_namespace
```
dhood: what is the point of adding a leading forward slash if the next line does `lstrip('/')`?

launch_ros/launch_ros/actions/node.py
wjwwood: 313bd70

launch_ros/package.xml
```diff
@@ -0,0 +1,20 @@
...
+  href="http://download.ros.org/schema/package_format2.xsd"
+  schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>launch_ros</name>
+  <version>0.4.0</version>
+  <description>ROS specific extensions to the launch tool.</description>
+  <maintainer email="william@osrfoundation.org">William Woodall</maintainer>
+  <license>Apache License 2.0</license>
+
+  <depend>launch</depend>
```
dhood: this also depends on rclpy, lifecycle_msgs, ament_index_python and osrf_pycommon

launch_ros/examples/lifecycle_pub_sub_launch.py
wjwwood: Checkout 46bd3e1

launch_ros/launch_ros/actions/node.py
wjwwood: No idea, 03e9d07

launch_ros/package.xml
wjwwood: Yes, but I don't think `osrf_pycommon` directly. cd0ebe2

launch_ros/launch_ros/default_launch_description.py
```diff
@@ -0,0 +1,143 @@
...
+        self.__rclpy_spin_thread = threading.Thread(target=self._run)
+        self.__rclpy_spin_thread.start()
+
+
+def get_default_launch_description(*, prefix_output_with_name=False):
+    """
+    Return a LaunchDescription to be included before user descriptions.
+
+    :param: prefix_output_with_name if True, each line of output is prefixed
+        with the name of the node as `[name] `, else it is printed unmodified
```
dhood: I think this should be more generic than referencing nodes since it will prefix all processes' output right?

launch_ros/launch_ros/default_launch_description.py
```diff
@@ -0,0 +1,143 @@
...
+    def _function(self, context: launch.LaunchContext):
+        try:
+            rclpy.init(args=context.argv)
+        except RuntimeError as exc:
+            if 'rcl_init called while already initialized' in str(exc):
+                pass
+            raise
+        self.__launch_ros_node = rclpy.create_node('launch_ros')
+        context.extend_globals({
+            'ros_startup_action': self,
```
dhood: what's the intended use of this?

launch_ros/launch_ros/default_launch_description.py
wjwwood: Correct b4b83da

launch_ros/launch_ros/default_launch_description.py
wjwwood: Future use, nothing for now.

launch_ros/launch_ros/events/lifecycle/lifecycle_node_matchers.py
```diff
@@ -0,0 +1,27 @@
...
+
+from typing import Callable
+from typing import Text
+
+if False:
+    # imports here would cause loops, but are only used as forward-references for type-checking
+    from ...actions import LifecycleNode  # noqa
+
+
+def matches_node_name(node_name: Text) -> Callable[['LifecycleNode'], bool]:
```
dhood: it seems like this isn't lifecycle specific (maybe it was at some point but it doesn't look like it's used anymore), and the module could be moved/renamed (maybe in followup)

wjwwood: Well, not lifecycle specific, but there are no events that target `Node` right now (there are some that target `ExecuteProcess` however), so I put it with the lifecycle events which it is used with. We may need to move it later.

launch_ros/launch_ros/actions/lifecycle_node.py
```diff
@@ -0,0 +1,123 @@
...
+        - :class:`launch.events.lifecycle.ChangeState`
+
+          - this event can be targeted to a single lifecycle node, or more than
+            one, or even all lifecycle nodes, and it causes the targeted nodes
+            to change state if possible, see its documentation for more details.
+        """
+        super().__init__(node_name=node_name, **kwargs)
+        self.__rclpy_subscription = None
+        # first item is unknown state
+        self.__current_state = next(iter(ChangeState.valid_states.values()))
```
dhood: seems like it'd be more appropriate to explicitly set `self.__current_state = ChangeState.valid_states[lifecycle_msgs.msg.State.PRIMARY_STATE_UNKNOWN]` (because we want to do this regardless of dict ordering)

launch_ros/launch_ros/actions/lifecycle_node.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def _call_change_state(self, request, context: launch.LaunchContext):
+        while not self.__rclpy_change_state_client.wait_for_service(timeout_sec=1.0):
+            if context.is_shutdown:
+                _logger.warn("Abandoning wait for the '{}' service.".format(
+                    self.__rclpy_change_state_client.srv_name))
+                return
+        response = self.__rclpy_change_state_client.call(request)
+        if not response.success:
+            _logger.warn("Failed to make transition '{}' for LifecycleNode '{}'".format(
```
dhood: maybe this should be an error instead of just a warning.

launch_ros/package.xml
wjwwood: You were right, I was wrong, here's the fixup: 15ac0c3

launch_ros/examples/lifecycle_pub_sub_launch.py
```diff
@@ -0,0 +1,106 @@
...
+import launch_ros.actions
+import launch_ros.events
+import launch_ros.events.lifecycle
+
+import lifecycle_msgs.msg
+
+
+def main(argv=sys.argv[1:]):
+    """Main."""
+    def _on_transition_event(context):
```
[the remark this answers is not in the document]
wjwwood: Right 838242e

launch_ros/launch_ros/actions/lifecycle_node.py
wjwwood: 65c2b53

launch_ros/launch_ros/actions/lifecycle_node.py
wjwwood: Fair enough, I think I was imagining that the "first" state might change and this would be more future proof, but that's silly (it was probably late): 6955bc8

launch_ros/launch_ros/actions/node.py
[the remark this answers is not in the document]
dhood: Even if `Node` doesn't inherit from ExecuteProcess we'd still probably benefit from distinguishing it from something that is actually a node (it is not, in any of these cases). For now, it's namespaced enough that I suppose we can lean on that.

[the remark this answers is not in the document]
wjwwood: That's part of the reason I use `launch.x.Y` in most of the demos, to reinforce those differences.

I also somewhat inconsistently use `Launch<thing>` and `<thing>`, e.g. `LaunchDescription` versus `Action` (and not `LaunchAction`). I don't have a good reason for the inconsistency, other than I felt that in context `launch.Action` was specific enough, but `launch.Description` was not (mostly due to where and how it's used in the examples.

It might be something we want to revisit in the near future.

Do you want me to change anything for this pr related to this conversation?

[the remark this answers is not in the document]
dhood: I think for this PR we're good, and then there's a chance we'll rename things in the future.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4664/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1552/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3856/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4737/)

wjwwood:
:shipit:

wjwwood:
Retroactive summary for this feature:

add ROS specific features to launch system
==========================================

### Purpose / Use Cases (Why implement this feature?)

This was implemented to provide ROS specific actions, substitutions, events, etc... to the launch system and to add default actions which allow monitoring of child nodes with lifecycle.

### Design (How does it work?)

It provides new actions (like `Node` or `LifecycleNode`) and other related new entities like substitutions and events, which a user my use in their launch description. These new items are often just syntactic sugar, building entirely on existing features in launch. This is by far not a complete set of new features for ROS, but just the first iteration.

It also adds some "default" ROS related actions that are (will be) included in the `ros2 launch` command. These default actions do things like create a `rclpy` node for launch, which is used to monitor the state of child nodes which have lifecycle. It uses this information to generate events in `launch` when these child nodes change state.

#### Inputs / Outputs / API (How to use it?)  

The API is anything in the `launch_ros` package:

https://github.com/ros2/launch/tree/baefc077b062ba5cfaff8a4dbecda7aaaeb4a05a/launch_ros/launch_ros

launch_ros/launch_ros/actions/node.py
```diff
@@ -0,0 +1,187 @@
...
+
+"""Module for the Node action."""
+
+import logging
+from typing import Dict
+from typing import Iterable
+from typing import List
+from typing import Optional
+from typing import Text
+from typing import Tuple
```
dirk-thomas: This line seems to result in a flake8 warning in the dev job: http://build.ros2.org/job/Ddev__launch_ros__ubuntu_bionic_amd64/21/consoleFull#console-section-17

@wjwwood Can you please double check this.

wjwwood: This pr is ancient, if it didn't flake then, then why do I need to check this?

dirk-thomas: The test fails in current `dev` builds but doesn't on ci.ros2.org. With double checki was referring to try to make it pass for both.

wjwwood: That sounds like a CI issue, if they don't agree, but either way, wouldn't make more sense to open a new issue against this repository saying "this doesn't pass linters in the dev job", and then we can triage it? I don't think going back to the last pr to touch that file makes sense, especially if it's this old...

wjwwood: Also, this code isn't in this repository any longer?

wjwwood: (I was trying to look into it)

dirk-thomas: The `dev` job uses the Debian package of `flake8` where as ci.ros.org uses the latest release from PyPI.

wjwwood: It's probably because what ever version of the linter that's on the dev job doesn't understand the Python 3.5 version of type annotations:

https://github.com/ros2/launch_ros/blob/2d5d3eac831788d54c41069378733837b48b44c8/launch_ros/launch_ros/actions/node.py#L175

dirk-thomas: Resolved by ros2/launch_ros#32.

Issue #78, opened 2018-06-14
Typing annotations of variables used in launch only available for python 3.6+


dhood:
http://mypy.readthedocs.io/en/latest/cheat_sheet_py3.html#variables

causes failures on xenial jobs, e.g.:

```
    from .launch_context import LaunchContext
E     File "/home/rosbuild/ci_scripts/ws/install/launch/lib/python3.5/site-packages/launch/launch_context.py", line 45
E       self._event_queue: asyncio.Queue = asyncio.Queue()
E                        ^
E   SyntaxError: invalid syntax
```

dhood:
@wjwwood putting this here for tracking/visibility, not trying to dump it and leave it. we can group-schedule it when it reaches priority.

PR #81, opened 2018-06-17
Use variable typing in comments to support python 3.5

dhood:
Fixes https://github.com/ros2/launch/issues/78

This fixes the build on xenial.

@wjwwood there are a few lines that are too long (https://ci.ros2.org/job/ci_linux/4703/testReport/) but I haven't got mypy setup to check what wrapping will be supported. What invocation do you use? I get `Can't find package 'launch'`  when calling `python3 -m mypy --package launch`...

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -21,7 +21,7 @@ import shlex
 import signal
 import threading
 import traceback
-from typing import Any
+from typing import Any  # noqa: F401
 from typing import cast
 from typing import Dict
 from typing import Iterable
@@ -52,7 +52,7 @@ from ..launch_context import LaunchContext
 from ..launch_description import LaunchDescription
 from ..some_actions_type import SomeActionsType
 from ..some_substitutions_type import SomeSubstitutionsType
-from ..substitution import Substitution
+from ..substitution import Substitution  # noqa: F401
 from ..substitutions import LaunchConfiguration
 from ..substitutions import PythonExpression
 from ..utilities import create_future
@@ -156,7 +156,7 @@ class ExecuteProcess(Action):
         super().__init__()
         self.__cmd = [normalize_to_list_of_substitutions(x) for x in cmd]
         self.__cwd = cwd if cwd is None else normalize_to_list_of_substitutions(cwd)
-        self.__env: Optional[Dict[List[Substitution], List[Substitution]]] = None
+        self.__env = None  # type: Optional[Dict[List[Substitution], List[Substitution]]]
         if env is not None:
             self.__env = {}
             for key, value in env.items():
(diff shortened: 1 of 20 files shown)


wjwwood:
Not sure let me check my integration into my editor.

wjwwood:
I'm using `--ignore-missing-imports`.

launch/launch/actions/execute_process.py
```diff
@@ -71,15 +71,15 @@ class ExecuteProcess(Action):
...
         self, *,
-        cmd: Iterable[SomeSubstitutionsType],
-        cwd: Optional[SomeSubstitutionsType] = None,
-        env: Optional[Dict[SomeSubstitutionsType, SomeSubstitutionsType]] = None,
-        shell: bool = False,
-        sigterm_timeout: SomeSubstitutionsType = LaunchConfiguration('sigterm_timeout', default=5),
-        sigkill_timeout: SomeSubstitutionsType = LaunchConfiguration('sigkill_timeout', default=5),
-        prefix: Optional[SomeSubstitutionsType] = None,
-        output: Optional[Text] = None,
-        log_cmd: bool = False,
```
wjwwood: I thought these kinds worked in Python3.5?

dhood: oh, maybe I went overkill. I'll check this afternoon.

dhood:
OK @wjwwood you were right and I had more changes than necessary before. It's been reduced to the minimal diff, which does include some code changes, if you could check them thanks (the changes are in individual commits).

launch/launch/launch_introspector.py
```diff
@@ -111,24 +114,26 @@ def format_event_handler(event_handler: EventHandler) -> List[Text]:
...
-                 for k, v in action.env.items()] + '}'),
-            action.shell,
+                 for k, v in typed_action.env.items()] + '}'),
+            typed_action.shell,
         )
         return [msg]
     elif is_a(action, RegisterEventHandler):
-        result = ["RegisterEventHandler('{}'):".format(action.event_handler)]
-        result.extend(indent(format_event_handler(action.event_handler)))
+        typed_action2 = cast(RegisterEventHandler, action)
```
dhood: mypy wasn't happy with reuse of the same variable name as above (even though they're in different branches)...

launch/launch/launch_introspector.py
```diff
@@ -100,7 +103,7 @@ def format_event_handler(event_handler: EventHandler) -> List[Text]:
     """Return a text representation of an event handler."""
     if hasattr(event_handler, 'describe'):
         # TODO(wjwwood): consider supporting mode complex descriptions of branching
-        description, entities = event_handler.describe()
+        description, entities = event_handler.describe()  # type: ignore
```
dhood: mypy was complaining about that `EventHandler` doesn't have a `describe` method: I think that ignoring it is the appropriate thing to do here.

dhood:
xenial [Build Status](https://ci.ros2.org/view/nightly/job/ci_linux/4716/)
bionic [Build Status](https://ci.ros2.org/job/ci_linux/4717/)

launch/launch/actions/timer_action.py
```diff
@@ -55,7 +55,7 @@ class TimerAction(Action):
     def __init__(
         self, *,
         period: Union[float, SomeSubstitutionsType],
-        actions: Iterable[LaunchDescriptionEntity],
+        actions: Iterable[LaunchDescriptionEntity]
```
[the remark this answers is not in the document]
dhood: Yeah, I know that we do it sometimes intentionally (thanks for the explanation though :) ) but it was causing syntax errors with coverage:
```
launch/actions/execute_process.py:37: in <module>
    from .timer_action import TimerAction
E     File "/home/dhood/ros2_ws/src/ros2/launch/launch/launch/actions/timer_action.py", line 59
E       ) -> None:
E       ^
E   SyntaxError: invalid syntax
```

(presumably something unique to `__init__`s)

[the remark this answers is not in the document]
dhood: this looks like the python ticket for the curious! (fixed in 3.6) https://bugs.python.org/issue9232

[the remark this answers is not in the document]
wjwwood: Hmm, weird. :)

PR #82, opened 2018-06-19
move suppression of KeyboardInterrupt

wjwwood:
Small follow up improvement that @dhood and I identified while reviewing. Should not affect behavior at all, but I'd appreciate someone trying it out locally (before and after).

Doesn't need to be in for Bouncy.

### DIFF
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -275,6 +275,8 @@ class LaunchService:
             while not run_loop_task.done():
                 try:
                     self.__loop_from_run_thread.run_until_complete(run_loop_task)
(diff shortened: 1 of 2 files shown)


wjwwood:
CI (`--packages-up-to launch`/`--packages-select launch`):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4743/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1611/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3931/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4803/)

PR #85, opened 2018-06-20
keep track of, and then close at exit, all asyncio loops

### LINKED ISSUE #84: Asyncio error on Xenial with (new) launch
```
$ ros2 launch demo_nodes_cpp add_two_ints.launch.py 
[INFO] [launch]: process[add_two_ints_server-1]: started with pid [17647]
[INFO] [launch]: process[add_two_ints_client-2]: started with pid [17648]
[INFO] [add_two_ints_server]: Incoming request
a: 2 b: 3
[INFO] [add_two_ints_client]: Result of add_two_ints: 5
[INFO] [launch]: process[add_two_ints_client-2]: process has finished cleanly
[INFO] [launch]: sending signal 'SIGINT' to process[add_two_ints_server-1]
signal_handler(2)
[INFO] [launch]: process[add_two_ints_server-1]: process has finished cleanly
Exception ignored in: <bound method BaseEventLoop.__del__ of <_UnixSelectorEventLoop running=False closed=True debug=False>>
  File "/usr/lib/python3.5/asyncio/base_events.py", line 431, in __del__
  File "/usr/lib/python3.5/asyncio/unix_events.py", line 58, in close
  File "/usr/lib/python3.5/asyncio/unix_events.py", line 139, in remove_signal_handler
  File "/usr/lib/python3.5/signal.py", line 47, in signal
TypeError: signal handler must be signal.SIG_IGN, signal.SIG_DFL, or a callable object
```

Seems like it's from https://bugs.python.org/issue23548 (you may have fixed this before @wjwwood , you are in the thread :smile: )

wjwwood:
Fixes #84

This is necessary only with Python < 3.6 (or at least with the Python 3.5.1 on Xenial). See #84 for more details.

### DIFF
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -248,11 +248,11 @@ class LaunchService:
 
         # Acquire the lock and initialize the asyncio loop.
         with self.__loop_from_run_thread_lock:
-            global _g_loops_used
             self.__loop_from_run_thread = osrf_pycommon.process_utils.get_loop()
-            _g_loops_used.add(self.__loop_from_run_thread)
             if self.__loop_from_run_thread is None:
                 raise RuntimeError('__loop_from_run_thread unexpectedly None')
+            global _g_loops_used
+            _g_loops_used.add(self.__loop_from_run_thread)
             if self.__debug:
                 self.__loop_from_run_thread.set_debug(True)
             self.__context._set_asyncio_loop(self.__loop_from_run_thread)
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -15,6 +15,7 @@
 """Module for the LaunchService class."""
 
 import asyncio
+import atexit
 import collections
 import logging
 import threading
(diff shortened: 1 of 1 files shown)


wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4763/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1628/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3949/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4820/)

launch/launch/launch_service.py
```diff
@@ -234,7 +248,9 @@ def run(self, *, shutdown_when_idle=True) -> None:
 
         # Acquire the lock and initialize the asyncio loop.
         with self.__loop_from_run_thread_lock:
+            global _g_loops_used
             self.__loop_from_run_thread = osrf_pycommon.process_utils.get_loop()
+            _g_loops_used.add(self.__loop_from_run_thread)
```
dhood: maybe better to move this line after the next check for `None` in case someone catches the exception and we end up trying to close a `None` loop. (doubt it would happen but seems more defensive)

launch/launch/launch_service.py
```diff
@@ -43,6 +44,19 @@
...
+# This atexit handler ensures all the loops are closed at exit.
+# This is only really required pre-3.6, see:
+#   https://github.com/ros2/launch/issues/84
+#   https://bugs.python.org/issue23548
+@atexit.register
+def close_loop():
+    global _g_loops_used
+    for loop in _g_loops_used:
+        if not loop.is_closed():
+            loop.close()
```
dirk-thomas: I would be good to double check the behavior on Windows since I have seen indefinite blocking when trying to close the loop.

wjwwood: I'll check it manually, but there is at least one automated test that runs the launch service and then exits (though `pytest` might influence the behavior if it also uses `asyncio`).

wjwwood: I'm not sure if this pr caused it (I don't think so), but I'm investigating an issue with `launch` on Windows anyways, so I'll make sure everything is working there: https://github.com/ros2/launch/issues/86

Issue #91, opened 2018-06-22
Reentrant call to stderr on interrupted Launch on Windows + Opensplice


dhood:
Using [last night's packaging job](https://ci.ros2.org/view/packaging/job/packaging_windows/1043/):

This only happens for rmw_opensplice_cpp:
```
C:\J\workspace\packaging_windows\ws>ros2 launch demo_nodes_cpp talker_listener.launch.py
[INFO] [launch]: process[talker.EXE-1]: started with pid [800]
[INFO] [launch]: process[listener.EXE-2]: started with pid [6560]
[INFO] [talker]: Publishing: 'Hello World: 1'
[INFO] [listener]: I heard: [Hello World: 1]
[INFO] [talker]: Publishing: 'Hello World: 2'
[INFO] [listener]: I heard: [Hello World: 2]
ssignal_handler(2)
[WARNING] [launch.LaunchService]: user interrupted with ctrl-c (SIGINT)
--- Logging error ---
Traceback (most recent call last):
  File "c:\python36\lib\logging\__init__.py", line 995, in emit
    stream.write(self.terminator)
RuntimeError: reentrant call inside <_io.BufferedWriter name='<stderr>'>

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "c:\python36\lib\logging\__init__.py", line 995, in emit
    stream.write(self.terminator)
  File "C:\J\workspace\packaging_windows\ws\install\Lib\site-packages\launch\utilities\signal_management.py", line 121, in __on_sigint
    __custom_sigint_handler(signum, frame)
  File "C:\J\workspace\packaging_windows\ws\install\Lib\site-packages\launch\launch_service.py", line 291, in _on_sigint
    _logger.warn(base_msg)
  File "c:\python36\lib\logging\__init__.py", line 1323, in warn
    self.warning(msg, *args, **kwargs)
  File "c:\python36\lib\logging\__init__.py", line 1318, in warning
    self._log(WARNING, msg, args, **kwargs)
  File "c:\python36\lib\logging\__init__.py", line 1442, in _log
    self.handle(record)
  File "c:\python36\lib\logging\__init__.py", line 1452, in handle
    self.callHandlers(record)
  File "c:\python36\lib\logging\__init__.py", line 1514, in callHandlers
    hdlr.handle(record)
  File "c:\python36\lib\logging\__init__.py", line 863, in handle
    self.emit(record)
  File "c:\python36\lib\logging\__init__.py", line 998, in emit
[...]

PR #92, opened 2018-06-23
Avoid reentrancy in signal handler

dhood:
@wjwwood what do you think of this to address https://github.com/ros2/launch/issues/91? I got the idea from [this blog post](https://utcc.utoronto.ca/~cks/space/blog/python/SIGCHLDSolution).

If it's alright I'll extend it to the other handlers.

### DIFF
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -19,6 +19,7 @@ import atexit
 import collections
 import logging
 import threading
+import signal
 from typing import Iterable
 from typing import List  # noqa: F401
 from typing import Optional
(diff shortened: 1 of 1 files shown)


launch/launch/launch_service.py
```diff
@@ -285,6 +286,12 @@ def exception_handler(loop, context):
 
             # Setup custom signal hanlders for SIGINT, SIGQUIT, and SIGTERM.
             def _on_sigint(signum, frame):
+                # Ignore additional interrupts until we finish processing this one.
+                prev_handler = signal.signal(signal.SIGINT, signal.SIG_IGN)
+                if prev_handler is signal.SIG_IGN:
+                    # This function has been called re-entrantly.
+                    return
```
wjwwood: Is this expected to occur and be ok? If yes and then no, should it raise an exception?

Is this the case where a signal is received between 288 and 289?

dhood: I am anticipating a race condition between 288 and 290, yeah. I haven't read anything to think that it's not possible.

I don't think that raising there is appropriate because the situation that caused #91 will then cause an exception, where I think that if we are already interrupting then there is no cause for concern. We print the message if we receive sigints later on but that's more for the information of the user from my understanding.

dhood:
This was responsible for instabilities in the nightlies last night because of a linter failure; addressed in https://github.com/ros2/launch/commit/ed75487baaaf94d7dade2d4ac1b4ee33df287986

PR #94, opened 2018-06-27
avoid SIGINT on Windows

wjwwood:
This is required to avoid a (non-fatal, but annoying) traceback on Windows when launch tries to send SIGINT to it's children.

This patch automatically escalates to SIGTERM (which does work), which is what the old launch did, and is also what eventually happens with launch now, but only after waiting five seconds. This removes the wait as well.

Eventually this might be fixed with some combination of `creationflags` and `signal.CTRL_C_EVENT`, but that's an ongoing issue with subprocess on Windows.

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -17,6 +17,7 @@
 import asyncio
 import logging
 import os
+import platform
 import shlex
 import signal
 import threading
@@ -240,6 +241,14 @@ class ExecuteProcess(Action):
                 typed_event.signal_name, self.process_details['name']
             ))
             return None
+        if platform.system() == 'Windows' and typed_event.signal_name == 'SIGINT':
+            # TODO(wjwwood): remove this when/if SIGINT is fixed on Windows
+            _logger.warn(
+                "'SIGINT' sent to process[{}] not supported on Windows, escalating to 'SIGTERM'"
+                .format(self.process_details['name']))
+            typed_event = SignalProcess(
+                signal_number=signal.SIGTERM,
+                process_matcher=lambda process: True)
         _logger.info("sending signal '{}' to process[{}]".format(
             typed_event.signal_name, self.process_details['name']
         ))


wjwwood:
Before:

```
C:\dev\ros2>ros2 launch demo_nodes_cpp add_two_ints.launch.py
[INFO] [launch]: process[add_two_ints_server.EXE-1]: started with pid [12376]
[INFO] [launch]: process[add_two_ints_client.EXE-2]: started with pid [7220]
[INFO] [add_two_ints_server]: Incoming request
a: 2 b: 3
[INFO] [add_two_ints_client]: Result of add_two_ints: 5
[INFO] [launch]: process[add_two_ints_client.EXE-2]: process has finished cleanly
[INFO] [launch]: sending signal 'SIGINT' to process[add_two_ints_server.EXE-1]
[ERROR] [asyncio]: Task exception was never retrieved
future: <Task finished coro=<LaunchService._process_one_event() done, defined at C:\dev\ros2\install\Lib\site-packages\launch\launch_service.py:176> exception=ValueError('Unsupported signal: 2',)>
  File "C:\dev\ros2\install\Lib\site-packages\launch\launch_service.py", line 178, in _process_one_event
  File "C:\dev\ros2\install\Lib\site-packages\launch\launch_service.py", line 197, in __process_event
  File "C:\dev\ros2\install\Lib\site-packages\launch\utilities\visit_all_entities_and_collect_futures_impl.py", line 38, in visit_all_entities_and_collect_futures
  File "C:\dev\ros2\install\Lib\site-packages\launch\action.py", line 39, in visit
    return self.execute(context)  # type: ignore
  File "C:\dev\ros2\install\Lib\site-packages\launch\actions\opaque_function.py", line 72, in execute
    return self.__function(context, *self.__args, **self.__kwargs)
  File "C:\dev\ros2\install\Lib\site-packages\launch\actions\execute_process.py", line 249, in __on_signal_process_event
    self._subprocess_transport.send_signal(typed_event.signal)
  File "c:\python36\lib\asyncio\base_subprocess.py", line 152, in send_signal
    self._proc.send_signal(signal)
  File "c:\python36\lib\subprocess.py", line 1131, in send_signal
    raise ValueError("Unsupported signal: {}".format(sig))
ValueError: Unsupported signal: 2
[ERROR] [launch]: process[add_two_ints_server.EXE-1] failed to terminate '5' seconds after receiving 'SIGINT', escalating to 'SIGTERM'
[INFO] [launch]: sending signal 'SIGTERM' to process[add_two_ints_server.EXE-1]
[ERROR] [launch]: process[add_two_ints_server.EXE-1] process has died [pid 12376, exit code 1, cmd 'C:\dev\ros2\install\lib\demo_nodes_cpp\add_two_ints_server.EXE'].
```

After:

```
C:\dev\ros2>ros2 launch demo_nodes_cpp add_two_ints.launch.py
[INFO] [launch]: process[add_two_ints_server.EXE-1]: started with pid [11484]
[INFO] [launch]: process[add_two_ints_client.EXE-2]: started with pid [1344]
[INFO] [add_two_ints_server]: Incoming request
a: 2 b: 3
[INFO] [add_two_ints_client]: Result of add_two_ints: 5
[INFO] [launch]: process[add_two_ints_client.EXE-2]: process has finished cleanly
[WARNING] [launch]: 'SIGINT' sent to process[add_two_ints_server.EXE-1] not supported on Windows, escalating to 'SIGTERM'
[INFO] [launch]: sending signal 'SIGTERM' to process[add_two_ints_server.EXE-1]
[ERROR] [launch]: process[add_two_ints_server.EXE-1] process has died [pid 11484, exit code 1, cmd 'C:\dev\ros2\install\lib\demo_nodes_cpp\add_two_ints_server.EXE'].
```

dhood:
For the [example counter](https://github.com/ros2/launch/blob/5526541eb711926f87445a3480d9ddec156e34c9/launch/examples/launch_counters.py#L111) that ignores sigterms, because we don't cancel [the sigterm timer](https://github.com/ros2/launch/blob/5526541eb711926f87445a3480d9ddec156e34c9/launch/launch/actions/execute_process.py#L282) that's already been scheduled, it's going to have a misleading error message [printed](https://github.com/ros2/launch/blob/5526541eb711926f87445a3480d9ddec156e34c9/launch/launch/actions/execute_process.py#L285) 5s after sigint which will again say that sigint is being escalated to sigterm. Shall we cancel that timer to prevent the second sigterm and the message?

nuclearsandwich:
I also tested this on Windows and had the same before and after output as William.

wjwwood:
We cannot cancel the `SIGTERM` timer from there because we don't know if `SIGINT` was sent by the user (using the `EmitEvent` action) or as part of the escalating shutdown.

I don't think it will be misleading because it will simply send SIGTERM twice, and the first time the warning message will indicate that is because it cannot send SIGINT as requested.

Also, the counter doesn't work right on Windows because you cannot ignore `SIGTERM` on Windows either. This is because Windows just emulates `SIGTERM` as a signal and instead just calls a terminate function on the process which cannot be handle that way I think. I'll have to rethink the behavior on Windows w.r.t. to shutting down processes programmatically and likely rewrite the `counter.py` example, but this is an incremental improvement only. It just avoids the traceback and ensures the process actually gets terminated.

wjwwood:
I still need a code review so that I can run CI.

wjwwood:
Thanks for the review @dhood!

CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4829/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1684/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4008/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4875/)

PR #121, opened 2018-07-25
implement Action conditions

### LINKED ISSUE #105: [launch] add concept of Conditions
The concept of a "Condition" is a predicate that applies to certain kinds of entities like (but not exclusively) `launch.Action`.

This is the equivalent of `if` and `unless` attributes in `roslaunch` from ROS 1:

https://wiki.ros.org/roslaunch/XML#if_and_unless_attributes

The proposed `launch.Condition` would:

- take a predicate which would be a list of substitutions
- convert the list of substitutions into a string when the associated entity is visited
- evaluate the resulting predicate string for true/false

There could be sub classes for convenience like `launch.IfCondition` and `launch.UnlessCondition`.

Conditions will need to be part of the interface of any entity which can be affected by them (i.e. either `Action` or `LaunchDescriptionEntity` should have a `condition` argument to the constructor.

One thing I haven't decided on is how to combine multiple conditions, the options are:

- only ever take one condition, require compound conditions to be evaluated with `Substitution`s like `launch.substitutions.PythonExpression`
- take a list of conditions, and do a boolean 'and' across all of them
- take only one condition, but have additional convenience classes based on `Condition`, like `CompoundCondition` or `AndCondition`/`OrCondition`.

wjwwood:
This fixes #105 by adding a Condition class, making Actions take them and evaluate them as the predicate for "executing" the action, and adding the `IfCondition` and `UnlessCondition` classes which can operate on `Substitution` classes.

See #105 for more design details and conceptual information. The idea that actions can be conditionally executed with a `Condition` is also in the architecture document (in the `doc` folder).

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -29,6 +29,7 @@ from typing import Iterable
 from typing import List
 from typing import Optional
 from typing import Text
+from typing import Tuple  # noqa: F401
 
 from osrf_pycommon.process_utils import async_execute_process
 from osrf_pycommon.process_utils import AsyncSubprocessProtocol
@@ -161,7 +162,7 @@ class ExecuteProcess(Action):
         super().__init__(**kwargs)
         self.__cmd = [normalize_to_list_of_substitutions(x) for x in cmd]
         self.__cwd = cwd if cwd is None else normalize_to_list_of_substitutions(cwd)
-        self.__env = None  # type: Optional[Dict[List[Substitution], List[Substitution]]]
+        self.__env = None  # type: Optional[List[Tuple[List[Substitution], List[Substitution]]]]
         if env is not None:
             self.__env = []
             for key, value in env.items():
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -71,16 +71,20 @@ class ExecuteProcess(Action):
     """Action that begins executing a process and sets up event handlers for the process."""
 
     def __init__(
-        self, *,
+        self,
+        *,
         cmd: Iterable[SomeSubstitutionsType],
         cwd: Optional[SomeSubstitutionsType] = None,
         env: Optional[Dict[SomeSubstitutionsType, SomeSubstitutionsType]] = None,
         shell: bool = False,
-        sigterm_timeout: SomeSubstitutionsType = LaunchConfiguration('sigterm_timeout', default=5),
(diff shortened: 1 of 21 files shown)


wjwwood:
CI (`launch` and `launch_ros` only):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4951/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1771/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4110/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4967/)

wjwwood:
New CI after fixing flake8 issue:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4958/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1778/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4118/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4974/)

mjcarroll:
This is the first time I've seem PEP484 in the wild, pretty nifty.

launch/launch/actions/execute_process.py
```diff
@@ -29,6 +29,7 @@
 from typing import List
 from typing import Optional
 from typing import Text
+from typing import Tuple  # noqa: F401
```
mjcarroll: Why is F401 needed here, looks like you are using Tuple?

wjwwood: Because we have to support python 3.5, we cannot use the type annotation for member variables syntax which was added in 3.6. So we use the type comment, which is the only place I use Tuple. So flake8 doesn’t count that as using it so I suppressed the unused import warning.

wjwwood: Because we have to support python 3.5, we cannot use the type annotation for member variables syntax which was added in 3.6. So we use the type comment, which is the only place I use Tuple. So flake8 doesn’t count that as using it so I suppressed the unused import warning.

wjwwood:
There is no example using it yet. I have a couple of related features (all part of “include another launch file into a namespace”) which I wanted to finish before updating or adding new examples.

wjwwood:
I really like the type annotations. I would use them everywhere but I caught at least ten to fifteen errors while writing the code before even running my tests.

wjwwood:
Would not*

Issue #112, opened 2018-07-12
[launch] prevent launch from hanging when unexpected exceptions occur internally


wjwwood:
A follow up of https://github.com/ros2/launch/issues/90, where an exception in the default ros launch description caused launch itself to hang, see:

https://github.com/ros2/launch/issues/90#issuecomment-400852504

PR #123, opened 2018-08-14
add ability to define and pass launch arguments to launch files

### LINKED ISSUE #107: [launch] add new concept for LaunchArgument
The idea behind the concept of `LaunchArgument` is that it should be something you can declare (perhaps as an Action) in a launch description which indicates some input configurations to the launch description.

If the launch description is included by another launch description then the arguments would be part of that action, but if the launch description is "run" by `ros2 launch` it would be possible to pass them as command line arguments.

This is partially equivalent to the `<arg>` tag in `roslaunch` from ROS 1:

https://wiki.ros.org/roslaunch/XML/arg

It should take a name, optionally a default value, optionally documentation for the argument, and possibly other things commonly associated with command line arguments, like being a flag rather than a key-value option or with special validation logic (is it the right type or in the right range)? Initially it should always have a type of string, and it can be evaluated by the things consuming it.

The result (default value or user provided) would be stored in a `LaunchConfiguration` of the same name.

wjwwood:
This pull request is ready for review.

To try it, you can use the example files in `ros2launch`, e.g.:

```
% ros2 launch --show-args ./path/to/ros2launch/examples/example.launch.py
...

% ros2 launch ./path/to/ros2launch/examples/example.launch.py
...

% ros2 launch ./path/to/ros2launch/examples/example.launch.py node_prefix:=foo_
...
```

fixes #107

### DIFF
--- a/launch/launch/action.py
+++ b/launch/launch/action.py
@@ -44,6 +44,11 @@ class Action(LaunchDescriptionEntity):
         """
         self.__condition = condition
 
+    @property
+    def condition(self) -> Optional[Condition]:
+        """Getter for condition."""
+        return self.__condition
+
     def describe(self) -> Text:
         """Return a description of this Action."""
         return self.__repr__()
--- a/launch/launch/actions/__init__.py
+++ b/launch/launch/actions/__init__.py
@@ -14,6 +14,7 @@
 
 """actions Module."""
 
+from .declare_launch_argument import DeclareLaunchArgument
 from .emit_event import EmitEvent
 from .execute_process import ExecuteProcess
 from .include_launch_description import IncludeLaunchDescription
@@ -24,6 +25,7 @@ from .set_launch_configuration import SetLaunchConfiguration
 from .timer_action import TimerAction
 
 __all__ = [
+    'DeclareLaunchArgument',
     'EmitEvent',
     'ExecuteProcess',
     'IncludeLaunchDescription',
--- a/launch/launch/actions/declare_launch_argument.py
+++ b/launch/launch/actions/declare_launch_argument.py
@@ -1,0 +1,124 @@
+# Copyright 2018 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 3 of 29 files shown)


wjwwood:
@nuclearsandwich c45a877 should fix Python 3.5 support. Why it fixes it 🤷‍♂️ .

dhood:
> Why it fixes it :man_shrugging:

I tracked this down once in the past: https://github.com/ros2/launch/pull/81/commits/72ae169959a8bbd1130aee0d00b4c80e0e4c15a4#r195963696

dhood:
edit (bad link): https://github.com/ros2/launch/pull/81/commits/72ae169959a8bbd1130aee0d00b4c80e0e4c15a4#r195963696

wjwwood:
Ok, this one is ready for review, CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5091/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1875/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4221/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5072/)

wjwwood:
So, I think that if you change the name of the executable that the example launch file is looking for or delete it (as you did), then you'll get the same behavior on master. So I don't really want to have to fix that on this pull request. It's already being tracked here: https://github.com/ros2/launch/issues/112 I hope to have time to look into that tomorrow.

Can you re-review with the above in mind, only testing out the new features (passing of arguments, etc...).

wjwwood:
The only difference with master is that I don't print out the traceback by default (only on debug), which I changed in this pr, but it still produces the same exception and runtime behavior.

wjwwood:
I looked into it and the fix was pretty simple, so I went ahead and committed it in this pr: b4969f4

Basically that just lets already running things shutdown when an unhandled exception is received.  I was never able to reproduce the "cannot even ctrl-c it" scenario, even before.

I also committed f077ab6 to avoid invalid state in the execute process action if the exception originated there (basically it registers event handlers, and then an exception occurs, making the event handlers invalid).

Also, I'd appreciate feedback on 4dbd9bb which is restoring the state of master which is that exceptions are reported as a single line and the traceback is only in the debug output.

PR #124, opened 2018-08-16
destroy node on shutdown to avoid hang

### LINKED ISSUE #89: Launching on arm64 with Fast-RTPS with fat archive from 2018-06-21 never quits
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu Bionic, arm64
- Installation type:
  - Fat archive from 2018-06-21, downloaded from here: https://ci.ros2.org/view/packaging/job/packaging_linux-aarch64/475/artifact/ws/ros2-package-linux-aarch64.tar.bz2
- DDS implementation:
  - Fast-RTPS
  - rclcpp

#### Steps to reproduce issue

```
$ ros2 launch share/demo_nodes_cpp/launch/topics/talker_listener.launch.py
```

Let it run for a few seconds, then Ctrl-C:

```
root@changeling:~/ros2-linux# ros2 launch share/demo_nodes_cpp/launch/topics/talker_listener.launch.py
[INFO] [launch]: process[talker-1]: started with pid [21936]
[INFO] [launch]: process[listener-2]: started with pid [21937]
[INFO] [talker]: Publishing: 'Hello World: 1'
[INFO] [listener]: I heard: [Hello World: 1]
[INFO] [talker]: Publishing: 'Hello World: 2'
[INFO] [listener]: I heard: [Hello World: 2]
[INFO] [talker]: Publishing: 'Hello World: 3'
[INFO] [listener]: I heard: [Hello World: 3]
^C[WARNING] [launch.LaunchService]: user interrupted with ctrl-c (SIGINT)
signal_handler(2)
signal_handler(2)
[INFO] [launch]: process[talker-1]: process has finished cleanly
[INFO] [launch]: process[listener-2]: process has finished cleanly
```

It hangs there forever, until the user hits Ctrl-C.

#### Expected behavior
Launch file terminates cleanly on first Ctrl-C

#### Actual behavior
Launch hangs around "forever".

wjwwood:
fixes #89 

At least it fixes the issue on bionic x86_64. I didn't try aarch64 yet, but I believe it's the same issue and so the same fix. If anyone has easy access to aarch64 and could test it, that would be great.

I cannot fully explain the reason for why this is necessary or why it results in the hang. There's more work to be done here (in my opinion) so that this doesn't happen to typical users. There's no reason that I should have to manually destroy the node.

Thanks to @nuclearsandwich for giving me the last hint I needed to find this workaround. We discussed that a potential better solution would be to have shutdown enforce the destruction of all nodes and services before shutdown finishes. That has it's own thread-safety issues and potential for deadlock, but I think that's what needs to happen to avoid this issue.

There might be another solution which address the deadlock directly or at least throws an error rather than deadlocking. I think this because I still cannot explain why this issue doesn't manifest on macOS or Xenial.

### DIFF
--- a/launch_ros/launch_ros/default_launch_description.py
+++ b/launch_ros/launch_ros/default_launch_description.py
@@ -76,6 +76,7 @@ class ROSSpecificLaunchStartup(launch.actions.OpaqueFunction):
     def _shutdown(self, event: launch.Event, context: launch.LaunchContext):
         self.__shutting_down = True
         self.__rclpy_spin_thread.join()
+        self.__launch_ros_node.destroy_node()
 
     def _run(self):
         executor = rclpy.get_global_executor()


nuclearsandwich:
I will give this a spin (pun extremely intended) on ARM tomorrow and add my review.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5061/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1854/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4201/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5052/)

PR #131, opened 2018-08-25
add class to provide some limited testing options

### LINKED ISSUE #127: Unit tests based on new launch API
While porting some ROS 1 packages to ROS 2 I wanted to also migrate the `rostest`s. Currently there is no API yet in `launch_testing` (beside the legacy API).

I tried two different approaches - the idea would be to move the "generic" stuff from one of these prototypes into `launch_testing` in order to be able to reuse it across tests:

1. Creating custom classes for the `LaunchDescription` as well as the `LaunchService` (https://github.com/GT-RAIL/async_web_server_cpp/pull/12/commits/1dad9c5ae6d4ca7521a5d7f6914d9b4231acd1cd). In this case the whole launch file uses those custom classes instead of the original ones. The custom description contains the logic for the return code handling and shutting down. The custom service contains the logic to use the return codes from the test actions as the global return code.

  The downside of this approach is that requires the launch file specifically being written for testing (by using these custom classes).

2. A custom class (`LaunchTestService`) is used in addition to the normal launch entities and the actions which are tests are being registered at that class and the class also wraps the `run` invocation on the "normal" launch service: (https://github.com/GT-RAIL/async_web_server_cpp/pull/12/commits/fc6f35adc5cb785a7b4234b32426c0a12fccbd74)

Any kind of feedback is highly appreciated.

dirk-thomas:
Closes #127.

The added class provides some limited ability to use new-style launch file to write tests. This allowed to port the following tests in Robot Web Tools:
* https://github.com/GT-RAIL/async_web_server_cpp/pull/12/commits/b2546f5aa59bc9138098ab49fe9f45e5ce2873b2
* https://github.com/GT-RAIL/rosauth/pull/18/commits/47b5dfbbfb1a42eb877f4ed2776d7768ec9a765e

### DIFF
--- a/launch_testing/launch_testing/__init__.py
+++ b/launch_testing/launch_testing/__init__.py
@@ -1,3 +1,25 @@
+# Copyright 2018 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from collections import OrderedDict
+
+from launch.actions import EmitEvent
+from launch.actions import ExecuteProcess
+from launch.actions import RegisterEventHandler
+from launch.event_handlers import OnProcessExit
+from launch.events import Shutdown
+
 # for backward compatibility
 from launch_testing.legacy import InMemoryHandler  # noqa: F401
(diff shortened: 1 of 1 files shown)


dirk-thomas:
> Long term it would be nice to have something that works with/like `ros2 launch`, but that will require some more thought and this is a good improvement for now.

Absolutely, there is much more work needed in this area e.g. allowing test processes to generate result file.

nuclearsandwich:
This PR has raised linter errors turning all nightlies unstable over the weekend. I've opened https://github.com/ros2/launch/pull/132 to resolve them.

dangyuluo:
Hi @dirk-thomas
I see you've started moving `launch_testing` towards the new launch framework and this ticker provides limited function(the test passed if all tested processed has `rc 0`). It would be nice to have regex feature like the legacy launch_testing. Is there any time table of this?

Or if you guys are too busy on other stuff, do you accept PR?

dirk-thomas:
We are not actively working on this part. PRs are always welcome.

PR #132, opened 2018-08-27
Fix linter errors from #131.

nuclearsandwich:
This fixes the linters which have been failing on nightlies over the weekend since #131. 

[Build Status](https://ci.ros2.org/job/ci_linux/5098/)

It may be that we don't care about these issues or have different preferences for resolving them in which case this PR may be amended to provide a different solution.

### DIFF
--- a/launch_testing/launch_testing/__init__.py
+++ b/launch_testing/launch_testing/__init__.py
@@ -21,12 +21,12 @@ from launch.event_handlers import OnProcessExit
 from launch.events import Shutdown
 
 # for backward compatibility
-from launch_testing.legacy import InMemoryHandler  # noqa: F401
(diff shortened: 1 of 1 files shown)


wjwwood:
Thanks!

dirk-thomas:
Nvm, after an update the config wasn't working anymore. It needs: `ignore = I202` and `import-order-style = google` to reproduce the behavior in my editor (basically our custom configuration).

nuclearsandwich:
Yeah the sort ordering was what I was most suspicious about. I was able to reproduce both errors locally running the tests via colcon test. I usually end up copying or linking the config from ament_flake8 into ros2 projects I'm working on so that editor based feedback matches the buildfarm linter configuration.

PR #133, opened 2018-08-28
add GroupAction for conditionally including other actions and scoping

### LINKED ISSUE #106: [launch] add new Action called GroupAction
Previously I had described a "Group" or "Group of Actions" as a separate entity from "Action", but since then I think we can just have a special Action that yields other actions and optionally does things like pushes/pops the launch configurations and set/unset specific launch configurations.

I believe the action I've described is capable of fulfilling the use cases I originally imagined for groups:

- push containing nodes into a namespace
  - would be done by modifying the `ros_namespace` `LaunchConfiguration` that affects only the actions it yields
    - see: https://github.com/ros2/launch/blob/b78d87f31bb995c7226df66826e1a57079ae3a3f/launch_ros/launch_ros/actions/node.py#L80-L81
- conditionally include other launch descriptions or actions
  - done with proposed Condition concept which would work on any Action, including this one
    - see: https://github.com/ros2/launch/issues/105
- scope arbitrary launch configurations
  - done by using the `LaunchContext` to push and pop them
    - see: https://github.com/ros2/launch/blob/b78d87f31bb995c7226df66826e1a57079ae3a3f/launch/launch/launch_context.py#L146-L157

We should also update the design document (https://github.com/ros2/design/pull/163).

wjwwood:
fixes #106 

See #106 for details of how it works, but in short it's an action that just yields other actions (sort of like a launch description, but without a source) and it can also have a condition for its inclusion (like all actions) and it can also optionally scope all changes to launch configurations made by actions within it.

This class is mostly syntactic sugar since it only uses other existing actions to be implemented.

### DIFF
--- a/launch/launch/actions/__init__.py
+++ b/launch/launch/actions/__init__.py
@@ -16,6 +16,7 @@
 
 from .emit_event import EmitEvent
 from .execute_process import ExecuteProcess
+from .group_action import GroupAction
 from .include_launch_description import IncludeLaunchDescription
 from .log_info import LogInfo
 from .opaque_function import OpaqueFunction
@@ -28,6 +29,7 @@ from .timer_action import TimerAction
 __all__ = [
     'EmitEvent',
     'ExecuteProcess',
+    'GroupAction',
     'IncludeLaunchDescription',
     'LogInfo',
     'OpaqueFunction',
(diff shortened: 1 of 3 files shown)


wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5102/)
* Linux @ Xenial [Build Status](https://ci.ros2.org/job/ci_linux/5103/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1884/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4230/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5081/)

Issue #117, opened 2018-07-12
[launch_ros] add ability to pass parameters to Node actions


wjwwood:
Currently you can pass remapping arguments to Node as a `dict`, but to set parameters in a node from a launch file you need to pass a normal argument which points to an existing `.yaml` file.

To pass parameters via the launch file (without predefining them in an external `.yaml` file) will ultimately require the ability to pass parameters via the command line one at a time, which as of right now is not implemented yet.

In the meantime, individually specified parameters in a launch file for a node could be written into a temporary `.yaml` file by `launch` and then passed to the node, deleting the file after the node exits. However, I think this would step on a user specified parameter `.yaml` file unless the node can take more than one right now (it was discussed, but I don't know off-hand if it was implemented). In either case, if launch also lets you specify one or more `.yaml` files to be used as parameters (in addition to, and at the same time as, individually specified parameters for a node), then launch could just combine all of the yaml files and parameter kay-value pairs into a single `.yaml` file, cascading them in order (overwriting keys that appear more than once, with last wins).

It might look something like this:

```python
launch_ros.actions.Node(...,
  parameters=[
    '/path/to/first_parameter_file.yaml',
    launch_ros.actions.node.NodeParameters({
      'foo': '1',
      // can have substitutions
      [LaunchConfiguration('ros_namespace'), '.my_node.executable_location']:
        [ExecutableInPackage('my_node', 'my_package')],
      ...,
    },
    [PackageShareDir('my_package'), Dir('config') / 'my_config_file.yaml'],
  ],
)
```

sloretz:
> In the meantime, individually specified parameters in a launch file for a node could be written into a temporary .yaml file by launch and then passed to the node, deleting the file after the node exits. However, I think this would step on a user specified parameter .yaml file unless the node can take more than one right now (it was discussed, but I don't know off-hand if it was implemented).

Multiple yaml files can be given on the command line. If two files specify a value for the same parameter then the value from the later file is used.

```
ros2 run demo_nodes_cpp talker __params:=generic-default-values.yaml __params:=more-specific-values.yaml
```

wjwwood:
Cool thanks!

sloretz:
Is this ticket complete now that #135 has been merged?

dhood:
nah, it didn't include passing parameters as dicts

dhood:
(which I'm working on now)

poonam1120:
Hi, 
[...]

PR #135, opened 2018-09-12
Add ability to pass parameter files to Node actions

dhood:
Connects to ros2/launch#117

Scoped out: passing parameters as a dictionary

Still writing tests, but example usage is:
``` python3
    parameters_file = pathlib.Path(__file__).resolve().parent / 'demo_parameters.yaml'
    return LaunchDescription([
        launch_ros.actions.Node(
            package='demo_nodes_cpp', node_executable='talker', output='screen',
            parameters=[
                str(parameters_file),  # as string
                parameters_file,  # as pathlib.Path
                [ThisLaunchFileDir(), os.sep, 'demo_parameters.yaml'],  # as substitution
            ],
        ),
    ])
```

### DIFF
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -16,6 +16,7 @@
 
 import logging
 import os
+import pathlib
 from typing import Dict  # noqa: F401
 from typing import Iterable
 from typing import List
@@ -113,7 +114,7 @@ class Node(ExecuteProcess):
             # All arguments are paths to files with parameters (or substitutions that evaluate
             # to paths).
             # TODO(dhood): add support for parameter dicts.
-            parameter_types = SomeSubstitutionsType_types_tuple
+            parameter_types = list(SomeSubstitutionsType_types_tuple) + [pathlib.Path]
             i = 0
             for param in parameters:
                 ensure_argument_type(param, parameter_types, 'parameters[{}]'.format(i), 'Node')
@@ -190,6 +191,8 @@ class Node(ExecuteProcess):
         if self.__parameters is not None:
             self.__expanded_parameters = []
             for param_file_path in self.__parameters:
+                if isinstance(param_file_path, pathlib.Path):
+                    param_file_path = str(param_file_path)
                 expanded_param_file_path = perform_substitutions(
                     context, normalize_to_list_of_substitutions(param_file_path))
                 if not os.path.isfile(expanded_param_file_path):
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -15,6 +15,7 @@
 """Module for the Node action."""
 
 import logging
+import os
 from typing import Dict  # noqa: F401
 from typing import Iterable
 from typing import List
@@ -26,7 +27,9 @@ from launch.action import Action
 from launch.actions import ExecuteProcess
 from launch.launch_context import LaunchContext
 from launch.some_substitutions_type import SomeSubstitutionsType
+from launch.some_substitutions_type import SomeSubstitutionsType_types_tuple
 from launch.substitutions import LocalSubstitution
+from launch.utilities import ensure_argument_type
 from launch.utilities import normalize_to_list_of_substitutions
 from launch.utilities import perform_substitutions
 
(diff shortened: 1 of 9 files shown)


dhood:
Putting this in review thanks!

A new package `test_launch_ros` has been added so that demo nodes can be launched in tests. It's in a separate package to not introduce a circular package dependency if the demo packages have launch files (`demo_nodes_py` doesn't at the moment, but `demo_nodes_cpp` does, for example). There is still a circular repo dependency.

launch_ros/launch_ros/actions/node.py
```diff
@@ -165,6 +187,20 @@ def _perform_substitutions(self, context: LaunchContext) -> None:
...
+        if self.__parameters is not None:
+            self.__expanded_parameters = []
+            for param_file_path in self.__parameters:
+                if isinstance(param_file_path, pathlib.Path):
+                    param_file_path = str(param_file_path)
+                expanded_param_file_path = perform_substitutions(
+                    context, normalize_to_list_of_substitutions(param_file_path))
+                if not os.path.isfile(expanded_param_file_path):
+                    _logger.warn(
+                        'Parameter file path is not valid: {}'.format(expanded_param_file_path))
```
wjwwood: Would it be more accurate here to say "is not a file"? Is not valid implies a range of things to me, including that it exists but the contents are malformed.

dhood:
A circular repo dependency already exists for this package, from what I can remember (because rcl and rcutils use launch_testing, perhaps?). That's not to say we should just make it worse, I agree. But, if this repo already needs special release attention, then it may be worth the additional effort to keep the tests co-located with what they're testing.
What do you think @wjwwood ? move them to the system_tests repo?

mikaelarguedas:
When discussing it at Bouncy release time, one suggestion was to split `launch` + `launch_testing` in another repo.
`launch_ros` and `ros2launch` could then be released after rclpy and friends removing the circular dependency

mikaelarguedas:
> one suggestion was to split launch + launch_testing in another repo.

At that point the new "test_launch_ros" could just go in the same repo as `launch_ros`.

I don't feel like this should hold this PR though, just should be address before the Crystal release.

dirk-thomas:
> I don't feel like this should hold this PR though, just should be address before the Crystal release.

During the Bouncy release it was discover that unnoticed we had introduced a circular repository dependency. We agreed to continue with the release with the conclusion that this should be resolved after the release.

The fact that we haven't followed up on this task shouldn't justify adding more of these undesired cases - and this patch introduces a new different cycle (as far as I can see).

So this problem can very well be resolved right now and doesn't need to keep building up technical debt for the next release. I agree that this PR is not "responsible" to resolve the existing cycle but it should not introduce another new one.

launch_ros/launch_ros/actions/node.py
```diff
@@ -165,6 +187,20 @@ def _perform_substitutions(self, context: LaunchContext) -> None:
...
+        if self.__parameters is not None:
+            self.__expanded_parameters = []
+            for param_file_path in self.__parameters:
+                if isinstance(param_file_path, pathlib.Path):
+                    param_file_path = str(param_file_path)
+                expanded_param_file_path = perform_substitutions(
+                    context, normalize_to_list_of_substitutions(param_file_path))
+                if not os.path.isfile(expanded_param_file_path):
+                    _logger.warn(
+                        'Parameter file path is not file: {}'.format(expanded_param_file_path))
```
mikaelarguedas: Nit: not **a** file ?

dhood:
`launch` + `launch_testing` in a repo separate to `launch_ros` + `ros2launch` + `test_launch_ros` sounds like a reasonable repo split to me and would fix both cycles FWICT. Happy to do it as a follow up to this and  https://github.com/ros2/launch/pull/137.

wjwwood:
> We agreed to continue with the release with the conclusion that this should be resolved after the release.

I don't remember agreeing to that, I think I just said it could be changed later.

Personally, I don't think these types of cycles are a problem. In the particular case of launch, anyone making changes to launch will find it more convenient and have less overhead when they're all in the same repository. The only downside I'm aware of is that the first release needs to be done in two steps. So that's a continuous benefit for all contributors weighed against an ideal and an extra step in the release process once or twice per year for a single person. As the single affected individual for launch at the moment, I think it's a worth while trade-off to keep things in the same repository.

Therefore, I would not actually split this repository up and I definitely would not do it as part of this pull request.

dhood:
alright, I am gonna go with @wjwwood's guidance on this as maintainer of the repo and merge this PR as is. Doesn't sound like there is followup repo shuffling planned for this repo: the repo dependency cycle will stay, with the reasoning that the overhead at release time is worth it in order to keep the packages co-located.

wjwwood:
At the very least, follow up discussion of breaking up the repository can take place on a new issue.

dhood:
_CI for this branch was included in CI listed in https://github.com/ros2/launch/pull/137#issuecomment-421113871_

PR #136, opened 2018-09-12
Return 1 on caught exceptions from launch

dhood:
Noticed this when writing tests that purposefully enter this case. (https://github.com/ros2/launch/pull/135/files#diff-25cebd7b45bcde9d178003dca93986abR30)


dhood:
just to check nothing was relying on that return code:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5168/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1942/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4293/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5148/)

PR #138, opened 2018-09-14
Support passing parameters as a dictionary

dhood:
Closes ros2/launch#117
Requires ros2/rcl#299

The dictionary can be nested (to specify different groups of parameters) and can have substitutions at each level.

A yaml file is written following [this structure](https://github.com/ros2/rcl/blob/c9fe312674f8c5a2c5b40b6208c830f4ca8c40ff/rcl_yaml_param_parser/README.md). It always specifies the namespace and node name (requires ros2/rcl#299). Scoped out: deleting the file(s).

Design decisions:
1. Typically "iterables" of substitutions will be concatenated into a single string. However, a user may have been trying to specify a parameter array with each element a substitution. To distinguish these two cases, only Tuples of substitutions will be concatenated into a string, not any Iterable. Lists of substitutions will be treated as a parameter array.
1. Parameter files only work if the node name is specified, so if a parameter dict is passed, the node name must be specified also (https://github.com/ros2/launch/issues/139).

### DIFF
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -91,13 +91,28 @@ class Node(ExecuteProcess):
         If no node_namespace is given, then the default namespace `/` is
         assumed.
 
+        The parameters are passed as a list, with each element either a yaml
+        file that contains parameter rules (string or pathlib.Path to the full
+        path of the file), or a dictionary that specifies parameter rules.
+        Keys of the dictionary can be strings or an iterable of Substitutions
+        that will be expanded to a string.
+        Values in the dictionary can be strings, integers, floats, or tuples
+        of Substitutions that will be expanded to a string.
+        Additionally, values in the dictionary can be lists of the
+        aforementioned types, or another dictionary with the same properties.
+        A yaml file with the resulting parameters from the dictionary will be
+        written to a temporary file, the path to which will be passed to the
+        node.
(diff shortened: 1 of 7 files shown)


launch_ros/launch_ros/actions/node.py
```diff
@@ -158,6 +182,70 @@ def node_name(self):
             raise RuntimeError("cannot access 'node_name' before executing action")
         return self.__final_node_name
 
+    def _create_params_file_from_dict(self, context, params):
+        with NamedTemporaryFile(mode='w', prefix='launch_params_', delete=False) as h:
+            param_file_path = h.name
+            # TODO(dhood): clean up generated parameter files.
```
wjwwood: Is the intention to merge this and follow up with a pull request to clean up the files or do you think we can do without it?

I know we talked about it before and I said it would be ok to not do that for now, but now I'm thinking we should really clean up after ourselves. What are your thoughts?

dhood: I agree that it's important also, was planning to leave it as followup to simplify reviewing of this PR. The tests are relying on the files being around atm so was going to also have an option to disable the cleanup.

dhood:
I'm gonna wait for https://github.com/ros2/rcl/pull/299 before I merge this so that it works for nodes in the root namespace. Will move this out of the review column.

Issue #102, opened 2018-07-11
[launch] improve test coverage for non-core classes


wjwwood:
Currently there are minimal tests for some of the core classes like `LaunchDescription`, `LaunchService`, and `Action`, but there's not really any tests for things like:

- `launch.LaunchInstrospector`
- `launch.Substitution`
- anything in the namespaces:
  - `launch.actions`
  - `launch.event_handlers`
  - `launch.events`
  - `launch.substitutions`
  - `launch.utilities`

jacobperron:
Tests added:
- [ ] launch.LaunchInstrospector
- [ ] launch.Substitution
- [ ] launch.actions
- [ ] launch.event_handlers
- [ ] launch.events
- [ ] launch.substitutions
- [ ] launch.utilities
    - [ ] class_tools_impl
    - [ ] create_future_impl
    - [ ] normalize_to_list_of_substitutions_impl
    - [ ] signal_management
    - [ ] ensure_argument_type_impl
    - [ ] perform_substitutions_impl
    - [ ] visit_all_entities_and_collect_futures_impl

PR #142, opened 2018-09-25
Add tests for class_tools_impl and fix is_a_subclass()

jacobperron:
In the previous implementation `is_a_subclass(int, int)` would return False. This is because `int.__class__` is of class `type`.
Wrapped the new solution in a try since `issubclass` can raise an exception when an object instance is passed, e.g. `issubclass(1, float)`.

Connects to #102

### DIFF
--- a/launch/test/launch/utilities/test_class_tools.py
+++ b/launch/test/launch/utilities/test_class_tools.py
@@ -1,0 +1,154 @@
+# Copyright 2018 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


launch/test/launch/utilities/test_class_tools.py
```diff
@@ -0,0 +1,150 @@
...
+               [],
+               (),
+               range(6),
+               bytes(7),
+               bytearray(),
+               memoryview(b'nine'),
+               set(),
+               frozenset(),
+               {},
+               MockParentClass()]
```
wjwwood: nitpick: I'd prefer this style:

```python
variable_name = [
    1,
    'foo',
    # ...,
]
```

And avoid vertical alignments.

wjwwood:
Feel free to either squash and merge directly or fixup the nitpick and then squash and merge without re-review (up to you whether or not to run CI again).

jacobperron:
Style fixed and squashed.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5276/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2032/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4392/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5255/)

Issue #104, opened 2018-07-11
[launch] provide default console and filesystem logging for executed processes


wjwwood:
Currently in `launch` an executed process does not have anything printed to the console and nothing is logged to a file.

Some of the examples will create custom event handlers for this:

https://github.com/ros2/launch/blob/b78d87f31bb995c7226df66826e1a57079ae3a3f/launch/examples/launch_counters.py#L51-L62

The "default launch description" provided by `launch_ros` also creates an event handler to cause executed processes to print to the console if configured to do so (with the `output=` option to `ExecuteProcess` and it's derived classes like `Node`):

https://github.com/ros2/launch/blob/b78d87f31bb995c7226df66826e1a57079ae3a3f/launch_ros/launch_ros/default_launch_description.py#L49-L65

But even this doesn't handle writing process output to a log file:

https://github.com/ros2/launch/blob/b78d87f31bb995c7226df66826e1a57079ae3a3f/launch_ros/launch_ros/default_launch_description.py#L65

Ideally, there would be some facility for this built into `launch` and `launch_ros` could additionally configure it to write files to something like `ROS_HOME` where as it might go to `~/.launch_logs` (or something like that) instead. The user could further configure this default event handler if desired.

The user would still be able to provide their own event handler for output from process, to be run along with the default or instead of it.

jacobperron:
Looking back at [roslaunch](http://wiki.ros.org/roslaunch) from ROS 1:

There are two lists of handlers for logging messages (for stdout and stderr) that are invoked every time a log function is called.
The provided functions for calling the handlers responsible for printing and logging are:

- [printlog](https://github.com/ros/ros_comm/blob/a925610b0d6120e40414298ad060490238e8aaed/tools/roslaunch/src/roslaunch/core.py#L110)
- [printlog_bold](https://github.com/ros/ros_comm/blob/a925610b0d6120e40414298ad060490238e8aaed/tools/roslaunch/src/roslaunch/core.py#L125)
- [printerrlog](https://github.com/ros/ros_comm/blob/a925610b0d6120e40414298ad060490238e8aaed/tools/roslaunch/src/roslaunch/core.py#L143)

Each of the "printlog" functions prints to the screen as well as logs to a file:

https://github.com/ros/ros_comm/blob/a925610b0d6120e40414298ad060490238e8aaed/tools/roslaunch/src/roslaunch/core.py#L110-L123

Pythons [logging module](https://docs.python.org/3/library/logging.html) is used for writing to a log file

[...]

PR #145, opened 2018-10-05
Add LaunchLogger class

jacobperron:
Supports logging messages from multiple processes and modules to the screen, a common log file, or both.
LaunchLogger is a singleton and when it is first instantiated a log file is created of the form 'DATETIME-launch-HOSTNAME-PID.log'.
Messages sent to the screen or the log file will have the format 'TIME [LEVEL] NAME: MESSAGE'.

Connects to #104 

Tests were added that capture and check stdout, but I could only get them working by supplying pytest with the `-s` option. I'm not sure why this is required.
E.g. `colcon test --pytest-args -s --packages-select launch`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5257/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2013/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4371/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5232/)


hidmic:
Running CI again:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6287/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2707/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5167/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6126/)

hidmic:
@wjwwood ready for re-review and merge! :)

launch/test/launch/test_logging.py
```diff
@@ -0,0 +1,237 @@
...
+    with pytest.raises(ValueError):
+        launch.logging.getOutputLoggers('some-proc', {'garbage': {'log'}})
+
+    with pytest.raises(ValueError):
+        launch.logging.getOutputLoggers('some-proc', {'stdout': {'garbage'}})
+
+
+@pytest.mark.skipif(
+    pytest.config.getoption('-s') != 'no',
+    reason='seems like "-s" option is required to capture stdout from logger'
```
wjwwood: Did you try to use `capsys` (I assume so since you used `capfd`)? From this doc it seems like it would work or that you could use `capsys` or `sys.stdout` conditionally by checking `capsys.disabled()`:

https://docs.pytest.org/en/latest/capture.html#accessing-captured-output-from-a-test-function

It would be nice to not skip these when `-s` is used, because I think that's the default behavior of our buildfarm, IIRC.

hidmic: > Did you try to use `capsys` (I assume so since you used `capfd`)?

Not really, I just stuck with what @jacobperron left there. Just tried `capsys` and it works like a charm. Thanks!

hidmic:
> I guess you were trying to emulate the `logging` module from Python in some cases, to have `launch.logging` be a drop in replacement of sorts for it.

That's correct. As a result of providing all necessary functionality (e.g. a [`reset()`](https://github.com/ros2/launch/pull/145/files#diff-108c43bc598cd250a943a8a073bca4f6R451) to prevent successive `LaunchService` instances from interacting with each other, not the same as `shutdown()` as this one can only be called once per process) and keeping the API simple (e.g. no need to configure each logger every time to output to screen and the launch log file) without preventing full use of the `logging` module capabilities. It seemed to me it was the most sane way to do it considering that `logging` state is global (e.g. having code affecting `logging.root` verbosity in `launch.logging` and user code seemed awkward).

An alternative, though it wouldn't allow for some of the functionality added now, would be to turn `launch.logging` into a set of functions to configure loggers. And leave everything else to be set through the `logging` module.

> This is pretty much what's happening right now anyways, for instance the only place that `launch.logging.getLogger()` is being used instead of `logging.getLogger` appears to be in the launch service, even though there are numerous other cases throughout the code that get the logger directly from Python.

I'm under the impression that you're not considering the changes added by #147 (which we should probably merge into this one). Using `launch.logging.getLogger()` instead of `logging.getLogger()` is a convenience to avoid having to explicitly attach screen and launch log file handlers to each and every logger instance. And no, we're not leveraging loggers' hierarchy for common handling because I was unable to use hierarchy relationships to correctly model process output configurations.

hidmic:
BTW thanks for the typing annotations :) Haven't had the time to setup `mypy` on Emacs yet.

wjwwood:
> I'm under the impression that you're not considering the changes added by #147 (which we should probably merge into this one).

Yup, that bit me again. I would support merging them.

So, you are using the custom `getLogger` in a few places, but I still think it's important for users to realize that this function is from `launch` and not python itself, so I stand by my request to make it `launch.logging.getLogger` everywhere in the other pr. Following that logic, I think the function doesn't need to look like the one from python's logging, and so I wouldn't change my proposal in #190, though could still be convinced to not go that way.

What do you want to do, leave it as is or make `launch.logging` conform to the rest of launch rather than emulate Python's logging module?

hidmic:
Ok, I think I see where you're going. To have `launch.logging` taking care of it's own subtree of loggers, right? I'll take your draft and move things around a bit.

hidmic:
@wjwwood Alright, done. I had to rebase to solve some merge conflicts. 

There're two things I'm not very fond of though:

1. [To have `launch_config()` affecting `logging.root` verbosity](https://github.com/ros2/launch/pull/145/files#diff-108c43bc598cd250a943a8a073bca4f6R119).
2. [To have `reset()` affecting all loggers known to `logging`](https://github.com/ros2/launch/pull/145/files#diff-108c43bc598cd250a943a8a073bca4f6R413).

I'd be inclined to push all `launch.logging` loggers below a common namespace like `launch.<logger-name>` to scope the effect of the aforementioned functions. Thoughts?

wjwwood:
I'm not sure why but there are still merge conflicts.

hidmic:
Uh. That's odd. I'll rebase again.

hidmic:
@wjwwood ping

hidmic:
> Yeah, I guess so, but couldn't you instead store the requested log level and apply it only to the loggers as they're created using launch.logging.* functions?

Hmm, the problem with that approach is that `launch_config()` may be called more than once with different levels or even coexist with levels set on a per logger basis by the user. Having a root logger helps implementing the fallback mechanism (i.e. if no level is set here, use the parent's).

> I don't understand the problem with this one.

That it's counter-intuitive if you're using the raw `logging` module too. A `launch.logging.reset()` will affect loggers created/retrieved by `logging.getLogger()`. It seems to my like the kind of caveat/feature/bug one may lose an hour on if the documentation was not read thoroughly.

> In the end, I think it won't matter that much as the default use case is running launch in it's own process by itself. 

That's a fair argument. Though we'll have a testing framework using `launch` as a module in short.

----

Anyways, I'll rebase, test and merge this to get things going. We can revisit it latter (even open issues if you think it's worth it). Thanks for the review @wjwwood !

hidmic:
Alright, [CI passed](https://github.com/ros2/launch/pull/145#issuecomment-468003671). I'll merge this one, we can discuss outstanding issues/caveats in future PRs.

PR #147, opened 2018-10-10
Integrate LaunchLogger

jacobperron:
Mostly find/replacing instances of Python's `logging` module with `LaunchLogger`. The key parts are:

execute_process.py, a logger is configured based on the supplied `output` arg:

https://github.com/ros2/launch/blob/d83812b3245d6eee701468ea4e7312f3a7b1d60c/launch/launch/actions/execute_process.py#L484-L485

launch_service.py, the callback is registered for the `OnProcessIO` event:

https://github.com/ros2/launch/blob/d83812b3245d6eee701468ea4e7312f3a7b1d60c/launch/launch/launch_service.py#L106-L111

---

Here's an example launch file where the talker node outputs to 'screen' and the listener node outputs to 'log' (might be nice to add an OpaqueFunction or something outputing plain stdout/stderr to 'both' for completeness):

```python
"""Launch a talker and a listener."""

import os
import sys

from launch import LaunchIntrospector

from launch_ros import get_default_launch_description
import launch_ros.actions


    ld = LaunchDescription([
        launch_ros.actions.Node(
            package='demo_nodes_cpp', node_executable='talker', output='screen',
            remappings=[('chatter', 'my_chatter')]),
        launch_ros.actions.Node(
            package='demo_nodes_cpp', node_executable='listener', output='log',
            remappings=[('chatter', 'my_chatter')]),
        ])

    print('Starting introspection of launch description...\n')

    print(LaunchIntrospector().format_launch_description(ld))

    return ld
```

Screen output:

```sh
Starting introspection of launch description...

<launch.launch_description.LaunchDescription object at 0x7fd7e8847c50>
├── ExecuteProcess(cmd=[ExecInPkg(pkg='demo_nodes_cpp', exec='talker'), LocalVar('remapping 1')], cwd=None, env=None, shell=False)
└── ExecuteProcess(cmd=[ExecInPkg(pkg='demo_nodes_cpp', exec='listener'), LocalVar('remapping 1')], cwd=None, env=None, shell=False)
1539131114.2102978 [INFO] [talker-1] process started with pid [23422]
1539131115.2265387 [INFO] [talker-1] [INFO] [talker]: Publishing: 'Hello World: 1'
1539131116.2265813 [INFO] [talker-1] [INFO] [talker]: Publishing: 'Hello World: 2'
1539131117.2271516 [INFO] [talker-1] [INFO] [talker]: Publishing: 'Hello World: 3'
1539131118.2269557 [INFO] [talker-1] [INFO] [talker]: Publishing: 'Hello World: 4'
1539131119.2268794 [INFO] [talker-1] [INFO] [talker]: Publishing: 'Hello World: 5'
^Cuser interrupted with ctrl-c (SIGINT)
1539131120.2120588 [INFO] [talker-1] signal_handler(2)
1539131120.2136695 [INFO] [talker-1] process has finished cleanly [pid 23422]
```

Log contents (located in `~/.ros/log`):

```sh
1539131114.210665 [INFO] [listener-2] process started with pid [23423]
1539131115.2269616 [INFO] [listener-2] [INFO] [listener]: I heard: [Hello World: 1]
1539131116.227143 [INFO] [listener-2] [INFO] [listener]: I heard: [Hello World: 2]
1539131117.2280726 [INFO] [listener-2] [INFO] [listener]: I heard: [Hello World: 3]
1539131118.2277687 [INFO] [listener-2] [INFO] [listener]: I heard: [Hello World: 4]
1539131119.2276838 [INFO] [listener-2] [INFO] [listener]: I heard: [Hello World: 5]
1539131120.213369 [INFO] [listener-2] signal_handler(2)
1539131120.213953 [INFO] [listener-2] process has finished cleanly [pid 23423]
```

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -71,10 +71,6 @@ _global_process_counter_lock = threading.Lock()
 _global_process_counter = 0  # in Python3, this number is unbounded (no rollover)
 
 
-def default_output_prefix(line: Text, action: 'ExecuteProcess') -> Text:
-    return '[{}] {}'.format(action.name, line)
-
-
 class ExecuteProcess(Action):
     """Action that begins executing a process and sets up event handlers for the process."""
 
@@ -92,7 +88,7 @@ class ExecuteProcess(Action):
             'sigkill_timeout', default = 5),
         prefix: Optional[SomeSubstitutionsType] = None,
         output: Text = 'log',
-        output_prefix: Optional[Callable[[Text, Action], Text]] = None,
+        output_format: Text = '[{this.name}] {line}',
         log_cmd: bool = False,
         **kwargs
     ) -> None:
@@ -165,10 +161,11 @@ class ExecuteProcess(Action):
             called 'launch-prefix'
         :param: output configuration for process output logging. Default is 'log' i.e.
             log both stdout and stderr to launch main log file and stderr to the screen.
-            See launch.logging.getOutputLoggers() documentation for further reference
+            See `launch.logging.getOutputLoggers()` documentation for further reference
             on all available options.
-        :param: output_prefix to preprocess each output line before logging it. Defaults
-            to prepending each line with the process name.
+        :param: output_format for logging each output line, supporting `str.format()`
+            substitutions with the following keys in scope: `line` to reference the raw
+            output line and `this` to reference this action instance.
         :param: log_cmd if True, prints the final cmd before executing the
             process, which is useful for debugging when substitutions are
             involved.
@@ -191,10 +188,7 @@ class ExecuteProcess(Action):
             LaunchConfiguration('launch-prefix', default='') if prefix is None else prefix
         )
         self.__output = output
-        # TODO(hidmic): consider implementing this using logging.LoggerAdapter instead
-        if output_prefix is None:
-            output_prefix = default_output_prefix
-        self.__output_prefix = output_prefix
+        self.__output_format = output_format
 
         self.__log_cmd = log_cmd
 
@@ -292,13 +286,17 @@ class ExecuteProcess(Action):
         self, event: ProcessIO
     ) -> Optional[SomeActionsType]:
         for line in event.text.decode().splitlines():
-            self.__stdout_logger.info(self.__output_prefix(line, self))
+            self.__stdout_logger.info(
+                self.__output_format.format(line=line, this=self)
+            )
 
     def __on_process_stderr(
         self, event: ProcessIO
     ) -> Optional[SomeActionsType]:
         for line in event.text.decode().splitlines():
-            self.__stderr_logger.info(self.__output_prefix(line, self))
+            self.__stderr_logger.info(
+                self.__output_format.format(line=line, this=self)
+            )
 
     def __on_shutdown(self, event: Event, context: LaunchContext) -> Optional[SomeActionsType]:
         return self.__shutdown_process(
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -31,6 +31,8 @@ from typing import Optional
 from typing import Text
 from typing import Tuple  # noqa: F401
 
+import launch.logging
+
 from osrf_pycommon.process_utils import async_execute_process
 from osrf_pycommon.process_utils import AsyncSubprocessProtocol
 
@@ -38,7 +40,6 @@ from .emit_event import EmitEvent
 from .opaque_function import OpaqueFunction
 from .timer_action import TimerAction
 
-from .. import logging
 from ..action import Action
 from ..event import Event
 from ..event_handler import EventHandler
@@ -373,7 +374,7 @@ class ExecuteProcess(Action):
             self.__context = context
             self.__action = action
             self.__process_event_args = process_event_args
-            self.__logger = logging.getLogger(process_event_args['name'])
+            self.__logger = launch.logging.getLogger(process_event_args['name'])
 
         def connection_made(self, transport):
             self.__logger.info(
@@ -502,11 +503,9 @@ class ExecuteProcess(Action):
         try:
             self.__completed_future = create_future(context.asyncio_loop)
             self.__expand_substitutions(context)
-            self.__logger = logging.getLogger(self.__name)
-            self.__logger.addHandler(logging.getScreenHandler())
-            self.__logger.addHandler(logging.getLogFileHandler())
+            self.__logger = launch.logging.getLogger(self.__name)
             self.__stdout_logger, self.__stderr_logger = \
-                logging.getOutputLoggers(self.__name, self.__output)
+                launch.logging.getOutputLoggers(self.__name, self.__output)
             context.asyncio_loop.create_task(self.__execute_process(context))
         except Exception:
             for event_handler in event_handlers:
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -39,10 +39,12 @@ from .timer_action import TimerAction
 from ..action import Action
 from ..event import Event
 from ..event_handler import EventHandler
+from ..event_handlers import OnProcessIO
 from ..event_handlers import OnShutdown
 from ..events import Shutdown
 from ..events.process import matches_action
 from ..events.process import ProcessExited
+from ..events.process import ProcessIO
 from ..events.process import ProcessStarted
 from ..events.process import ProcessStderr
 from ..events.process import ProcessStdin
@@ -51,6 +53,7 @@ from ..events.process import ShutdownProcess
 from ..events.process import SignalProcess
 from ..launch_context import LaunchContext
 from ..launch_description import LaunchDescription
+from ..launch_logger import LoggerLevel
 from ..launch_logger import LaunchLogger
 from ..some_actions_type import SomeActionsType
 from ..some_substitutions_type import SomeSubstitutionsType
@@ -153,11 +156,11 @@ class ExecuteProcess(Action):
         :param: prefix a set of commands/arguments to preceed the cmd, used for
             things like gdb/valgrind and defaults to the LaunchConfiguration
             called 'launch-prefix'
-        :param: output either 'log', 'screen', or 'both'; if 'screen' stderr is directed
-            to stdout and stdout is printed to the screen; if 'log' stderr is
-            directed to the screen and both stdout and stderr are directed to
-            a log file; if 'both' stdout and stderr are directed to screen and a log file;
-            the default is 'log'
+        :param: output either 'log', 'screen', or 'both'; if 'screen' stdout and
+            stderr are directed to the screen; if 'log' stderr is directed to the
+            screen and both stdout and stderr are directed to a log file; if 'both'
+            stdout and stderr are directed to screen and a log file; the default is
+           'log'
         :param: log_cmd if True, prints the final cmd before executing the
             process, which is useful for debugging when substitutions are
             involved.
@@ -265,17 +268,28 @@ class ExecuteProcess(Action):
         self._subprocess_transport.send_signal(typed_event.signal)
         return None
 
-    def __on_process_stdin_event(
+    def __on_process_input(
         self,
-        event: Event,
-        context: LaunchContext
-    ) -> Optional[LaunchDescription]:
+        event: ProcessIO
+    ) -> Optional[SomeActionsType]:
         self.__logger.warning(
             self.__name,
             "in ExecuteProcess('{}').__on_process_stdin_event()".format(id(self)),
         )
         cast(ProcessStdin, event)
         return None
+
+    def __on_process_output(
+        self, event: ProcessIO
+    ) -> Optional[SomeActionsType]:
+        name = self.process_details['name']
+        level = LoggerLevel.ERROR if event.from_stderr else LoggerLevel.INFO
+        output = '\n'.join(['---', *map(
+            lambda line: ' ' + line, event.text.decode().splitlines()
+        ), '---'])
+        self.__logger.log(name, level, 'process has {} output:\n{}'.format(
+            'stderr' if event.from_stderr else 'stdout', output
+        ))
 
     def __on_shutdown(self, event: Event, context: LaunchContext) -> Optional[SomeActionsType]:
         return self.__shutdown_process(
@@ -420,7 +434,6 @@ class ExecuteProcess(Action):
                 env=env,
                 shell=self.__shell,
                 emulate_tty=False,
-                stderr_to_stdout=(self.__output == 'screen'),
             )
         except Exception:
             self.__logger.error(name, 'exception occurred while executing process:\n{}'.format(
@@ -467,9 +480,11 @@ class ExecuteProcess(Action):
                 matcher=lambda event: is_a_subclass(event, SignalProcess),
                 entities=OpaqueFunction(function=self.__on_signal_process_event),
             ),
-            EventHandler(
-                matcher=lambda event: is_a_subclass(event, ProcessStdin),
-                entities=OpaqueFunction(function=self.__on_process_stdin_event),
+            OnProcessIO(
+                target_action=self,
+                on_stdin=self.__on_process_input,
+                on_stdout=self.__on_process_output,
+                on_stderr=self.__on_process_output,
             ),
             OnShutdown(
                 on_shutdown=self.__on_shutdown,
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -22,6 +22,7 @@ import signal
 import threading
 import traceback
 from typing import Any  # noqa: F401
+from typing import Callable
 from typing import cast
 from typing import Dict
 from typing import Iterable
@@ -36,6 +37,8 @@ from osrf_pycommon.process_utils import AsyncSubprocessProtocol
 from .emit_event import EmitEvent
 from .opaque_function import OpaqueFunction
 from .timer_action import TimerAction
+
+from .. import logging
 from ..action import Action
 from ..event import Event
 from ..event_handler import EventHandler
@@ -53,8 +56,6 @@ from ..events.process import ShutdownProcess
 from ..events.process import SignalProcess
 from ..launch_context import LaunchContext
 from ..launch_description import LaunchDescription
-from ..launch_logger import LoggerLevel
-from ..launch_logger import LaunchLogger
 from ..some_actions_type import SomeActionsType
 from ..some_substitutions_type import SomeSubstitutionsType
 from ..substitution import Substitution  # noqa: F401
@@ -67,6 +68,10 @@ from ..utilities import perform_substitutions
 
 _global_process_counter_lock = threading.Lock()
 _global_process_counter = 0  # in Python3, this number is unbounded (no rollover)
+
+
+def default_output_prefix(line: Text, action: 'ExecuteProcess') -> Text:
+    return '[{}] {}'.format(action.name, line)
 
 
 class ExecuteProcess(Action):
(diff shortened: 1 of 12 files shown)


wjwwood:
My two wishes for the user facing part of the logging example above would be to:

- drop the timestamp from the console output but leave it in the file output, by default (maybe an option to turn on timestamps in the console), and
- drop the "severity" or "log level" (e.g. `[INFO]`) from output from the subprocesses
  - perhaps achievable from the "not set" severity or by defining a new severity specifically for subprocess output that the formatters can handle in a special way

Perhaps in a second pass (different pull request) we should try to capture things like `user interrupted with ctrl-c (SIGINT)`, which is supposed to be using the logger already:

https://github.com/ros2/launch/blob/a46aabdde1f58c7e2e5c056c0c4a1ea830faac5c/launch/launch/launch_service.py#L299

hidmic:
> drop the "severity" or "log level" (e.g. [INFO]) from output from the subprocesses 

@wjwwood do you mean dropping subprocesses' log severity in the output or propagating it? If the former, wouldn't that drop information?

wjwwood:
> >    drop the "severity" or "log level" (e.g. [INFO]) from output from the subprocesses

> @wjwwood do you mean dropping subprocesses' log severity in the output or propagating it? If the former, wouldn't that drop information?

I'm not talking about changing the output from the subprocess or trying to configure it or anything, it's just that right now, every line of output from the subprocess (whether that be printf or one of our logging functions) is put through our log info function again in launch. So actually if the subprocess printed an info, then it would come out like `[INFO] ... [INFO] ... subprocess messaage`, e.g. from the above output I want `1539131115.2269616 [INFO] [listener-2] [INFO] [listener]: I heard: [Hello World: 1]` to become `[listener-2] [INFO] [listener]: I heard: [Hello World: 1]` or maybe just `[INFO] [listener]: I heard: [Hello World: 1]` though that then relies on the subprocesses to make their own output distinguishable.

wjwwood:
> Should we tackle this one in the short term now that we know SOSS wants it?

I think that makes sense.

hidmic:
Alright, taking over.

hidmic:
@wjwwood it looks like @jacobperron already applied above changes to #145.

hidmic:
I did notice however that there's an inconsistency in the `output` argument meaning for `ExecuteProcess` action and the `LaunchLogger`. 

Also, IMHO logging like this makes it very hard to collect process output if actually meaningful i.e. not just logs. So... I have a proposal :). 

Say we add an optional `output_format` argument to the `ExecuteProcess` action to allow the launch API user to provide an equivalent format string to that used by the launched process for its own logging e.g. `'[{severity}] [{name}] {message}'` or `'--{severity}@{name}-- {message}'`. With that information, we then attempt to parse the data _back_ from each stdout/stderr line we get: from `--INFO@exec-- some info` we go to `{'severity': 'INFO', 'name': 'exec', 'message': 'some info'}` (you've probably noticed by now that we'll have to put some rules for attribute matching in place). If we succeed, we forward a `LogRecord` to the logger such that:

```
[listener-2] [INFO] [listener]: I heard: [Hello World: 1] 
[listener-2] [WARN] [listener]: Something's fishy in here
```

becomes:

```
1539131115.2269616 [INFO] [listener-2.listener] I heard: [Hello World: 1]
1539131115.2269616 [WARN] [listener-2.listener] Something's fishy in here
```

Lines that do not match we collate them into stdout/stderr blocks, something like:

```
1539131115.2269616 [INFO] [listener-2] process has stdout output:
---
 hello!
 argh!?
---
1539131116.2269616 [INFO] [listener-2.listener] I heard: [Hello World: 1]
```

That way we could deal, upon user request, with any kind of logging system. Moreover, with that in place we can add some nice heuristics e.g. if `output_format == 'ros'` we lookup `RCUTILS_CONSOLE_FORMAT` in the process environment, falling back to `rcutils` default format string otherwise. Making such a heuristic the default for `Node` actions would likely take care of most ROS launch use cases. Thoughts @wjwwood @jacobperron ?

jacobperron:
@hidmic What is the use-case for having a different output formats between processes? I can see value in letting users specify a format for the logger as a whole, but is a mixed-format log worth the effort? IMO, I'd like a consistent format in a single log file to make parsing more straight forward.

It might make sense to direct output from different processes (or subsets of processes) into separate log files. And log format could be specified per log file (or screen).

hidmic:
@jacobperron I'm not sure I follow you. I'd expect all ROS-aware processes to log pretty much the same way, yeah. Nothing to change there. But `launch` can execute processes that are not ROS-aware nor use `rcutils`. It could even be the case that some processes are third party and one's not willing or not able to change the logs format. Thus, the `output_format` argument for them. Meanwhile, all ROS aware processes can use the `rcutils` heuristic described above.

This is consistent with your idea of having a single kind of log format (that of launch) to ease parsing. Regardless of what you're running (e.g. a rails server instance? sure! :)). 

Sure, we can also preclude support for all possible cases and just implement this logic for the `Node` action using `rcutils` logging case. That'd be fine too I think.

jacobperron:
@hidmic Sorry, I misunderstood your proposal. The optional `output_format` argument for non-ROS processes makes sense to me :)

hidmic:
@wjwwood Alright, that's a very explicit and strong take on how this should look like! :) 

> I'm not convinced about having to parse the output from all the subprocesses, aside from performance concerns, I'm worried about the transparency for the user (assuming the runner of a launch file may have not written it), and I'd prefer to leave the output from the subprocesses as unmodified as possible.

The idea is to allow the user to make that call himself/herself instead i.e. by not specifying an output format. Concerns about performance are perfectly valid.

> Then if they do not match the pattern, then they're being printed in a different pattern (block output you were talking about), which is inconsistent

The rationale is to make it look like log forwarding across process boundaries. If you look at the examples above, parsed log output gets its logger name prepended with the process name using the same hierarchical naming convention that the `logging` module uses. And since you are effectively working with child loggers of the process action logger, you can use all the `logging` filtering and handling functionality.

But still, it's a fair objection.

> I find the suggestion of output blocks to be too much additional markup to be nice to read

One of the reasons why I thought the block output may be practical is if you want to copy or even post-process the output. Having prefixes in the middle is painful. So, I'm OK with optionally dropping prefixes at the expense of disambiguation, which I do think is inconvenient but one could argue that if you want to use the output of several subprocess, you may as well dump it into a file.

> this is the style used in catkin_tools when streaming compiler output

Hmm, I don't think that's a comparable use case. Aside from helping one figure out why a build broke, compiler output is not usually post-processed or used for anything (at least not that I'm aware of, maybe in some testing setup?).

> still allows you to capture subprocess outputs separately, or even their respective stdout and stderr separately if you'd like, or you can choose not to and let the process log itself to a file

That's a fair solution to the problem of disambiguation I have to agree with.

-----

I think my approach was a bit more oriented towards _everything is a log_, instead of focusing on stdout/stderr regardless of what's coming out. It's fine to go with the latter. And yes, it probably scales better with launched process count.

So let me get this straight, you'd want:

1. [x] Optional timestamps for screen logs.
2. [x] Optional log file for all subprocesses output combined, with timestamps.
3. [ ] Optional `output_prefix` for subprocess output logs including the process name and maybe output origin as well (not yet optional nor can include output origin).
4. [ ] Optional log files on a per subprocess basis: one for stdout, one for stderr and one for their combination.

(2) is right now generated if at least one subprocess has `'log'` output. How do you want to disambiguate it from (4)? Another output type? A launch service wide setting for (2) instead? Or should we just generate both? The latter sounds a bit too much, but I don't have a strong opinion on this regard really.

If that's OK, I'll rebase and update these PRs to do (3) and (4) too.

wjwwood:
> Alright, that's a very explicit and strong take on how this should look like! :)

As I said, it's ideal for me, but that need not be the final say. I was just trying to clearly articulate my vision for it and why I think it is a good direction. :)

> The idea is to allow the user to make that call himself/herself instead i.e. by not specifying an output format.

Yeah, but the user can do that at various stages, either in the process itself, or after launch has run, or by writing their own launch event handler to do so. I don't see that we need to facilitate that for them at this time, but I could be convinced otherwise.

> The rationale is to make it look like log forwarding across process boundaries.

I understand the idea behind it, but parsing and the mutating stdout isn't the best route imo. Instead, `/rosout` or an equivalent is probably better, where you're given the "raw" log information as a "struct" and you can construct a string using that as you like.

From my perspective, this feature is about capturing all stdout/stderr without assuming what the content of the data is.

> If you look at the examples above, parsed log output gets its logger name prepended with the process name using the same hierarchical naming convention that the logging module uses. And since you are effectively working with child loggers of the process action logger, you can use all the logging filtering and handling functionality.

I did see that, but I don't think it's a great idea, mostly because this new "logger" (e.g. `listener-2.listener`) doesn't actually exist, so if you looked for it in `/rosout` or tried to configure it with log4cxx maybe, or tried to do it with env vars to the process, then it wouldn't work. I don't actually consider logging in the subprocess as a child of the launch logger, but that's just how I see it.

> One of the reasons why I thought the block output may be practical is if you want to copy or even post-process the output. Having prefixes in the middle is painful. So, I'm OK with optionally dropping prefixes at the expense of disambiguation, which I do think is inconvenient but one could argue that if you want to use the output of several subprocess, you may as well dump it into a file.

The prefixes don't see that bad to me, especially since you can expect them for all lines (either something like `[INFO]` for launch logging, or `[<process>]` for all subprocesses), but if you have trouble with them, then instead you should probably post process the split output from each process separately, which would not contain anything but what the process produced itself.

> Hmm, I don't think that's a comparable use case. Aside from helping one figure out why a build broke, compiler output is not usually post-processed or used for anything (at least not that I'm aware of, maybe in some testing setup?).

I think it's valuable from the perspective that it works well for humans that are watching the thing run in real-time and for them to be able to understand what's going on. I wasn't thinking it was a good example for post processing, but that being said I don't think it would be hard to parse it in post processing.

>  - Optional log file for all subprocesses output combined, with timestamps.

Not exactly, it would actually be all logging output from launch, including combined output from subprocess which were configured to print to the screen. Basically anything that passes through the Python logging calls within launch (as opposed to things printed with `print()`, which hopefully at this point we're minimizing within launch itself).

> - Optional output_prefix for subprocess output logs including the process name and maybe output origin as well (not yet optional nor can include output origin).

Yeah, that sounds fine, though it should probably have a default, and you could implement it as a transfer function rather than just a prefix, e.g. takes a callback with a signature like: `def callback(line: str, process_action: ExecuteProcess, context: LaunchContext) -> str`.

That would prevent us from needing a substitution framework (e.g. `prefix='[{process_name}] `), and it would leave the door open to more sophisticated output processing like you described.

> (2) is right now generated if at least one subprocess has 'log' output. How do you want to disambiguate it from (4)? Another output type? A launch service wide setting for (2) instead? Or should we just generate both? The latter sounds a bit too much, but I don't have a strong opinion on this regard really.

As I see it, there are really four considerations for each process:

- should it's stdout and stderr be forwarded to the console (and therefore to launch's main log file)?
  - I think this is a per action setting
- if printed to console, should the output be prefixed?
  - This could be a global (i.e. for all processes that print to the console, use this prefix) or per action for me, not sure what's best
- should it's stdout and stderr be logged to its own file?
  - I think this should be per action
- if stdout and stderr is being logged separately, should it be combined, separated, or both?
  - I think this should be per action too

`colcon` generates everything by default (combined, separate, etc. and optionally prints to screen in various ways), so I don't know if it would be too much as a default to capture launch's output in one file, and all stdout/stderr from each subprocess each in their own file as well.

As for how to actually configure it, I'm not sure. It could be one a few choices as peers to the existing `log` and `screen` options, or it could be separate configurations, allowing for any combination without having an option for each combinatorial.

hidmic:
Alright! Finally found the time to wrap up. Things have changed quite a bit i.e. from singleton wrapping the `logging` module to an extension of the `logging` module itself, but I think I've covered all the features described above and solved all pending TODOs. I spent some time thinking how to allow easy configuration. What you see is the result of weighing several options, hope you find it reasonable.

Output of @jacobperron example now shows:

*On screen*
```
[INFO] [talker-1]: process started with pid [22214]
[INFO] [listener-2]: process started with pid [22215]
[talker-1] [INFO] [talker]: Publishing: 'Hello World: 1'
[talker-1] [INFO] [talker]: Publishing: 'Hello World: 2'
[talker-1] [INFO] [talker]: Publishing: 'Hello World: 3'
[talker-1] [INFO] [talker]: Publishing: 'Hello World: 4'
[talker-1] [INFO] [talker]: Publishing: 'Hello World: 5'
^C[WARNING] [launch]: user interrupted with ctrl-c (SIGINT)
[talker-1] [INFO] [rclcpp]: signal_handler(signal_value=2)
[INFO] [talker-1]: process has finished cleanly [pid 22214]
[INFO] [listener-2]: process has finished cleanly [pid 22215]
```

*On file*

```
1550866401.8164542 [INFO] [talker-1]: process started with pid [22214]
1550866401.8168678 [INFO] [listener-2]: process started with pid [22215]
1550866402.8387372 [listener-2] [INFO] [listener]: I heard: [Hello World: 1]
1550866403.8386636 [listener-2] [INFO] [listener]: I heard: [Hello World: 2]
1550866404.8386359 [listener-2] [INFO] [listener]: I heard: [Hello World: 3]
1550866405.8386660 [listener-2] [INFO] [listener]: I heard: [Hello World: 4]
1550866406.8386412 [listener-2] [INFO] [listener]: I heard: [Hello World: 5]
1550866407.2243264 [WARNING] [launch]: user interrupted with ctrl-c (SIGINT)
1550866407.8163753 [INFO] [talker-1]: process has finished cleanly [pid 22214]
1550866407.8166566 [INFO] [listener-2]: process has finished cleanly [pid 22215]
1550866407.8168473 [listener-2] [INFO] [rclcpp]: signal_handler(signal_value=2)
```

This example however is not exercising all the other, newer options.

launch/launch/actions/execute_process.py
```diff
@@ -180,15 +189,12 @@ def __init__(
...
-        allowed_output_options = ['log', 'screen']
-        if self.__output not in allowed_output_options:
-            raise ValueError(
-                "output argument to ExecuteProcess is '{}', expected one of [{}]".format(
-                    self.__output,
-                    allowed_output_options,
-                )
-            )
+        self.__output = output
+        # TODO(hidmic): consider implementing this using logging.LoggerAdapter instead
```
wjwwood: What's the reason to not do it now? Time, complexity, or not sure if it's the right thing to do? I don't mind deferring it, I'm just curious.

launch/launch/actions/execute_process.py
```diff
@@ -37,13 +37,17 @@
 from .emit_event import EmitEvent
 from .opaque_function import OpaqueFunction
 from .timer_action import TimerAction
+
+from .. import logging
```
wjwwood: This is a bit confusing. i kept thinking you were calling the python logging module below. I'd recommend either renaming this module (locally with `as ....` or at the module itself), or consider using non-relative import for this particular case, i.e. `launch.logging.*` everywhere.

launch/launch/actions/execute_process.py
wjwwood: My preference would be a non-relative import, as the alternative names I'd suggest are just as long, i.e. `launch.logging.*` over `launch_logging.*`.

launch/launch/launch_service.py
```diff
@@ -89,8 +89,9 @@ def __init__(
         self.__debug = debug
 
         # Setup logging
-        self.__logger = logging.getLogger('launch.LaunchService')
-
+        self.__logger = logging.getLogger('launch')
```
wjwwood: Why did you change the logger name here? Just to be shorter? I'm not necessarily for one or the other, I'm just curious because it seems aesthetic.

launch/launch/utilities/signal_management.py
```diff
@@ -143,6 +142,9 @@ def __on_sigterm(signum, frame):
             # Windows does not support SIGQUIT
             signal.signal(signal.SIGQUIT, __on_sigquit)
     except ValueError:
-        _logger.error("failed to set signal handlers in 'launch.utilities.signal_management.py'")
-        _logger.error('this function must be called in the main thread')
+        logger = logging.getLogger(__name__)
```
wjwwood: Is this the same as it was before? (`launch.utilities.signal_management`)

launch/launch/utilities/signal_management.py
```diff
@@ -143,6 +142,9 @@ def __on_sigterm(signum, frame):
...
+        logger.error(
+            "failed to set signal handlers in 'launch.utilities.signal_management.py'",
```
wjwwood: Why is this multi line is it longer now? Was it too long before? Why does it have a trailing `,`?

hidmic:
@wjwwood this PR was originally done against #145. The missing module is there.

wjwwood:
@hidmic sorry, it was long enough ago that I forgot that we split them up. I should have dug more, but thanks for the pointer :)

launch/launch/actions/execute_process.py
hidmic: > What's the reason to not do it now?
I was unsure if that's what you wanted. Right now we're using a transfer callback w/ no launch context. Using the logger adapter would instead provide a substitution framework, with it's advantages and restrictions.

> My preference would be a non-relative import

Non-relative import it is then.

launch/launch/actions/execute_process.py
hidmic: Non-relative import it is then.

launch/launch/launch_service.py
hidmic: Purely aesthetic :)

launch/launch/utilities/signal_management.py
hidmic: So, `__name__` seems to be populated w.r.t. the `__main__` script. A script would normally include `launch`, that may in turn include `.utilities.signal_management` and thus the full name will `launch.utilities.signal_management`.

launch/launch/utilities/signal_management.py
hidmic: It got changed at some point before I made these changes. I'll restore it.

launch/launch/actions/execute_process.py
wjwwood: Weird, I posted the comment in the wrong place 🙃, sorry for the confusion.

launch/launch/actions/execute_process.py
wjwwood: Sounds good.

launch/launch/launch_service.py
wjwwood: That's fine, just curious.

launch/launch/utilities/signal_management.py
wjwwood: My main concern is that if launch is embedded into another Python project or something like that, then `__name__` might change, making the logger name unpredictable. I have a slight preference for the hard coded string over `__name__`, but I'll leave it up to you. I don't consider this blocking.

launch/launch/actions/execute_process.py
wjwwood: Re: the TODO

I don't mind, but I would like to remove or rephrase the TODO so it has a clear "do X after condition Y". Otherwise it will be here forever most likely.

hidmic: Alright, I'll take a shot at substitutions.

launch/launch/utilities/signal_management.py
hidmic: Hmm, that's true. Though it's not strictly unpredictable, if one embeds `launch` in a e.g. `hate_pypi` package, the logger will be named after the module the action is in. That changed. Because `launch` became a module in a package.

------

Honestly, I'd like to provide a hint as to which action _instance_ is actually logging what instead. But for a first shot I think I'm happy with this.

launch/launch/actions/execute_process.py
hidmic: Alright, not even a `LoggerAdapter`. This is much much simpler and will probably do for a long while. I thought at some point of enabling arbitrary output preprocessing and filtering but it was overkill.

hidmic:
Merging into the main PR to ease reviewing.

PR #154, opened 2018-11-13
Ensure that shutdown event is handled correctly

mjcarroll:
There is a potential race condition in bewteen when the shutdown event
is emitted and the rest of the shutdown handling code.  This introduces
an additional await to ensure that the event is emitted before
proceeding.

### DIFF
--- a/test_launch_ros/test/test_launch_ros/actions/test_node.py
+++ b/test_launch_ros/test/test_launch_ros/actions/test_node.py
@@ -57,7 +57,7 @@ class TestNode(unittest.TestCase):
     def test_launch_invalid_node(self):
         """Test launching an invalid node."""
         node_action = launch_ros.actions.Node(
-            package='nonexistent_package', node_executable='node', output='screen'),
+            package='nonexistent_package', node_executable='node', output='screen')
         self._assert_launch_errors([node_action])
 
     def test_launch_node(self):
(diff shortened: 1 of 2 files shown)


mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5555/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2202/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4577/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5438/)

test_launch_ros/test/test_launch_ros/actions/test_node.py
```diff
@@ -57,7 +57,7 @@ def _assert_type_error_creating_node(self, *, parameters=None, remappings=None):
     def test_launch_invalid_node(self):
         """Test launching an invalid node."""
         node_action = launch_ros.actions.Node(
-            package='nonexistent_package', node_executable='node', output='screen'),
+            package='nonexistent_package', node_executable='node', output='screen')
```
wjwwood: We think this was always causing a subtle bug (`node_action` was a single item tuple rather than just a single item), but was being hidden because the tests didn't complete due to the race condition being fixed in this pull request.

mjcarroll:
And the windows rebuild after umlaut-gate: [Build Status](https://ci.ros2.org/job/ci_windows/5444/)

PR #156, opened 2018-11-23
signit executes the shutdown event synchronously

joncppl:
Recently my launch scripts have been hanging on signint (control-c).
After a little digging it seems to exist since https://github.com/ros2/launch/pull/154 .

Having the signint-triggered shutdown event execute synchronously fixes the hanging.
This makes sense to me, although I admit my understanding of Python's async is rather limited.

### DIFF
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -299,7 +299,7 @@ class LaunchService:
                 base_msg = 'user interrupted with ctrl-c (SIGINT)'
                 if not sigint_received:
                     _logger.warn(base_msg)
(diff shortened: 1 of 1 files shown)


mjcarroll:
I believe that this makes sense, given the changes that I introduced.  I'm also going to get @wjwwood to take a look to make sure that it makes sense to him as well.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5670/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2277/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4687/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5562/)

PR #165, opened 2019-01-17
Adds example tests for launch descriptions

### LINKED ISSUE #166: [launch] Propose launch-based testing API through example
## Feature request

Propose an API for launch-based testing through user stories.

#### Feature description

Write down testing API use examples to:

- Test an existing launch file through composition.
- Run a gtest/pytest program in test fixture context.

In addition, all legacy launch testing use cases, as can be found in [ros2/launch](https://github.com/ros2/launch) and [ros2/demos](https://github.com/ros2/demos), must be captured.

hidmic:
Closes #166. Still a work in progress, but it should help getting the gist of it. I _think_ something like this should cover most if not all use cases moving forward. Please, shove this PR with some more use cases if you have them.

### DIFF
--- a/launch_testing/examples/test_pub_sub_launch.py
+++ b/launch_testing/examples/test_pub_sub_launch.py
@@ -1,0 +1,230 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Some example tests around ROS 2 talker/listener nodes.
+
+.. note::
+
+To maximize code reuse, this test assumes that execution and executable
+abstractions are decoupled (see https://github.com/ros2/launch/issues/114).
+"""
+
+import sys
+import pytest
+
+from launch import LaunchDescription
+from launch.actions import Execute
+from launch.events.process import ProcessStarted
+from launch.launch_description_sources import \
+    get_launch_description_from_python_launch_file
+from launch_ros import get_default_launch_description
+from launch_ros.executables import Node
+from launch_ros.executables import ComposableNode
+from launch_ros.executables import ComposableNodeProcess
+import launch_ros.events.lifecycle as lifecyle
+
+from launch_testing import TestLaunchService
+from launch_testing.actions import PyTest
+from launch_testing.actions import GTest
+from launch_testing.actions import Assert
+from launch_testing.actions import AssertSequenceOnce
+
+import launch_testing.predicates as predicates
+import launch_testing.variables as variables
+
+
+@pytest.fixture
+def ros_test_launch_service() -> TestLaunchService:
+    """
+    A test fixture providing a test-aware launch service (i.e. one
+    that listens to events fired by `launch_testing.actions.Assert`
+    and `launch_testing.actions.Test` actions and reacts accordingly).
+    """
+    ls = TestLaunchService()
+    # TODO(hidmic): consider integrating ROS specific actions and the
+    # launch service in some way to ease launch file composition without
+    # repetition).
+    ls.include_launch_description(
+        get_default_launch_description(prefix_output_with_name=False)
+    )
+    return ls
+
+
+@pytest.fixture
+def pub_sub_launch_description() -> LaunchDescription:
+    """
+    A test fixture providing the basic pub/sub launch description.
+    """
+    # TODO(hidmic): implement functionality to lookup launch files
+    # on a per package basis
+    launch_path = get_launch_path(
+        package_name='launch_ros', launch_file_path='pub_sub_launch.py'
+    )
+    return get_launch_description_from_python_launch_file(launch_path)
+
+
+def test_pub_sub_launch(ros_test_launch_service, pub_sub_launch_description):
+    """
+    Tests an existing simple pub/sub system.
+    """
+    # Launches a pytest that also happens to be a node with
+    # a 30s timeout.
+    pub_sub_launch_description.add_action(
+        # TODO(hidmic): allow launching non-installed
+        # executables within the current package.
+        PyTest(Node(
+            node_executable='pub_test.py',
+            remappings=[('chatter', 'my_chatter')]
+        ), timeout=30)
+    )
+    # Launches a gtest that also happens to be a node with
+    # a 30s timeout.
+    pub_sub_launch_description.add_action(
+        GTest(Node(
+            node_executable='sub_test',
+            remappings=[('chatter', 'my_chatter')]
+        ), timeout=30)
+    )
+
+    ros_test_launch_service.include_launch_description(pub_sub_launch_description)
+
+    # TODO(hidmic): implement launch_testing specific pytest plugin to aggregate
+    # all test result information at the launch system-level and below.
+    assert ros_test_launch_service.run() == 0
+
+
+@pytest.fixture
+def lifecycle_pub_sub_launch_description() -> LaunchDescription:
+    """
+    A test fixture providing the lifecycle pub/sub launch description.
+    """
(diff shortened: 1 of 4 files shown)


hidmic:
BTW don't even try to run this, it won't work :)

launch_testing/examples/test_pub_sub_launch.py
```diff
@@ -0,0 +1,230 @@
...
+
+def test_intra_comm_pub_sub(ros_test_launch_service):
+    """
+    Tests a intra process communication between composable
+    pub/sub nodes via gtest.
+    """
+    # Launches a gtest that happens to be a container
+    # for composable nodes.
+    ld = LaunchDescription([
+        GTest(ComposableNodeProcess(
```
dirk-thomas: I don't see how a composable node process can be a gtest. Which part is actually a gtest in this configuration?

launch_testing/examples/test_pub_sub_launch.py
```diff
@@ -0,0 +1,230 @@
...
+def test_pub_sub_launch(ros_test_launch_service, pub_sub_launch_description):
+    """
+    Tests an existing simple pub/sub system.
+    """
+    # Launches a pytest that also happens to be a node with
+    # a 30s timeout.
+    pub_sub_launch_description.add_action(
+        # TODO(hidmic): allow launching non-installed
+        # executables within the current package.
+        PyTest(Node(
```
dirk-thomas: I don't think `Node` is the write argument for `PyTest` (or `GTest` below). The actual Python script containing tests or the executable containing a gtest main function are not necessarily nodes.

launch_testing/examples/test_pub_sub_launch.py
hidmic: Here's the use case I came up with. 

Say you want to test a pointcloud postprocessing pipeline that runs in a single process through composable nodes for improved performance. Testing from outside the process may yield different results, specially if the pipeline is time critical. So the test has to be run from within the process.

In that scenario, I envisioned a container process that's actually a gtest with a fixture that takes care of pipeline setup and teardown. With that in mind composition (actually, it's closer to decoration) instead of duplication makes sense because the two traits of the executable (being a gtest and being composable node) are orthogonal.

launch_testing/examples/test_pub_sub_launch.py
hidmic: That's exactly the point I want to make. Both the pytest script and the gtest executable could be nodes, containers of composable nodes, plain ROS-agnostic programs, etc. So we'd want to reuse what we have in terms of doing file lookups, argument passing, output handling, etc. If you think about it, all the `PyTest` and `GTest` actions have to do is execute whatever is being passed in a single host process (for starters) assuming it's what we say it is, pass specific command line arguments and collect whatever output it's expected the executable will generate. 

So above, where it says `Node`, it could say `Process` (i.e. any non ROS aware executable).

dirk-thomas: > So above, where it says `Node`, it could say `Process` (i.e. any non ROS aware executable).

Maybe I am missing how this is supposed to work.

A Python script being a node will have a `__main__` which get the node running. On the other hand a Python file containing unit tests invoked by `pytest` doesn't. Also such a test file wouldn't be represented by `Process` (since it doesn't have a `__main__`). In fact a Python test file can't even be invoked - it commonly doesn't have an executable flag. So e.g. `node_executable` doesn't apply here - neither can you pass command line arguments of any kind.

wjwwood: Specifically for pytest, I think Dirk might be right. For gtest, I do think it makes sense to have an argument that inherits from the proposed `Executable` or `ExecuteProcess` (as `Node` does).

hidmic: >A Python script being a node will have a __main__ which get the node running. On the other hand a Python file containing unit tests invoked by pytest doesn't. Also such a test file wouldn't be represented by Process (since it doesn't have a __main__).

That's mostly true, though not strictly. The following script can be made executable to run all the tests in it:

```python
#! /usr/bin/env python                                                          
                                                                                
import os                                                                       
import pytest                                                                   
                                                                                
                                                                                
def test_stuff():                                                               
    assert False                                                                
                                                                                
                                                                                
pytest.main([os.path.realpath(__file__)])
```

But, set aside above's unusual setup, you're right about `pytest` test files not being executable in general. And we do have a problem with a command line arguments, though [not insurmountable](https://docs.pytest.org/en/latest/reference.html#_pytest.hookspec.pytest_addoption).

--------

The initial idea was to use the fact that the testing frameworks 

So, as I see it, this leaves us with either of two options (that I'm going to sketch for both `gtest` and `pytest` as I'd rather have a consistent API):

1. We derive a _test_ version of each executable kind we have and we think of testing frameworks as testing protocols (e.g. how to execute, how to collect results, etc.):

  ```python
  NodeTest(
       path='path/to/test.py', protocol='pytest', 
       timeout=20, remapping=[('chatter', 'my_chatter')]
   ),
  ComposableNodeTest(
       container_executable='intra_proc_comm_test_container',
       composable_node_descriptions=[
           ComposableNode(
               package_name='composition',
               node_plugin_name='talker_component',
               name='my_talker_component'
           ),
           ComposableNode(
               package_name='composition',
               node_plugin_name='listener_component'
               name='my_listener_component'
           ),
        protocol='gtest', timeout=10
     ]
  )
  ```
  
   
2. We outright ban test parameterization at the launch level.

  ```python
  PyTest(path='path/to/test.py', timeout=10),
  GTest(path='path/to/gtest', timeout=10)
  ```

(1) is somewhat complex, but it enables full launch API reuse. (2) is easy to use and to implement, though a bit limiting in my opinion.

I guess the question is whether we want to provide all the tools to parameterize tests or if we just to want to cover the basic use cases and then leave the users on their own to come up with custom ways of doing so.

launch_testing/examples/test_pub_sub_launch.py
```diff
@@ -0,0 +1,230 @@
...
+    # once (40s total timeout).
+    lifecycle_pub_sub_launch_description.add_action(
+        AssertSequenceOnce([
+            variables.EventEmitted(
+                event_type=lifecycle.ChangeState
+            ),
+            variables.EventEmitted(
+                event_type=ProcessStarted
+            )
+        ], timeout=[30, 10])
```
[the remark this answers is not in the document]
hidmic: > This is interesting, and I like it, but the the generic name really needs a comment

Sure. I'm open to any naming changes. How would you briefly state that you're asserting that N conditions occur in the right sequence/order?

> Also, would it be valuable to ensure that the events came from a specific process?

For sure. Don't take these examples as a completely exhaustive demonstration of all the (future) API features. In this case, we could as easily add some matching to only check for the source (i.e. the executable) of interest.

> In which case, would it make sense to couple this assertion to the thing being tested?

I did play with this idea. It wasn't clear to me either how to deal with multiple assertions of different kinds, potentially targeting multiple actions. That's why I dropped it in favor of using the target action instances directly in the predicates being asserted. We could have something similar to what you describe though, by deferring variable binding. E.g. instead of:

```python
Assert(len(predicates.regex_match(
    variables.Output(target), pattern='^Oops!.*'
)) > 10, timeout=10)
```

we could write:

```python
Test(target=Node(), description=[
    Assert(len(predicate.regex_match(
        variables.Output(), pattern='Oops!.*'
    )) > 10, timeout=10)
])
```

It may get somewhat confusing though if you start mixing both on the same launch file. The former is the most general one, so I don't think we can replace it with the latter.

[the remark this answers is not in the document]
wjwwood: > Sure. I'm open to any naming changes. How would you briefly state that you're asserting that N conditions occur in the right sequence/order?

I don't have an alternative in mind, but I'll keep thinking on it.

The rest makes sense to me, you can realize those cases when/if we need them.

launch_testing/examples/test_pub_sub_launch.py
dirk-thomas: At the moment I am not aware of any use case where we need `ComposableNodeTest`. In the future there could certainly be test using that approach. For now I would rather focus on the cases which we do have a use case for.

hidmic: Alright, went with the simple one.

hidmic:
Future of this sample API depends on we go about other alternatives that are being proposed concurrently.

hidmic:
Alright, it seems we won't be pursing this route and thus I'll close it for now. We can resurrect it if necessary.

Issue #159, opened 2019-01-08
Remove legacy API


sloretz:
The legacy API should be removed.

<details><summary>Places where legacy API is used</summary>

```
./ros2/system_tests/test_cli_remapping/test/test_cli_remapping.py:from launch.legacy import LaunchDescriptor
./ros2/system_tests/test_cli_remapping/test/test_cli_remapping.py:from launch.legacy.exit_handler import primary_exit_handler
./ros2/system_tests/test_cli_remapping/test/test_cli_remapping.py:from launch.legacy.launcher import DefaultLauncher
./ros2/rcutils/test/test_logging_long_messages.py:from launch.legacy import LaunchDescriptor
./ros2/rcutils/test/test_logging_long_messages.py:from launch.legacy.exit_handler import ignore_exit_handler
./ros2/rcutils/test/test_logging_long_messages.py:from launch.legacy.launcher import DefaultLauncher
./ros2/rcutils/test/test_logging_long_messages.py:from launch.legacy.output_handler import ConsoleOutput
./ros2/rcutils/test/test_logging_output_format.py:from launch.legacy import LaunchDescriptor
./ros2/rcutils/test/test_logging_output_format.py:from launch.legacy.exit_handler import ignore_exit_handler
./ros2/rcutils/test/test_logging_output_format.py:from launch.legacy.launcher import DefaultLauncher
./ros2/rcutils/test/test_logging_output_format.py:from launch.legacy.output_handler import ConsoleOutput
./ros2/launch/launch_testing/launch_testing/legacy/__init__.py:from launch.legacy.output_handler import LineOutput
./ros2/launch/launch_testing/launch_testing/__init__.py:from launch_testing.legacy import create_handler  # noqa: F401
./ros2/launch/launch_testing/launch_testing/__init__.py:from launch_testing.legacy import get_default_filtered_patterns  # noqa: F401
./ros2/launch/launch_testing/launch_testing/__init__.py:from launch_testing.legacy import get_default_filtered_prefixes  # noqa: F401
./ros2/launch/launch_testing/launch_testing/__init__.py:from launch_testing.legacy import get_rmw_output_filter  # noqa: F401
./ros2/launch/launch_testing/launch_testing/__init__.py:from launch_testing.legacy import InMemoryHandler  # noqa: F401
./ros2/launch/launch_testing/launch_testing/__init__.py:from launch_testing.legacy import UnmatchedOutputError  # noqa: F401
[...]

PR #167, opened 2019-01-29
Migrates legacy launch API tests

hidmic:
Connected to #159. This pull request adds the bare minimum functionality to migrate all tests in ROS 2 core packages that make use of the legacy launch API. May eventually face deprecation depending on the path taken after #165. 

Some `launch.legacy` tests still remain to be ported, unclear if these should in the first place as such additions may as well fall under the improving test coverage for `launch` umbrella.

### DIFF
--- a/launch/launch/action.py
+++ b/launch/launch/action.py
@@ -56,7 +56,21 @@ class Action(LaunchDescriptionEntity):
     def visit(self, context: LaunchContext) -> Optional[List[LaunchDescriptionEntity]]:
         """Override visit from LaunchDescriptionEntity so that it executes."""
         if self.__condition is None or self.__condition.evaluate(context):
-            return cast(Optional[List[LaunchDescriptionEntity]], self.execute(context))
+            try:
+                return cast(Optional[List[LaunchDescriptionEntity]], self.execute(context))
+            finally:
+                from .events import ExecutionComplete  # noqa
+                future = self.get_asyncio_future()
+                if future is not None:
+                    future.add_done_callback(
+                        lambda _: context.emit_event_sync(
+                            ExecutionComplete(action=self)
+                        )
+                    )
+                else:
+                    context.emit_event_sync(
+                        ExecutionComplete(action=self)
+                    )
         return None
 
     def execute(self, context: LaunchContext) -> Optional[List['Action']]:
--- a/launch/launch/action.py
+++ b/launch/launch/action.py
@@ -60,17 +60,15 @@ class Action(LaunchDescriptionEntity):
                 return cast(Optional[List[LaunchDescriptionEntity]], self.execute(context))
             finally:
                 from .events import ExecutionComplete  # noqa
-                future = self.get_asyncio_future()
-                if future is not None:
-                    future.add_done_callback(
-                        lambda _: context.emit_event_sync(
-                            ExecutionComplete(action=self)
+                event = ExecutionComplete(action=self)
+                if context.would_handle_event(event):
+                    future = self.get_asyncio_future()
+                    if future is not None:
+                        future.add_done_callback(
+                            lambda _: context.emit_event_sync(event)
                         )
-                    )
-                else:
-                    context.emit_event_sync(
-                        ExecutionComplete(action=self)
-                    )
+                    else:
+                        context.emit_event_sync(event)
         return None
 
     def execute(self, context: LaunchContext) -> Optional[List['Action']]:
--- a/launch/launch/actions/opaque_coroutine.py
+++ b/launch/launch/actions/opaque_coroutine.py
@@ -25,7 +25,10 @@ from typing import Optional
 from typing import Text
 
 from ..action import Action
+from ..event import Event
+from ..event_handlers import OnShutdown
 from ..launch_context import LaunchContext
+from ..some_actions_type import SomeActionsType
 from ..utilities import ensure_argument_type
 
 
@@ -56,9 +59,11 @@ class OpaqueCoroutine(Action):
         """Constructor."""
         super().__init__(**left_over_kwargs)
         if not asyncio.iscoroutinefunction(coroutine):
-            raise TypeError("OpaqueCoroutine expected a couroutine function for 'couroutine', got '{}'".format(
-                type(function)
-            ))
+            raise TypeError(
+                "OpaqueCoroutine expected a couroutine for 'couroutine', got '{}'".format(
+                    type(coroutine)
+                )
+            )
         ensure_argument_type(
             args, (collections.abc.Iterable, type(None)), 'args', 'OpaqueCoroutine'
         )
@@ -74,6 +79,12 @@ class OpaqueCoroutine(Action):
         self.__ignore_context = ignore_context  # type: bool
         self.__future = None  # type: Optional[asyncio.Future]
 
+    def __on_shutdown(self, event: Event, context: LaunchContext) -> Optional[SomeActionsType]:
+        """Cancel ongoing coroutine upon shutdown."""
+        if self.__future is not None:
+            self.__future.cancel()
+        return None
+
     def execute(self, context: LaunchContext) -> Optional[List[Action]]:
         """Execute the action."""
         args = self.__args
@@ -81,6 +92,9 @@ class OpaqueCoroutine(Action):
             args = [context, *self.__args]
         self.__future = context.asyncio_loop.create_task(
             self.__coroutine(*args, **self.__kwargs)
+        )
+        context.register_event_handler(
+            OnShutdown(on_shutdown=self.__on_shutdown)
         )
         return None
 
--- a/launch/launch/actions/opaque_coroutine.py
+++ b/launch/launch/actions/opaque_coroutine.py
@@ -41,11 +41,23 @@ class OpaqueCoroutine(Action):
     .. code-block:: python
 
         async def coroutine(
-            context: LaunchContext,  # iff ignore_context is False
+            context: LaunchContext,
             *args,
             **kwargs
         ):
             ...
+
+    if ignore_context is False on construction (currently the default), or
+
+    .. code-block:: python
+
+        async def coroutine(
+            *args,
+            **kwargs
+        ):
+            ...
+
+    if ignore_context is True on construction.
     """
 
     def __init__(
@@ -60,7 +72,7 @@ class OpaqueCoroutine(Action):
         super().__init__(**left_over_kwargs)
         if not asyncio.iscoroutinefunction(coroutine):
             raise TypeError(
-                "OpaqueCoroutine expected a couroutine for 'couroutine', got '{}'".format(
+                "OpaqueCoroutine expected a coroutine for 'coroutine', got '{}'".format(
                     type(coroutine)
                 )
             )
--- a/launch/launch/actions/opaque_coroutine.py
+++ b/launch/launch/actions/opaque_coroutine.py
@@ -1,0 +1,89 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Module for the OpaqueCoroutine action."""
+
+import asyncio
+import collections.abc
+from typing import Any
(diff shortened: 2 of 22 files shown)


hidmic:
I think it makes sense to get rid of the legacy API regardless of what comes out of the launch based testing framework work. Unrelated functionality I had to add I think could still be useful outside testing, do tell if you think otherwise.

launch_testing/launch_testing/__init__.py
```diff
@@ -17,52 +17,152 @@
...
+            if self.__tests[test_name] == 'armed':
+                self.__tests[test_name] = 'dropped'
+        return EmitEvent(event=Shutdown(reason=reason))
+
+    def _succeed(self, test_name):
+        self.__tests[test_name] = 'succeeded'
+        if all(status == 'succeeded' for status in self.__tests.values()):
+            return EmitEvent(event=Shutdown(reason='all tests finished'))
+
+    def add_fixture_action(self, launch_description, action, required=False):
```
wjwwood: Passing the launch_description in here was a bit off-putting to me at first, but I see the reason for it now, though I wish it were more transparent what was happening. If you look just at the calling code (like at one of the converted tests), you don't get a since as to what is being done to the launch description, you can assume it's adding the fixture/test/output_test or whatever, but I hadn't imagined it would add more actions until reading the code for it here.

A possible alternative would be to make this function simply generate and return a list of actions to be added, rather than adding them directly. Though I'm not sure if the launch description is required within this function in any other cases. For example, it might look like:

```
my_test_exec_action = ExecuteProcess(...)
fixture_action = lts.create_fixture_action(my_test_exec_action)
ld.add_action(fixture_action)
```

In that example `fixture_action` might be just the original action or a `GroupAction` that consists of several things.

I'm not 100% sold on my approach either, as I wouldn't be able to know without an example or very good docs if I should add the original action or just the result of `create_fixture_action()` instead.

launch_testing/launch_testing/__init__.py
```diff
@@ -17,52 +17,152 @@
...
+                        lambda *args: self._succeed(test_name)
+                    )
+                ))
+            )
+        self._arm(test_name)
+        return action
+
+    def add_output_test(self, launch_description, action, output_file,
+                        filtered_prefixes=None, filtered_patterns=None,
+                        filtered_rmw_implementation=None):
```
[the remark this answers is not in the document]
wjwwood: Here and elsewhere.

launch_testing/launch_testing/__init__.py
hidmic: > Passing the launch_description in here was a bit off-putting to me at first...

It sort of is to me as well, but as I said before, I was trying to minimize the amount of code changes (because it's not at all unlikely we end up ditching this in the near future) and thus I kept the API they way it was. 

> A possible alternative would be to make this function simply generate and return a list of actions to be added, rather than adding them directly.

That is an option, sure. Another easy one (easy to use I mean) would be to have a `LaunchTestDescription` along with the `LaunchTestService` and make them wrap the `LaunchDescription` and `LaunchService` respectively. Not sure if it makes sense to spend time polishing this API though.

launch_testing/launch_testing/__init__.py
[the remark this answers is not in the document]
hidmic: Fair enough, done.

launch/launch/action.py
```diff
@@ -56,7 +56,21 @@ def describe(self) -> Text:
...
+                if future is not None:
+                    future.add_done_callback(
+                        lambda _: context.emit_event_sync(
+                            ExecutionComplete(action=self)
+                        )
+                    )
+                else:
+                    context.emit_event_sync(
+                        ExecutionComplete(action=self)
+                    )
```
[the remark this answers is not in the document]
hidmic: Sounds good. Done.

launch/launch/action.py
```diff
@@ -56,7 +56,21 @@ def describe(self) -> Text:
     def visit(self, context: LaunchContext) -> Optional[List[LaunchDescriptionEntity]]:
         """Override visit from LaunchDescriptionEntity so that it executes."""
         if self.__condition is None or self.__condition.evaluate(context):
-            return cast(Optional[List[LaunchDescriptionEntity]], self.execute(context))
+            try:
+                return cast(Optional[List[LaunchDescriptionEntity]], self.execute(context))
+            finally:
+                from .events import ExecutionComplete  # noqa
+                future = self.get_asyncio_future()
```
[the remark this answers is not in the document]
hidmic: Hmm correct me if I'm wrong, but I was expecting the future, if any, to be always the same and to be resolved by the action. The rationale is, if there are ongoing tasks (i.e. `future is not None`) emit the event once those tasks are done (i.e. `future.done() is True`), otherwise emit right away.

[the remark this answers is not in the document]
wjwwood: You're right, I read `self.get_asyncio_future()` as one of the future _creation_ functions, but it's actually an accessor.

hidmic:
Running CI for all related PRs.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6114/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2589/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5043/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5984/)

hidmic:
@mjcarroll @sloretz @wjwwood I'm a bit unsure on how to handle bad exit codes upon shutdown in a true cross platform way, thus I'm [blatantly ignoring them](https://github.com/ros2/launch/pull/167/files#diff-95fa932db2912f5a4060524ea72cd0cfR79). Not very proud, but I can't figure out how to do better than that.

hidmic:
Alright, this PR collection is mostly passing, but for odd behavior we're seeing with @mjcarroll on Windows whenever we attempt to make FastRTPS talk with Opensplice or Connext. Would it make sense to merge these PRs and open a followup issue? It'd make `master` unstable on Windows, though I really wonder why that isn't the case already.

wjwwood:
> Would it make sense to merge these PRs and open a followup issue? It'd make `master` unstable on Windows, though I really wonder why that isn't the case already.

I think making master unstable (even just on Windows), is not really a good idea.

hidmic:
So, the real question is why it isn't. Could it be that the existing test is not actually testing anything?

hidmic:
Ok, it looks like the current legacy launch test for node name lookups across middlewares is passing even though the executable under test is exiting with a non zero return code. See below for an instance of this:

```bash
C:\Users\Ekumen\Workspaces\ros2>C:/Python37/python.exe -u -m pytest -s C:/Users/Ekumen/Workspaces/ros2/build/test_rclcpp/test_node_name__rmw_fastrtps_cpp__rmw_opensplice_cpp_Release.py
============================================================================================== test session starts ==============================================================================================
platform win32 -- Python 3.7.2, pytest-4.2.1, py-1.7.0, pluggy-0.8.1
rootdir: C:\Users\Ekumen\Workspaces\ros2, inifile:
plugins: rerunfailures-6.0, repeat-0.7.0, cov-2.6.1
collected 1 item

build\test_rclcpp\test_node_name__rmw_fastrtps_cpp__rmw_opensplice_cpp_Release.py (node_with_name) pid 448: ['C:/Users/Ekumen/Workspaces/ros2/build/test_rclcpp/Release/node_with_name.exe', 'rmw_fastrtps_cpp'] (stderr > stdout, all > console)
(node_name_list) pid 10780: ['C:/Users/Ekumen/Workspaces/ros2/build/test_rclcpp/Release/node_name_list.exe', 'node_with_name_rmw_fastrtps_cpp'] (stderr > stdout, all > console)
[node_name_list] Waiting for node with name: node_with_name_rmw_fastrtps_cpp
[node_with_name] Starting node with name: node_with_name_rmw_fastrtps_cpp
[node_name_list]
[node_name_list] -
[node_name_list] - node_name_list
...
[node_name_list]
[node_name_list] -
[node_name_list] - node_name_list
[node_name_list] not found expected node name
(node_name_list) rc 1
() tear down
(node_with_name) signal SIGTERM
(node_with_name) rc 1
.

=========================================================================================== 1 passed in 11.84 seconds ===========================================================================================
```

I'd bet the issue is in the `default_exit_handler` implementation: 

https://github.com/ros2/launch/blob/master/launch/launch/legacy/exit_handler.py#L26-L51

That condition on the `if` clause always evaluates to `False` on Windows and thus return codes are always ignored.

hidmic:
Also, we haven't seen any issues on other tests because they are currently disabled, for instance:

https://github.com/ros2/system_tests/blob/4424f64f183a2a6c952af8d4124114f619fdf909/test_communication/CMakeLists.txt#L161-L172

hidmic:
So, we have a problem with FastRTPS talking to Opensplice and Connext on Windows. Looks like we have an [issue](https://github.com/ros2/rmw_fastrtps/issues/246) tracking it. But since this predates these changes (which have only revealed more issues), in the interest of time I'd say: let's disable the offending test just as we did before, and follow up the that issue. We can then come back and re-enable all tests. 

@mjcarroll @sloretz @ivanpauno thoughts?

hidmic:
[CI is green](https://github.com/ros2/launch/pull/167#issuecomment-460348251)! Mind to take a shot at this. We can re-run CI on all platforms right after.

launch/launch/actions/opaque_coroutine.py
```diff
@@ -0,0 +1,103 @@
...
+class OpaqueCoroutine(Action):
+    """
+    Action that adds a Python coroutine to the launch run loop.
+
+    The signature of a coroutine should be:
+
+    .. code-block:: python
+
+        async def coroutine(
+            context: LaunchContext,  # iff ignore_context is False
```
mjcarroll: I understand what `iff` is in this context, but it may be helpful to be a little bit more explicit.  It may also be worth updating the docs for `OpaqueFunction` so that they are consistent in this regard.

launch/launch/actions/opaque_coroutine.py
```diff
@@ -0,0 +1,103 @@
...
+        args: Optional[Iterable[Any]] = None,
+        kwargs: Optional[Dict[Text, Any]] = None,
+        ignore_context: bool = False,
+        **left_over_kwargs
+    ) -> None:
+        """Constructor."""
+        super().__init__(**left_over_kwargs)
+        if not asyncio.iscoroutinefunction(coroutine):
+            raise TypeError(
+                "OpaqueCoroutine expected a couroutine for 'couroutine', got '{}'".format(
```
mjcarroll: couroutine -> coroutine

launch/launch/event_handlers/on_execution_complete.py
```diff
@@ -0,0 +1,122 @@
...
+from ..launch_context import LaunchContext
+from ..launch_description_entity import LaunchDescriptionEntity
+from ..some_actions_type import SomeActionsType
+
+
+class OnExecutionComplete(EventHandler):
+    """
+    Convenience class for handling an action completion event.
+
+    It may be configured to only handle the completion of a specific action,
```
mjcarroll: Do we have an example or test for handling specific vs all?  I can't seem to find it.

hidmic: Good point, we do not. Test added.

launch/launch/actions/opaque_coroutine.py
hidmic: Oops, good catch!

launch/launch/actions/opaque_coroutine.py
hidmic: That makes sense. I've updated the documentation, but I'm not sure what you mean about changing `OpaqueFunction` docs.

hidmic:
Grrr something changed on `actions` and `sros2`, tests are failing again.

hidmic:
@mjcarroll CI got unstable because (a) I had to rebase `ros2/system_tests` branch after https://github.com/ros2/system_tests/pull/330 and (b) there's a test failure coming from `sros2` and it's not specific to this set of PRs (see https://ci.ros2.org/job/ci_osx/5150/ or https://ci.ros2.org/job/ci_linux/6268/). So if we run CI again, it'll go yellow again.

We can wait till (b) gets fixed, but I'm afraid more things will get in the way after that. It's fine by me, but that means more delay for these PRs.

hidmic:
@mjcarroll could you take a look at the dependency fixing PRs?

mjcarroll:
Done!

wjwwood:
There was a bug fix in this pr that was unrelated to the original change. @stonier opened a pr for just this change here:

https://github.com/ros2/launch/pull/220

I declined it so that we could put it through our typical patch release process.

The change is small, @nuclearsandwich FYI.

nuclearsandwich:
> I declined it so that we could put it through our typical patch release process.
> 
> The change is small, @nuclearsandwich FYI.

Thanks for the ping. I don't mind if the original PR gets re-opened or if we open a new PR explicitly preserving the original commit authors as long as we run Crystal CI on it before merging and releasing.

Issue #160, opened 2019-01-11
Launch composable C++ nodes


sloretz:
## Feature request

This is a feature request for launching multiple composable nodes into the same process. The acceptance criteria for this ticket is limited to launching dynamically composable rclcpp non-lifecycle nodes into a container process.

#### Feature description

`launch_ros` must pass arguments to the composable nodes where "arguments" means:

* the information that can be passed via the command line of any `rcl` based node
  * parameters
  * remap rules
  * node name
  * namespace
  * log level
* information that is specific to the client library `rclcpp`
  * use_intra_process_comms
  * start_parameter_services

An exception is `use_global_arguments` which should always be `false` so a composed node is not affected by arguments passed to the container process.

Arguments does not include environment variables like `RMW_IMPLEMENTATION` which apply to the whole container process.

##### launch_ros

See slide 8 of https://roscon.ros.org/2018/presentations/ROSCon2018_launch.pdf

```python
        ExecuteComposableNodeProcess(
            package='rclcpp_components', container_executable='rclcpp_node_container', output='screen',
            composable_node_descriptions=[
                ComposableNodeDescription(
                    package_name='demo_nodes_cpp', node_plugin_name='talker',
                    name='my_talker'),
                ComposableNodeDescription(
                    package_name='demo_nodes_cpp', node_plugin_name='listener'
                    name='my_listener')
            ]
        )
```

`ExecuteComposableNodeProcess` should start a container executable with a defined API. This allows a user to create a custom container process (e.g. create a container with a custom executor). The API may be a config file and/or a ROS API. The acceptance criteria for this ticket is limited to a ROS service API that can load and unload composable nodes.

##### rclcpp

There must be a new ROS package (`rclcpp_components`?) with a container process for rclcpp nodes. The acceptance criteria for this task is to be able to use either a single threaded executor or multithreaded executor in a container process. For example, this could be two container processes, or one with a CLI argument to choose the executor.

Launching a composable node requires a common API for constructing a node.
One option is to mimic ROS 1 nodelets which required a default constructor on the node and had an `init()` member function to initialize the nodes with parameters.
Another option is having a constructor with one argument and and a factory class to call it.
Regardless this task should require ros2/rclcpp#492 so there is one object `NodeArguments/NodeOptions` so new arguments to a node don't require changing the signature of a function.

* https://github.com/ros2/rclcpp/tree/wjwwood/components_wip
[...]

PR #168, opened 2019-01-29
Examples for launching composable nodes

sloretz:
Part of #160, opening for discussion

This adds examples for launching composable nodes. One is the same as slide 8 of https://roscon.ros.org/2018/presentations/ROSCon2018_launch.pdf with `ExecuteComposableNodeProcess` renamed to `ComposableNodeContainer` and `ComposableNodeDescription` renamed to `ComposableNode`. The other example assumes an `OnProcessStart` event handler is created.

### DIFF
--- a/launch_ros/composition_by_event_launch.py
+++ b/launch_ros/composition_by_event_launch.py
@@ -1,0 +1,69 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


launch_ros/composition_by_event_launch.py
```diff
@@ -0,0 +1,69 @@
...
+    ld.add_action(launch.actions.RegisterEventHandler(launch.event_handlers.OnProcessStart(
+        target_action=container_action,
+        on_start=[
+            launch_ros.actions.ComposableNode(
+                package_name='composition', node_plugin_name='talker',
+                name='my_talker', target_container=container_action),
+            launch_ros.descriptions.ComposableNode(
+                package_name='composition', node_plugin_name='listener'
+                name='my_listener', target_container=container_action)
+            ]
```
wjwwood: I'm curious about this, maybe a better way to do this would be something like:

```python
on_start=[
  launch_ros.actions.LoadComposableNodes(nodes=[
    launch_ros.descriptions.ComposableNode(...),
    ...
  ], target_container=container_action),
]
```

The benefit to this is that you don't need the (seems to be optional) `target_container` for the `ComposableNode` class.

sloretz: `LoadComposableNodes` in 406b4da9c8e413031acfb5b913615b82fd1171a9

sloretz:
I'll close this PR for now and reopen when an implementation is ready

PR #172, opened 2019-02-01
Refactor arg print functions

pbaughman:
Make it easier to print args if you already have a LaunchDescription object and don't need to parse a file

I'm working on a tool that uses arguments in the same way as the ros2 launch CLI tool but I only have LaunchDescription objects, not python launch files.  This PR makes it easy to print the same message as ros2 launch <file> --show-args, but starting with a LaunchDescription object instead of python file


wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6107/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2582/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5035/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5976/)

PR #173, opened 2019-02-01
Add normalize_remap_rule and types

sloretz:
I want to reuse the types and type checking on arguments to `Node` in `ComposableNode`. This PR does that for the `remappings` argument. I'd like to do the same thing for `parameters`, but thought I would split it up to get feedback sooner.

### DIFF
--- a/launch_ros/launch_ros/remap_rule_type.py
+++ b/launch_ros/launch_ros/remap_rule_type.py
@@ -1,0 +1,27 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Module for remap rule types."""
+
+from typing import Iterable
+from typing import Tuple
+
+from launch.some_substitutions_type import SomeSubstitutionsType
+from launch.substitution import Substitution
+
+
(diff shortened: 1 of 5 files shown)


launch_ros/launch_ros/remap_rule_type.py
```diff
@@ -0,0 +1,27 @@
...
+from typing import Iterable
+from typing import Tuple
+
+from launch.some_substitutions_type import SomeSubstitutionsType
+from launch.substitution import Substitution
+
+
+SomeRemapRule = Tuple[SomeSubstitutionsType, SomeSubstitutionsType]
+SomeRemapRules = Iterable[SomeRemapRule]
+RemapRule = Tuple[Tuple[Substitution, ...], Tuple[Substitution, ...]]
```
sloretz: This is `Tuple[Tuple[` instead of `Tuple[List[` to make a normalized remap rule immutable. 

There isn't anything called `description` yet, so currently I'm thinking a `launch_ros.description.ComposableNode` instance is an immutable description of some end state. Unlike an action instance which is only performed once, a description could be given to multiple actions.  Defining the type as `Tuple` is a way to prevent remap rules on it from being modified.

wjwwood: Re: Tuple vs List, immutable is fine by me.

Re: `launch_ros.description.*`, is there a description base class for these? If not, I had imagined they would not be in a subpackage together. If so, then I think a module called `launch_ros.description` with a subpackage called `launch_ros.descriptions.*` would be most appropriate.

Otherwise, I think `launch_ros.ComposableNode` is just fine.

sloretz: > Re: launch_ros.description.*, is there a description base class for these? If not, I had imagined they would not be in a subpackage together.

Roger, `launch_ros.ComposableNode` it is

sloretz:
Added tests in 1e0616c 

CI (testing `launch_ros`, `test_launch_ros`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6109/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2584/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5038/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5979/)

PR #177, opened 2019-02-06
launch: ensure event handlers add event to context locals

kyrofa:
There is some inconsistency in what event handlers add to the context's `locals`. Some add the event, others don't, and [the documentation for `EventHandler`](https://github.com/ros2/launch/blob/master/launch/launch/event_handler.py#L35) implies that they all should (this was reaffirmed during a conversation with @wjwwood in IRC).

First of all, this PR adds tests for all event handlers (only `OnProcessStart` had any) in order to ensure behavior is not changed. In the process, a few inconsistencies were found regarding exception types which were resolved.

Finally, a small backward-compatible `EventHandler` API modification was made to ensure all event handlers properly added the event to the context's locals as well as properly handled `handle_once=True` (few did).

### DIFF
--- a/launch/launch/event_handler.py
+++ b/launch/launch/event_handler.py
@@ -36,9 +36,6 @@ class EventHandler:
     via the context's locals, e.g. `context.locals.event`
     As another example, getting the name of the event as a Substitution:
     `launch.substitutions.LocalSubstitution('event.name')`.
-
-    Child classes should implement the _handle(event: Event, context: 'LaunchContext')
-    function instead of overriding the handle() function.
     """
 
     def __init__(
@@ -111,8 +108,4 @@ class EventHandler:
         context.extend_locals({'event': event})
         if self.handle_once:
             context.unregister_event_handler(self)
-
-        try:
-            return self._handle(event, context)
-        except AttributeError:
-            return self.entities
+        return self.entities
--- a/launch/launch/event_handler.py
+++ b/launch/launch/event_handler.py
@@ -28,7 +28,7 @@ if False:
     from .launch_context import LaunchContext  # noqa
 
 
-class EventHandler:
+class BaseEventHandler:
     """
     Base class for event handlers, which handle events in the launch system.
 
@@ -38,31 +38,17 @@ class EventHandler:
     `launch.substitutions.LocalSubstitution('event.name')`.
     """
 
-    def __init__(
-        self,
-        *,
-        matcher: Callable[[Event], bool],
-        entities: Optional[SomeActionsType] = None,
-        handle_once: bool = False
-    ) -> None:
+    def __init__(self, *, matcher: Callable[[Event], bool], handle_once: bool = False):
         """
         Constructor.
 
         :param: matcher is a callable that takes an event and returns True if
             the event should be handled by this event handler, False otherwise.
-        :param: entities is an LaunchDescriptionEntity or list of them, and is
-            returned by handle() unconditionally if matcher returns True.
         :param: handle_once is a flag that, if True, unregisters this EventHandler
             after being handled once.
         """
         self.__matcher = matcher
-        self.__entities = entities
         self.__handle_once = handle_once
-
-    @property
-    def entities(self):
-        """Getter for entities."""
-        return self.__entities
 
     @property
     def handle_once(self):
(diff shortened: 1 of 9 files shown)


launch/launch/event_handler.py
```diff
@@ -36,6 +36,9 @@ class EventHandler:
     via the context's locals, e.g. `context.locals.event`
     As another example, getting the name of the event as a Substitution:
     `launch.substitutions.LocalSubstitution('event.name')`.
+
+    Child classes should implement the _handle(event: Event, context: 'LaunchContext')
+    function instead of overriding the handle() function.
```
kyrofa: The easiest thing to do would have been for child classes to override `self.entities` instead, but most of them also needed access to the context, and I didn't want to break the API. Alternative ideas are welcome.

launch/launch/event_handlers/on_process_exit.py
```diff
@@ -54,7 +54,7 @@ def __init__(
         """Constructor."""
         from ..actions import ExecuteProcess  # noqa
         if not isinstance(target_action, (ExecuteProcess, type(None))):
-            raise RuntimeError("OnProcessExit requires an 'ExecuteProcess' action as the target")
+            raise TypeError("OnProcessExit requires an 'ExecuteProcess' action as the target")
```
kyrofa: Some classes used `TypeError` for this, others used `RuntimeError`. I made them all consistent and used `TypeError`.

launch/launch/event_handler.py
wjwwood: I'm fine with not changing entities, but see my review for a subtle change to the approach documented here.

kyrofa:
@wjwwood it would be an easy change to make, but I do have a concern on which I'd like to get your thoughts.

I'm a firm believer in making APIs easy to use correctly and hard to incorrectly. That's easier in C++ than in Python, but I'll ignore that :wink: . If I was looking at the `EventHandler`'s public API today, it's pretty easy to use. However, the addition of a `base_handle()` function to its public API completely throws that off. As a caller, do I use `handle()` or `base_handle()`? Using the wrong one will cause things to work incorrectly, but they look equivalently-valid to call.

Now, there's admittedly the same concern to be had for child classes: which function are they supposed to implement? In this situation (coming from a C++ background) I tend to favor having the children implement more private-y functions in order to keep the public API clean, but it sounds like you may prefer the opposite?

kyrofa:
Note that we could get away with only overriding and not adding to the API at all as long as each base class called `super().handle()` and ignored the return value (which should be None in most cases anyway), but that feels a tad odd given how the base class' `handle()` function looks like it's supposed to be returning something useful.

wjwwood:
I'd prefer no private functions, and just a better naming scheme for the "call this to handle events" method and the "override this to have custom handling logic without messing up base class logic".

Is it possible to restructure to use super some how?

kyrofa:
Totally, take a look at what I just pushed, see what you think.

launch/launch/event_handlers/on_process_exit.py
```diff
@@ -88,8 +88,10 @@ def __init__(
             else:
                 self.__actions_on_exit = [on_exit]
 
-    def _handle(self, event: Event, context: LaunchContext) -> Optional[SomeActionsType]:
+    def handle(self, event: Event, context: LaunchContext) -> Optional[SomeActionsType]:
         """Handle the given event."""
+        super().handle(event, context)
```
wjwwood: This is a bit weird, but I guess it's ok. Maybe I'd feel better if we captured the return and made sure it was none, or maybe even if there was just a comment explaining what is happening.

What if the base class had a `pre_handle()` method, that did everything except return entities which the derived classes could call with super?

kyrofa:
@wjwwood I tried a different refactor, what do you think about this? I still want to double-check that I didn't mess up the docs, but wanted to get your thoughts on the approach.

kyrofa:
Took another look, docs look fine.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6138/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2610/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5064/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6005/)

Issue #176, opened 2019-02-05
Add actions for running gtest and pytests


ivanpauno:
## Feature request
Add two actions, PyTest and GTest, to launch tests. See #161.

#### Feature description
Both of the actions will take two arguments:
- path: Path of the test to execute.
- timeout: float [seconds], indicating after how much time the test is killed.

Example usage will be like: https://github.com/ros2/launch/blob/5518345495df34ec11c4767a0d777811edce152d/launch_testing/examples/test_pub_sub_launch.py#L74-L83
Other frameworks could be added using the same parameters.

#### Implementation considerations
- It will use an underlying timeout mechanism of the test framework if it exists. If not, it will use a TimerAction for killing it.
- It will wrap the testing framework details (e.g. command line arguments).
- In a first iteration, it will not handle substitutions.

Proposed steps would be:
1. [x] Add a GTest action.
2. [x] Add a PyTest action.
3. [x] Add substitution support to GTest.
4. [x] Add substitution support to PyTest.

hidmic:
Addressed by #178.

PR #178, opened 2019-02-07
Add test actions

ivanpauno:
Connected to #176. This pull request add two actions GTest and PyTest, to run tests.
I'm doing this over #167, as running the old test with the legacy API generated a run-time asyncio error ('event loop is closed'), because of bad test interaction.
Also migrated to ament_cmake package, in order to build the dummy cpp test (maybe it exists a better way of doing this).


launch_testing/dummy_tests/locking.py
```diff
@@ -12,12 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
```
hidmic: @ivanpauno please update the copyright year.

launch_testing/launch_testing/__init__.py
```diff
@@ -22,8 +22,13 @@
 from launch.event_handlers import OnProcessIO
 from launch.events import Shutdown
 
+from . import actions
 from .output import create_output_check
 
+__all__ = [
+    'actions',
+]
+
```
hidmic: @ivanpauno I'm not sure we want __all__ to behave like this (see [here](http://xion.io/post/code/python-all-wild-imports.html)). Maybe we should add the class defined below and the `output` module as well?

launch_testing/launch_testing/actions/gtest.py
```diff
@@ -0,0 +1,48 @@
+# Copyright 2018 Open Source Robotics Foundation, Inc.
```
hidmic: @ivanpauno `2018` -> `2019`

launch_testing/launch_testing/actions/pytest.py
```diff
@@ -0,0 +1,48 @@
...
+        Constructor.
+
+        timeout argument is passed to :class:`launch_testing.Test`.
+        The other arguments are passed to :class:`launch.ExecuteProcess`, so
+        see the documentation for the class for additional details.
+
+        :param: path the path of the test to be executed.
+        """
+        cmd = [FindExecutable(name='python3'), '-m pytest', path]
+        super().__init__(cmd=cmd, shell=True, **kwargs)
```
hidmic: @ivanpauno I _think_ that using `FindExecutable` precludes the need for `shell=True`.

launch_testing/launch_testing/actions/test.py
```diff
@@ -0,0 +1,76 @@
...
+
+        Delegated to :meth:`launch.actions.ExecuteProcess.execute`.
+        """
+        actions = super().execute(context)
+        if not self.__timeout:
+            return actions
+        # Setup a timer to send us a SIGKILL if the test locks
+        sigkill_timer = TimerAction(period=self.__timeout, actions=[
+            EmitEvent(event=SignalProcess(
+                signal_number=signal.SIGKILL,
```
hidmic: @ivanpauno maybe we could send a `SIGINT` instead? It's slightly more polite :)

launch_testing/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
+find_package(ament_cmake REQUIRED)
+
+ament_export_dependencies(ament_cmake)
+
+ament_python_install_package(${PROJECT_NAME})
+
+add_executable(locking dummy_tests/locking.cpp)
```
hidmic: @ivanpauno consider relocating your dummy applications below `test/`.

launch_testing/test/launch_testing/test_gtest.py
```diff
@@ -0,0 +1,35 @@
...
+
+from pathlib import Path
+
+from launch import LaunchDescription
+from launch import LaunchService
+
+from launch_testing.actions import GTest
+
+
+def test_gtest():
```
hidmic: @ivanpauno consider adding tests for real `googletest` and `pytest` instances, here and below respectively.

launch_testing/launch_testing/actions/test.py
ivanpauno: Sounds good, but maybe it should be escalated to SIGKILL after an extra timeout. WDYT?

hidmic: That's `ExecuteProcess` action behavior upon shutdown already. Maybe we could try to reuse its `__shutdown_process()` method somehow?

ivanpauno:
@wjwwood I'm having a problem here, in a test that is locking forever.
I'm adding two actions to run gtest and pytest, which after a timeout are killed. I'm reusing the on_shutdown  actions of ExecuteProcess, in order to not re-implement the same thing again.
The first timer is ending fine, SIGINT is being emited, and the gtest action is ending fine. Then the two other timers (for escalating to SIGTERM and SIGKILL) are cancelled, but the loop continues waiting for them and never ends. I don't know if this behavior is ok, can you check?

How to repro:
- I added an extra on_process_exit handler to avoid the error in test_gtest.py, comment it:
https://github.com/ros2/launch/blob/cea2cb23b8bacd2f2b6b5529fd3246095118045b/launch_testing/test/launch_testing/test_gtest.py#L41-L43
- It's better only running the test_gtest.py, and not all the tests.

launch_testing/launch_testing/actions/pytest.py
ivanpauno: It fails without shell=True

wjwwood:
Those timers do need to hold up shutdown, but not all timers need to do so (see: https://github.com/ros2/launch/pull/181), but `ExecuteProcess` action cancels those timers when the process exits before they expire, preventing them from holding up shutdown when the process exits quickly. Perhaps you're missing that part.

I haven't had time to fully review this, so I can't speak to how you're reusing the parts of execute process, though I'm a bit curious why you're not inheriting from it if you're both running a subprocess and wanting it to be killed gracefully (with escalating signals).

hidmic:
@wjwwood the added `Test` action and their subclass do inherit from `ExecuteProcess`. @ivanpauno showed me the issue he's describing above. It's not that timers are holding up the loop, but the `shutdown_on_idle` feature that (it seems to me) has a race. 

- The executable exits and a `ProcessExit` event is sent.
- The timers are canceled (i.e. their cancelling future is resolved).
- The **main loop awakes first** and handles the event sent.
- The main loop goes on and awaits for more events since the timers' completion futures have not been resolved yet.
- The timers' coroutines are awaken and their completion futures resolved.
- The main loop never wakes up because no event was emitted nor will be emitted.

hidmic:
On a side note, may I ask you guys  @mjcarroll @sloretz (and maybe @wjwwood if you feel like it) to review this? I've touched it, so I can't review it anymore.

wjwwood:
All the waiting occurs here:

https://github.com/ros2/launch/blob/d5ace5084262de96115ea7d68b1c43160b0bf3e7/launch/launch/launch_service.py#L230-L237

Do you see debug output for "still waiting on ..."? I'm not initially convinced by your bullet point scenario above that there is a race, but it's been a while since I've steeped myself in this code, so maybe.

hidmic:
But that's if you've received a shutdown event. In the case described above, we were hoping the loop would trigger the shutdown itself once it detects it's idle:

https://github.com/ros2/launch/blob/d5ace5084262de96115ea7d68b1c43160b0bf3e7/launch/launch/launch_service.py#L210-L214

But it seems to be getting stuck in:

https://github.com/ros2/launch/blob/d5ace5084262de96115ea7d68b1c43160b0bf3e7/launch/launch/launch_service.py#L239-L240

hidmic:
@mjcarroll ping

hidmic:
@mjcarroll can we get this one in? I need to the test action concept for https://github.com/ApexAI/apex_rostest/issues/19. 

Running CI just in case:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6391/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2783/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5258/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6230/)

hidmic:
And that's why running CI is important :)

hidmic:
Alright, CI is passing. Merging! Thanks for taking a look @mjcarroll !

Issue #183, opened 2019-02-13
timer_action use of global variable prevents multiple launches from same process


pbaughman:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 16.04
- Installation type:
  - source
- Version or commit hash:
  - 34f10c69ea0f666348e11c2e5c7d66e947430730
- DDS implementation:
  - 
- Client library (if applicable):
  - 

#### Steps to reproduce issue

```
import sys                                                      
                                                                
import launch                                                   
import launch.actions                                           
import launch_ros.actions                                       
                                                                
def generate_launch_description():                              
    return launch.LaunchDescription([                           
        launch_ros.actions.Node(                                
            package="demo_nodes_py",                            
            node_executable='talker',                           
        ),                                                      
                                                                
        launch.actions.TimerAction(                             
            period="1",                                         
            actions=[                                           
                launch.actions.Shutdown(reason="timeout")       
            ]                                                   
        )                                                       
    ])                                                          
                                                                
if __name__ == "__main__":                                      
                                                                
    ls = launch.LaunchService()                                 
    ls.include_launch_description(generate_launch_description())
    assert 0 == ls.run()                                        
                                                                
    ls = launch.LaunchService()                                 
    ls.include_launch_description(generate_launch_description())
    assert 0 == ls.run()                                        
```

#### Expected behavior
The talker node should be launched, shutdown, launched, and shutdown again

#### Actual behavior
The talker node is launched, shutdown, launched, then hangs around forever

#### Additional information
The problem is that the [timer_action assumes there will only ever be one launch context](https://github.com/ros2/launch/blob/master/launch/launch/actions/timer_action.py#L139-L150).  The second time through, the event handler that handles the TimerEvent callback isn't registered on the context.

The most obvious time this becomes a problem is when writing tests, but any application that uses multiple launch descriptions that contain timers will have a problem with this.  I ran into this while writing pytests for https://github.com/ros2/launch/pull/181 .  The first test would work just fine, but all subsequent tests would fail because the timers weren't working correctly

pbaughman:
I have some time to fix this issue, too.  The _event_handler_has_been_installed bool should probably be scoped to the context, not to a module level global.

Since it's python, does anybody object to me just sticking a field onto the launch context object.  Since this is more of an implementation detail, I think that's preferable to sticking it into the context __globals dictionary

Something like
```
timer_action.py def execute( . . . .
. . .

[...]

PR #181, opened 2019-02-12
Timers will cancel on shutdown

pbaughman:
- By default, a shutdown event will cancel a TimerAction
 - The timers used to send SIGKILL and SIGTERM to processes don't cancel on shutdown

From issue https://github.com/ros2/launch/issues/180

The only TimerActions I found in launch are the sigterm timer and sigkill timer in launch/actions/execute_process.py.  I've set these up to not cancel on shutdown

@wjwwood I'm a little concerned about the thread safety note [here](https://github.com/ros2/launch/blob/master/launch/launch/actions/timer_action.py#L103-L111).  I didn't deep dive the thread model for event EventHandler actions.  Do I need to be more cautious about calling self.cancel from an OpaqueFunction in TimerAction.execute?

#### Todo:
- [x] Make sure this works for ctrl+c, otherwise it's weird

PR #184, opened 2019-02-13
Fixes timer global init of event handler

pbaughman:
Fixes https://github.com/ros2/launch/issues/183

 - Use an attribute on the context to see if we need to install the timer
   callback event handler
 - Add a regression test that fails (hangs, actually) before the fix was applied

I manually verified that a launch description containing multiple timers would only register the TimerEvent handler once.

### DIFF
--- a/launch/launch/actions/timer_action.py
+++ b/launch/launch/actions/timer_action.py
@@ -135,7 +135,7 @@ class TimerAction(Action):
             return None
 
         # Once per context, install the general purpose OnTimerEvent event handler.
-        if not hasattr(context, "_TimerAction__event_handler_has_been_installed"):
+        if not hasattr(context, '_TimerAction__event_handler_has_been_installed'):
             from ..actions import OpaqueFunction
             context.register_event_handler(EventHandler(
                 matcher=lambda event: is_a_subclass(event, TimerEvent),
@@ -145,7 +145,7 @@ class TimerAction(Action):
                     )
                 ),
             ))
-            setattr(context, "_TimerAction__event_handler_has_been_installed", True)
+            setattr(context, '_TimerAction__event_handler_has_been_installed', True)
 
         # Capture the current context locals so the yielded actions can make use of them too.
         self.__context_locals = dict(context.get_locals_as_dict())  # Capture a copy
--- a/launch/launch/actions/timer_action.py
+++ b/launch/launch/actions/timer_action.py
@@ -41,7 +41,6 @@ from ..utilities import is_a_subclass
 from ..utilities import normalize_to_list_of_substitutions
 from ..utilities import perform_substitutions
 
-_event_handler_has_been_installed = False
 _logger = logging.getLogger('launch.timer_action')
 
 
@@ -136,8 +135,7 @@ class TimerAction(Action):
             return None
 
         # Once globally, install the general purpose OnTimerEvent event handler.
-        global _event_handler_has_been_installed
-        if not _event_handler_has_been_installed:
+        if not hasattr(context, "_TimerAction__event_handler_has_been_installed"):
             from ..actions import OpaqueFunction
             context.register_event_handler(EventHandler(
                 matcher=lambda event: is_a_subclass(event, TimerEvent),
@@ -147,7 +145,7 @@ class TimerAction(Action):
                     )
                 ),
             ))
-            _event_handler_has_been_installed = True
+            setattr(context, "_TimerAction__event_handler_has_been_installed", True)
 
         # Capture the current context locals so the yielded actions can make use of them too.
         self.__context_locals = dict(context.get_locals_as_dict())  # Capture a copy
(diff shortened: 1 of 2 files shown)


launch/launch/actions/timer_action.py
```diff
@@ -147,7 +145,7 @@ def execute(self, context: LaunchContext) -> Optional[List['Action']]:
                     )
                 ),
             ))
-            _event_handler_has_been_installed = True
+            setattr(context, "_TimerAction__event_handler_has_been_installed", True)
```
wjwwood: Wouldn't it be better to set this to `False` in the context and then set it to `True` here? Rather than externally adding a private member to the context.

launch/test/launch/test_timer_action.py
```diff
@@ -0,0 +1,40 @@
...
+            )
+        ])
+
+    ls = launch.LaunchService()
+    ls.include_launch_description(generate_launch_description())
+    assert 0 == ls.run(shutdown_when_idle=False)  # Always works
+
+    ls = launch.LaunchService()
+    ls.include_launch_description(generate_launch_description())
+    assert 0 == ls.run(shutdown_when_idle=False)  # Hangs forever before BUG183 is fixed
```
wjwwood:
```suggestion
    # Next line hangs forever before https://github.com/ros2/launch/issues/183 was fixed.
    assert 0 == ls.run(shutdown_when_idle=False)
```

launch/launch/actions/timer_action.py
wjwwood: Please make this suggested change.

wjwwood:
Other than one small implementation detail, this lgtm.

launch/launch/actions/timer_action.py
pbaughman: I thought about that and did it this way because a TimerAction object is required to know about a LaunchContext in order to function, but a LaunchContext shouldn't have to know about a TimerAction in order to do its job.  Also, by making the name start with __TimerAction_ and using setattr, I prevent it from conflicting with any of the Launch Context's mangled names in the future.

If I add a member to the LaunchContext, what do you recommend I name it?  If I prefix it with `__`  then it may be difficult to access from a TimerAction due to name mangling.  Is single underbar ok?

wjwwood: > I thought about that and did it this way because a TimerAction object is required to know about a LaunchContext in order to function, but a LaunchContext shouldn't have to know about a TimerAction in order to do its job.

That's reasonable to me, perhaps we could use something like the context's locals (we could us a global to prevent it from being scoped, see: https://github.com/ros2/launch/blob/34f10c69ea0f666348e11c2e5c7d66e947430730/launch/launch/launch_context.py#L94-L102) to store this information? I suppose that's not much better than just add a member to the context class itself.

After considering your point, I'd be ok with you just leave it as-is.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6186/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2641/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5095/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6039/)

pbaughman:
Hah - my different version of flake8 strikes again. . .
```
./test/launch/test_timer_action.py:19:1: E302 expected 2 blank lines, found 1
def test_multiple_launch_with_timers():
^

./test/launch/test_timer_action.py:27:24: Q000 Remove bad quotes
                period="1",
                       ^

./test/launch/test_timer_action.py:29:52: Q000 Remove bad quotes
                    launch.actions.Shutdown(reason="Timer expired")
                                                   ^

./launch/actions/timer_action.py:138:33: Q000 Remove bad quotes
        if not hasattr(context, "_TimerAction__event_handler_has_been_installed"):
                                ^

./launch/actions/timer_action.py:148:30: Q000 Remove bad quotes
            setattr(context, "_TimerAction__event_handler_has_been_installed", True)
                             ^

1     E302 expected 2 blank lines, found 1
4     Q000 Remove bad quotes

143 files checked
5 errors

'E'-type errors: 1
'Q'-type errors: 4
```

wjwwood:
That might also be that you're missing some of the flake8 extensions we use, but I'm not sure. I think the quote style one is an extension (again not 100% sure, but I think I remember that being the case).

wjwwood:
I'll rerun CI quickly if you can fix those up, so we can unblock #181.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6187/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2642/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5096/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6040/)

PR #190, opened 2019-02-28
convert function names to snake_case

wjwwood:
This makes it more consistent with the rest of the code in launch, see my review of #145 for more discussion.


hidmic:
Since #145 has been merged, I'll close this one out. Thanks for drafting and reviewing along @wjwwood !

PR #192, opened 2019-03-01
Add normalize_parameters and evaluate_paramters

sloretz:
Probably complete, but in progress while I self review. I wrote this a while ago and am just now coming back to it.

Like #173 but for parameters. This refactors the type checking and evaluation of parameters in `Node` into new functions `normalize_parameters()` and `evaluate_parameters()`. The reason is to make this logic reusable for composable nodes #160.

I made some changes along the way

* parameter dictionaries are flattened to one level with names containing `.`
  * Reason: python type hints can't yet do recursive types
  * `rcl_yaml_param_parser` supports this just fine.
* Lists of parameters may be any `Sequence`, not just `List`
  * Reason: I wanted to use make evaluated parameters `normalize_parameters()` immutable, meaning `tuple` is used for the list of parameters. I didn't succeed since they still use dictionaries, but this is a little closer.
* Parameter types are checked in `Node.__init__()` instead of when the action is executed
  * Reason: Not really intentional; it just made sense to call `normalize_parameters()` there
  * Of course substitutions are still evaluatated when the action is executed
* Types of lists output by `evaluated_parameters()` are made uniform
  * Reason: Composable nodes will need to know the list type so it can be passed in a ros service.
  * if a list contains int and float, the list type is float
  * if a list contains dissimilar types, the list type is `str`

### DIFF
--- a/launch_ros/launch_ros/utilities/normalize_parameters.py
+++ b/launch_ros/launch_ros/utilities/normalize_parameters.py
@@ -33,7 +33,6 @@ from launch.utilities import ensure_argument_type
 from launch.utilities import normalize_to_list_of_substitutions
 
 from ..parameters_type import ParameterFile
-from ..parameters_type import ParameterName
 from ..parameters_type import Parameters
 from ..parameters_type import ParametersDict
 from ..parameters_type import ParameterValue
@@ -84,11 +83,11 @@ def _normalize_parameter_array_value(value: SomeParameterValue) -> ParameterValu
             return tuple(normalize_to_list_of_substitutions(cast(SomeSubstitutionsType, value)))
 
         if target_type == float:
-            return tuple([float(s) for s in value])
+            return tuple(float(s) for s in value)
         elif target_type == int:
-            return tuple([int(s) for s in value])
+            return tuple(int(s) for s in value)
         elif target_type == bool:
-            return tuple([bool(s) for s in value])
+            return tuple(bool(s) for s in value)
         else:
             output_value = []  # type: List[Tuple[Substitution, ...]]
             for subvalue in value:
--- a/launch_ros/launch_ros/utilities/normalize_parameters.py
+++ b/launch_ros/launch_ros/utilities/normalize_parameters.py
@@ -14,7 +14,6 @@
 
 """Module with utility for normalizing parameters to a node."""
 
-from collections.abc import Iterable
 from collections.abc import Mapping
 from collections.abc import Sequence
 import pathlib
@@ -24,6 +23,7 @@ from typing import Optional
 from typing import Sequence as SequenceTypeHint
 from typing import Tuple  # noqa
 from typing import Union  # noqa
+from typing import Set  # noqa
 
 from launch.some_substitutions_type import SomeSubstitutionsType
 from launch.some_substitutions_type import SomeSubstitutionsType_types_tuple
@@ -47,57 +47,48 @@ def _normalize_parameter_array_value(value: SomeParameterValue) -> ParameterValu
         raise TypeError('Value {} must be a sequence'.format(repr(value)))
 
     # Figure out what type the list should be
-    target_type = None  # type: Optional[type]
+    has_types = set()  # type: Set[type]
     for subvalue in value:
         allowed_subtypes = (float, int, str, bool) + SomeSubstitutionsType_types_tuple
         ensure_argument_type(subvalue, allowed_subtypes, 'subvalue')
 
         if isinstance(subvalue, Substitution):
-            subtype = Substitution  # type: type
+            has_types.add(Substitution)
+        elif isinstance(subvalue, str):
+            has_types.add(str)
+        elif isinstance(subvalue, bool):
+            has_types.add(bool)
+        elif isinstance(subvalue, int):
+            has_types.add(int)
+        elif isinstance(subvalue, float):
+            has_types.add(float)
+        elif isinstance(subvalue, Sequence):
+            has_types.add(tuple)
         else:
-            subtype = type(subvalue)
+            raise RuntimeError('Failed to handle type {}'.format(repr(subvalue)))
 
-        if target_type is None:
-            target_type = subtype
-
-        if subtype == float and target_type == int:
-            # If any value is a float, convert all integers to floats
-            target_type = float
-        elif subtype == int and target_type == float:
-            # If any value is a float, convert all integers to floats
-            pass
-        elif subtype == str and target_type == Substitution:
-            # If any value is a single substitution then result is a single string
-            target_type = Substitution
-        elif subtype == Substitution and target_type == str:
-            # If any value is a single substitution then result is a single string
-            target_type = Substitution
-        elif subtype != target_type:
-            # If types don't match, assume list of strings
-            target_type = str
-
-    if target_type is None:
-        # No clue what an empty list's type should be
-        return []
-    elif target_type == Substitution:
-        # Keep the list of substitutions together to form a single string
+    if {int} == has_types:
+        # all integers
+        return tuple(int(e) for e in value)
+    elif has_types in ({float}, {int, float}):
+        # all were floats or ints, so return floats
+        return tuple(float(e) for e in value)
+    elif Substitution in has_types and has_types.issubset({str, Substitution, tuple}):
+        # make a list of substitutions forming a single string
         return tuple(normalize_to_list_of_substitutions(cast(SomeSubstitutionsType, value)))
-
-    if target_type == float:
-        return tuple(float(s) for s in value)
-    elif target_type == int:
-        return tuple(int(s) for s in value)
-    elif target_type == bool:
-        return tuple(bool(s) for s in value)
+    elif {bool} == has_types:
+        # all where bools
+        return tuple(bool(e) for e in value)
     else:
-        output_value = []  # type: List[Tuple[Substitution, ...]]
-        for subvalue in value:
-            if not isinstance(subvalue, Iterable) and not isinstance(subvalue, Substitution):
-                # Convert simple types to strings
-                subvalue = str(subvalue)
-            # Make everything a substitution
-            output_value.append(tuple(normalize_to_list_of_substitutions(subvalue)))
-        return tuple(output_value)
+        # Should evaluate to a list of strings
+        # Normalize to a list of lists of substitutions
+        new_value = []  # type: List[SomeSubstitutionsType]
+        for element in value:
+            if isinstance(element, float) or isinstance(element, int) or isinstance(element, bool):
+                new_value.append(str(element))
+            else:
+                new_value.append(element)
+        return tuple(normalize_to_list_of_substitutions(e) for e in new_value)
 
 
 def normalize_parameter_dict(
@@ -116,10 +107,12 @@ def normalize_parameter_dict(
     Normalized values that were lists will have all subvalues converted to the same type.
     If all subvalues are int or float, then the normalized subvalues will all be float.
     If the subvalues otherwise do not all have the same type, then the normalized subvalues
-    will be lists of Substitution that will result in string parameters.
+    will be lists of Substitution that will result in a list of strings.
 
     Values that are a list of strings will become a list of strings when normalized and evaluated.
-    Values that are a list of :class:`Substitution` will become a single string.
+    Values that are a list which has at least one :class:`Substitution` and all other elements
+    are either strings or a list of substitutions will become a single list of substitutions that
+    will evaluate to a single string.
     To make a list of strings from substitutions, each item in the list must be a list or tuple.
 
     Normalized values that contained nested dictionaries will be collapsed into a single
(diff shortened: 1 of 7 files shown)


sloretz:
CI (testing launch, launch_ros, and test_launch_ros)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6304/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2720/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5180/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6141/)

test_launch_ros/test/test_launch_ros/test_normalize_parameters.py
```diff
@@ -0,0 +1,173 @@
...
+
+def test_dictionary_with_int():
+    orig = [{'foo': 1, 'fiz': [2, 3, 4]}]
+    norm = normalize_parameters(orig)
+    expected = ({'foo': 1, 'fiz': (2, 3, 4)},)
+    assert evaluate_parameters(LaunchContext(), norm) == expected
+
+
+def test_dictionary_with_int_and_float():
+    orig = [{'foo': 1, 'fiz': [2, 3.1, 4]}]
```
jacobperron: With knowledge of the implementation, I think it wouldn't be a bad idea to add a test case where the `float` is the first element in the list.

test_launch_ros/test/test_launch_ros/test_normalize_parameters.py
```diff
@@ -0,0 +1,173 @@
...
+    assert evaluate_parameters(LaunchContext(), norm) == expected
+
+    orig = [{'foo': [[TextSubstitution(text='fiz')], [TextSubstitution(text='buz')]]}]
+    norm = normalize_parameters(orig)
+    expected = ({'foo': ('fiz', 'buz')},)
+    assert evaluate_parameters(LaunchContext(), norm) == expected
+
+
+def test_dictionary_with_mixed_substitutions_and_strings():
+    orig = [{'foo': [TextSubstitution(text='fiz'), 'bar']}]
```
jacobperron: We should add a test case with a string as the first element followed by a substitution.

test_launch_ros/test/test_launch_ros/test_normalize_parameters.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+def test_dictionary_with_bytes():
+    orig = [{'foo': 1, 'fiz': bytes([0xff, 0x5c, 0xaa])}]
+    norm = normalize_parameters(orig)
+    expected = ({'foo': 1, 'fiz': bytes([0xff, 0x5c, 0xaa])},)
+    assert evaluate_parameters(LaunchContext(), norm) == expected
+
+
+def test_dictionary_with_dissimilar_array():
```
jacobperron: I'm curious to know if a dissimilar array with a substitution type will also work. In particular:

```
[True, 1, TextSubstitution(text='foo')]
```
and
```
[TextSubstitution(text='foo'), True, 1]
```

launch_ros/launch_ros/utilities/normalize_parameters.py
```diff
@@ -0,0 +1,191 @@
...
+            if subtype == float and target_type == int:
+                # If any value is a float, convert all integers to floats
+                target_type = float
+            elif subtype == int and target_type == float:
+                # If any value is a float, convert all integers to floats
+                pass
+            elif subtype == str and target_type == Substitution:
+                # If any value is a single substitution then result is a single string
+                target_type = Substitution
+            elif subtype == str and target_type == Substitution:
```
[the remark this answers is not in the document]
sloretz: Whoops, fixed in d66bcb33570ca70161884d1ca776dbd01913fc97. I didn't use the github suggestion because I think it doesn't add the `signed-off-by`

launch_ros/launch_ros/utilities/normalize_parameters.py
```diff
@@ -0,0 +1,191 @@
...
+from ..parameters_type import ParameterFile
+from ..parameters_type import Parameters
+from ..parameters_type import ParametersDict
+from ..parameters_type import ParameterValue
+from ..parameters_type import SomeParameters
+from ..parameters_type import SomeParametersDict
+from ..parameters_type import SomeParameterValue
+
+
+def _normalize_parameter_array_value(value: SomeParameterValue) -> ParameterValue:
```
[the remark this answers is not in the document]
sloretz: I think I would need to make it more more specific to make mypy happy `Union[Sequence[str], Sequence[int], Sequence[Substitution]]`. Since the function is prefixed with leading underscore I'll leave it as is for now.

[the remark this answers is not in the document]
wjwwood: Would it be `Sequence[SomeSubstitutions]`? I.e. would you want in the end a single string or a list of strings?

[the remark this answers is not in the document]
sloretz: > Would it be Sequence[SomeSubstitutions]?  I.e. would you want in the end a single string or a list of strings?

For value or on the return type? For `value`, the types `int` and `float` could passed in via a dictionary in the launch description. For the return type the values need to be given to composable nodes via a ROS service, which means the types have to be kept to know which type to set in the [parameter message](https://github.com/ros2/rcl_interfaces/blob/e631d8f93a48f2aeaa192ecc8ed9bf218330c69a/composition_interfaces/srv/LoadNode.srv#L16). `Sequence[Substitution]` wouldn't work because it [only outputs `Text`](https://github.com/ros2/launch/blob/632ef1f3eca7dbc54fbed85ab65a5efc512d83ed/launch/launch/substitution.py#L37).

launch_ros/launch_ros/utilities/normalize_parameters.py
```diff
@@ -0,0 +1,191 @@
...
+    """Normalize substitutions while preserving the type if it's not a substitution."""
+    if isinstance(value, Sequence):
```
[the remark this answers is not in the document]
sloretz: Removed indentation in 1e5d3dcf0ebcdd5bdd714d7ebb17864bd0494d95

launch_ros/launch_ros/utilities/normalize_parameters.py
```diff
@@ -0,0 +1,191 @@
...
+def normalize_parameter_dict(
+    parameters: SomeParametersDict, *,
+    _prefix: Optional[SequenceTypeHint[Substitution]] = None
+) -> ParametersDict:
+    """
+    Normalize types used to store parameters in a dictionary.
+
+    The parameters are passed as a dictionary that specifies parameter rules.
+    Keys of the dictionary can be strings, a Substitution, or an iterable of Substitution.
+    A normalized keys will be a tuple of substitutions.
```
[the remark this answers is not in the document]
sloretz: Fixed in 21be882ed654f0f0e9053414b87d6bbbf5746cfa

launch_ros/launch_ros/utilities/normalize_parameters.py
```diff
@@ -0,0 +1,191 @@
...
+    Values in the dictionary can be strings, integers, floats, substututions, lists of
```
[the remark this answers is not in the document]
sloretz: Fixed in d835512ae133f5dbdbe0c279ffef4a2e39459634

launch_ros/launch_ros/utilities/normalize_parameters.py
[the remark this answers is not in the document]
wjwwood: I was speaking about `value` because I thought that's what @jacobperron was talking about.

I meant as one of the choices in the union you mentioned, not just `Sequence[SomeSubstitutionsType]` (and I do mean that instead of `Sequence[Substitution]`).

I don't have a strong opinion, but I was extrapolating from the function name (contains `array`) and @jacobperron's question. Currently `SomeParameterValue` -> `Union[SomeSubstitutionsType, _SingleValueType, _MultiValueType]`, and `SomeSubstitutionsType` is a list of substitutions, but they're meant to be evaluated and then concatenated, so they'd equate to a "single value type" in the end. If you want to represent a "multi value type" (what I assume is a list of values) then you'd need a list of `SomeSubstitutionsType`.

Maybe there's nothing to be done here, I was just trying to make sure that you're not making the assumption that `SomeSubstitutionsType` ends up being a list of values after evaluations.

test_launch_ros/test/test_launch_ros/test_normalize_parameters.py
sloretz: b1f22861830e10212d8176e9a069226d78077400

test_launch_ros/test/test_launch_ros/test_normalize_parameters.py
sloretz: Added in 46b149528115662912592690150d5af577226867

test_launch_ros/test/test_launch_ros/test_normalize_parameters.py
sloretz: <details><summary>Good catch, there's a bug</summary>

```
test/test_launch_ros/test_normalize_parameters.py:180: in test_dictionary_with_dissimilar_array
    norm = normalize_parameters(orig)
../../../../build/launch_ros/launch_ros/utilities/normalize_parameters.py:183: in normalize_parameters
    normalized_params.append(normalize_parameter_dict(param))
../../../../build/launch_ros/launch_ros/utilities/normalize_parameters.py:162: in normalize_parameter_dict
    normalized[tuple(name)] = _normalize_parameter_array_value(value)
../../../../build/launch_ros/launch_ros/utilities/normalize_parameters.py:84: in _normalize_parameter_array_value
    return tuple(normalize_to_list_of_substitutions(cast(SomeSubstitutionsType, value)))
../../../../build/launch/launch/utilities/normalize_to_list_of_substitutions_impl.py:42: in normalize_to_list_of_substitutions
../../../../build/launch/launch/utilities/normalize_to_list_of_substitutions_impl.py:42: in <listcomp>
../../../../build/launch/launch/utilities/normalize_to_list_of_substitutions_impl.py:36: in normalize
E   TypeError: Failed to normalize given item of type '<class 'bool'>', when only 'str' or 'launch.Substitution' were expected.
```
<details>

sloretz: Added tests and fixed bug in a4ac02b6c423f3dccd3d90be97d805b9e97239db

launch_ros/launch_ros/utilities/normalize_parameters.py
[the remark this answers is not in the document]
sloretz: Ah you're right. Added `Sequence[SomeSubstitutionsType]` (and mypy appeasement required afterwards) to `SomeParameterValue` in 65f23a6cb5b70e4a162af6a6da233dbb851d60c0. The normalized type `ParameterValue` already had the right type for list of lists of substitutions.

sloretz:
CI (Testing launch_ros and test_launch_ros)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6322/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2736/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5204/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6172/)

launch_ros/launch_ros/utilities/normalize_parameters.py
```diff
@@ -0,0 +1,185 @@
...
+        return tuple(normalize_to_list_of_substitutions(cast(SomeSubstitutionsType, value)))
+    elif {bool} == has_types:
+        # all where bools
+        return tuple(bool(e) for e in value)
+    else:
+        # Should evaluate to a list of strings
+        # Normalize to a list of lists of substitutions
+        new_value = []  # type: List[SomeSubstitutionsType]
+        for element in value:
+            if isinstance(element, float) or isinstance(element, int) or isinstance(element, bool):
```
dirk-thomas: This logic could be simplified to:

```
if isinstance(element, (float, int, bool)):
```

sloretz: ros2/launch_ros#18

launch_ros/launch_ros/utilities/normalize_parameters.py
```diff
@@ -0,0 +1,185 @@
...
+from typing import Set  # noqa
+
+from launch.some_substitutions_type import SomeSubstitutionsType
+from launch.some_substitutions_type import SomeSubstitutionsType_types_tuple
+from launch.substitution import Substitution
+from launch.substitutions import TextSubstitution
+from launch.utilities import ensure_argument_type
+from launch.utilities import normalize_to_list_of_substitutions
+
+from ..parameters_type import ParameterFile
```
dirk-thomas: This line seems to result in a `flake8` warning in the dev job: http://build.ros2.org/job/Ddev__launch_ros__ubuntu_bionic_amd64/21/consoleFull#console-section-17

@sloretz Can you please double check this.

sloretz: That's odd, this PR is from March. Is this a flake8 regression? `ParameterFile` is used in a type annotation. https://github.com/ros2/launch_ros/blob/2d5d3eac831788d54c41069378733837b48b44c8/launch_ros/launch_ros/utilities/normalize_parameters.py#L175


Maybe using the [PEP526 typing syntax](https://www.python.org/dev/peps/pep-0526/) will workaround the issue?

dirk-thomas: The `dev` job users the Debian package of `flake8` which is different from the latest release we use on ci.ros2.org.

Issue #189, opened 2019-02-19
[launch_ros] use a custom context to avoid interferring with user init/shutdown of rclpy


wjwwood:
#### Feature description

Right now `launch_ros` (or more specifically `ros2 launch`) will init rclpy, create a node, and then later shutdown rclpy. It also expects that shutdown is not called prematurely.

If the user did their own init or shutdown, it might interfere with this. So `launch_ros` should instead use a custom context and do its own init and shutdown there.

pbaughman:
@wjwwood I actually have a version of this done, but I'm waiting for ros2 crystal internally to test it (the version we have doesn't let you pass arguments to init yet).  I will push it for review, but I haven't been able to test it yet

wjwwood:
That would be great!

wjwwood:
This was done in #193.

PR #193, opened 2019-03-02
Plumb rclpy.init context to get_default_launch_description

pbaughman:
From https://github.com/ros2/launch/issues/189

This lets you pass a launch context to get_default_launch_description into get_default_launch_description and plumbs it all the way down to the ROSSpecificLaunchStartup launch action.

I couldn't find where rclpy.shutdown is called, so I didn't hook that up anywhere.

I also had to change `rclpy.get_global_executor` to constructing my own SingleThreadedExecutor because the executor needs to be built with the same context as the node.

#### This Incomplete!
This is about 50% of what's needed for issue 189.  The other half is to make ros2 launch actually use this plumbing.  A question for @wjwwood - is it necessary to shut-down the context when we're done, or is it sufficient to just [pass a context here?](https://github.com/ros2/launch/blob/632ef1f3eca7dbc54fbed85ab65a5efc512d83ed/ros2launch/ros2launch/api/api.py#L140-L141)

### DIFF
--- a/launch_ros/launch_ros/default_launch_description.py
+++ b/launch_ros/launch_ros/default_launch_description.py
@@ -122,10 +122,10 @@ def get_default_launch_description(*, prefix_output_with_name=False, rclpy_conte
     :param: prefix_output_with_name if True, each line of output is prefixed
         with the name of the process as `[process_name] `, else it is printed
         unmodified
-
-    :param: rclpy_context Provide a context other than the default rclpy context to pass down
-    to rclpy.init.  The context is expected to have already been initialized by the caller using
-    rclpy.init
+    :param: rclpy_context Provide a context other than the default rclpy context
+        to pass down to rclpy.init.
+        The context is expected to have already been initialized by the caller
+        using rclpy.init().
     """
     default_ros_launch_description = launch.LaunchDescription([
         # ROS initialization (create node and other stuff).
--- a/launch_ros/launch_ros/default_launch_description.py
+++ b/launch_ros/launch_ros/default_launch_description.py
@@ -96,7 +96,9 @@ class ROSSpecificLaunchStartup(launch.actions.OpaqueFunction):
 
     def _function(self, context: launch.LaunchContext):
         try:
-            rclpy.init(args=context.argv, context=self.__rclpy_context)
+            if self.__rclpy_context is None:
+                # Initialize the default global context
+                rclpy.init(args=context.argv)
         except RuntimeError as exc:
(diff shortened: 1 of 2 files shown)


launch_ros/launch_ros/default_launch_description.py
```diff
@@ -94,12 +96,12 @@ def _run(self):
 
     def _function(self, context: launch.LaunchContext):
         try:
-            rclpy.init(args=context.argv)
+            rclpy.init(args=context.argv, context=self.__rclpy_context)
```
wjwwood: This is only needed if the context is not already initialized. I'd recommend requiring that the context that was passed in is initialized, and also require that shutting it down is the responsibility of the caller.

ros2launch/ros2launch/api/api.py
```diff
@@ -137,8 +139,14 @@ def parse_launch_arguments(launch_arguments: List[Text]) -> List[Tuple[Text, Tex
 def launch_a_python_launch_file(*, python_launch_file_path, launch_file_arguments, debug=False):
     """Launch a given Python launch file (by path) and pass it the given launch file arguments."""
     launch_service = launch.LaunchService(argv=launch_file_arguments, debug=debug)
+    context = rclpy.context.Context()
+    rclpy.init(args=launch_file_arguments, context=context)
```
pbaughman: @wjwwood I'm afraid I'm not up on what kind of arguments we're expecting to pass to the context here so I'm not able to test this very well.  It looks like the launch_file_arguments get passed all the way through to become LaunchContext.argv which gets passed to rclpy.init up in ROSSpecificLaunchStartup so this should do the same thing, but I'm afraid I haven't checked

wjwwood: I think this is right.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6329/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2743/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5212/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6179/)

wjwwood:
The test failures look unrelated/flaky:

- the `test_cli` one that is present on Linux and Windows: https://github.com/ros2/build_cop/issues/166
- the `gtest_waitable` one that is present on macOS is listed here: https://github.com/ros2/build_cop/issues/168

PR #196, opened 2019-03-07
Replace 'Coroutine' import to support Python v3.5.2

### LINKED ISSUE #195: Support Python 3.5.2 (Xenial)?
#### Feature description

Until recently, `launch` was compatible with Python 3.5.2, which is the current default version for `python3` on Ubuntu Xenial, but now is broken.

#167, introduced the following line:

https://github.com/ros2/launch/blob/d5ace5084262de96115ea7d68b1c43160b0bf3e7/launch/launch/actions/opaque_coroutine.py#L20

`Coroutine` is not available in the `typing` module until Python 3.5.3 ([reference](https://stackoverflow.com/questions/44651115/python-typing-module-missing-the-coroutine-class-in-python-3-5#comment76294521_44651115)).

I think it's an easy fix as we can use something like [TypeVar('Coroutine')](https://docs.python.org/release/3.5.2/library/typing.html#typing.TypeVar) as a replacement.
But, I'm not sure if we plan to support Xenial in Dashing and/or if we'd like to keep `launch` backwards compatible (given that it's an easy fix).

jacobperron:
Resolves #195

### DIFF
--- a/launch/launch/actions/opaque_coroutine.py
+++ b/launch/launch/actions/opaque_coroutine.py
@@ -17,12 +17,12 @@
 import asyncio
 import collections.abc
 from typing import Any
-from typing import Coroutine
 from typing import Dict
 from typing import Iterable
 from typing import List
 from typing import Optional
 from typing import Text
+from typing import TypeVar
 
 from ..action import Action
 from ..event import Event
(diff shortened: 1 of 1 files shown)


dirk-thomas:
This is not necessary since master / Dashing don't target Xenial.

PR #197, opened 2019-03-08
Add composable node launch actions

hidmic:
Connected to #160. This pull request adds `ComposableNodeContainer` and `LoadComposableNodes` actions to `launch_ros`.


launch_ros/launch_ros/descriptions/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""descriptions Module."""
+
+from .composable_node import ComposableNode
+
+__all__ = [
+    'ComposableNode',
```
sloretz: @wjwwood mentioned making this `launch_ros.ComposableNode` instead of having a descriptions module. https://github.com/ros2/launch/pull/173#discussion_r253234298

I'm fine with either if @wjwwood is on board.

launch_ros/launch_ros/descriptions/composable_node.py
```diff
@@ -0,0 +1,113 @@
...
+
+    @property
+    def node_namespace(self) -> Optional[List[Substitution]]:
+        """Get node namespace as a sequence of substitutions to be performed."""
+        return self.__node_namespace
+
+    @property
+    def parameters(self) -> Optional[Parameters]:
+        """Get node parameter YAML files or dicts with substitutions to be performed."""
+        return self.__parameters
```
sloretz: Since everything is a reference and this can contain dictionaries which are mutable, I had been thinking about the description doing a `copy.deepcopy` to guard against a buggy action modifying a description that was used multiple places. I'm don't think it needs to be done, but thought I'd mention it in case that makes someone think of something.

launch_ros/launch_ros/utilities/to_parameters_list.py
```diff
@@ -0,0 +1,54 @@
...
+
+    :param parameters: parameters as either paths to parameter files or name/value pairs.
+        See `evaluate_parameters()` documentation for further reference.
+    :returns: a list of parameters
+    """
+    parameters = []  # type: List[rclpy.parameter.Parameter]
+    for params_set_or_path in evaluated_parameters:
+        if isinstance(params_set_or_path, pathlib.Path):
+            with open(str(params_set_or_path), 'r') as f:
+                params_set = yaml.load(f)  # type: Dict[str, EvaluatedParameterValue]
```
sloretz: I think if loaded from a yaml file there can be nested dictionaries, so the name should be joined with a `.` from each level. A hack might be to call `normalize_parameters/evaluate_parameters` on the loaded yaml to flatten it

launch_ros/launch_ros/descriptions/__init__.py
hidmic: Oh, I see. The stubs did not match the examples so I chose to go with the later. But I can back up.

launch_ros/launch_ros/utilities/to_parameters_list.py
hidmic: Hmm, that's a very good point. I'll flatten it.

launch_ros/launch_ros/descriptions/composable_node.py
hidmic: +1 to keep descriptions immutable, though the same could apply to actions in general and we don't try to enforce it elsewhere.

hidmic: Well, we do try by using properties, but that's as far as we go.

launch_ros/launch_ros/descriptions/__init__.py
wjwwood: I don't have a strong opinion about it, I'll leave it up to you guys based on what you think and my previous comment.

hidmic:
@sloretz is there an implementation to test this against?

launch_ros/launch_ros/actions/load_composable_nodes.py
```diff
@@ -0,0 +1,198 @@
...
+        *,
+        composable_nodes_description: List[ComposableNode],
+        target_container: ComposableNodeContainer,
+        in_parallel: Optional[SomeSubstitutionsType] = None,
+        **kwargs,
+    ) -> None:
+        """
+        Construct a LoadComposableNodes action.
+
+        The container node is expected to provide a `~/_container/load` service for loading
```
mjcarroll: did we have somewhere that defined this service name?  I hadn't been using it, but will update accordingly.

sloretz: ros2/design#206 has `~/_containter/load_node` instead

hidmic: My bad, I'll fix it.

hidmic: Fixed in https://github.com/ros2/launch/pull/197/commits/b28bf50a56c63124d081626d4000b82466bc1c08.

launch_ros/launch_ros/actions/load_composable_nodes.py
```diff
@@ -0,0 +1,202 @@
...
+    ) -> None:
+        """
+        Load composable nodes in parallel.
+
+        :param composable_nodes_description: descriptions of composable nodes to be loaded
+        :param context: current launch context
+        """
+        for composable_node_description in composable_nodes_description:
+            context.add_completion_future(
+                context.asyncio_loop.run_in_executor(
```
sloretz: I think `asyncio` loops are always single threaded, so this will still load the nodes sequentially. I'm not sure about the order they'll be run, but I imagine it's the same order that `run_in_executor` is called.

launch_ros/launch_ros/actions/load_composable_nodes.py
```diff
@@ -0,0 +1,202 @@
...
+            ]
+        if composable_node_description.extra_arguments is not None:
+            request.extra_arguments = [
+                param.to_parameter_msg() for param in to_parameters_list(
+                    context, evaluate_parameters(
+                        context, composable_node_description.extra_arguments
+                    )
+                )
+            ]
+        response = self.__rclpy_load_node_client.call(request)
```
sloretz: It would be handy to emit an event with the id the container gave the node. It is needed to unload a node later, though I haven't figured out how to plumb that to a possible `UnloadComposableNode` action.

launch_ros/launch_ros/actions/load_composable_nodes.py
```diff
@@ -0,0 +1,202 @@
...
+from .composable_node_container import ComposableNodeContainer
+
+from ..descriptions import ComposableNode
+from ..utilities import evaluate_parameters
+from ..utilities import to_parameters_list
+
+_logger = logging.getLogger(name='launch_ros')
+
+
+class LoadComposableNodes(Action):
```
sloretz: I was expecting this to be singular `LoadComposableNode`. I can't think of a downside to this approach though.

launch_ros/launch_ros/actions/composable_node_container.py
```diff
@@ -0,0 +1,69 @@
...
+from ..descriptions import ComposableNode
+
+
+class ComposableNodeContainer(Node):
+    """Action that executes a container ROS node for composable ROS nodes."""
+
+    def __init__(
+        self,
+        *,
+        composable_nodes_description: Optional[List[ComposableNode]] = None,
```
sloretz: recommend `composable_node_descriptions` instead of `composable_nodes_description`

launch_ros/launch_ros/actions/load_composable_nodes.py
hidmic: That's true as long as `_load_node()` remains synchronous, or in other words, as long as we keep using services synchronously. And of course all this is assuming that once the service response comes along, the composed node has been fully brought up already.

IMHO the neatest solution would be to turn `rclpy` service methods into coroutines. Alternatively, we could spin up a pool of threads, but I'm less fond of it. What do you think?

launch_ros/launch_ros/actions/composable_node_container.py
hidmic: Sure thing.

launch_ros/launch_ros/actions/load_composable_nodes.py
hidmic: `LoadComposableNode` sounds good too.

launch_ros/launch_ros/actions/load_composable_nodes.py
hidmic: Hmm, what's your specific use case? Something akin to ROS capabilities? I'd defer that to another PR, it sounds to me like a problem on its own.

sloretz: Specifically I'm looking for a way to test the `unload_node` service on the container process @mjcarroll is working on. +1 to deferring this and `UnloadComposableNode` to another PR.

launch_ros/launch_ros/actions/load_composable_nodes.py
sloretz: Ah, I meant I expected an action `LoadComposableNode` that loaded only one composable node instead of taking a list of them. No need to change the class name like was done in fffac89. The plural name fits better since it can load multiple composable nodes.

launch_ros/launch_ros/actions/load_composable_nodes.py
sloretz: > neatest solution would be to turn rclpy service methods into coroutines

Agreed that would be the neatest. I think the future returned from [`call_async`](https://github.com/ros2/rclpy/blob/28a77736f1d5ffd13a1dc065e066fab3a89f27f2/rclpy/rclpy/client.py#L113) could be `await`'d in an asyncio loop, but that future expects an rclpy executor to be spinning somewhere. I recommend removing the parallel option for now, with a plan to add it later.

hidmic: Fair enough. Want me to comment it out w/ a TODO or drop it altogether?

launch_ros/launch_ros/actions/load_composable_nodes.py
hidmic: Argh, I misunderstood you.

wjwwood:
Sorry for disrupting this pr, but since we moved several packages to ros2/launch_ros this pr will need to be moved over there. However, I cannot move pr's (unfortunately) like I did the issues.

@hidmic I assume you'll be able to make the move yourself?

hidmic:
I'll take care. I'll close this one and re-open @sloretz. I won't forget about your comments! :)

launch_ros/launch_ros/actions/load_composable_nodes.py
sloretz: Now that I'm re-reading ros2/launch#168 I see the example action is `LoadComposableNodes` plural. I have the memory of a goldfish :man_shrugging:

PR #198, opened 2019-03-11
Fix event loop closed bug

### LINKED ISSUE #185: Asyncio traceback when closing
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - from source
- Version or commit hash:
  - https://github.com/ros2/launch/commit/67c2006628044a5264452f211c73ad0ac21b8b12
- DDS implementation:
  - Fast-RTPS
  - N/A

#### Steps to reproduce issue
Example launch file with the issue:

```python
import signal

from launch.actions import EmitEvent
from launch.actions import ExecuteProcess
from launch.events.process import matches_action
from launch.events.process import SignalProcess
from launch.substitutions import FindExecutable


def run_ls():
    exec_action = ExecuteProcess(
        cmd=[FindExecutable(name='python3'), '-c', '"while(1): pass"'],
        shell=True)
    kill_timer = TimerAction(
            period=5.0,
            actions=[
                EmitEvent(
                    event=SignalProcess(
                        signal_number=signal.SIGTERM,
                        process_matcher=matches_action(exec_action)))
                ])
    ld = LaunchDescription([
        exec_action,
        kill_timer,
    ])
    ls = LaunchService(debug=True)
    ls.include_launch_description(ld)
    assert 0 == ls.run()


if __name__ == '__main__':
    run_ls()
    run_ls()
```
I ran it doing python3 _SCRIPT_NAME_

#### Expected behavior
End after the five seconds timeout without an exception.

#### Actual behavior
After ending, this error happens:

```
Exception ignored in: <bound method BaseSubprocessTransport.__del__ of <_UnixSubprocessTransport closed pid=3649 returncode=-15 stdin=<_UnixWritePipeTransport closing fd=7 open> stdout=<_UnixReadPipeTrans
port fd=8 open> stderr=<_UnixReadPipeTransport fd=10 open>>>
  File "/usr/lib/python3.6/asyncio/base_subprocess.py", line 132, in __del__
    self.close()
  File "/usr/lib/python3.6/asyncio/base_subprocess.py", line 106, in close
    proto.pipe.close()
  File "/usr/lib/python3.6/asyncio/unix_events.py", line 624, in close
    self.write_eof()
  File "/usr/lib/python3.6/asyncio/unix_events.py", line 610, in write_eof
    self._loop.call_soon(self._call_connection_lost, None)
  File "/usr/lib/python3.6/asyncio/base_events.py", line 580, in call_soon
    self._check_closed()
  File "/usr/lib/python3.6/asyncio/base_events.py", line 366, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
```

#### Additional information

This is strange, because all is working fine and an exception is being raised when closing from asyncio lib. I couldn't find a simpler repro. I had a similar problem when working with #178. I only could repro this error when using shell=True option in ExecuteProcess.

hidmic:
Fixes #185. Apparently, `atexit` registered functions get run before the python's gc runs objects destructors and thus asyncio loops were being closed before `ExecuteProcess` related awaitables were terminated.

~This PR also fixes an issue introduced by #167 on Python 3.5 (in Xenial).~

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -359,6 +359,9 @@ class ExecuteProcess(Action):
             self.__sigterm_timer.cancel()
         if self.__sigkill_timer is not None:
             self.__sigkill_timer.cancel()
+        # Close subprocess transport if any.
+        if self._subprocess_transport is not None:
+            self._subprocess_transport.close()
         # Signal that we're done to the launch system.
         self.__completed_future.set_result(None)
 


dirk-thomas:
> This PR also fixes an issue introduced by #167 on Python 3.5 (in Xenial).

This looks like a duplicate of #196 and shouldn't be necessary since Xenial / Python 3.5 isn't a supported platform on master anymore.

hidmic:
Fair enough, I'll drop that commit.

wjwwood:
This patch is definitely more defensive, but I'd be interested to see if the `atexit` handler is even needed anymore, see:

https://github.com/ros2/launch/blob/632ef1f3eca7dbc54fbed85ab65a5efc512d83ed/launch/launch/launch_service.py#L52-L55

hidmic:
> This patch is definitely more defensive, but I'd be interested to see if the atexit handler is even needed anymore

It is on Python 3.5, that bug is still there. One could argue, as Dirk did, that we're not targeting Ubuntu Xenial anymore and (transitively) neither Python 3.5, and thus we should probably drop that handler. I didn't do it to keep this PR small and the discussion brief :)

PR #200, opened 2019-03-11
fix test_env_testing test

dirk-thomas:
Before: [Build Status](https://ci.ros2.org/job/ci_linux/6369/)
After: [Build Status](https://ci.ros2.org/job/ci_linux/6370/)

Fixes ros2/build_cop#126.

### DIFF
--- a/launch_testing/test/test_env_passing.py
+++ b/launch_testing/test/test_env_passing.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import copy
 import os
 import sys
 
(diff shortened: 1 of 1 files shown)


launch_testing/test/test_env_passing.py
```diff
@@ -26,22 +26,26 @@ def test_env():
...
-                os.path.abspath(
-                    os.path.dirname(__file__)),
-                'check_env.py')],
-        name='test_env',
-        env=sub_env,
-    ))
-    launch_service = LaunchService()
-    launch_service.include_launch_description(ld)
-    return_code = launch_test.run(launch_service)
+    env = os.environ.copy()
```
wjwwood: It's odd that you can use `.copy()` here but `copy.deepcopy` is used in the test. Could you not use `.copy()` in the test (removing the need for the copy import), or if there is a reason you cannot, then should you not use `copy.deepcopy` to store the old environment too?

dirk-thomas: > It's odd that you can use .copy() here

Why is that odd? It is just a normal method on a dictionary: https://docs.python.org/3/library/stdtypes.html#dict.copy

> Could you not use .copy() in the test (removing the need for the copy import)

I am not sure I understand what you are requesting? The copy method on a `dict` is available independently of the import of the `copy` module.

> should you not use copy.deepcopy to store the old environment too?

Using shallow or deep copy doesn't make a different for `os.environ` since it maps strings to strings. If you prefer a deepcopy (in case something has mutated the `os.environ` before with non-string key/values) I can change it to use `copy.deepcopy`.

wjwwood: > > It's odd that you can use .copy() here

> Why is that odd? It is just a normal method on a dictionary: https://docs.python.org/3/library/stdtypes.html#dict.copy

If you take it out of context like that I think you will miss the point.

What I said was:

> It's odd that you can use .copy() here **but copy.deepcopy is used in the test.**

What I meant was that it's odd that it is sufficient to use `.copy()` in your additions when the test is already using a deepcopy. I was already aware that `.copy()` was a method for dictionaries.

My point was that either `.copy()` is sufficient (in which case the test should use `.copy()` as well) or there was some reason that deepcopy was used (of which I am not aware), in which case your additions would most likely need to use `deepcopy` as well.

> > Could you not use .copy() in the test (removing the need for the copy import)

> I am not sure I understand what you are requesting?

I am asking why deepcopy was used in the existing test but you decided to instead use `.copy` for essentially the same operation. And if there is no reason for the original code to use deepcopy, then I'm requesting that you use just one or the other everywhere, probably just `.copy()`.

> The copy method on a dict is available independently of the import of the copy module.

I didn't intend to imply that it was, in fact I implied the opposite.

The intended implication was that if you were able to remove the deepcopy already in the test, then you would be able to remove the `import copy` since it was not required for `dict.copy()`.

> Using shallow or deep copy doesn't make a different for os.environ since it maps strings to strings.

Based on that it sound like `.copy` is sufficient and `deepcopy` is not required.

> If you prefer a deepcopy (in case something has mutated the os.environ before with non-string key/values) I can change it to use copy.deepcopy.

I have no reason to believe deepcopy is necessary, in which case I'd suggest just using `.copy` everywhere and removing (the now unused) `import copy`.

dirk-thomas: `dict.copy()` instead of `copy.deepcopy()` it is in both places then: cfb2e94c383c868510ad9df55cdb9fdb277eba76

[Build Status](https://ci.ros2.org/job/ci_linux/6377/)

PR #202, opened 2019-03-15
Replace characters with marker when there is a decoding error

jacobperron:
Resolves ros2/build_cop#174

Linux: [Build Status](https://ci.ros2.org/view/packaging/job/packaging_linux/1400/)


jacobperron:
> Just out of curiosity: what characters are in these events which can't be decoded?

I don't actually know. I just took a stab at this solution given the error message.

Issue #208, opened 2019-03-20
merging with ApexAI/apex_rostest and repository reorganization


wjwwood:
See also: https://github.com/ApexAI/apex_rostest/issues/18

We plan to integrate the work done in https://github.com/ApexAI/apex_rostest into this repository and at the same time separate the ROS specific components of launch into a new separate repository (https://github.com/ros2/launch_ros).

This is the planned layout:

- repository ros2/launch:
  - package `launch`
  - package `launch_testing`
    - will be combined with the package `apex_launchtest` from the repository ApexAI/apex_rostest
    - will also contain a CLI for running tests, name is TBD
  - package `launch_testing_ament_cmake` (new)
    - will be based on the package `apex_launchtest_cmake` from the repository ApexAI/apex_rostest
    - will provide cmake integration like add_launch_test(...) (name TDB) which implicitly calls the test runner rather than pytest
      - in the future may optionally invoke pytest instead (still working on integration)
    - will depend on the package `launch_testing`
- repository ros2/launch_ros (new)
  - package `launch_ros`
  - package `test_launch_ros`
  - package `ros2launch`
  - package `launch_ros_testing` (new)
    - will be based on the package `apex_launchtest_ros` from the repository ApexAI/apex_rostest
    - also, may be based on the launch_testing package in ros2/launch repository as well (depends on what we find)
- repository ros2/rostest (new)
  - package `rostest` (new)
    - will contain a placeholder CLI called `rostest` (or `ros2 test`?) which will alias to the CLI in `launch_testing`
      - will allow for future ROS specific options for manual test running
    - may contain cmake macros like `add_rostest(...)` to allow for future ROS specific integration
    - will serve as the "single point of entry" for ROS 2 users, i.e. they will `<test_depend>rostest</test_depend>`
    - will contain high level documentation and examples

----

Action items:

- [x] Import ApexAI/apex_rostest into ros2/launch @wjwwood
  - [x] Import code to sub folder called `apex_rostest`
  - [x] Add `apex_rostest/AMENT_IGNORE` to prevent them from getting picked up by CI
  - [x] open pull request to this repository https://github.com/ros2/launch/pull/209
- [x] Separate ROS and ROS agnostic code
  - [x] create ros2/launch_ros @wjwwood
  - [x] create first pr on that repository with all history from this repository @wjwwood (https://github.com/ros2/launch_ros/pull/1)
    - delete packages that are not ros specific
[...]

PR #209, opened 2019-03-20
Import apex rostest

wjwwood:
See: https://github.com/ros2/launch/issues/208


wjwwood:
CI (full run):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6473/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2839/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5314/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6288/)

wjwwood:
I'll be merging this without the DCO checker passing because the commits predate our policy, however, I will ask @dejanpan and @pbaughman to just confirm in a comment, that they would have signed-off for the commits that lack the sign-off in the commit message.

apex_rostest/README.md
```diff
@@ -0,0 +1,129 @@
+# apex_launchtest
```
hidmic: Just as a reminder for our future selves, we need to find a new home for this repo level README.

apex_rostest/apex_launchtest/setup.cfg
```diff
@@ -0,0 +1,4 @@
+[coverage:run]
```
hidmic: Hmm, we don't do coverage checks, do we? Just wondering if it'd make sense to bring that improvement along with the repo.

apex_rostest/apex_launchtest/setup.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python
```
hidmic: Shouldn't be `python3` instead?

wjwwood: It probably shouldn't have a shebang.

apex_rostest/apex_launchtest/setup.cfg
wjwwood: We don't collect coverage stats right now in our CI, but it would be nice to have this at some point. I don't know, however, if putting it in the setup.cfg makes sense, as it should be possible to run it with and without. I don't know what the recommended strategy is.

wjwwood:
We should address the comments you raised when merging it with the existing code, i.e. after this import pr.

apex_rostest/apex_launchtest/setup.cfg
pbaughman: My understanding is that a bunch of different tools may look for a setup.cfg file and should each ignore configuration sections not meant for them.  If this is causing a problem with another tool that expects its setup.cfg file to be free of `[coverage:run]` sections, we could also move it to a file called `.coveragerc`

See https://coverage.readthedocs.io/en/v4.5.x/config.html for how 'coverage' uses configuration files

pbaughman:
@wjwwood - I confirm that I would've signed off on the commits that lack the sign-off in the commit message

wjwwood:
Thanks @pbaughman for the confirmation.

The CI failures are the know `test_cli.test_param_yaml` tests.

@dejanpan just waiting for you to 👍 the commits that weren't signed-off at the time.

dejanpan:
@wjwwood - I confirm that I would've signed off on the commits that lack the sign-off in the commit message

apex_rostest/apex_launchtest_cmake/cmake/add_apex_launchtest.cmake
```diff
@@ -0,0 +1,125 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#  This file contains modified code from the following open source projects
+#  published under the licenses listed below:
+
+# Software License Agreement (BSD License)
```
dirk-thomas, line 18: This doesn't match the license declared in the manifest?

Issue #126, opened 2018-08-22
LuanchService.shutdown() fails to terminate run loop


crdelsey:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 16.04
- Python Version:
 - Python 3.5.2
- Installation type:
  - source
- Version or commit hash:
  - 0.6.0
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - 

#### Steps to reproduce issue

Run the following python code
```python
#!/usr/bin/env python

import sys
import time
import threading
from launch import LaunchDescription
from launch import LaunchService
import launch_ros.actions

def main():
    ld = LaunchDescription([
        launch_ros.actions.Node(
            package='demo_nodes_cpp', node_executable='listener', output='screen'),
            ])
    ls = LaunchService()
    ls.include_launch_description(ld)
    t = threading.Thread(target=ls.run)
    t.start()
    time.sleep(5) # wait for listener to fully start
    ls.shutdown()
    t.join()

if __name__=="__main__":
    main()

```

#### Expected behavior
I expect the listener executable to terminate and the python code to exit.

#### Actual behavior
The python script produces the following output but never terminates.
```
[INFO] [launch]: process[listener-1]: started with pid [20208]
Starting shutdown
[INFO] [launch]: sending signal 'SIGINT' to process[listener-1]
[ERROR] [launch]: process[listener-1] failed to terminate '5' seconds after receiving 'SIGINT', escalating to 'SIGTERM'
[INFO] [launch]: sending signal 'SIGTERM' to process[listener-1]
[ERROR] [launch]: process[listener-1] failed to terminate '10.0' seconds after receiving 'SIGTERM', escalating to 'SIGKILL'
[INFO] [launch]: sending signal 'SIGKILL' to process[listener-1]
```

#### Additional information

The listener executable in fact terminates when the SIGINT is sent and remains as a zombie process.
```
20208 pts/20   Z+     0:00 [listener] <defunct>
```

wjwwood:
I think this issue comes from using `asyncio.subprocess` from a thread, if I run on my macOS machine with Python 3.7 I get:

```
% python3 ./test.py
[ERROR] [launch]: exception occurred while executing process[listener-1]:
Traceback (most recent call last):
  File "/Users/william/ros2_ws/build/launch/launch/actions/execute_process.py", line 417, in __execute_process
    stderr_to_stdout=(self.__output == 'screen'),
[...]

PR #210, opened 2019-03-20
Add async run method to LaunchService

hidmic:
This pull request enables asynchronous launch runs. This popped up in a discussion I had with @wjwwood as an alternative to pushing a launch run to a background thread (see #126), which is unsupported (explicitly after these changes).

Consider the following snippet for an example of the new async API:

```python
#!/usr/bin/env python3

import asyncio
import osrf_pycommon

import launch_ros.actions


def main():
    ld = LaunchDescription([
        launch_ros.actions.Node(
            package='demo_nodes_cpp',
            node_executable='listener',
            output='screen'
        ),
    ])

    loop = osrf_pycommon.process_utils.get_loop()
    ls = LaunchService(debug=True)
    ls.include_launch_description(ld)
    print(">>> Do some work")
    launch_task = loop.create_task(ls.run_async())
    print(">>> Sleep time!")
    loop.run_until_complete(asyncio.sleep(5, loop=loop))
    print(">>> I'm awake!")
    if not launch_task.done():
        print(">>> Wrapping up")
        asyncio.ensure_future(ls.shutdown(), loop=loop)
        loop.run_until_complete(launch_task)
    print(">>> Bye bye")
    return launch_task.result()


if __name__=="__main__":
    main()
```

Hopefully, this will simplify launch based fixtures for testing.

cc @pbaughman for awareness.

### DIFF
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -17,11 +17,13 @@
 import asyncio
 import atexit
 import collections.abc
+import contextlib
 import logging
 import os
 import signal
 import threading
 import traceback
+from typing import Coroutine
 from typing import Iterable
 from typing import List  # noqa: F401
 from typing import Optional
@@ -91,14 +93,14 @@ class LaunchService:
             level=logging.DEBUG if debug else logging.INFO
         )
         self.__debug = debug
+        self.__argv = argv if argv is not None else []
 
         # Setup logging
         self.__logger = launch.logging.get_logger('launch')
+
         # Install signal handlers if not already installed, will raise if not
         # in main-thread, call manually in main-thread to avoid this.
         install_signal_handlers()
-
-        self.__argv = argv if argv is not None else []
 
         # Setup context and register a built-in event handler for bootstrapping.
         self.__context = LaunchContext(argv=self.__argv)
@@ -108,10 +110,6 @@ class LaunchService:
         # Setup storage for state.
         self._entity_future_pairs = \
             []  # type: List[Tuple[LaunchDescriptionEntity, asyncio.Future]]
-
-        # Used to prevent run() being called from multiple threads.
-        self.__running_lock = threading.Lock()
-        self.__running = False
 
         # Used to allow asynchronous use of self.__loop_from_run_thread without
         # it being set to None by run() as it exits.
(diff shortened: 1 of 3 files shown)


hidmic:
Rebased to solve merge conflicts.

launch/test/launch/test_launch_service.py
```diff
@@ -14,21 +14,14 @@
 
 """Tests for the LaunchService class."""
 
+import asyncio
+import osrf_pycommon
```
wjwwood: `osrf_pycommon` isn't a system library, so it belongs in a separate block.

launch/test/launch/test_launch_service.py
```diff
@@ -72,19 +65,19 @@ class MockEvent:
...
+        handled_events.get(block=True, timeout=5.0)
+        # Emit and then check for a second event.
+        ls.emit_event(MockEvent())
+        handled_events.get(block=True, timeout=5.0)
+        # Shutdown (generates a third event) and join the thread.
+        ls.shutdown()
 
-    # First event (after including description of event handler).
-    handled_events.get(block=True, timeout=5.0)
+    threading.Thread(target=perform_test_sequence).start()
```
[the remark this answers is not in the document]
hidmic: Oops, fixed!

launch/test/launch/test_launch_service.py
hidmic: True that, and `asyncio` shouldn't be there anymore. Fixed.

hidmic:
> I am however concerned about regressions, because this code is so tricky (the current implementation included)

I share your concerns. 

> Do you want to move forward with getting this merged, or do you only want to merge it if we're going to be using it with the testing framework.

There's not point in having this PR collecting dust IMHO. If we regard this as a good addition, I think we should move forward and endure the pain, regardless of what happens with `launch_testing`. If at some point it comes down to having a stable release or not, we can always revert the patch.

hidmic:
Hmm, Windows has failing tests, but the same show up in an unrelated CI run (https://github.com/ros2/rcl/pull/477). I'll check if the issue happens on `master` already.

In the meantime, @wjwwood @dirk-thomas PTAL.

launch/launch/launch_service.py
```diff
@@ -207,187 +315,112 @@ def _is_idle(self):
...
+        asynchronous runs (see `run_async()` documentation).
 
-        return self.__return_code
+        :param: shutdown_when_idle if True (default), the service will shutdown when idle
+        """
+        # Make sure this has not been called from any thread but the main thread.
+        if threading.current_thread() is not threading.main_thread():
+            raise RuntimeError(
+                'LaunchService can only be run in the main thread.'
+            )
```
ivanpauno: run_async is already doing this check.

launch/launch/launch_service.py
```diff
@@ -176,6 +174,116 @@ def _is_idle(self):
...
+            on_sigterm(_on_sigterm)
+
+            def _on_sigquit(signum, frame, prev_handler):
+                # Ignore additional signals until we finish processing this one.
+                current_handler = signal.signal(signal.SIGQUIT, signal.SIG_IGN)
+                if current_handler is signal.SIG_IGN:
+                    # This function has been called re-entrantly.
+                    return
+                self.__logger.error('user interrupted with ctrl-\\ (SIGQUIT), terminating...')
+                _on_sigterm(signum, frame, prev_handler)
```
ivanpauno: Why is `SIGTERM` disabled when `SIGQUIT` is sent?
I know we were doing that before too.

hidmic: That is a good point, though, since the side effect of a SIGQUIT is also deferred termination of the run loop task, ignoring SIGTERM seems innocuous to me. Thoughts?

launch/launch/launch_service.py
hidmic: True that. Removed in 9b16ec3c6b3214f307b162e297f64426341d842a.

launch/launch/launch_service.py
ivanpauno: Yes, I agree it's innocuous.

hidmic:
Alright, going in.

hidmic:
I'll keep an eye on this one, just in case nightlies go red (for some unexpected reason).

PR #212, opened 2019-03-21
fix treating stderr output separate from stdout

dirk-thomas:
Without the patch the underlaying transport doesn't even open a separate `stderr` fd and therefore all output is redirected to `stdout`. As a consequence all higher level handling for `stderr` won't be effective.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6503/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2854/) (known flaky test `test_params_yaml`)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5329/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6303/) (known flaky test `test_params_yaml`)

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -465,6 +465,7 @@ class ExecuteProcess(Action):
                 env=env,
                 shell=self.__shell,
                 emulate_tty=False,
+                stderr_to_stdout=False,
             )
         except Exception:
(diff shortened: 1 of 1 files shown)


hidmic:
It would've been really nice to have a test covering the bugfix. I'll try to add one next time.

PR #218, opened 2019-03-26
Support external ExecuteProcess output overrides.

### LINKED ISSUE #214: provide an option to change the default output setting
The current default is [log](https://github.com/ros2/launch/blob/86f663db2ff878b5541bf4a51a23f04eac606f5f/launch/launch/actions/execute_process.py#L92) and defined on a per process level. It would be more convenient if the default would be `both` since the common user expects to see the output of the processes.

Independent of the default value there should be a command line option to change the global default.

Follow up of #145.

hidmic:
Fixes #214 by means of an `$OVERRIDE_LAUNCH_PROCESS_OUTPUT` envvar. 

Lacking a test to keep this PR simple, we can add one if necessary.

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -165,10 +165,11 @@ class ExecuteProcess(Action):
         :param: prefix a set of commands/arguments to preceed the cmd, used for
             things like gdb/valgrind and defaults to the LaunchConfiguration
             called 'launch-prefix'
-        :param: output configuration for process output logging. Default is 'log' i.e.
-            log both stdout and stderr to launch main log file and stderr to the screen.
-            See `launch.logging.get_output_loggers()` documentation for further reference
-            on all available options.
+        :param: output configuration for process output logging. Defaults to 'log'
+            i.e. log both stdout and stderr to launch main log file and stderr to
+            the screen. Overridden externally by the $OVERRIDE_LAUNCH_PROCESS_OUTPUT
+            envvar value. See `launch.logging.get_output_loggers()` documentation
+            for further reference on all available options.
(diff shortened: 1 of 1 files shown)


launch/launch/actions/execute_process.py
```diff
@@ -165,10 +165,11 @@ def __init__(
         :param: prefix a set of commands/arguments to preceed the cmd, used for
             things like gdb/valgrind and defaults to the LaunchConfiguration
             called 'launch-prefix'
-        :param: output configuration for process output logging. Default is 'log' i.e.
-            log both stdout and stderr to launch main log file and stderr to the screen.
-            See `launch.logging.get_output_loggers()` documentation for further reference
-            on all available options.
+        :param: output configuration for process output logging. Defaults to 'log'
+            i.e. log both stdout and stderr to launch main log file and stderr to
+            the screen. Overridden externally by the $OVERRIDE_LAUNCH_PROCESS_OUTPUT
```
dirk-thomas: Nitpick: a new sentence should start on a newline. Same below.

Same comment about the `$` as in https://github.com/ros2/launch_ros/pull/10#discussion_r269235964

hidmic: Sure, I tend to add the `$` to emphasize it's an envvar. FIxed.

wjwwood:
In the future, please run CI on changes like this, even if they seem trivial.

hidmic:
My bad! Did it fail on `master`?

wjwwood:
No, just try not to do so, in case it does. If it just touches comments, that's fine, if it touches only docs and you run tests locally that's ok too, but in a case like there where you touch Python code, definitely always run the CI, even if you think the code isn't covered by any tests.

PR #225, opened 2019-04-15
Add default value to EnvironmentVariable substitution

ivanpauno:
It will be needed for the front-end substitution `$(env env-var [default-value])`.
See [article]( https://github.com/ros2/design/blob/d3a35d7ea201721892993e85e28a5a223cdaa001/articles/151_roslaunch_xml.md).

### DIFF
--- a/launch/launch/substitutions/environment_variable.py
+++ b/launch/launch/substitutions/environment_variable.py
@@ -30,8 +30,12 @@ class EnvironmentVariable(Substitution):
     If the environment variable is not found, it returns empty string.
     """
 
-    def __init__(self, name: SomeSubstitutionsType, *,
-                 default_value: SomeSubstitutionsType = '') -> None:
+    def __init__(
+        self,
+        name: SomeSubstitutionsType,
+        *,
+        default_value: SomeSubstitutionsType = ''
+    ) -> None:
         """Constructor."""
         super().__init__()
 
@@ -46,7 +50,7 @@ class EnvironmentVariable(Substitution):
 
     @property
     def default_value(self) -> List[Substitution]:
-        """Getter for default."""
+        """Getter for default_value."""
         return self.__default_value
 
     def describe(self) -> Text:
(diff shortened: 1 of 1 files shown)


launch/launch/substitutions/environment_variable.py
```diff
@@ -30,23 +30,32 @@ class EnvironmentVariable(Substitution):
     If the environment variable is not found, it returns empty string.
     """
 
-    def __init__(self, name: SomeSubstitutionsType) -> None:
+    def __init__(self, name: SomeSubstitutionsType, *,
+                 default_value: SomeSubstitutionsType = '') -> None:
```
wjwwood: Please wrap all the arguments or none of them, also, never indent more than one time between two lines. Also for multi line function/method declarations, always end the closing `)` on its own line at the same indention level as the opening `(`.

You may have multiple arguments per line, like this:

```python
    def __init__(
        self, name: SomeSubstitutionsType, *,
        default_value: SomeSubstitutionsType = ''
    ) -> None:
```

But I would recommend doing one per line unless you have an unreasonable number of arguments with short names, e.g. `xx, xy, xz, yx, yy, yz, zx, zy, zz` or similar:

```python
    def __init__(
        self,
        name: SomeSubstitutionsType,
        *,
        default_value: SomeSubstitutionsType = ''
    ) -> None:
```

For example as to why, in the above it's easy to see that `default_value` is a required keyword because the `*,` is more visible, whereas it's easily overlooked in the current code and in the previous example.

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6698/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3014/)
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/5503/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/6529/)

ivanpauno:
Unrelated failures in CI/checks, so I'm merging.

Issue #163, opened 2019-01-14
XML/YAML Front-end


hidmic:
## Feature request

This is a feature request to implement a declarative markup (XML, YAML, TOML, etc.) front-end for launch descriptions. Acceptance criteria is limited to TBD.

#### Feature description

- Optionally narrow down feature set for a first iteration.
- Draft a format standardization document in https://github.com/ros2/design.
- Gather feedback from potential users (i.e. community and customers).
- Implement parser, from markup based description into `LaunchDescription` instances.
- Update `ros2launch` CLI to also understand declarative launchfiles using the parser.

#### Implementation considerations

**ros2launch**

Consider refactoring to ease extension through integration of external parsers (using Python modules?).

**launch_\[markup_fmt\]**

Weigh different markup formats. Some tentative XML can be found in https://github.com/ros2/design/pull/163. Some draft YAML can be found [here](https://gist.github.com/mjcarroll/1609c19433868a0fd863fbfd2a1506b1).

With the current launch API being already quite declarative, a 1-on-1 mapping may be possible (though this is still a conjecture).

dirk-thomas:
Also take the existing ROS 1 roslaunch XML into account: http://wiki.ros.org/roslaunch/XML

hidmic:
A first XML launch format specification can be found in ros2/design#207. Several approaches on how to implement launch front ends (in general, not specific to XML) are described in ros2/design#208. 

@wjwwood and I think that [ADP](https://github.com/ros2/design/pull/208/files#diff-1acde8639704c397705390e306dba4bfR190) offers the best tradeoff between complexity, scalability and extensibility, and thus we'd rather pursue that one. As a summary of what a first Python implementation of such a design would entail, consider the following todo list:

1. [ ] Define an abstract `launch_frontend.Entity` class. (~1 day for 1 FTE)
2. [ ] Implement (and likely refine base API of) a `launch_xml.Entity` class to wrap XML tags (as represented by any [standard xml parsing module](https://docs.python.org/3/library/xml.html)). Thoroughly document the mapping from XML schema to object abstraction. (~4 days for 1 FTE)
3. [ ] Implement a `launch_frontend.expose_action(*, aliases)` decorator for functions or `launch.Action` subclasses that have a `parse(*, entity: launch_frontend.Entity, parser: launch_frontend.Parser) -> launch.Action` class method that keeps the mapping from known actions aliases to parsing functions. (~1 day for 1 FTE)
[...]

PR #226, opened 2019-04-15
Add frontend module in launch, launch_xml and launch_yaml packages

ivanpauno:
See https://github.com/ros2/design/pull/207 and https://github.com/ros2/design/pull/208.
Targets all the tasks of https://github.com/ros2/launch/issues/163#issuecomment-464763427 (except the last three).

This is already functional.

TODOs (in importance order):
- Write launch_yaml readme :heavy_check_mark:
- Write more action and substitution parsing functions. :heavy_check_mark:
- Write more xml and yaml examples to exercise the written code. :heavy_check_mark:
- launch_frontend package should probably be merged with launch, as they are completely coupled.
 Action and Substitution parsing methods should be moved to the classes (see 
 https://github.com/ros2/launch_ros/pull/23) (nice to have). :heavy_check_mark:
- Create a `TestEntity` to unit test the parsing methods (nice to have).
- See if we can write a decorator that infers the parsing function from the constructor typing annotations.
 That is a really tricky tasks (nice to have).

P.S.:
`launch_frontend/launch_frontend/parse_substitution.py` (later modified by me) and `launch_frontend/launch_frontend/grammar.lark` were contributed by @hidmic.

Depends on https://github.com/ros2/launch/pull/235 (after https://github.com/ros2/launch/pull/226/commits/3c1c1e9d349e8b1f1c8fd19a0c8e2bb56c0f72a9) (already merged).


launch_xml/README.md
```diff
@@ -0,0 +1,67 @@
+# launch_xml
+
+This package provides an abstraction of the XML tree.
+
+## XML front-end mapping rules
```
hidmic: @ivanpauno we should also document the attribute lookup behavior (first check attributes in tags, then check attributes as children, then fail).

launch_frontend/launch_frontend/parser.py
```diff
@@ -0,0 +1,64 @@
...
+    shell = str_to_bool(getattr(entity, 'shell', None))
+    prefix = getattr(entity, 'launch-prefix', None)
+    output = getattr(entity, 'output', None)
+    args = getattr(entity, 'args', None)
+    args = args.split(' ') if args else []
+    if not type(args) == list:
+        args = [args]
+    # TODO(ivanpauno): How will predicates be handle in env?
+    # Substitutions aren't allowing conditions now.
+    env = get_dictionary_from_key_value_pairs(getattr(entity, 'env', None))
```
hidmic: @ivanpauno why not:
```suggestion
    env = getattr(entity, 'env', None)
    if env is not None:
       env = {e.name: e.value for e in env}
```
?

launch_frontend/launch_frontend/parser.py
```diff
@@ -0,0 +1,64 @@
...
+def parse_executable(entity: Entity):
+    """Parse executable tag."""
+    cmd = entity.cmd
+    cwd = getattr(entity, 'cwd', None)
+    name = getattr(entity, 'name', None)
+    shell = str_to_bool(getattr(entity, 'shell', None))
+    prefix = getattr(entity, 'launch-prefix', None)
+    output = getattr(entity, 'output', None)
+    args = getattr(entity, 'args', None)
+    args = args.split(' ') if args else []
```
hidmic: @ivanpauno so here we have a rather subtle issue. If an attribute isn't found during parsing, we should not be passing `None` to the constructor if it defines a suitable default for that attribute. Same care applies to `str_to_bool` coercion of a `None` variable to a `False` boolean.

launch_xml/README.md
```diff
@@ -0,0 +1,67 @@
...
+### Built-in substitutions
+
+See [this](https://github.com/ros2/design/blob/d3a35d7ea201721892993e85e28a5a223cdaa001/articles/151_roslaunch_xml.md) document.
+
+Additional substitution, for handling lists:
+
+`$(list [sep=,] a,b,c,d)`
+: Substituted by a python list, splited by the separator that follows `sep=`.
+  Default separator is `,`.
+  This substitution is evaluated instantaneously, and not in a lazy way.
```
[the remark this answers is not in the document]
ivanpauno: Yes. I think the more general way of supporting list in xml is inferring it from the text. We could have a different notation for this special case.
Another way of handling it is: when accessing the attribute `attr` if the attribute `attr-sep` exists, return `attr` split with `attr-sep` when accessing it.
e.g.:
```xml
<root>
  <tag attr="1,2,3"/>
  <tag2 attr="1,2,3" attr-sep=","/>
</root>
```
```python3
e.tag == '1,2,3'
e.tag2 == ['1', '2', '3']
```
In this last case, we should change the attribute `sep` of `param` to `value-sep` in the xml schema. In my original proposal, that attribute was not more needed.

launch_frontend/launch_frontend/parser.py
hidmic: Still applies, IMHO.

launch_xml/README.md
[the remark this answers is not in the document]
hidmic: SGTM!

launch_frontend/launch_frontend/parser.py
```diff
@@ -0,0 +1,59 @@
...
+        return True
+    raise RuntimeError('Expected "true" or "false", got {}'.format(string))
+
+
+def parse_executable(entity: Entity):
+    """Parse executable tag."""
+    cmd = entity.cmd
+    cwd = getattr(entity, 'cwd', None)
+    name = getattr(entity, 'name', None)
+    shell = str_to_bool(getattr(entity, 'shell', 'false'))
```
hidmic: @ivanpauno do we also need other `str_to_*()` functions?

ivanpauno: Probably. When handling node parameters, some kind of automatic conversion of string to int, float and bool will be needed. But it will be an special case for that parsing function.

launch_frontend/test/launch_frontend/test_substitutions.py
```diff
@@ -0,0 +1,112 @@
...
+    assert subst[3].name[0].perform(None) == 'something '
+    assert subst[3].name[1].perform(None) == '10'
+    assert subst[3].default_value[0].perform(None) == ''
+
+
+def test_double_quoted_nested_substitution():
+    subst = default_parse_substitution(
+        r'$(env "asd_bsd_qsd_$(test "asd_bds")" "$(env DEFAULT)_qsd")'
+    )
+    assert len(subst) == 1
```
hidmic: @ivanpauno missing test?

launch_xml/README.md
```diff
@@ -0,0 +1,95 @@
...
+This package provides an abstraction of the XML tree.
+
+## XML front-end mapping rules
+
+### Accessing xml attributes
+
+When having an xml tag like:
+
+```xml
+<tag attr="2"/>
```
hidmic: @ivanpauno nit: can we change `attr` to some other string to avoid `get_attr('attr', ...)` below?

launch_xml/setup.py
```diff
@@ -0,0 +1,34 @@
+from setuptools import find_packages
+from setuptools import setup
+
+setup(
+    name='launch_xml',
+    version='0.7.3',
```
hidmic: @ivanpauno same about the version.

launch_xml/test/launch_xml/let_var.xml
```diff
@@ -0,0 +1,4 @@
+<launch>
+    <let name="var1" value="asd"/>
+    <let name="var2" value="2 $(var var1)"/>
+</launch>
```
hidmic: @ivanpauno missing new line

launch_frontend/launch_frontend/parser.py
```diff
@@ -0,0 +1,111 @@
...
+    @classmethod
+    def parse_description(cls, entity: Entity) -> launch.LaunchDescription:
+        """Parse a launch description."""
+        if entity.type_name != 'launch':
+            raise RuntimeError('Expected \'launch\' as root tag')
+        actions = [cls.parse_action(child) for child in entity.children]
+        return launch.LaunchDescription(actions)
+
+    @classmethod
+    def load(
```
hidmic: @ivanpauno meta: for testing, it might make sense to have `load` take a plain string and maybe `load_from` take a file path or object. That way, tests and the description that's being validated can be kept together.

launch_yaml/README.md
```diff
@@ -0,0 +1,103 @@
...
+'str', 'int', 'float', 'bool', 'list[int]', 'list[float]', 'list[bool]', 'list[str]'
+```
+A combination of them can be specified with a tuple. e.g.: `('int', 'str')`.
+In that case, conversions are tried in order and the first successful conversion is returned.
+`types` can also be set to `guess`, which works in the same way as passing:
+
+```python
+'int', 'float', 'bool', 'list[int]', 'list[float]', 'list[bool]', 'list[str]', 'str'
+```
+
```
hidmic: @ivanpauno meta: IMHO, type deduction documentation is design document material.

launch_yaml/launch_yaml/entity.py
```diff
@@ -0,0 +1,103 @@
...
+    """Single item in the intermediate YAML front_end representation."""
+
+    def __init__(
+        self,
+        element: dict,
+        *,
+        type_name: Optional[Text] = None,
+        parent: 'Entity' = None
+    ) -> Text:
+        """Construnctor."""
```
hidmic: @ivanpauno typo `Construnctor` -> `Constructor`

launch_yaml/launch_yaml/entity.py
```diff
@@ -0,0 +1,103 @@
...
+        if type_name is None:
+            if len(element) != 1:
+                raise RuntimeError('Expected a len 1 dictionary')
+            self.__type_name = list(element.keys())[0]
+            self.__element = element[self.__type_name]
```
hidmic: @ivanpauno mind to explain the logic here a bit? Is this to cope with the root dictionary?

launch_yaml/setup.py
```diff
@@ -0,0 +1,34 @@
+from setuptools import find_packages
+from setuptools import setup
+
+setup(
+    name='launch_yaml',
+    version='0.7.3',
```
hidmic: @ivanpauno same thing about the version.

launch_yaml/test/launch_yaml/test_executable.py
```diff
@@ -0,0 +1,48 @@
...
+from pathlib import Path
+
+from launch import LaunchService
+
+from launch_frontend import Parser
+
+
+def test_executable():
+    """Parse executable yaml example."""
+    root_entity, parser = Parser.load(str(Path(__file__).parent / 'executable.yaml'))
```
hidmic: @ivanpauno same comment about parametrizing our tests by the launch description markup as in https://github.com/ros2/launch_ros/pull/23#discussion_r284335310.

launch_frontend/launch_frontend/parse_substitution.py
```diff
@@ -0,0 +1,108 @@
...
+    single_quoted_template = template
+    double_quoted_template = template
+
+
+grammar_file = os.path.join(os.path.dirname(__file__), 'grammar.lark')
+parser = Lark.open(grammar_file, start='template')
+transformer = ExtractSubstitution()
+
+
+def default_parse_substitution(string_value):
```
hidmic: Hmm we probably need someone else to review the grammar.

hidmic:
> launch_frontend package should probably be merged with launch, as they are completely coupled.
Action and Substitution parsing methods should be moved to the classes (see
ros2/launch_ros#23) (nice to have).

I fully agree and also think we need to devise a way to reuse parsing code in derived actions and substitutions.

hidmic:
> See if we can write a decorator that infers the parsing function from the constructor typing annotations.
That is a really tricky tasks (nice to have).

Yeah, that *is* tricky. I'll try to take a stab at it at some point, should be possible though.

launch_frontend/launch_frontend/action_parse_methods.py
```diff
@@ -0,0 +1,117 @@
...
+        raise RuntimeError("if and unless conditions can't be usede simultaneously")
+    if if_cond is not None:
+        kwargs['condition'] = IfCondition(predicate_expression=if_cond)
+    if unless_cond is not None:
+        kwargs['condition'] = UnlessCondition(predicate_expression=unless_cond)
+
+    return launch.actions.ExecuteProcess(
+        cmd=cmd_list,
+        **kwargs
+    )
```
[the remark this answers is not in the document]
ivanpauno: If we finally modify the `expose_*` decorators to automatically infer the parsing function from the constructor, this will automatically be solved (i.e.: the decorator should not only check the type annotation of the constructor arguments, but it should also check the derived class constructor arguments).
For the moment, I will let this as-is.

launch_frontend/launch_frontend/type_utils.py
```diff
@@ -0,0 +1,155 @@
...
+
+"""Module which implements get_typed_value function."""
+
+from typing import Any
+from typing import Text
+from typing import Tuple
+from typing import Union
+
+
+types_guess__ = (
```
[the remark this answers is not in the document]
ivanpauno: These aren't exactly the supported types, but how `types='guess'` works.
I would let it as-is.

launch_frontend/launch_frontend/type_utils.py
```diff
@@ -0,0 +1,155 @@
...
+    elif isinstance(types, Text):
+        types = [types]
+
+    typed_value = None
+    for x in types:
+        type_obj, is_list = extract_type(x)
+        if type_obj is bool:
+            def type_obj(x):
+                if x.lower() in ('true', 'false'):
+                    return x.lower() == 'true'
```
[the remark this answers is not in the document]
ivanpauno: 1 and 0 too

launch_frontend/launch_frontend/type_utils.py
```diff
@@ -0,0 +1,155 @@
...
+        types = types_guess__
+    elif isinstance(types, Text):
+        types = [types]
+
+    typed_value = None
+    for x in types:
+        type_obj, is_list = extract_type(x)
+        if type_obj is bool:
+            def type_obj(x):
+                if x.lower() in ('true', 'false'):
```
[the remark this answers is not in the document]
ivanpauno: `myfalsething` raises `ValueError` and doesn't qualify as `False`.

launch_frontend/test/launch_frontend/test_substitutions.py
ivanpauno: Thanks! It was actually a problematic case

launch_frontend/launch_frontend/parser.py
ivanpauno: If I'm understanding well, you want to be able to pass the xml document in an string (and not the file name). Am I right?
Unfortunately, in some cases (like xml), I will have to dump it in a file and read it again ... but that's not much problem.

launch_yaml/README.md
ivanpauno: If you want, I can update https://github.com/ros2/design/pull/208 and https://github.com/ros2/design/pull/207, and create a new PR for yaml design doc.

launch_yaml/launch_yaml/entity.py
ivanpauno: Yeah this is a little confusing ...
- The usual loading is with `type_name=None`. Each yaml launch description is expected to only have one root element, if not an error is raised. Then, I store the (only) key as the name, and the value as the element tree. Having this constructor is also a good idea when using the `children` property, if one of the children is ill formed an error will be raised.
- The "unusual" loading, is with `typename='somestringhere'`. When using `get_attr('sth', types='list[Entity]')`, you have a dict with a list inside. Each of the list elements has to be loaded with the same type_name (the key of the dict).

The real solution would be to always specify the `type_name` argument, and have some formal checking before to check if the yaml file is ill formed or not.

launch_frontend/test/launch_frontend/test_substitutions.py
```diff
@@ -0,0 +1,136 @@
...
+    assert len(subst[0].default_value) == 2
+    assert subst[0].default_value[0].name[0].perform(context) == 'DEFAULT'
+    assert subst[0].default_value[0].default_value[0].perform(context) == ''
+    assert subst[0].default_value[1].perform(context) == '_qsd'
+
+
+def test_combining_quotes_nested_substitution():
+    subst = default_parse_substitution(
+        '$(env "asd_bsd_qsd_$(test \'asd_bds\')" \'$(env DEFAULT)_qsd\')'
+    )
```
ivanpauno: @hidmic I added this new test case. For having a more uniform behavior, I think this should look like:
`'$(env "asd_bsd_qsd_$(test \\\'asd_bds\\\')" \'$(env DEFAULT)_qsd\')'`
(this is not a raw a string, I'm escaping both the \ and ')

For that, I would change the following in the grammar:
```
SINGLE_QUOTED_STRING: (/[^'\\"$]|\$(?=!\()|(?<=\\)\$/ | "\\\"" | "\\'" | "\\")+
SINGLE_QUOTED_RSTRING: (/[^ '\\"$\(\)]|(?<=\\)\(|(?<=\\)\)|\$(?=!\()|(?<=\\)\$/ | "\\\"" | "\\'" | "\\")+

DOUBLE_QUOTED_STRING: (/[^"\\'$]|\$(?=!\()|(?<=\\)\$/ | "\\\"" | "\\'" | "\\")+
DOUBLE_QUOTED_RSTRING: (/[^ "\\'$\(\)]|(?<=\\)\(|(?<=\\)\)|\$(?=!\()|(?<=\\)\$/ | "\\\"" | "\\'" | "\\")+
```

Basically, I'm stopping allowing any kind of quote inside `SINGLE_QUOTED_*` and `DOUBLE_QUOTED_*`, and only allowing it if it is escaped.

Do you agree with that change?

launch_frontend/launch_frontend/action_parse_methods.py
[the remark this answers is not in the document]
ivanpauno: Finally, I have corrected it!

launch_yaml/launch_yaml/entity.py
ivanpauno: I finally changed it, enforcing passing `type_name`.

launch_frontend/launch_frontend/parser.py
ivanpauno: I still only have one load method, but I put all the descriptions in place in the tests now.

launch_yaml/test/launch_yaml/test_executable.py
ivanpauno: But, where should I put them? Is ok putting them in launch_frontend? And having launch_xml and launch_yaml as test dependencies.
Or maybe, a new package only for the tests.

launch_frontend/test/launch_frontend/test_substitutions.py
hidmic: I'm fine with it, though it may be counter intuitive for some accustomed to combine quotes. If we do it, we should clearly document it somewhere.  @wjwwood any thoughts on this i.e. on forcing quote escaping everywhere if one wants to keep it raw?

launch_yaml/README.md
hidmic: Eventually, yeah, but let's hold that for a minute. At least until we've settled here.

launch_frontend/launch_frontend/action_parse_methods.py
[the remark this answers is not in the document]
hidmic: > Finally, I have corrected it!

Awesome! And yes, automatic inference may reduce the need for this feature somewhat but we don't plan to work on it soon and we don't know if it's feasible for every use case.

launch_yaml/test/launch_yaml/test_executable.py
hidmic: Hmm, good point. Ok, nevermind my comment, it makes it more obscure than what it's worth.

ivanpauno:
Rebased with master again.

launch_frontend/launch_frontend/parser.py
```diff
@@ -0,0 +1,113 @@
...
+
+    extensions_loaded = False
+    frontend_parsers = None
+
+    @classmethod
+    def load_parser_extensions(cls):
+        """Load parser extension, in order to get all the exposed substitutions and actions."""
+        if cls.extensions_loaded is False:
+            for entry_point in iter_entry_points('launch_frontend.parser_extension'):
+                entry_point.load()
```
hidmic: @ivanpauno meta: we can leave it as is for a first iteration, but I'm not entirely convinced that working with importation side effects is in our best interest in the long run. I wonder if having some sort of index at each package root module that its classes can expose/register themselves into would be cleaner than a having a global one. Maybe @wjwwood can weigh in too on this.

launch_frontend/launch_frontend/action_parse_methods.py
```diff
@@ -0,0 +1,131 @@
...
+    # Conditions won't be allowed in the `env` tag.
+    # If that feature is needed, `set_enviroment_variable` and
+    # `unset_enviroment_variable` actions should be used.
+    env = entity.get_attr('env', types='list[Entity]', optional=True)
+    if env is not None:
+        # TODO(ivanpauno): Change `ExecuteProcess` api. `additional_env`
+        # argument is supposed to be a dictionary with `SomeSubstitutionType`
+        # keys, but `SomeSubstitutionType` is not always hashable.
+        # Proposed `additional_env` type:
+        #   Iterable[Tuple[SomeSubstitutionType, SomeSubstitutionsType]]
```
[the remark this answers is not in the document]
ivanpauno: See https://github.com/ros2/launch/issues/260.

launch_frontend/launch_frontend/entity.py
```diff
@@ -0,0 +1,97 @@
...
+            - 'list[int]'
+            - 'list[float]'
+            - 'list[bool]'
+
+        Types that can not be combined with the others:
+            - 'guess'
+            - 'list[Entity]'
+
+        'guess' work in the same way as:
+            ('int', 'float', 'bool', 'list[int]', 'list[float]', 'list[bool]', 'list[str]', 'str')
```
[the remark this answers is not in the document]
ivanpauno: Because the type that is specified most of the times is `str` (substitutions are `str`).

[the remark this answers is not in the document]
hidmic: Same, making `str` the default.

launch_frontend/launch_frontend/parse_substitution.py
```diff
@@ -0,0 +1,108 @@
...
+    single_quoted_arguments = arguments
+    double_quoted_arguments = arguments
+
+    def substitution(self, args):
+        assert len(args) >= 1
+        name = args[0]
+        assert isinstance(name, Token)
+        assert name.type == 'IDENTIFIER'
+        # TODO(hidmic): Lookup and instantiate Substitution
+        if name.value not in substitution_parse_methods:
```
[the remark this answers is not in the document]
ivanpauno: Could you expand your idea? I don't understand the benefits of using an `instantiate_substitution` function (it will finally access the internal data).

launch_frontend/package.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>launch_frontend</name>
+  <version>0.8.1</version>
+  <description>The ROS launch frontend.</description>
```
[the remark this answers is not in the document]
ivanpauno: https://github.com/ros2/launch/blob/188c773482a255088b084f7bc18c6dd5b9e05f31/launch/package.xml#L4-L6

launch_frontend/launch_frontend/parser.py
```diff
@@ -0,0 +1,113 @@
...
+        action, kwargs = action_parse_methods[entity.type_name](entity, cls)
+        return action(**kwargs)
+
+    @classmethod
+    def parse_substitution(cls, value: Text) -> launch.SomeSubstitutionsType:
+        """Parse a substitution."""
+        return default_parse_substitution(value)
+
+    @classmethod
+    def parse_description(cls, entity: Entity) -> launch.LaunchDescription:
```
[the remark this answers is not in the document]
ivanpauno: `parse_description`, `parse_action` should. `cls` is finally passed to the action parsing methods as `parser` argument. The parsing methods could call `parse.substitution`, which can be overridden by the derived class.

launch_frontend/launch_frontend/parser.py
```diff
@@ -0,0 +1,113 @@
...
+        """Parse a launch description."""
+        if entity.type_name != 'launch':
+            raise RuntimeError('Expected \'launch\' as root tag')
+        actions = [cls.parse_action(child) for child in entity.children]
+        return launch.LaunchDescription(actions)
+
+    @classmethod
+    def load(
+        cls,
+        file: Union[str, io.TextIOBase],
```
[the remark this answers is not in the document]
ivanpauno: `file` isn't a builtin in python3 (it was in python 2).

launch_frontend/launch_frontend/type_utils.py
```diff
@@ -0,0 +1,155 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
```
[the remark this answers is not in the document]
ivanpauno: The problem is with `'list[int]'`, etc. Using `typing` objects doesn't help much in the implementation (i.e.: I don't have an easy way to extract if it was a list, and what is the item type).

ivanpauno:
@hidmic Addressed most of your comments in https://github.com/ros2/launch/pull/226/commits/61093a53a83a391043d0ac252a3445ea522cbca7.
Also, rebased with master.

launch_frontend/launch_frontend/parse_substitution.py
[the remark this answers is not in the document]
hidmic: It will access data structures eventually, yes, but you stabilize an API in case anything changes in the future (e.g. how we instantiate, how lookup, what the data structure is, etc.).

launch_frontend/launch_frontend/action_parse_methods.py
```diff
@@ -0,0 +1,126 @@
...
+    # `unset_enviroment_variable` actions should be used.
+    env = entity.get_attr('env', types='list[Entity]', optional=True)
+    if env is not None:
+        env = {e.get_attr('name'): parser.parse_substitution(e.get_attr('value')) for e in env}
+        kwargs['additional_env'] = env
+    args = entity.get_attr('args', optional=True)
+    # `args` is supposed to be a list separated with ' '.
+    # All the found `TextSubstitution` items are split and
+    # added to the list again as a `TextSubstitution`.
+    # TODO(ivanpauno): Change `ExecuteProcess` api from accepting
```
hidmic: @ivanpauno same as the other TODO, I think we should file this or it'll get lost.

launch_frontend/launch_frontend/entity.py
```diff
@@ -0,0 +1,96 @@
...
+        Text,
+        int,
+        float,
+        List[Text],
+        List[int],
+        List[float],
+        List['Entity']
+    ]]:
+        """
+        Access an element of the entity.
```
hidmic: @ivanpauno `element` -> `attribute`?

launch_frontend/launch_frontend/entity.py
```diff
@@ -0,0 +1,96 @@
...
+
+        If `optional` argument is `True`, will return `None` instead of raising `AttributeError`.
+
+        :param name: name of the attribute
+        :param types: type of the attribute to be read. Default to 'str'
+        :param optional: when `True`, it doesn't raise an error when the attribute is not found.
+            It returns `None` instead. Defaults to `False`
+        :raises `AttributeError`: Attribute not found. Only possible if `optional` is `False`
+        :raises `TypeError`: Attribute found but it is not of the correct type.
+            Only happens in froentends that do type checking
```
hidmic: @ivanpauno consider `froentends` -> `frontend implementations`.

launch_frontend/launch_frontend/entity.py
```diff
@@ -0,0 +1,96 @@
...
+        :raises `ValueError`: Attribute found but can't be coerced to one of the types.
+            Only happens in froentends that do type coercion
```
hidmic: @ivanpauno consider `froentends` -> `frontend implementations`.

launch_frontend/launch_frontend/parse_substitution.py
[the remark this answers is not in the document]
hidmic: Still applies.

launch_frontend/launch_frontend/parser.py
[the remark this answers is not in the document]
hidmic: Ok, fair enough.

launch_frontend/launch_frontend/parser.py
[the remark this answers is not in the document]
hidmic: Right, my point being that these should be instance methods.

launch_frontend/launch_frontend/type_utils.py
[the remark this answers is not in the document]
hidmic: Right, but you can check for equality just like you do with strings. And then you don't have to do substring checks.

launch_frontend/package.xml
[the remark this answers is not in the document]
hidmic: That's simply wrong :)

launch_frontend/launch_frontend/parser.py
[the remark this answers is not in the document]
ivanpauno: The `Parse` instances don't have a state now, but it could have it in the future, so I think it is a good idea. I will change `parse_action`, `parse_substitution`, `parse_description` to instance methods.
`load`, `load_parser_extensions` and `load_parser_implementations` will be continue being class methods.
I will stop returning an object of `Parse` type in the `load` method (ie: a subclass of `Parse`) , and start returning an instance of it.
Note: I will also rename `load_parser_extensions` as `load_launch_extensions`.

launch/launch/launch_description_sources/any_launch_description_source.py
```diff
@@ -0,0 +1,69 @@
...
+from .python_launch_file_utilities import InvalidPythonLaunchFileError
+from ..launch_description_source import LaunchDescriptionSource
+from ..some_substitutions_type import SomeSubstitutionsType
+
+
+class AnyLaunchDescriptionSource(LaunchDescriptionSource):
+    """
+    Encapsulation of a launch file, which can be loaded during launch.
+
+    This try to load the file as a python launch file, or as a frontend launch file.
```
hidmic: @ivanpauno consider:
```suggestion
    This launch description source will attempt to load the file at the given location as a python launch file first, and as a declarative (markup based) launch file if the former fails.

File: launch/launch/launch_description_sources/any_launch_description_source.py [hunk 35]
```diff
@@ -0,0 +1,69 @@
...
+        Constructor.
+
+        The given file path should be to a launch file of any style.
+        The path should probably be absolute, since the current working
+        directory will be wherever the launch file was run from, which might
+        change depending on the situation.
+        The path can be made up of Substitution instances which are expanded
+        when :py:meth:`get_launch_description()` is called.
+
+        :param launch_file_path: the path to the launch file
```
hidmic: @ivanpauno missing period at the end of the sentence.

launch/launch/launch_description_sources/any_launch_description_source.py
```diff
@@ -0,0 +1,69 @@
...
+    ) -> None:
+        """
+        Constructor.
+
+        The given file path should be to a launch file of any style.
+        The path should probably be absolute, since the current working
+        directory will be wherever the launch file was run from, which might
+        change depending on the situation.
+        The path can be made up of Substitution instances which are expanded
+        when :py:meth:`get_launch_description()` is called.
```
hidmic: @ivanpauno consider moving this all these details to the parameter documentation itself.

launch/launch/launch_description_sources/any_launch_description_source.py
```diff
@@ -0,0 +1,69 @@
...
+        self,
+        launch_file_path: SomeSubstitutionsType,
+    ) -> None:
+        """
+        Constructor.
+
+        The given file path should be to a launch file of any style.
+        The path should probably be absolute, since the current working
+        directory will be wherever the launch file was run from, which might
+        change depending on the situation.
```
hidmic: @ivanpauno hmm, why not simply stating the outcome of passing a relative path?

launch/launch/launch_description_sources/frontend_launch_description_source.py
```diff
@@ -0,0 +1,57 @@
...
+    ) -> None:
+        """
+        Constructor.
+
+        The given file path should be to a launch frontend style file (like xml or yaml).
+        The path should probably be absolute, since the current working
+        directory will be wherever the launch file was run from, which might
+        change depending on the situation.
+        The path can be made up of Substitution instances which are expanded
+        when :py:meth:`get_launch_description()` is called.
```
hidmic: @ivanpauno same as above.

launch/launch/launch_description_sources/xml_launch_description_source.py
```diff
@@ -0,0 +1,46 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
```
hidmic: @ivanpauno hmm, shouldn't this source live in `launch_xml`?

launch/launch/launch_description_sources/yaml_launch_description_source.py
```diff
@@ -0,0 +1,46 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
```
hidmic: @ivanpauno same, shouldn't this source live in `launch_yaml`?

launch/launch/launch_description_sources/yaml_launch_description_source.py
```diff
@@ -0,0 +1,46 @@
...
+    ) -> None:
+        """
+        Constructor.
+
+        The given file path should be to a launch YAML style file (`.launch.yaml`).
+        The path should probably be absolute, since the current working
+        directory will be wherever the launch file was run from, which might
+        change depending on the situation.
+        The path can be made up of Substitution instances which are expanded
+        when :py:meth:`get_launch_description()` is called.
```
hidmic: @ivanpauno same on documentation.

launch_frontend/launch_frontend/action_parse_methods.py
```diff
@@ -0,0 +1,155 @@
...
+from .entity import Entity
+from .expose import expose_action
+from .parser import Parser
+
+
+def parse_action(entity: Entity, parser: Parser):
+    """
+    Parse action.
+
+    This is only intented for code reusage, and it's not exposed.
```
hidmic: @ivanpauno consider:
```suggestion
    This is only intended for code reuse, and it's not exposed.
```

launch_frontend/launch_frontend/action_parse_methods.py
```diff
@@ -0,0 +1,155 @@
...
+    args = entity.get_attr('arg', types='list[Entity]', optional=True)
+    if args is not None:
+        args = [
+            (
+                parser.parse_substitution(e.get_attr('name')),
+                parser.parse_substitution(e.get_attr('value'))
+            )
+            for e in args
+        ]
+        kwargs['launch_arguments'] = args
```
hidmic: @ivanpauno consider collapsing above two lines in one:

```python
        kwargs['launch_arguments'] = [
            (
                parser.parse_substitution(e.get_attr('name')),
                parser.parse_substitution(e.get_attr('value'))
            )
            for e in args
        ]
```

launch_frontend/launch_frontend/parse_substitution.py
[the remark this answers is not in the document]
hidmic: I'm fine with bringing another reviewer but I still think this applies.

launch_frontend/launch_frontend/type_utils.py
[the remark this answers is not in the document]
hidmic: I still wonder if it wouldn't be better to use `typing` objects instead of plain `str`ings.

launch_xml/test/launch_xml/test_executable.py
```diff
@@ -0,0 +1,58 @@
...
+    assert(executable.cwd[0].perform(None) == '/')
+    assert(executable.name[0].perform(None) == 'my_ls')
+    assert(executable.shell is True)
+    assert(executable.output == 'log')
+    key, value = executable.additional_env[0]
+    key = key[0].perform(None)
+    value = value[0].perform(None)
+    assert(key == 'var')
+    assert(value == '1')
+    # assert(executable.prefix[0].perform(None) == 'time')
```
hidmic: @ivanpauno remove?

launch_xml/test/launch_xml/test_include.py
```diff
@@ -0,0 +1,44 @@
...
+            <include file="some_file"/>
+        </launch>
+        """  # noqa: E501
+    xml_file = textwrap.dedent(xml_file)
+    root_entity, parser = Parser.load(io.StringIO(xml_file))
+    ld = parser.parse_description(root_entity)
+    include = ld.entities[0]
+    assert isinstance(include, IncludeLaunchDescription)
+    assert isinstance(include.launch_description_source, AnyLaunchDescriptionSource)
+    # TODO(ivanpauno): Load something really
```
hidmic: @ivanpauno do you think we can do it for this PR?

launch_yaml/launch_yaml/parser.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @classmethod
+    def load(
+        cls,
+        stream: Union[str, io.TextIOBase],
+    ) -> (Entity, 'Parser'):
+        """Load a YAML launch file."""
+        if isinstance(stream, str):
+            stream = open(stream, 'r')
+        """Return entity loaded with markup file."""
```
hidmic: @ivanpauno hmm, a leftover?

launch_xml/test/launch_xml/test_include.py
wjwwood: It would be good to see it work.

launch_frontend/launch_frontend/parser.py
hidmic: It's not so much about performance (though I had not event considered that, thanks for the reference) but the way actions, substitutions and parsers are discovered. It's currently loading a global collection inside `launch` upon entry point loading. 

I was wondering if it wouldn't make sense to move away from that and make exposure somewhat more explicit, and maybe introducing namespaces in the process (which we do not have, we just fail upon tag name collision). Anyways, just a thought. No need to address this now, really.

launch_frontend/launch_frontend/type_utils.py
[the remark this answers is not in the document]
hidmic: So, between using `typing` classes and using a stringification of their names, I'm slightly inclined towards the first one. Collection type information is readily available and the API looks cleaner that way IMO. No need to do substring checks anymore.

launch/launch/__init__.py
```diff
@@ -38,7 +40,9 @@
     'actions',
     'conditions',
     'events',
+    'launch_frontend',
```
hidmic: @ivanpauno can't we just call it `frontend`? To go from `launch.launch_frontend` to `launch.frontend`.

launch/launch/actions/execute_process.py
```diff
@@ -225,6 +230,71 @@ def __init__(
...
+                    text = [TextSubstitution(text=item) for item in text]
+                    new_args.extend(text)
+                else:
+                    new_args.append(arg)
+            args = new_args
+        else:
+            args = []
+        cmd_list.extend(args)
+        kwargs['cmd'] = cmd_list
+        _, action_kwargs = super(ExecuteProcess, ExecuteProcess).parse(entity, parser)
```
hidmic: @ivanpauno consider moving this to the top instead of default initializing `kwargs`.

launch/launch/actions/execute_process.py
```diff
@@ -225,6 +230,71 @@ def __init__(
...
+        # added to the list again as a `TextSubstitution`.
+        if args is not None:
+            args = parser.parse_substitution(args)
+            new_args = []
+            for arg in args:
+                if isinstance(arg, TextSubstitution):
+                    text = arg.text
+                    text = shlex.split(text)
+                    text = [TextSubstitution(text=item) for item in text]
+                    new_args.extend(text)
```
hidmic: @ivanpauno can't we collapse this as:

```python
new_args.extend([TextSubstitution(text=token) for token in shlex.split(args.text)])
```

launch/launch/actions/execute_process.py
```diff
@@ -225,6 +230,71 @@ def __init__(
         self.__stdout_buffer = io.StringIO()
         self.__stderr_buffer = io.StringIO()
 
+    @staticmethod
+    def parse(
+        entity: Entity,
+        parser: Parser,
+        optional_cmd: bool = False
```
hidmic: @ivanpauno meta: let's revisit this mechanism for reuse after this PR.

launch/launch/actions/group_action.py
```diff
@@ -54,6 +58,16 @@ def __init__(
         else:
             self.__launch_configurations = {}
 
+    @staticmethod
+    def parse(entity: Entity, parser: Parser):
+        """Return `GroupAction` action and kwargs for constructing it."""
+        _, kwargs = super(GroupAction, GroupAction).parse(entity, parser)
```
hidmic: @ivanpauno I think I've seen enough of this idiom to think it'd be better to use class methods, which then allows:
```python
super(GroupAction, cls).parse(entity, parser)
```
which IMHO is slightly less terse.

launch/launch/actions/include_launch_description.py
```diff
@@ -69,6 +74,26 @@ def __init__(
...
         self.__launch_arguments = launch_arguments
 
+    @staticmethod
+    def parse(entity: Entity, parser: Parser):
+        """Return `IncludeLaunchDescription` action and kwargs for constructing it."""
+        _, kwargs = super(
+            IncludeLaunchDescription,
+            IncludeLaunchDescription
+        ).parse(entity, parser)
+        file = parser.parse_substitution(entity.get_attr('file'))
```
hidmic: @ivanpauno nit: I know I brought this up before, and that `file` is no longer a Python keyword, but I can count 3 syntax highlighters that think otherwise. Can we rename it?

launch/launch/launch_description_sources/any_launch_description_source.py
```diff
@@ -0,0 +1,66 @@
...
+from ..some_substitutions_type import SomeSubstitutionsType
+
+
+class AnyLaunchDescriptionSource(LaunchDescriptionSource):
+    """
+    Encapsulation of a launch file, which can be loaded during launch.
+
+    This launch description source will attempt to load the file at the given location as a python
+    launch file first, and as a declarative (markup based) launch file if the former fails.
+    It is recommended to use a specific `LaunchDescriptionSource` subclasses when possible.
```
hidmic: @ivanpauno consider
```suggestion
    It is recommended to use specific `LaunchDescriptionSource` subclasses when possible.
```

launch/launch/launch_description_sources/any_launch_description_source.py
```diff
@@ -0,0 +1,66 @@
...
+            launch_description = get_launch_description_from_python_launch_file(location)
+        except (InvalidPythonLaunchFileError, SyntaxError):
+            pass
+        try:
+            root_entity, parser = Parser.load(location)
+            launch_description = parser.parse_description(root_entity)
+        except RuntimeError:
+            pass
+        if launch_description is None:
+            raise RuntimeError('Can not load launch file')
```
hidmic: @ivanpauno consider 
```suggestion
            raise RuntimeError('Cannot load launch file')
```

launch/launch/launch_frontend/entity.py
```diff
@@ -0,0 +1,104 @@
...
+            - `bool`
+            - `List[str]`
+            - `List[int]`
+            - `List[float]`
+            - `List[bool]`
+
+        Types that can not be combined with the others:
+            - `List[Entity]`
+
+        `types = None` work in the same way as:
```
hidmic: @ivanpauno consider
```suggestion
        `types=None` implies `types=(...)`
```

launch/launch/launch_frontend/entity.py
```diff
@@ -0,0 +1,104 @@
...
+            `(int, float, bool, List[int], List[float], List[bool], List[str], str)`
+        `List[Entity]` will return a list of more entities.
```
hidmic: @ivanpauno consider
```suggestion
        `List[Entity]` will return the list of entities that are identified by the given attribute.
```

launch/launch/launch_frontend/entity.py
```diff
@@ -0,0 +1,104 @@
...
+
+        See the frontend documentation to see how `list` and `List[Entity]` look like for each
+        frontend implementation.
+
+        If `optional` argument is `True`, will return `None` instead of raising `AttributeError`.
```
hidmic: @ivanpauno consider
```suggestion
        If `optional` is `True` and the attribute cannot be found, `None` will be returned instead of raising `AttributeError`.
```

launch/launch/substitutions/environment_variable.py
```diff
@@ -43,6 +46,17 @@ def __init__(
         self.__name = normalize_to_list_of_substitutions(name)
         self.__default_value = normalize_to_list_of_substitutions(default_value)
 
+    @staticmethod
+    def parse(data: Iterable[SomeSubstitutionsType]):
+        """Parse `EnviromentVariable` substitution."""
+        if not data or len(data) > 2:
+            raise TypeError('env substitution expects 1 or 2 arguments')
+        kwargs = {}
+        kwargs['name'] = data[0]
```
hidmic: @ivanpauno nit: you can also do

```python
kwargs = {'name': data[0]}
```

launch/launch/substitutions/find_executable.py
```diff
@@ -39,6 +42,15 @@ def __init__(self, *, name: SomeSubstitutionsType) -> None:
         from ..utilities import normalize_to_list_of_substitutions  # import here to avoid loop
         self.__name = normalize_to_list_of_substitutions(name)
 
+    @staticmethod
+    def parse(data: Iterable[SomeSubstitutionsType]):
+        """Parse `FindExecutable` substitution."""
+        if not data or len(data) > 1:
```
hidmic: @ivanpauno consider
```suggestion
        if len(data) != 1:
```

launch/launch/substitutions/find_executable.py
```diff
@@ -39,6 +42,15 @@ def __init__(self, *, name: SomeSubstitutionsType) -> None:
...
+            raise AttributeError('find-exec substitution expects 1 argument')
+        kwargs = {}
+        kwargs['name'] = data[0]
+        return FindExecutable, kwargs
```
hidmic: @ivanpauno nit: you can also just do

```python
return FindExecutable, {'name': data[0]}
```

launch/launch/substitutions/launch_configuration.py
```diff
@@ -62,6 +64,17 @@ def __init__(
                 normalize_to_list_of_substitutions(
                     str_normalized_default)  # type: List[Substitution]
 
+    @staticmethod
+    def parse(data: Iterable[SomeSubstitutionsType]):
+        """Parse `FindExecutable` substitution."""
+        if not data or len(data) > 2:
```
hidmic: @ivanpauno consider
```suggestion
        if len(data) < 1 or len(data) > 2:
```

launch_xml/launch_xml/entity.py
```diff
@@ -0,0 +1,117 @@
...
+            else:
+                sep = self.__xml_element.attrib[name_sep]
+                value = self.__xml_element.attrib[name].split(sep)
+        if value is None:
+            if not optional:
+                raise attr_error
+            else:
+                return None
+        try:
+            value = get_typed_value(value, types)
```
hidmic: @ivanpauno hmm why are we mixing YAML and XML here?

launch_xml/test/launch_xml/executable.xml
```diff
@@ -0,0 +1,5 @@
+<launch>
+    <executable cmd="ls" cwd="/" name="my_ls" args="-l -a -s" shell="true" output='log' launch-prefix='$(env LAUNCH_PREFIX)'>
```
hidmic: @ivanpauno nit: I wonder if just collapsing `args` into `cmd` wouldn't be simpler: `ls -l -a -s`

launch/launch/actions/execute_process.py
ivanpauno: is not the same

launch/launch/actions/group_action.py
ivanpauno: Yeah, I don't know why I didn't use classmethods.
You can use `super()` without any arguments in them, which looks better.

launch_xml/launch_xml/entity.py
ivanpauno: I'm only doing the conversions from string to int/float/bool like yaml, and allowing the yaml ways of escaping them. If not it was imposible to specify an string param of value "1" (now you can do that passing "'1'".

Maybe this is not the best solution, if you have any other ideas please tell me.

Maybe, one option is to add an optional type field in `param`. If you don't use it, the automatic type conversions are done as before ("1" is an int "'1'" doesn't escape the quotes, you see them).
If you want a string param containing 1 you could do:
```xml
<param name="asd" value="1" type="str"/>
```

I like this last option. It should be possible to use the `type` attribute in yaml too, but it hasn't much sense for the user.

hidmic: My point is that with `get_typed_value()`'s current implementation, you can pass a list of floats using `[1, 2, 3]`. And then we also have value separators attributes.

ivanpauno: I restricted that conversion in https://github.com/ros2/launch/pull/226/commits/18cf4a06acaa0b5002e7b5c5f4fe90e03b08651d.
For parameters inside substitutions, you can still pass a list like yaml (becuase of https://github.com/ros2/launch_ros/pull/31).
So this example is still working:
https://github.com/ros2/launch_ros/blob/713fe0034d89c412dd0efbd327ce31aecdb359c0/test_launch_ros/test/test_launch_ros/launch_frontend/test_node_frontend.py#L31L50

ivanpauno:
Finally, CI (only fastrtps, up to launch launch_ros launch_yaml launch_xml test_launch_ros):
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/7492/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3630/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6142/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7322/)

ivanpauno:
Again, now that CI problems have been solved:
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/7505/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3645/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6150/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7331/)

ivanpauno:
Checking if windows is happy after https://github.com/ros2/launch/pull/226/commits/f15b0f21739d9c2b4514a658e73287238d7642c9:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7332/)

ivanpauno:
CI ...
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/7512/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3648/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6152/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7335/)

launch/launch/frontend/entity.py
```diff
@@ -0,0 +1,87 @@
...
+        `types` states the expected types of the attribute. Type coercion or type checking is
+        applied depending on the particular frontend.
+
+        The allowed types are:
+            - a scalar type: `str`, `int`, `float`, `bool`
+            - a uniform list: `List[str]`, `List[int]`, `List[float]`, `List[bool]`
+            - a non-uniform list.
+                Can specify some of the scalar types for its members, e.g.: List[Union[int, str]].
+                `list` or `List`, means that any of the scalar types are allowed.
+            - an union of the above.
```
hidmic: @ivanpauno consider: 

```
The allowed types are:
  - a scalar type i.e. `str`, `int`, `float`, `bool`;
  - a uniform list i.e `List[str]`, `List[int]`, `List[float]`, `List[bool]`;
  - a non-uniform list of known scalar types e.g. `List[Union[int, str]]`;
  - a non-uniform list of any scalar type i.e. `list` or `List`;
  - a `Union` of any of the above.
```

launch/launch/frontend/entity.py
```diff
@@ -0,0 +1,87 @@
...
+
+        `types = None` works in the same way as:
+            `Union[int, float, bool, list, str]`
+
+        See the frontend documentation to see how `list` and `List[Entity]` look like for each
```
hidmic: @ivanpauno consider
```suggestion
        Check the documentation of each specific frontend implementation to see how `list` and `List[Entity]` look like.
```

launch/launch/frontend/type_utils.py
```diff
@@ -0,0 +1,254 @@
...
+    int, float, bool, str
+)
+
+__TypesForGuessTuple = (
+    int, float, bool, list, str
+)
+
+
+def check_is_list(data_type: Any) -> bool:
+    """Check if `data_type` was origined from `typing.List`."""
```
hidmic: @ivanpauno consider
```suggestion
    """Check if `data_type` is based on a `typing.List`."""
```

launch/launch/frontend/type_utils.py
```diff
@@ -0,0 +1,254 @@
...
+    return hasattr(data_type, '__origin__') and \
+        data_type.__origin__ in (list, List)  # On Linux/Mac is List, on Windows is list.
+
+
+def check_is_union(data_type: Any) -> bool:
+    """Check if `data_type` was origined from `typing.Union`."""
```
hidmic: @ivanpauno consider
```suggestion
    """Check if `data_type` is based on a `typing.Union`."""
```

launch/launch/frontend/type_utils.py
```diff
@@ -0,0 +1,254 @@
...
+    return hasattr(data_type, '__origin__') and \
+        data_type.__origin__ is Union
+
+
+def check_is_list_entity(data_type: Any) -> bool:
+    """
+    Return `True` if `data_type` is `List[launch.frontend.Entity]`.
```
hidmic: @ivanpauno consider
```suggestion
    Check if `data_type` is a `typing.List` with elements of `launch.frontend.Entity` type or derived types.
```

launch/launch/frontend/type_utils.py
```diff
@@ -0,0 +1,254 @@
...
+
+    It returns also `True` the member type is a subclass of `launch.frontend.Entity`.
+    """
+    return check_is_list(data_type) and \
+        issubclass(data_type.__args__[0], Entity)
+
+
+def get_tuple_of_types(data_type: Any) -> Tuple:
+    """Convert typing.Union to tuple of types. If not, return `(data_type,)`."""
```
hidmic: @ivanpauno consider
```suggestion
    """
    Normalize `data_type` to a tuple of types. 
    
    If `data_type` is based on a `typing.Union`, return union types. 
    Otherwise, return a `(data_type,)` tuple.
    """
```

launch/launch/frontend/type_utils.py
```diff
@@ -0,0 +1,254 @@
...
+    if check_is_union(data_type):
+        return data_type.__args__
+    else:
+        return (data_type,)
+
+
+def check_valid_scalar_type(data_type: Any) -> bool:
+    """Check if it is a valid scalar type."""
```
hidmic: @ivanpauno consider
```suggestion
    """Check if `data_type` is a valid scalar type."""
```

launch/launch/frontend/type_utils.py
```diff
@@ -0,0 +1,254 @@
...
+def extract_type(data_type: Any) -> Tuple[Any, bool]:
+    """
+    Extract type information from type object.
+
+    :param data_type: It can be:
+        - a scalar type: `str`, `int`, `float`, `bool`
+        - a uniform list: `List[str]`, `List[int]`, `List[float]`, `List[bool]`
+        - a non-uniform list.
+            Can specify some of the scalar types for its members, e.g.: List[Union[int, str]].
+            `list` or `List`, means that any of the scalar types are allowed.
```
hidmic: @ivanpauno see similar docstring suggestion above.

launch/launch/frontend/type_utils.py
```diff
@@ -0,0 +1,254 @@
...
+    else:
+        return x
+
+
+def scalar_type_key(data_type: Any) -> int:
+    """Get key. Used for sorting the scalar data_types."""
+    keys = {
+        int: 0,
+        float: 1,
+        bool: 2,
```
hidmic: @ivanpauno meta: I wonder if `bool` should come before `int`

launch_xml/README.md
```diff
@@ -0,0 +1,98 @@
...
+`data_type` can also be set to `None`, which works in the same way as passing:
+
+```python
+Union[int, float, bool, list, str]
+```
+
+For handling lists, the `*-sep` attribute is used. e.g.:
+
+```xml
+<tag value="2,3,4" sth-sep=","/>
```
hidmic: @ivanpauno shouldn't it be `value-sep` instead?

launch/launch/frontend/type_utils.py
ivanpauno: No, if not `0` will be converted to bool. If both are possible, it should be an int.
And maybe we should just delete that option. So the rules do not overlap.

ivanpauno:
@hidmic I've addressed all the comments in both PRs. I will run an extra CI after an approval.

launch/launch/frontend/type_utils.py
hidmic: > If both are possible, it should be an int.

Ok, fair enough, Python coercion will do its job.

ivanpauno:
Luckly, last CI:
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/7531/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3668/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6168/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7360/)

PR #227, opened 2019-04-16
Add additional_env argument to ExecuteProcess class

ivanpauno:
See https://github.com/ros2/launch_ros/issues/13 discussion.

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -83,6 +83,7 @@ class ExecuteProcess(Action):
         name: Optional[SomeSubstitutionsType] = None,
         cwd: Optional[SomeSubstitutionsType] = None,
         env: Optional[Dict[SomeSubstitutionsType, SomeSubstitutionsType]] = None,
+        additional_env: Optional[Dict[SomeSubstitutionsType, SomeSubstitutionsType]] = None,
         shell: bool = False,
         sigterm_timeout: SomeSubstitutionsType = LaunchConfiguration(
             'sigterm_timeout', default=5),
@@ -152,7 +153,11 @@ class ExecuteProcess(Action):
         :param: name the label used to represent the process, as a string or a
             Substitution to be resolved at runtime, defaults to the basename of
             the executable
-        :param: env dictionary of environment variables to be used
+        :param: env dictionary of environment variables to be used, starting from
+            a clean environment. If 'None', the current environment is used.
+        :param: additional_env dictionary of environment variables to be added.
+            If 'env' was None, they are added to the current environment.
+            If not, 'env' is updated with additional_env.
         :param: shell if True, a shell is used to execute the cmd
         :param: sigterm_timeout time until shutdown should escalate to SIGTERM,
             as a string or a list of strings and Substitutions to be resolved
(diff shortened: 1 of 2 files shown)


launch/launch/actions/execute_process.py
```diff
@@ -183,13 +188,15 @@ def __init__(
...
-        if env is not None:
-            self.__env = []
-            for key, value in env.items():
-                self.__env.append((
-                    normalize_to_list_of_substitutions(key),
-                    normalize_to_list_of_substitutions(value)))
+        if env is None:
+            env = dict(os.environ)
+        if additional_env is not None:
+            env.update(additional_env)
```
hidmic: @ivanpauno won't you be mutating `env` if it was user provided dictionary?

launch/launch/actions/execute_process.py
```diff
@@ -183,13 +188,15 @@ def __init__(
...
         self.__cwd = cwd if cwd is None else normalize_to_list_of_substitutions(cwd)
-        self.__env = None  # type: Optional[List[Tuple[List[Substitution], List[Substitution]]]]
-        if env is not None:
-            self.__env = []
-            for key, value in env.items():
-                self.__env.append((
-                    normalize_to_list_of_substitutions(key),
-                    normalize_to_list_of_substitutions(value)))
+        if env is None:
+            env = dict(os.environ)
```
hidmic: @ivanpauno this patch is effectively changing behavior: when no `env` is provided, it is no longer inherited but forced to be the one it was at construction time.

launch/launch/actions/execute_process.py
ivanpauno: https://github.com/ros2/launch_ros/issues/13#issuecomment-483319933

launch/launch/actions/execute_process.py
ivanpauno: Good catch, I will change it again.

launch/launch/actions/execute_process.py
```diff
@@ -190,6 +195,14 @@ def __init__(
                 self.__env.append((
                     normalize_to_list_of_substitutions(key),
                     normalize_to_list_of_substitutions(value)))
+        self.__additional_env = \
+            None  # type: Optional[List[Tuple[List[Substitution], List[Substitution]]]]
```
wjwwood: Please use the normal style of variable typing, the comment based style was only require for Python 3.5 which we don't support any more.

wjwwood: I believe it would be:

```python
self.__additional_env: Optional[List[Tuple[List[Substitution], List[Substitution]]]] = None
```

launch/launch/actions/execute_process.py
```diff
@@ -429,13 +442,17 @@ def __expand_substitutions(self, context):
...
-        env = None
+        env = {}
         if self.__env is not None:
-            env = {}
             for key, value in self.__env:
                 env[''.join([context.perform_substitution(x) for x in key])] = \
                     ''.join([context.perform_substitution(x) for x in value])
-
+        else:
+            env = dict(os.environ)
```
wjwwood: The problem I have with the current logic is that `None` is never passed to the subprocess call, under any condition (as far as I can tell).

I think that if `self.__env` is `None` and `self.__additional_env` is also `None`, then `env` in this scope should be `None` as well.

wjwwood: It will be effectively the same thing, but I'd really like a way to actually defer to the behavior of subprocess when env is None, which is currently not possible.

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6​71​3/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3027/)
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/5516/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/6543/)

ivanpauno:
A lot of unrelated failures on CI and github check. Merging.

dirk-thomas:
@ivanpauno It doesn't look like anyone approved this PR yet?

ivanpauno:
> @ivanpauno It doesn't look like anyone approved this PR yet?

Sorry for the confusion. @hidmic @wjwwood please let me know if you have any further comments, and I will open a follow-up PR addressing them.

wjwwood:
I don't have any further comments, but yes please do wait for an approval before merging in the future, even if just addressing feedback.

launch/test/launch/test_execute_process.py
```diff
@@ -14,21 +14,39 @@
...
-    ld = LaunchDescription([
-        ExecuteProcess(
-            cmd=[sys.executable, 'TEST_PROCESS_WITH_ENV'], output='screen',
-            env={'TEST_PROCESS_WITH_ENV': 'Hello World'},
-        )
-    ])
+    os.environ['TEST_CHANGE_CURRENT_ENV'] = '1'
+    additional_env = {'TEST_PROCESS_WITH_ENV': 'Hello World'}
+    executable = ExecuteProcess(
+        cmd=[sys.executable, 'TEST_PROCESS_WITH_ENV'],
```
hidmic: @ivanpauno it'd be nice for the test to execute something like `python -c "print('$TEST_PROCESS_WITH_ENV')"` in a shell and then check that the `stdout` of that action is as expected.

PR #215, opened 2019-03-22
Merge apex_launchtest functionality into launch_testing

hidmic:
This pull request merges functionality from `apex_launchtest` and `apex_launchtest_cmake` packages into `launch_testing` and `launch_testing_ament_cmake` respectively. Some renaming and style corrections were applied as well.

Additionally, a new package `test_launch_testing` is introduced to avoid `launch_testing` from being an `ament_cmake` type of package.

Connected to #208.

PR #229, opened 2019-04-30
Fix proc lookup for processes with multiple command-line arguments

pbaughman:
This is built on top of https://github.com/ros2/launch/pull/215.  Only the last commit is new

Taken from here: https://github.com/ApexAI/apex_rostest/pull/43

Previously looking up processes via strings was broken in cases where you wanted to disambiguate processes with multiple command-line arguments.  This PR fixes the issue by making cmd_args an array of strings, and attempting to match all args against the launched processes's action `process_details['cmd'][1:]`

### DIFF
--- a/launch_testing/launch_testing/util/proc_lookup.py
+++ b/launch_testing/launch_testing/util/proc_lookup.py
@@ -68,7 +68,7 @@ def _str_name_to_process(info_obj, proc_name, cmd_args):
         elif cmd_args is NO_CMD_ARGS:
             return len(proc.process_details['cmd']) == 1
         else:
-            return cmd_args in proc.process_details['cmd'][1:]
+            return all(arg in proc.process_details['cmd'][1:] for arg in cmd_args)
 
     matches = [proc for proc in info_obj.processes()
                if name_match_fn(proc) and cmd_match_fn(proc)]
@@ -97,6 +97,12 @@ def resolveProcesses(info_obj, *, process=None, cmd_args=None, strict_proc_match
     :returns: A list of one or more matching processes taken from the info_obj.  If no processes
     in info_obj match, a NoMatchingProcessException will be raised.
     """
+    # Old versions of this let you pass a string for cmd_args, but there was no way to search
+    # for multiple arguments when you did that.  For backwards compatibility, we'll wrap string
+    # cmd_args in a list
+    if isinstance(cmd_args, str):
+        cmd_args = [cmd_args]
+
     if process is None:
         # We want to search all processes
         all_procs = info_obj.processes()
--- a/launch_testing/launch_testing/util/proc_lookup.py
+++ b/launch_testing/launch_testing/util/proc_lookup.py
@@ -97,12 +97,6 @@ def resolveProcesses(info_obj, *, process=None, cmd_args=None, strict_proc_match
     :returns: A list of one or more matching processes taken from the info_obj.  If no processes
     in info_obj match, a NoMatchingProcessException will be raised.
     """
-    # Old versions of this let you pass a string for cmd_args, but there was no way to search
-    # for multiple arguments when you did that.  For backwards compatibility, we'll wrap string
-    # cmd_args in a list
-    if isinstance(cmd_args, str):
-        cmd_args = [cmd_args]
-
     if process is None:
         # We want to search all processes
         all_procs = info_obj.processes()
@@ -122,6 +116,13 @@ def resolveProcesses(info_obj, *, process=None, cmd_args=None, strict_proc_match
     elif isinstance(process, str):
         # We want to search one (or more) processes that match a particular string.  The 'or more'
         # part is controlled by the strict_proc_matching argument
+
+        # Old versions of this let you pass a string for cmd_args, but there was no way to search
+        # for multiple arguments when you did that.  For backwards compatibility, we'll wrap string
+        # cmd_args in a list
+        if isinstance(cmd_args, str):
+            cmd_args = [cmd_args]
+
         matches = _str_name_to_process(info_obj, process, cmd_args)
         if len(matches) == 0:
             names = ', '.join(sorted(_proc_to_name_and_args(p) for p in info_obj.processes()))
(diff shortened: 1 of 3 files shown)


hidmic:
@pbaughman mind to rebase?

pbaughman:
@hidmic Done!

launch_testing/launch_testing/util/proc_lookup.py
```diff
@@ -68,7 +68,7 @@ def cmd_match_fn(proc):
         elif cmd_args is NO_CMD_ARGS:
             return len(proc.process_details['cmd']) == 1
         else:
-            return cmd_args in proc.process_details['cmd'][1:]
```
pbaughman: This was just wrong all along - but ever since we started matching process using the launch.actions themselves (a much better idea) nothing inside launch_testing actually uses this so it wasn't noticed.

I stumbled on this while trying to do some tooling that would match tested processes up to documentation, so I need string -> launch action lookup to work correctly.  The new tests should cover the correct behavior well enough to prevent regression

hidmic:
Running CI (packages up to `test_communication`, only linux for now):

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/7053/)

pbaughman:
@hidmic I just noticed this comment in test_io_handler_and_assertions.py:
```
        # TODO(hidmic): add NO_CMD_ARGS test again when and if its implementation
        # improves beyond checking argument count.                               
        # assertInStdout(self.proc_output, txt, 'terminating_proc', NO_CMD_ARGS) 
```
Should I fix that as part of this PR too ?  I'm not sure I understand the comment.  NO_CMD_ARGS was meant to be used to match processes that had zero command line arguments, since "cmd_args=None" is used to match any (including zero) command-line arguments

hidmic:
@pbaughman sure, we can do that in a follow up PR. I added that comment back when I realized that the `NO_CMD_ARGS` wouldn't server our purposes anymore (since IIRC now the actual command is `python`).

hidmic:
Hmm we have a couple issues on Linux ARM64 that may be related to this PR. Mind to check?

pbaughman:
@hidmic It looks like the issue is that no output was captured from test_executable_0-1.  This change affects how processes are looked-up via string name and arguments, but the test that failed uses the ExecuteProcess object to directly look up processes.  We can hold off on this PR to be sure, but I suspect you'll see the same failure pop up elsewhere.

Error came from here: https://github.com/ros2/launch/blob/26e46af8fbe88105bdb3411434d5360fe94377a5/launch_testing/launch_testing/util/proc_lookup.py#L118

New code only runs down here:
https://github.com/ros2/launch/blob/26e46af8fbe88105bdb3411434d5360fe94377a5/launch_testing/launch_testing/util/proc_lookup.py#L125

I will move the backward-compatibility logic for string cmd_args down into the section that deals with string processes to make it completely clear that it doesn't affect the other cases.

launch_testing/test/launch_testing/test_resolve_process.py
```diff
@@ -66,3 +71,120 @@ def test_unlaunched_process_lookup_with_substitutions(self):
...
+
+
+class TestStringProcessResolution(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        # Run a launch so we can get some real looking proc_info and proc_output objects to test
+        # against
+        proc_command = os.path.join(
+            ament_index_python.get_package_prefix('launch_testing'),
```
hidmic: @pbaughman Windows is barfing at this one, as we're trying to run a Python script that is not properly wrapped to be treated as an executable on said platform (as it would be if we were installing it as a `console_script`). A workaround is to use `python` as your actual command, as in https://github.com/ros2/launch/blob/master/launch_testing/test/launch_testing/test_io_handler_and_assertions.py#L35.

launch_testing/test/launch_testing/test_resolve_process.py
```diff
@@ -66,3 +71,120 @@ def test_unlaunched_process_lookup_with_substitutions(self):
...
+            cmd_args=['--two-arg', 'arg_two']
+        )
+
+        assert found_proc
+
+    def test_proc_string_lookup_no_args(self):
+        found_proc = launch_testing.util.resolveProcesses(
+            self.proc_info,
+            process='good_proc',
+            cmd_args=launch_testing.util.NO_CMD_ARGS
```
hidmic: @pbaughman and if you do the above, this test will fail as the executable becomes `python path/to/script.py` and `NO_CMD_ARGS` no longer applies. Same thing happened to me on `test_io_handler_and_assertions.py`

launch_testing/test/launch_testing/test_resolve_process.py
pbaughman: I understand.  I'll be back on Tuesday to fix this.  Maybe I'll serialize a known proc_info object instead of generating one on the fly

pbaughman:
@hidmic Ok, sorry that took so long.  I didn't have any luck with serializing the proc_info object, but I realized it's not necessary.  We're not asserting anything about the output for the NO_ARGS_PROC, so we can just run 'sys.executable' without any args and check that we're still able to look the process up.

Hopefully this changed fixed things on Windows!

pbaughman:
@hidmic Hey, can we do something with this MR?

pbaughman:
It looks like the launch_testing failures on windows have been resolved

hidmic:
Yeap! This PR is ready to go but we're right in the middle of the Dashing release, so let's put this one on hold for a little bit longer. It'll go into the next patch release for sure.

pbaughman:
@hidmic What's the timeline for the next patch?  AFAIK, this issue is the only thing preventing us from switching from apex_launchtest over to launch_testing

hidmic:
Thanks for the bump @pbaughman! Re-running CI before merging, just in case. About the next patch release, it's happening next week. Hopefully we can get this one in today.

pbaughman:
@hidmic poke

hidmic:
@pbaughman I'm checking MacOS CI, as it keeps failing for apparently unrelated reasons.

hidmic:
Alright, CI's green, going in!

PR #228, opened 2019-04-17
Make launchtest junit XML match pytest XML more closely

pbaughman:
- Use the `classname` attribute in the `testcase` element instead of jamming everything into `name`
 - Use the `message` attribute for `skipped` elements
 - Handle the case where @unittest.skip("Skip Message") is added to the generate_launch_description function

Signed-off-by: Pete Baughman <pete.baughman@apex.ai>

I noticed some of our internal tooling was choking on apex_launchtest XML, so I've made the generated XML match pytest's XML more closely

Example pytest XML using same classnames and test names as the good_proc.test.py example, but with empty test methods:
```
<?xml version="1.0" encoding="utf-8"?>
<testsuite errors="0" failures="0" name="pytest" skipped="1" tests="4" time="0.110">
  <testcase classname="example_pytest.TestGoodProcess" file="example_pytest.py" line="55" name="test_count_to_four" time="0.001"/>
  <testcase classname="example_pytest.TestProcessOutput" file="example_pytest.py" line="68" name="test_exit_code" time="0.001">
    <skipped message="Example of a skip" type="pytest.skip">example_pytest.py:68: Example of a skip</skipped>
  </testcase>
  <testcase classname="example_pytest.TestProcessOutput" file="example_pytest.py" line="75" name="test_full_output" time="0.001"/>
  <testcase classname="example_pytest.TestProcessOutput" file="example_pytest.py" line="85" name="test_out_of_order" time="0.001"/>
</testsuite>
```

Example launchtest XML:
```
<?xml version="1.0" encoding="us-ascii"?>
<testsuites errors="0" failures="0" name="apex_launchtest" tests="4">
  <testsuite errors="0" failures="0" name="good_proc.test.py" skipped="1" tests="4" time="4.111">
    <testcase classname="TestGoodProcess" name="test_count_to_four" time="4.11"/>
    <testcase classname="TestProcessOutput" name="test_full_output" time="0.0"/>
    <testcase classname="TestProcessOutput" name="test_exit_code" time="0.0">
      <skipped message="Example of a skip"/>
    </testcase>
    <testcase classname="TestProcessOutput" name="test_out_of_order" time="0.0"/>
  </testsuite>
</testsuites>
```
It's not spot-on, but it's sufficient for our internal test reporting tooling.  I'm open to feedback about what parts of the pytest junit XML are important.

This PR matches is my internal PR https://github.com/ApexAI/apex_rostest/pull/41 on the Apex AI repo

PR #231, opened 2019-05-07
Sync missing parametrization commits

pbaughman:
When the parametrization stuff moved from apex_rostest to ros2:launch, it was WIP.  It looks like two commits didn't make it over to launch.  I noticed this [here](https://github.com/ros2/launch/pull/228#discussion_r281209795)

This PR adds the two missing commits.  The changes are:
1) Give individual test cases names that contain the value of the parameter used for the test (if parametrized).
2) Give better error messages when parametrization argument binding goes wrong

Side-note:  After this PR and the process lookup PR, I'm going to blow away the apex version of launch_testing and switch to the ros2 version, so there shouldn't be any more upstream syncing issues.

### DIFF
--- a/launch_testing/launch_testing/loader.py
+++ b/launch_testing/launch_testing/loader.py
@@ -52,8 +52,10 @@ class TestRun:
         # If we're parametrized, extend the test names so we can tell more easily what
         # params they were run with
         if self.param_args:
-            for tc in itertools.chain(_iterate_tests_in_test_suite(pre_shutdown_tests),
-                                      _iterate_tests_in_test_suite(post_shutdown_tests)):
+            for tc in itertools.chain(
+                _iterate_tests_in_test_suite(pre_shutdown_tests),
+                _iterate_tests_in_test_suite(post_shutdown_tests)
+            ):
                 test_method = getattr(tc, tc._testMethodName)
                 new_name = tc._testMethodName + self._format_params()
                 setattr(tc, '_testMethodName', new_name)
--- a/launch_testing/launch_testing/loader.py
+++ b/launch_testing/launch_testing/loader.py
@@ -14,6 +14,7 @@
 
 import functools
 import inspect
+import itertools
 import unittest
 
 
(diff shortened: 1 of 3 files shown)


launch_testing/launch_testing/loader.py
```diff
@@ -48,6 +49,16 @@ def __init__(self,
         self.pre_shutdown_tests = pre_shutdown_tests
         self.post_shutdown_tests = post_shutdown_tests
 
+        # If we're parametrized, extend the test names so we can tell more easily what
+        # params they were run with
+        if self.param_args:
+            for tc in itertools.chain(_iterate_tests_in_test_suite(pre_shutdown_tests),
+                                      _iterate_tests_in_test_suite(post_shutdown_tests)):
```
dirk-thomas: Commonly we try to avoid this kind of alignment and wrap after the open parenthesis. I just took the liberty to apply that change to get this merged quickly: 7baa83e2bb1e4d3c9d5df60f6ede09206b6da7e5

launch_testing/launch_testing/test_runner.py
```diff
@@ -223,9 +222,9 @@ def run(self):
         # We will return the results as a {test_run: TestResult)}
         results = {}
 
-        for index, run in enumerate(self._test_runs):
+        for run in self._test_runs:
             if len(self._test_runs) > 1:
-                print('Starting test run {}'.format(index + 1))
+                print('\n***** Starting test run {} *****'.format(run))
```
dirk-thomas: Please don't use "random" starts to make a certain print statement stand out like this. If multiple parts in the code do that the output becomes even more cluttered and difficult to read. I just took the liberty to apply that change to get this merged quickly: 6071175be995f11d6ee6b8e3017402877418b4cf

pbaughman: @dirk-thomas Can we at least keep the extra line of whitespace please?

dirk-thomas:
Linux testing `--packages-above launch_testing`: [Build Status](https://ci.ros2.org/job/ci_linux/7013/)

launch_testing/launch_testing/test_runner.py
dirk-thomas: Sure: b974f6498ae2da1fc4b95d2fc105159f66a69354

pbaughman: Thanks.

There's probably a future issue to clean up the output from launch_testing to make it easier to read.  One of the beefs I had with rostest was it was often difficult to visually parse what was going on.  

Right now the output from unittest gets intermingled with the output from launch so it's hard to tell what's a test result and what's a message about a process starting or stopping.  Also we don't intercept stdout in the passing case like pytest or nosetest does.  It's probably lower priority, but I'll open an issue and self-assign it

PR #234, opened 2019-05-13
Moved some common code to LaunchDescriptionSource

ivanpauno:
Replaces #232 (see discussion there).

The idea of this, is to have some common code in `LaunchDescriptionSource`. The only derived class now is `PythonLaunchDescriptionSource`, but we will soon have `XMLLaunchDescriptionSource` and `YAMLLaunchDescriptionSource` which will need similar logic.

In those cases, the only that will be needed is to pass a different `get_method` to `LaunchDescriptionSource` constructor.

Note: `LaunchDescriptionSource` can still be used to wrap a `LaunchDescription`, as it is used in some places. This doesn't break api, but it adds new features to `LaunchDescriptionSource`.

### DIFF
--- a/launch/launch/launch_description_source.py
+++ b/launch/launch/launch_description_source.py
@@ -14,10 +14,17 @@
 
 """Module for the LaunchDescriptionSource class."""
 
+import traceback
 from typing import Optional
+from typing import Text
+
+import launch.logging
 
 from .launch_context import LaunchContext
 from .launch_description import LaunchDescription
+from .some_substitutions_type import SomeSubstitutionsType
+from .utilities import normalize_to_list_of_substitutions
+from .utilities import perform_substitutions
 
 
 class LaunchDescriptionSource:
@@ -26,8 +33,10 @@ class LaunchDescriptionSource:
     def __init__(
         self,
         launch_description: Optional[LaunchDescription] = None,
-        location: str = '<string>',
+        location: SomeSubstitutionsType = '<string>',
         method: str = 'unspecified mechanism from a script',
+        *,
+        get_method: Optional[callable] = None
     ) -> None:
         """
         Constructor.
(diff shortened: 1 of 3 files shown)


ivanpauno:
CI, just to keep the things going:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7088/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3314/)
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/5810/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/6904/)

launch/launch/launch_description_source.py
```diff
@@ -51,21 +70,43 @@ def try_get_launch_description_without_context(self) -> Optional[LaunchDescripti
...
+            self.__launch_description = \
+                self.__get_method(self.__expanded_location)
         return self.__launch_description
 
     @property
     def location(self) -> str:
-        """Getter for self.__location."""
-        return self.__location
+        """
+        Get the location of the Python launch file as a string.
```
hidmic: @ivanpauno consider:
```suggestion
        Get the location of the launch description as a string.
```

launch/launch/launch_description_source.py
```diff
@@ -26,8 +33,10 @@ class LaunchDescriptionSource:
     def __init__(
         self,
         launch_description: Optional[LaunchDescription] = None,
-        location: str = '<string>',
+        location: SomeSubstitutionsType = '<string>',
         method: str = 'unspecified mechanism from a script',
+        *,
+        get_method: Optional[callable] = None
```
hidmic: @ivanpauno why not use method overloading in derived classes instead?

ivanpauno: Sounds good

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7​11​3/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3331/)
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/5​82​6/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/6​93​2/)

PR #235, opened 2019-05-13
Allow SomeSubstitutionsType in variable_name of LaunchConfiguration substitution

ivanpauno:
Found this while working in launch_frontend.
I don't see a reason for not allowing `SomeSubstitutionsType` there.

### DIFF
--- a/launch/launch/substitutions/launch_configuration.py
+++ b/launch/launch/substitutions/launch_configuration.py
@@ -26,7 +26,6 @@ from .substitution_failure import SubstitutionFailure
 from ..launch_context import LaunchContext
 from ..some_substitutions_type import SomeSubstitutionsType
 from ..substitution import Substitution
-from ..utilities import ensure_argument_type
 
 
 class LaunchConfiguration(Substitution):
@@ -40,8 +39,6 @@ class LaunchConfiguration(Substitution):
     ) -> None:
         """Constructor."""
         super().__init__()
-
-        ensure_argument_type(variable_name, str, 'variable_name', 'LaunchConfiguration')
 
         from ..utilities import normalize_to_list_of_substitutions
         self.__variable_name = normalize_to_list_of_substitutions(variable_name)


ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7056/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3287/)
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/5783/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/6875/)

ivanpauno:
Added bug label, because I think not allowing `SomeSubstitutionsType` was a leftover of an old change.

launch/launch/substitutions/launch_configuration.py
```diff
@@ -41,8 +40,6 @@ def __init__(
         """Constructor."""
         super().__init__()
 
-        ensure_argument_type(variable_name, str, 'variable_name', 'LaunchConfiguration')
```
hidmic: @ivanpauno shouldn't we ensure it's `SomeSubstitutionsType` anyways?

ivanpauno: We aren't usually ensuring that arguments are of an specific type. e.g.:
https://github.com/ros2/launch/blob/d2ea92cc1136eb655f3a61f3a01cd2ae2740a5aa/launch/launch/substitutions/environment_variable.py#L26-L44
I don't know why we should do it here and not in other places.

CI is already green or with unrelated errors. Could you please check macOS failures too? They seem to be unrelated.

hidmic: Alright, fair enough.

hidmic:
Test failures on Mac OS look like the `message_filters` ones, though I don't know if those are supposed to be fixed.

ivanpauno:
MacOS faliures were unrelated (see https://github.com/ros2/build_cop/issues/189). Merging

PR #236, opened 2019-05-15
Enable reuse of launch testing functionality

hidmic:
Connected to https://github.com/ros2/rostest/pull/1. This pull request enables reuse of the `launch_test` CLI and `add_launch_test` CMake macro functionality, to be leveraged by the `ros2test` and `ros_testing` packages.

Incidentally, a bug was found during the testing phase and thus this PR includes a fix for it, in it's own separate commit.

### DIFF
--- a/launch_testing/launch_testing/asserts/assert_output.py
+++ b/launch_testing/launch_testing/asserts/assert_output.py
@@ -45,7 +45,7 @@ def get_matching_function(expected_output):
         return lambda expected, actual: expected in actual
     elif hasattr(expected_output, 'search'):
         return lambda expected, actual: (
-            expected.match(actual.replace(os.linesep, '\n')) is not None
+            expected.search(actual.replace(os.linesep, '\n')) is not None
         )
     raise ValueError('Unknown format for expected output')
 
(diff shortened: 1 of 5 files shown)


launch_testing/launch_testing/asserts/assert_output.py
```diff
@@ -45,7 +45,7 @@ def _match(expected, actual):
         return lambda expected, actual: expected in actual
     elif hasattr(expected_output, 'search'):
         return lambda expected, actual: (
-            expected.match(actual.replace(os.linesep, '\n')) is not None
+            expected.search(actual.replace(os.linesep, '\n')) is not None
```
pbaughman: I encountered this same issue while trying to improve the test coverage on launch_testing.  A simple test that catches this issue and prevents regression would be something like

```
def test_works_with_regex(self):                 
    assertInStdout(                              
        self.proc_output,                        
        re.compile(r"Called with arguments \S+"),
        "terminating_proc-2"                     
    )                                            
```

You can add it to the bottom of `test_io_handler_and_assertions.py`  You'll also need to `import re` at the top

hidmic: Sounds good!

PR #238, opened 2019-05-15
Change EXIT_FORCED value in linux to ease assertExitCodes calls

ivanpauno:
Follow-up of #215


ivanpauno:
@hidmic Closing, as I don't need this for https://github.com/ros2/system_tests/pull/367 and related. Please, reopen if you think that it's still a valuable change.

PR #240, opened 2019-05-16
Stop randomizing ROS_DOMAIN_ID by default in launch tests

hidmic:
This pull request drops the `--disable-ros-isolation` option in `launch_test` in favor of `--isolated`, effectively negating the default behavior.

Running CI (only on Linux, testing `test_communication` and `launch_testing` only)

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/7117/)

### DIFF
--- a/launch_testing/launch_testing/launch_test.py
+++ b/launch_testing/launch_testing/launch_test.py
@@ -59,7 +59,7 @@ def main():
                         action='store_true',
                         default=False,
                         help=('Isolate tests using a custom ROS_DOMAIN_ID.'
-                              'Useful for test paralellization.'))
+                              'Useful for test parallelization.'))
 
     parser.add_argument(
         'launch_arguments',
--- a/launch_testing/launch_testing/launch_test.py
+++ b/launch_testing/launch_testing/launch_test.py
@@ -54,10 +54,12 @@ def main():
                         default=False,
                         help='Show arguments that may be given to the test file.')
 
-    parser.add_argument('--disable-ros-isolation',
+    # TODO(hidmic): Provide this option for rostests only.
+    parser.add_argument('-i', '--isolated',
                         action='store_true',
                         default=False,
-                        help='Do not set a ROS_DOMAIN_ID.  Useful for debugging ROS tests')
+                        help=('Isolate tests using a custom ROS_DOMAIN_ID.'
+                              'Useful for test paralellization.'))
 
     parser.add_argument(
         'launch_arguments',
@@ -85,7 +87,7 @@ def main():
         _logger_.setLevel(logging.DEBUG)
         _logger_.debug('Running with verbose output')
 
-    if not args.disable_ros_isolation:
+    if args.isolated:
         domain_id = get_coordinated_domain_id()  # Must copy this to a local to keep it alive
         _logger_.debug('Running with ROS_DOMAIN_ID {}'.format(domain_id))
         os.environ['ROS_DOMAIN_ID'] = str(domain_id)


launch_testing/launch_testing/launch_test.py
```diff
@@ -54,10 +54,12 @@ def main():
...
                         help='Show arguments that may be given to the test file.')
 
-    parser.add_argument('--disable-ros-isolation',
+    # TODO(hidmic): Provide this option for rostests only.
+    parser.add_argument('-i', '--isolated',
                         action='store_true',
                         default=False,
-                        help='Do not set a ROS_DOMAIN_ID.  Useful for debugging ROS tests')
+                        help=('Isolate tests using a custom ROS_DOMAIN_ID.'
+                              'Useful for test paralellization.'))
```
dirk-thomas: Spelling: parallelization?

dirk-thomas:
An example for the unintended cross talk: https://ci.ros2.org/job/ci_osx/5825/testReport/junit/(root)/projectroot/test_get_node_names__rmw_fastrtps_cpp/

launch_testing/launch_testing/launch_test.py
hidmic: Oh my. Fixed in https://github.com/ros2/launch/pull/240/commits/cbfa7dee7c7641a4363e3ea9dc26bfa27d0d8406.

hidmic:
Alright, everything's passing though I don't know how we can verify if cross talk went away. @dirk-thomas ?

dirk-thomas:
> I don't know how we can verify if cross talk went away.

From reading the patch I am confident that it is resolved. :boat:

pbaughman:
Can someone help me understand why this was needed?

Also, is there a way to turn on isolation bank on from the CMakeLists file?  This change will make all of our CI test jobs fail by default.

dirk-thomas:
> Can someone help me understand why this was needed?

If the user sets a specific `ROS_DOMAIN_ID` the expectation is that it is being used - even in tests. Using random domain ids leads to cross talk with other users intentionally choosing a different id.

> is there a way to turn on isolation bank on from the CMakeLists file?

Not at the moment. Please provide a pull request to add such a CMake option for your use case.

pbaughman:
```
Please provide a pull request to add such a CMake option for your use case.
```
I will do this, but I'm feeling a little burned here because I specifically made ros domain isolation to cover our use case, made it work the same way as rostest from ros1, waited for the interested parties to review and approve it, and now it's been turned off in a PR that has only been open for 5 hours.

I really wish equivalent functionality could've been provided here instead of regressing the tool.

dirk-thomas:
The merged PR adding the randomization broke tests on our infrastructure as well as affected local testing. Therefore the change was effectively reverted to restore the previously working state.

I am sorry that it add additional effort for you but restoring the stability is more important than keeping recently merged new features which turned out to be responsible for that.

pbaughman:
Would you accept having domain isolation on by default in the add_launch_test cmake function if ROS_DOMAIN_ID isn't already set?

dirk-thomas:
> Would you accept having domain isolation on by default in the add_launch_test cmake function if ROS_DOMAIN_ID isn't already set?

That sounds reasonable to me :+1:

tfoote:
When making the update I'd like to see a prominent notice in the usage that the isolation only works for single hosts. If we want to make it really isolated it should be checking to see if there's any other DDS participant visible via discovery or provide another mechanism that is visible outside of the current machine. 

And for the sake of determinism it might make more sense to just use an iterator starting at the base port instead of starting with a random number. The DOMAIN_ID shouldn't make a difference but using a random one will likely have us end up with test failures that are hard to debug. It won't attempt to solve the multi machine problem but that's not really solved by the current implementation. Randomly assigning inside of 100 domains is approximately 50% chance of two of the IDs colliding by the time that you add the 10th machine to the network.

pbaughman:
@tfoote. I'm a little perplexed by your second paragraph.  If you're concerned about multiple machines on the same network colliding, having every launch_testing always start at the same port and counting up takes your 50% chance of collision with ten machines up to 100% with two machines... That's much worse.  What am I not understanding here?

Is there a way to configure DDS to only talk on one host?

Anyway, I'm open to other suggestions on how to not totally brick doing 'colcon test' in parallel, but nobody has proposed any other ideas yet.  In ROS1, rostest did this right out of the box by using anonymous ROS masters.

tfoote:
I think that the biggest difference between ROS 1 picking a random port and picking a random DOMAIN_ID is that the range of possible DOMAIN_IDs is much smaller. This makes the likelyhood of collision over 2 orders of magnitude higher. With that lower probability of collision it takes 34 tests in parallel to reach a 1% chance of collision. Which is actually not great. 

Using the deterministic port selection will not break colcon test in parallel for a single machine. It will only break them if two are running simultaneously on separate machines on the same network. We work very hard to make tests reliable and reproducible and this randomness will add uncertainty about ever test failure as to whether it might be cross talk. Whereas if it reliably fails if not properly isolated then we know what the issues is and how to resolve it. 

To that end we should find ways to actively isolate machines and networks. I'm not familiar with whether we have enough control with the DDS configuration files to isolate the system. But setting up virtual networks for most CI machines shouldn't be too hard. 

Especially if we think about being able to run parallel tests on multiple machines. If they each run `-j8` we're now talking about guaranteed collisions at just over 12 machines if random happens to give us perfectly spaced starting numbers exactly 9 apart. If there's just 2 machines running with -j8 there's a ~75% chance of collision and 3 machines the odds approach 95% of a collision. 

Odds are that many of these collisions won't be an issue so we'll be having collisions but only occasionally will the cause failures since they would need to be running similar tests. Which will just lead us to have non-repeatable flakey tests. But randomly selecting out of 100 is not really ieffective isolation and I'd advocate for finding us solutions that will provide deterministic isolation instead of relying on luck. 

In general I'd suggest that if there's any chance of collision we should not be running anything in parallel. Flaky tests are much more expensive to verify and if the tests are correct but getting cross talk from another tests, that's now a complete waste of time to debug. So I guess my point is that this is not really isolation in any sense and we shouldn't pretend it is such because it works most of the time with two computers on the network. The fact that we've seen this collision enough to detect it and debug it shows quite how prevalent it is already.

pbaughman:
@tfoote I'm going to go ahead with the proposed solution above, and we can tweak the details of the iterator in there.  I'll do some tests to figure out the performance loss of using a regular iterator vs a random start.  We can also be more verbose in the output about what domain is being used.

Our CI jobs don't have crosstalk, and we absolutely need a way to run 'colcon test' in parallel and almost all of our integration tests use the same topics and data types so the tests are pretty much guaranteed to conflict without this feature.

I think your suggestion of setting up networks to isolate CI machines is the correct solution.

PR #244, opened 2019-05-18
Convert retreived path to CMake path for use.

nuclearsandwich:
While working I encountered an error due to mixed CMake and Windows
style paths from this package (Example below).

With this change the path will be converted to a CMake path as soon as
it is validated.

```
CMake Error at cmake/add_launch_test.cmake:67 (if):
  Syntax error in cmake code at

    C:/Users/Administrator/jacob_ws/src/ros2/launch/launch_testing_ament_cmake/cmake/add_launch_test.cmake:67

  when parsing string

    c:\Users\Administrator\jacob_ws\install\launch_testing/share/launch_testing/examples/args.test.py

  Invalid character escape '\U'.
Call Stack (most recent call first):
  cmake/add_launch_test.cmake:105 (parse_launch_test_arguments)
  CMakeLists.txt:28 (add_launch_test)
```

### DIFF
--- a/launch_testing_ament_cmake/CMakeLists.txt
+++ b/launch_testing_ament_cmake/CMakeLists.txt
@@ -23,7 +23,6 @@ if(BUILD_TESTING)
   if(NOT LAUNCH_TESTING_INSTALL_PREFIX)
       message(FATAL_ERROR "launch_testing package not found")
   endif()
-  file(TO_CMAKE_PATH "${LAUNCH_TESTING_INSTALL_PREFIX}" LAUNCH_TESTING_INSTALL_PREFIX)
 
   # Test argument passing.  This test won't pass unless you give it an argument
   add_launch_test(
--- a/launch_testing_ament_cmake/CMakeLists.txt
+++ b/launch_testing_ament_cmake/CMakeLists.txt
@@ -23,6 +23,7 @@ if(BUILD_TESTING)
   if(NOT LAUNCH_TESTING_INSTALL_PREFIX)
       message(FATAL_ERROR "launch_testing package not found")
   endif()
+  file(TO_CMAKE_PATH "${LAUNCH_TESTING_INSTALL_PREFIX}" LAUNCH_TESTING_INSTALL_PREFIX)
 
   # Test argument passing.  This test won't pass unless you give it an argument
   add_launch_test(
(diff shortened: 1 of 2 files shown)


nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7147/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3358/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5851/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6973/)

launch_testing_ament_cmake/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ if(BUILD_TESTING)
   if(NOT LAUNCH_TESTING_INSTALL_PREFIX)
       message(FATAL_ERROR "launch_testing package not found")
   endif()
+  file(TO_CMAKE_PATH "${LAUNCH_TESTING_INSTALL_PREFIX}" LAUNCH_TESTING_INSTALL_PREFIX)
```
dirk-thomas: Would it make not sense to perform this step within the `add_launch_test` function called below?

nuclearsandwich: My thought was that converting the value immediately guarantees that it's a valid CMake path for any subsequent uses. I'm wasn't sure whether `file(TO_CMAKE_PATH ...)` would correctly handle a path with mixed path segments but a small emperical example suggests that it would.

I think both approaches have their advantages and I don't have an opinion on which advantages are preferred.

jacobperron: I think it's better to place it inside `add_launch_test`. Then other packages using the function won't fall victim to the same issue. My two cents.

nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7152/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3363/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5856/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6979/)

nuclearsandwich:
The second commit moves the conversion into the add_launch_test function. I tried at first to do it inside the argument parsing macro but the macro invocation itself has errors if the string contains unescaped backslashes at that point. So it has to be just before it is called.

hidmic:
Just saw this on CI: https://ci.ros2.org/job/ci_windows/6980/testReport/junit/(root)/projectroot/C__J_workspace_ci_windows_ws_install_share_launch_testing_examples_args_test_py/. Could be it a regression related to this patch?

nuclearsandwich:
This test failure is indeed introduced by this change.

PR #246, opened 2019-05-20
Stop using 'launch' logger

### LINKED ISSUE #242: launch.actions.LogInfo prints to the screen twice
I have this demo I was making to answer a question, and noticed that the `LogInfo` action prints to the screen twice. Here's the demo:

```python
import launch


        launch.actions.DeclareLaunchArgument('msg', default_value='hello world'),
        launch.actions.LogInfo(msg=launch.substitutions.LaunchConfiguration('msg')),
    ])

```

When I run that I get:

```
% ros2 launch ./test.launch.py
[INFO] [launch]: All log files can be found below /home/william/.ros/log/2019-05-17-13-28-06-638843-parallels-Parallels-Virtual-Platform-29924
[INFO] [launch.actions.log_info]: hello world
[INFO] [launch.actions.log_info]: hello world
```

I haven't had time to investigate, but I thought I'd go ahead and report it while it was fresh.

@hidmic any ideas? I already poked @jacobperron about it.

hidmic:
Fixes #242. Since most classes in the `launch` package use the logger named after the module they are in, the `'launch'` logger would route all their traffic due to the hierarchical log propagation feature that Python's `logging` module comes with.

Alternatively, we _could_ disable propagation altogether, but that closes the door to creative uses of that feature in `launch`. Having said that, I don't have a strong preference.

### DIFF
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -89,7 +89,7 @@ class LaunchService:
         self.__debug = debug
 
         # Setup logging
-        self.__logger = launch.logging.get_logger('launch')
+        self.__logger = launch.logging.get_logger(__name__)
         # Install signal handlers if not already installed, will raise if not
         # in main-thread, call manually in main-thread to avoid this.
         install_signal_handlers()


wjwwood:
With this the output changes to:

```
% ros2 launch ./test.launch.py
[INFO] [launch.launch_service]: All log files can be found below /home/william/.ros/log/...
[INFO] [launch.launch_service]: Default logging verbosity is set to INFO
[INFO] [launch.actions.log_info]: hello world
```

Is there a configuration where it would look more like:


```
% ros2 launch ./test.launch.py
[INFO] [launch]: All log files can be found below /home/william/.ros/log/...
[INFO] [launch.user]: hello world
```

(or something similar)

Basically, I don't understand why using `'launch'` in the `LaunchService` would require the `LogInfo` action to print twice. Also, I see the value in have the module name as part of internal log messages, but for commonly made messages like where the log file is, etc., the user doesn't case which module it comes from and in fact would probably prefer the logger name didn't change even if we rearrange the internals of the library. They just care that the message is from `launch` itself rather than their launch file or one of their applications.

hidmic:
> Basically, I don't understand why using `'launch'` in the `LaunchService` would require the `LogInfo` action to print twice.

It's a side effect of the `LogInfo` logger being called `'launch.actions.log_info'`. Hierarchies in Python's module system and parenthood relationships in logging are expressed in the same way, thus messages logged to `'launch.*'` are also routed to `'launch'`.

Of course, it doesn't _have to_ be that way, but as it stands, the `'launch'` logger will always print the content of all and since all its children have the same handlers you'll see duplication (and it happens that all loggers that use `__name__` inside the `launch` package are its children). 

To achieve exactly what you describe, we must disable log propagation by default everywhere and use some well known loggers e. g. `'launch.user'`. I'm OK with it, as long we understand the restrictions we introduce.

hidmic:
As to why we reuse handlers like we do instead of using `logging` hierarchies: I couldn't find a reasonable, consistent way to reconcile that functionality with the configuration flexibility we wanted.

wjwwood:
Is it possible to configure it such that a single logger (say `launch.action.log_info`) does not propagate (or is filtered during propagation), but the rest continue to work as is?

Sorry I know I'm asking noob questions, I just wanted to make we were aiming for the same end goal, it it's not possible to get there, then that's fine.

hidmic:
> Is it possible to configure it such that a single logger (say launch.action.log_info) does not propagate (or is filtered during propagation), but the rest continue to work as is?

Yes! But how would you like this to work? To keep the `'launch'` logger we have to disable propagation everywhere we use a `__name__` logger. Then we can have a `'launch.user'` logger for the `LogInfo` action with disabled propagation as well. If we are OK with that (i.e. blocking propagation by default), let's do that then.

> Sorry I know I'm asking noob questions, I just wanted to make we were aiming for the same end goal,

Not a noob question, we have to agree on something :)

wjwwood:
> Yes! But how would you like this to work?

Like this :)

```
% ros2 launch ./test.launch.py
[INFO] [launch]: All log files can be found below /home/william/.ros/log/...
[INFO] [launch.user]: hello world
```

(or `launch.actions.log_info`, not sure which is clearer)


> If we are OK with that (i.e. blocking propagation by default), let's do that then.

It just means some extra boilerplate every place we add a logger within launch right? I think that's ok, if you're ok with it.

hidmic:
> It just means some extra boilerplate every place we add a logger within launch right?

Either that or disabling propagation within `launch.logging` itself (and thus applying to all loggers).

wjwwood:
Well, let's do it then. I think you have a better inclination on how we should do it (when/where to disable propagation and what we need to do instead), so if you can make it look/feel like the example I gave above, then I'll approve it.

hidmic:
Closing in favor of #250. Thanks for the discussion @wjwwood !

PR #249, opened 2019-05-21
Do not check declared arguments of IncludeLaunchDescription subentities

### LINKED ISSUE #248: [launch] Unexpected failure when providing arguments to an included launch description
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - Binaries
- Version or commit hash:
  - Crystal
- DDS implementation:
  - N/A
  - N/A

#### Steps to reproduce issue

Create a simple python ROS2 package (I'm naming it `launch_test`) containing these two launch files:

```
# launch/parent.launch.py

import launch

        launch.actions.DeclareLaunchArgument('my_arg')
    ])
```

```
# launch/child.launch.py

import launch
from launch.actions.include_launch_description import IncludeLaunchDescription
from launch.substitutions.this_launch_file_dir import ThisLaunchFileDir

from ament_index_python import get_package_share_directory

        IncludeLaunchDescription(
            PythonLaunchDescriptionSource([get_package_share_directory('launch_test'), '/parent.launch.py']),
            launch_arguments={'my_arg': 'my_value'}.items()
        )
    ])
```

Then build/install this package to a workspace and run

```
$ ros2 launch -d launch_test child.launch.py
```

#### Expected behavior

The `parent.launch.py` launch description should recognize that it's been provided with the `my_arg` argument when it was included by `child.launch.py`.

#### Actual behavior

An exception occurs:

```
RuntimeError: Included launch description missing required argument 'my_arg' (description: 'no description given'), given: []
```

It seems that the launch service doesn't recognize that `parent.launch.py` has been provided with the launch argument by `child.launch.py`.

#### Additional information

It's possible to work around this by giving a default value for `my_arg` in `parent.launch.py`, but that's undesirable if I want to enforce that the argument has provided externally (whether it's provided from the command line or by a child launch file should be irrelevant, ideally).

ivanpauno:
Closes #248

### DIFF
--- a/launch/launch/launch_description.py
+++ b/launch/launch/launch_description.py
@@ -97,6 +97,10 @@ class LaunchDescription(LaunchDescriptionEntity):
                     entity._conditionally_included |= entity.condition is not None
                     declared_launch_arguments.append(entity)
                 else:
+                    from .actions import IncludeLaunchDescription
+                    if isinstance(entity, IncludeLaunchDescription):
+                        # Do not check launch arguments of included descriptions.
+                        continue
                     process_entities(
                         entity.describe_sub_entities(), _conditional_inclusion=False)
                     process_entities(
--- a/launch/test/launch/actions/test_include_launch_description.py
+++ b/launch/test/launch/actions/test_include_launch_description.py
@@ -19,6 +19,7 @@ import os
 from launch import LaunchContext
 from launch import LaunchDescription
 from launch import LaunchDescriptionSource
+from launch import LaunchService
 from launch.actions import DeclareLaunchArgument
 from launch.actions import IncludeLaunchDescription
 from launch.actions import SetLaunchConfiguration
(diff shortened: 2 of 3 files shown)


hidmic:
This patch is odd, can you explain the rationale and the implied limitations?

ivanpauno:
> This patch is odd, can you explain the rationale and the implied limitations?

Yes. When an `IncludeLaunchDescription` action is visited, it's checked if the non-optional non-default arguments are satisfied or not:

https://github.com/ros2/launch/blob/a0be2bf295d34099b7e51e9dcc1d2b1a1a4fac15/launch/launch/actions/include_launch_description.py#L100-L130

That checks the declared arguments in the `LaunchDescription` using the `get_launch_arguments` method:

https://github.com/ros2/launch/blob/a0be2bf295d34099b7e51e9dcc1d2b1a1a4fac15/launch/launch/launch_description.py#L60-L107

That recursively checks all the `DeclareLaunchArgument` action in subentities. The funny problem, is when you have a `IncludeLaunchDescription` as a subentity of the description being included. The declared launch arguments in the included launch description are automatically being recognized as launch arguments of the "parent" launch description. That's a problem, if you only wanted to pass an argument to the included launch description (e.g.: see https://github.com/ros2/launch/issues/248).

This patch ignore the `DeclareLaunchArgument` subentities of an `IncludeLaunchDescription` subentity (sorry for the confusing explanation).

What changes:
Before, launch arguments of included launch descriptions were automatically recognized as launch arguments of the "parent" description (the one which is including it).
Now, if you want that behavior, you should declare it again. I think it worked in that way in ROS1, and really, the more natural way of thinking it.

Extra note:
I really wouldn't allow declaring launch arguments in subentities of launch description entities (basically delete lines 99 to 103 in `launch/launch/launch/launch_description.py`). But that's a bigger change, an the `DeclareLaunchArgument` action would still be executed. The proposed patch is less intrusive (and less problematic).

hidmic:
> Now, if you want that behavior, you should declare it again. I think it worked in that way in ROS1, and really, the more natural way of thinking it.

That sounds reasonable to me. Does this patch enforce that behavior? In other words, does the example in #248 work as expected with this patch? A test would be nice :)

> I really wouldn't allow declaring launch arguments in subentities of launch description entities

Hmm, did you mean to not pick up launch argument in subentities? Propagating arguments through hierarchies of arbitrarily nested launch descriptions is a typical use case in ROS 1, and useful in general.

ivanpauno:
> That sounds reasonable to me. Does this patch enforce that behavior? In other words, does the example in #248 work as expected with this patch? A test would be nice :)

Yes, it worked as expected after the patch. I will add a test.

> Hmm, did you mean to not pick up launch argument in subentities? Propagating arguments through hierarchies of arbitrarily nested launch descriptions is a typical use case in ROS 1, and useful in general.

I don't have a real use case in mind, but sounds good.

wjwwood:
I think the fix is too broad, maybe instead it needs to be "ignore non-optional launch arguments if the action which included it will not be passing it either". Or maybe, when you reach an include you need to delegate the checking to it, so it can check itself and its children (ignore it's own arguments if needed), and then aggregate the result.

wjwwood:
I'm going to take a stab at fixing this as I described, since it's a bit involved. It might not be possible, so I'll take a shot at it first.

wjwwood:
I opened https://github.com/ros2/launch/pull/252, which is a good deal more complicated, but it can handle an important case which is that a non-optional `DeclareLaunchArgument` in an included launch file will complain (raise an exception) if visited from a parent `IncludeLaunchDescription` which doesn't pass that argument.

If I'm correct, I think this pr's solution would miss that case, but I haven't tried it yet.

ivanpauno:
Running CI (only fastrtps):

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/7230/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3424/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5920/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7052/)

wjwwood:
@ivanpauno does this need another CI or are we good?

ivanpauno:
> @ivanpauno does this need another CI or are we good?

Flake8 is passing locally now, and the other errors are unrelated. Merging!

PR #252, opened 2019-05-23
fix detection of arguments passed when including a launch description

wjwwood:
Potential alternative to https://github.com/ros2/launch/pull/249 and fix for #248.

@mxgrey FYI

I used this pair of launch files to test it:

```python
# test.launch.py
import launch


        launch.actions.DeclareLaunchArgument('other2', default_value='pingpong'),
        launch.actions.IncludeLaunchDescription(
            launch.launch_description_sources.PythonLaunchDescriptionSource('./other.launch.py'),
            # Does not work, missing 'other2':
            # launch_arguments={'other': 'foobar'}.items(),

            # Does not work, missing 'other' and 'other2':
            # launch_arguments={'msg': 'world hello'}.items(),

            # Works:
            launch_arguments={
                'other': 'foobar',
                'other2': launch.substitutions.LaunchConfiguration('other2')
            }.items(),
        ),
    ])
```

```python
# other.launch.py
import launch


        launch.actions.DeclareLaunchArgument('msg', default_value='hello world'),
        launch.actions.DeclareLaunchArgument('other'),
        launch.actions.DeclareLaunchArgument('other2'),
        launch.actions.LogInfo(msg=launch.substitutions.LaunchConfiguration('msg')),
        launch.actions.LogInfo(msg=launch.substitutions.LaunchConfiguration('other')),
        launch.actions.LogInfo(msg=launch.substitutions.LaunchConfiguration('other2')),
    ])
```

I haven't made a test case yet, but I wanted to get this up for review asap. I'll keep working on a test case.

This change is different from #249 in that it uses the arguments passed to the "parent" `IncludeLaunchDescription` action, whether that be the one used by `ros2 launch` or the one the user used to nest a launch file into another one, when evaluating whether or not a non-optional `DeclareLaunchArgument` action has been satisfied. Whereas #249 just ignores nested `DeclareLaunchArguments` (if I'm understanding the issue correctly).

### DIFF
--- a/launch/launch/actions/include_launch_description.py
+++ b/launch/launch/actions/include_launch_description.py
@@ -106,19 +106,30 @@ class IncludeLaunchDescription(Action):
 
         # Do best effort checking to see if non-optional, non-default declared arguments
         # are being satisfied.
-        argument_names = [
-            perform_substitutions(context, normalize_to_list_of_substitutions(arg_name))
-            for arg_name, arg_value in self.launch_arguments
-        ]
-        declared_launch_arguments = launch_description.get_launch_arguments()
+        declared_launch_arguments = launch_description.get_launch_arguments(
+            include_launch_description_parent=self
+        )
         for argument in declared_launch_arguments:
             if argument._conditionally_included or argument.default_value is not None:
                 continue
+            argument_names: List[str] = []
+            if argument._include_launch_description_parent is not None:
+                argument_names = [
+                    perform_substitutions(context, normalize_to_list_of_substitutions(arg_name))
+                    for arg_name, arg_value
+                    in argument._include_launch_description_parent.launch_arguments
+                ]
             if argument.name not in argument_names:
+                launch_file_location = os.path.abspath(self.__launch_description_source.location)
                 raise RuntimeError(
-                    "Included launch description missing required argument '{}' "
-                    "(description: '{}'), given: [{}]"
-                    .format(argument.name, argument.description, ', '.join(argument_names))
+                    "Included launch description (located at '{}') missing required argument '{}' "
+                    "(argument description: '{}'), arguments that were given: [{}]"
+                    .format(
+                        launch_file_location,
+                        argument.name,
+                        argument.description,
+                        ', '.join(["'{}'".format(a) for a in argument_names])
+                    )
                 )
 
         # Create actions to set the launch arguments into the launch configurations.
--- a/launch/launch/actions/declare_launch_argument.py
+++ b/launch/launch/actions/declare_launch_argument.py
@@ -17,6 +17,7 @@
 from typing import List
 from typing import Optional
 from typing import Text
+from typing import TYPE_CHECKING
 
 import launch.logging
 
@@ -26,6 +27,9 @@ from ..some_substitutions_type import SomeSubstitutionsType
 from ..substitution import Substitution
 from ..utilities import normalize_to_list_of_substitutions
 from ..utilities import perform_substitutions
+
+if TYPE_CHECKING:
+    from ..actions import IncludeLaunchDescription  # noqa
 
 
 class DeclareLaunchArgument(Action):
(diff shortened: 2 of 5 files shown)


launch/launch/actions/include_launch_description.py
```diff
@@ -106,19 +106,30 @@ def visit(self, context: LaunchContext) -> List[LaunchDescriptionEntity]:
...
+        )
         for argument in declared_launch_arguments:
             if argument._conditionally_included or argument.default_value is not None:
                 continue
+            argument_names: List[str] = []
+            if argument._include_launch_description_parent is not None:
+                argument_names = [
+                    perform_substitutions(context, normalize_to_list_of_substitutions(arg_name))
+                    for arg_name, arg_value
+                    in argument._include_launch_description_parent.launch_arguments
```
ivanpauno: The problem that I see here is the following:
An argument declared in a nested `IncludeLaunchDescription`, it's automatically an argument of the parent launch_description. So I expect that if the included launch description has a "grandparent", which was included passing the remaining `launch_arguments` that weren't solved by the "parent", it should work. But it isn't working in that way, because it is not checking if `argument._include_launch_description_parent` has a parent and recursively loading its arguments. Clearer example [here](https://gist.github.com/ivanpauno/c3840eec2cea59c8a1e32b96b1f26f0e).

wjwwood: I see your point... this is also connected to whether or not we want to require explicitly passing an argument from the parent launch file into the included one or not.

We may not have time to address all of these cases before the distro freeze. In light of that, @ivanpauno do you want to just use #249? Or do you think we should go forward with this pr even though it doesn't support the case you put into your example?

ivanpauno: Between this as-is and #249, I prefer the second one.
#249 solves #248, and it also have a consistent behavior in the example I give above ([here again](https://gist.github.com/ivanpauno/c3840eec2cea59c8a1e32b96b1f26f0e)). A call to `ros2 launch --show-args` only shows the arguments you can really set (in the example above, `other2`). So calling `ros2 launch package_name test.launch.py problematic_argument:='asd'` will fail when including `other.launch.py` from `test.launch.py`, which is at least consistent with the idea of not detecting nested arguments in `IncludeLaunchDescription` actions.

If the final goal is to automatically recognize the nested arguments in others `IncludeLaunchDescription` actions, we will not break existing code when introducing that (duplicated declarations aren't a problem).

I would like to hear another opinion before continuing with #249, maybe @hidmic?

hidmic: Hmm, on one hand we have this PR, that allows implicit propagation of arguments, which wasn't supported on ROS1 but IMHO is a really cool feature i.e. it reduces the maintenance burden for large hierarchies of launch files, specially if including third party ones. However, IIUC it's currently limited to a depth of 1, which is at very least unexpected. So I'd be inclined to go with #249, being as limited as ROS 1, and hopefully deliver this PR's functionality in an upcoming patch release.

mxgrey:
Just my two cents: 

I think it can be kind of scary to allow arguments to propagate automatically through launch file hierarchies, because it's possible that there could be collisions between different launch file argument keys, and that could produce surprising results. So I wouldn't personally want that behavior to be automatic.

Would it be possible to take advantage of the Pythonic pattern of argument forwarding, like what can be done with `**kwargs`? If we give users an explicit way to say "I'm comfortable with forwarding all the unknown arguments given to this launch file into this other launch file that I'm including"? Is it possible to produce a Substitution that acts as a dictionary of all the launch argument key/value pairs whose keys were not explicitly declared by the current launch file?

I suppose that would take considerable effort, so it probably can't be rushed through for the Dashing release, but I think it's what I would personally consider ideal, if it's feasible at all.

wjwwood:
What I had planned originally was to have an "opt-in" for "pass local arguments into the included launch file", similar to this pr from ROS 1: https://github.com/ros/ros_comm/pull/710

That way you'd normally have to explicitly pass arguments from one launch file to the next (avoiding most collisions), but if you had a situation where you had a deep hierarchy and needed to pass several parameters from top to bottom, a flag like this can be really useful.

wjwwood:
Deferring this until after Dashing, thanks @ivanpauno!

hidmic:
@wjwwood @ivanpauno backlog?

PR #255, opened 2019-06-04
Line buffering of logger output

### LINKED ISSUE #254: Handle split launched processes logged output
## Bug report

**Required Info:**

- Operating System:
  - Windows Server 2019 Datacenter (akin to Windows 10)
- Installation type:
  - From binaries.
- Version or commit hash:
  - [Dashing pre-release](https://ci.ros2.org/job/packaging_windows/1388/)

#### Steps to reproduce issue

Run any launch file, e.g. the in lifecycle demos:
```
ros2 launch lifecycle lifecycle_demo.launch.py
```

#### Expected behavior
Processes output is logged appropriately, line by line.

#### Actual behavior

Processes output is logged with additional splits, not only at newlines, e.g.:

```
[lifecycle_talker.EXE-1] [
[lifecycle_talker.EXE-1] INFO] [lc_talker]: Lifecycle publisher is currently inactive. Messages are not published.
[lifecycle_talker.EXE-1] [
[lifecycle_talker.EXE-1] WARN] [LifecyclePublisher]: Trying to publish message on the topic '/lifecycle_chatter', but the publisher is not activated
```

#### Additional information

Most likely, this is just `launch.actions.ExecuteProcess` implementation assuming that each `launch.events.ProcessIO` event comes with an integer count of lines, see [here](https://github.com/ros2/launch/blob/master/launch/launch/actions/execute_process.py#L317).

ivanpauno:
Fixes #254.

Tested both on Windows and Linux, it is working ok.

PS: I don't like much the idea of line buffering stderr, but I don't see another solution.

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -330,10 +330,9 @@ class ExecuteProcess(Action):
                 else:
                     last_line = line
                     break
+        self.__stdout_buffer = io.StringIO()
         if last_line is not None:
-            self.__stdout_buffer = io.StringIO(last_line)
-        else:
-            self.__stdout_buffer = io.StringIO()
+            self.__stdout_buffer.write(last_line)
 
     def __on_process_stderr(
         self, event: ProcessIO
@@ -351,10 +350,9 @@ class ExecuteProcess(Action):
                 else:
                     last_line = line
                     break
+        self.__stderr_buffer = io.StringIO()
         if last_line is not None:
-            self.__stderr_buffer = io.StringIO(last_line)
-        else:
-            self.__stderr_buffer = io.StringIO()
+            self.__stderr_buffer.write(last_line)
 
     def __flush_buffers(self, event, context):
         with self.__stdout_buffer as buf:
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -15,6 +15,7 @@
 """Module for the ExecuteProcess action."""
 
 import asyncio
+import io
 import os
 import platform
 import shlex
@@ -221,8 +222,8 @@ class ExecuteProcess(Action):
         self.__sigterm_timer = None  # type: Optional[TimerAction]
         self.__sigkill_timer = None  # type: Optional[TimerAction]
         self.__shutdown_received = False
-        self.__stdout_buffer = ''
-        self.__stderr_buffer = ''
+        self.__stdout_buffer = io.StringIO()
+        self.__stderr_buffer = io.StringIO()
 
     @property
     def output(self):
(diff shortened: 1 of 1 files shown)


launch/launch/actions/execute_process.py
```diff
@@ -316,46 +317,58 @@ def __on_process_stdin(
...
-        if output[-1] != 'a':
-            new_buffer = output[-1][:-1]
-        output[0] = self.__stdout_buffer + output[0]
-        output = output[:-1]
-        self.__stdout_buffer = new_buffer
-        for line in output:
-            self.__stdout_logger.info(
-                self.__output_format.format(line=line, this=self)
-            )
+        with self.__stdout_buffer as buf:
```
hidmic: @ivanpauno what that does the context manager do here?

launch/launch/actions/execute_process.py
```diff
@@ -316,46 +317,58 @@ def __on_process_stdin(
...
+            buf.write(event.text.decode(errors='replace'))
+            buf.seek(0)
```
hidmic: @ivanpauno hmm, why not:
```py
buf.write(event.text.decode(errors='replace'))
buf.seek(0)
for line in buf.readline():
     if not line.endswith(os.linesep):
          buf.truncate(0)
          buf.write(line)
          break
     self.__logger.info(line)
```
instead? That way you get to reuse the buffers and it's somewhat simpler to read.

launch/launch/actions/execute_process.py
ivanpauno: I suppose it calls StringIO.close(). But I don't think it's really needed.

launch/launch/actions/execute_process.py
ivanpauno: Done. The idea is to truncate in both situations, if not the StringIO buffer could grow indefinitely.

ivanpauno:
Running CI (fastrtps, up to `launch`):

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/7298/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3473/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5​97​3/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7118/)

launch/launch/actions/execute_process.py
```diff
@@ -314,18 +317,52 @@ def __on_process_stdin(
...
+        with self.__stdout_buffer as buf:
+            line = buf.getvalue()
+            if line != '':
+                self.__stdout_logger.info(
+                    self.__output_format.format(line=line, this=self)
+                )
+        with self.__stderr_buffer as buf:
+            line = buf.getvalue()
+            if line != '':
+                self.__stdout_logger.info(
```
dirk-thomas: Shouldn't this be logged on `__stderr_logger`?

hidmic: True that.

ivanpauno:
@hidmic I think this is ready to go. If you approve, I will merge this without running CI again, as we aren't going to detect nothing new after https://github.com/ros2/launch/pull/255/commits/569d7ef0e102ae8cdcfee0187c7a996ef5161b40 change.

jacobperron:
> I will merge this without running CI again, as we aren't going to detect nothing new after 569d7ef change.

@ivanpauno 
Famous last words :laughing: 

The last commit seems to cause test failures for Windows: https://ci.ros2.org/job/ci_windows/7129/testReport/junit/launch.test.launch/test_execute_process/test_execute_process_with_env_None_expected0_/

PR #256, opened 2019-06-05
Remove launch_testing dependency on ros domain ID

### LINKED ISSUE #253: remove conceptual dependency on ROS_DOMAIN_ID in launch_testing
Currently `launch_testing` has options/features related to `ROS_DOMAIN_ID`:

https://github.com/ros2/launch/blob/066009b097e06dad7f981de55b9ee05c83cf6e0b/launch_testing/launch_testing/launch_test.py#L51

This should not be the case, it is inappropriate for anything in this repository to depend on, reference, or really even mention anything from ROS.

There's a TODO, which I linked to, but I opened this issue to raise the severity. This needs to be changed.

@hidmic @pbaughman please don't let things like this leak into `launch` or `launch_testing` in the future.

ivanpauno:
Fixes #253.

Depends on https://github.com/ros2/ros_testing/pull/2.


ivanpauno:
cc @pbaughman. I will wait you move the ros domain coordinator to the right place, and then remove it from https://github.com/ros2/ros_testing/pull/2 and add the correct import.

pbaughman:
@hidmic @ivanpauno Please find the new domain_coordinator package here: https://github.com/ros2/ament_cmake_ros/pull/3

ivanpauno:
CI results [here](https://github.com/ros2/ros_testing/pull/2#issuecomment-500484400).

PR #259, opened 2019-06-07
Fix build_cop #214

ivanpauno:
truncate method doesn't change the position in the stream (https://docs.python.org/3/library/io.html#io.IOBase.truncate).

The error was introduced in https://github.com/ros2/launch/pull/255 after https://github.com/ros2/launch/pull/255/commits/3b0f53b36bab7ff0fa747df3502164812412603b.

Fixes ros2/build_cop#214

I was sure I tested it correctly, but it seems that I didn't :man_facepalming:.

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -328,6 +328,7 @@ class ExecuteProcess(Action):
             else:
                 last_line = line
                 break
+        self.__stdout_buffer.seek(0)
         self.__stdout_buffer.truncate(0)
         if last_line is not None:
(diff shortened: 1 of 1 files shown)


ivanpauno:
Full CI:
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/7314/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3488/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5991/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7144/)

Issue #188, opened 2019-02-18
Launch not wiring print statements to stdout in realtime


stonier:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - binaries
- Version or commit hash:
  - Crystal
- DDS implementation:
  - Default
- Client library (if applicable):
  - rclpy 

#### Steps to reproduce issue

* Put a `cout` equivalent of the logged message in [demo_nodes_cpp/src/topics/talker.cpp](https://github.com/ros2/demos/blob/master/demo_nodes_cpp/src/topics/talker.cpp#L51)
* Put a `print` equivalent of the logged message in [demo_nodes_py/demo_nodes_py/talker.py](https://github.com/ros2/demos/blob/master/demo_nodes_py/demo_nodes_py/topics/talker.py#L34)

Add them both to a launcher and call `ros2 launch ...` on it:

```
    launch_description.add_action(
        launch_ros.actions.Node(
            package='demo_nodes_cpp', node_executable='talker', output='screen'),
    )
    launch_description.add_action(
        launch_ros.actions.Node(
            package='demo_nodes_py', node_executable='talker', output='screen'),
    )
```

#### Expected behavior

Both logged messages and cout/print messages make it in realtime to stdout.

#### Actual behavior

All make it there except the python `print` statements. They only show on stdout after interrupting the launch with CTRL-C.

wjwwood:
I think the difference is that we intentionally flush the out when logging:

https://github.com/ros2/rcutils/blob/a7581a333444c2804f2b7c27dd0da1b2946bebff/src/logging.c#L702

There is inherent buffing of stdout when a process is not connected to a tty.

If you changed this:

```python
import launch
import launch_ros.actions

def generate_launch_description():
    launch_description = launch.LaunchDescription()
    launch_description.add_action(
        launch_ros.actions.Node(
            package='demo_nodes_cpp', node_executable='talker', output='screen'),
    )
    launch_description.add_action(
        launch_ros.actions.Node(
            package='demo_nodes_py', node_executable='talker', output='screen',
            env={PYTHONUNBUFFERED : "1"}),
    )
    return launch_description
```

I expect that you'll see prompt printing.

----

It's possible that the `emulate_tty` option in `osrf_pycommon`'s API could address this, but I didn't enable it by default. You could do it here:

[...]

PR #265, opened 2019-06-22
[execute_process] emulate_tty configurable and defaults to true

stonier:
Attempting resolution of #188. Can test with the example included.

@wjwwood is this in line with what you were expecting?

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -593,13 +593,6 @@ class ExecuteProcess(Action):
                 ))
         except KeyError:
             emulate_tty = self.__emulate_tty
-        # guard against windows - tty emulation is not yet working
-        #     https://github.com/ros2/launch/issues/268
-        if emulate_tty and platform.system() == 'Windows':
-            self.__logger.warning(
-                'tty emulation not yet supported on windows, disabling'
-            )
-            emulate_tty = False
         try:
             transport, self._subprocess_protocol = await async_execute_process(
                 lambda **kwargs: self.__ProcessProtocol(
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -22,6 +22,7 @@ import shlex
 import signal
 import threading
 import traceback
+import yaml
 from typing import Any  # noqa: F401
 from typing import Callable
 from typing import cast
@@ -95,6 +96,7 @@ class ExecuteProcess(Action):
             'sigterm_timeout', default=5),
         sigkill_timeout: SomeSubstitutionsType = LaunchConfiguration(
             'sigkill_timeout', default=5),
+        emulate_tty: bool = True,
         prefix: Optional[SomeSubstitutionsType] = None,
         output: Text = 'log',
         output_format: Text = '[{this.name}] {line}',
@@ -173,6 +175,8 @@ class ExecuteProcess(Action):
             as a string or a list of strings and Substitutions to be resolved
             at runtime, defaults to the LaunchConfiguration called
             'sigkill_timeout'
+        :param: emulate_tty emulate a tty (terminal), defaults to
+            the LaunchConfiguration called 'emulate_tty'
         :param: prefix a set of commands/arguments to preceed the cmd, used for
             things like gdb/valgrind and defaults to the LaunchConfiguration
             called 'launch-prefix'
@@ -211,6 +215,7 @@ class ExecuteProcess(Action):
         self.__shell = shell
         self.__sigterm_timeout = normalize_to_list_of_substitutions(sigterm_timeout)
         self.__sigkill_timeout = normalize_to_list_of_substitutions(sigkill_timeout)
+        self.__emulate_tty = emulate_tty
         self.__prefix = normalize_to_list_of_substitutions(
             LaunchConfiguration('launch-prefix', default='') if prefix is None else prefix
         )
@@ -578,6 +583,16 @@ class ExecuteProcess(Action):
                 ', '.join(cmd), cwd, 'True' if env is not None else 'False'
             ))
         try:
+            emulate_tty = yaml.safe_load(
+                context.launch_configurations['emulate_tty']
+            )
+            if type(emulate_tty) is not bool:
+                raise TypeError(
+                    "emulate_tty is not boolean [{}]".format(type(emulate_tty))
+                )
+        except KeyError:
+            emulate_tty = self.__emulate_tty
+        try:
             transport, self._subprocess_protocol = await async_execute_process(
                 lambda **kwargs: self.__ProcessProtocol(
                     self, context, process_event_args, **kwargs
@@ -586,7 +601,7 @@ class ExecuteProcess(Action):
                 cwd=cwd,
                 env=env,
                 shell=self.__shell,
-                emulate_tty=False,
+                emulate_tty=emulate_tty,
                 stderr_to_stdout=False,
             )
         except Exception:
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -597,7 +597,7 @@ class ExecuteProcess(Action):
         #     https://github.com/ros2/launch/issues/268
         if emulate_tty and platform.system() == 'Windows':
             self.__logger.warning(
-                "tty emulation not yet supported on windows, disabling"
+                'tty emulation not yet supported on windows, disabling'
             )
             emulate_tty = False
         try:
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -593,6 +593,13 @@ class ExecuteProcess(Action):
                 ))
         except KeyError:
             emulate_tty = self.__emulate_tty
+        # guard against windows - tty emulation is not yet working
+        #     https://github.com/ros2/launch/issues/268
+        if emulate_tty and platform.system() == 'Windows':
+            self.__logger.warning(
+                "tty emulation not yet supported on windows, disabling"
+            )
+            emulate_tty = False
         try:
             transport, self._subprocess_protocol = await async_execute_process(
                 lambda **kwargs: self.__ProcessProtocol(
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -33,12 +33,12 @@ from typing import Text
 from typing import Tuple  # noqa: F401
 from typing import Union
 
-import yaml
-
 import launch.logging
 
 from osrf_pycommon.process_utils import async_execute_process
 from osrf_pycommon.process_utils import AsyncSubprocessProtocol
+
+import yaml
 
 from .emit_event import EmitEvent
 from .opaque_function import OpaqueFunction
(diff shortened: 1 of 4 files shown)


stonier:
* @wjwwood is this in line with what you were expecting?
* could use some inspiration on the testing front

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7419/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3570/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6086/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7242/)

wjwwood:
@stonier thanks for updating the pr, but it looks like flake8 is complaining still in our CI. :(

stonier:
Your CI looks to be somewhat ahead of my bionic installation (I have pytest 3.3.2, CI has 4.6.3) which passes the flake 8 test. I've taken a stab at guessing what the fix should be - can you trigger the CI run?

clalancette:
I've just rebased this branch, and added some fixes for the flake8 errors; it seems to pass for me locally now.  Here's new CI:


* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7426/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3576/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6093/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7248/)

wjwwood:
Thanks @clalancette and @stonier, yeah I think our CI uses flake8 from pypi (latest).

wjwwood:
It does appear, however that a related test is failing on Windows.

stonier:
Just for reference (because that log is a PITA to read):

```
UnicodeEncodeError: 'charmap' codec can't encode character '\\u2713' in position 154: character maps to <undefined>
....
  File "C:\\J\\workspace\\ci_windows\\ws\\src\\ros2\\launch\\launch\\launch\\launch_service.py", line 184, in __process_event\n
    "processing event: '{}' \u2713 '{}'".format(event, event_handler))\n
```

https://github.com/ros2/launch/blob/81476cbe2d31aa9752e5ad61d0d386a355bdf172/launch/launch/launch_service.py#L184

wjwwood:
Weird, that's a debug message. Not sure why that would try to print only after these changes... I don't see a failure on master. I wonder if this is a red herring and there's another issue? I'll look into it quickly.

launch/test/launch/test_emulate_tty.py
```diff
@@ -0,0 +1,53 @@
...
+        ld.add_action(launch.actions.SetLaunchConfiguration('emulate_tty', test_input))
+    ld.add_action(
+        launch.actions.RegisterEventHandler(
+            launch.event_handlers.OnProcessExit(on_exit=on_exit.handle)
+        )
+    )
+    ls = launch.LaunchService()
+    ls.include_launch_description(ld)
+    ls.run()
+    assert on_exit.returncode == expected
```
wjwwood: This is actually what is failing. The other output is a problem, but not the reason for the failure. See:

```
test\launch\test_emulate_tty.py:53: in test_emulate_tty
    assert on_exit.returncode == expected
E   assert 0 == 1
E    +  where 0 = <test_emulate_tty.OnExit object at 0x000002340EE160B8>.returncode
```

Basically `isatty()` is returning false on Windows (kind of expected) for one or more of your parameterized cases.

stonier: Aye, thanks for posting the details. I got to the same conclusions after doing some more reading last night.

How would you like to handle it? Options:

1. Dig deep, get tty emulation actually working and with unicode on windows (winpty?)
2. Refuse to attempt tty emulation (with warning?) on windows, adjust test expectations for windows with comments, raise an issue
3. Adjust test expectations for windows with comments (but perhaps there are risks with having gone down the fork of acting on tty emulation which cause problems), raise an issue
4. Other?

I think I'd lean towards 2. to avoid introducing unknown problems and because I don't have the resources to chase down the windows problems.

wjwwood: Number 2 or 3 is fine by me. The underlying API already will silently degrade when `import pty` fails, as it should on Windows.

stonier:
Implemented option 2. 

This closes #188, triggers new issue #268.

clalancette:
Another CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7479/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3619/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6134/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7314/)

stonier:
Pushed a commit for the linting. Worth noting though, it's something of a challenge getting the linting to work - I haven't been able to do so yet. Without going into the CI console log to try and reproduce the environment setup step-by-step are there more convenient instructions somewhere?

It's worth noting that on a dashing-bionic install:

* System dependencies defined by `ament_cmake_flake8` do not create the same environment as the CI tests (e.g. does not even bring in `python3-flake8-docstrings`)
* If someone does happen to have `python3-flake8-docstrings` installed on their system, the tests fail in a multitude of ways

wjwwood:
> are there more convenient instructions somewhere?

This part of the setup instructions have you install `flake8-docstrings` and some other things, from pip:

https://index.ros.org/doc/ros2/Installation/Dashing/Linux-Development-Setup/#install-development-tools-and-ros-tools

Which is what our CI does.

Our binary instructions exclude things needed to run tests, because the assumption is that if you're installing the binaries that you are building on top of it and will install your own test dependencies if you want to run linters on your packages. We also build from source when drafting changes to the core packages, since that is more likely to be up-to-date than the binary packages and because that's what we need to do in order to emulate CI's checks.

You could also use our docker images if you want a known good environment in which to test.

Sorry you had issues getting the linters to behave, that can be frustrating.

> System dependencies defined by ament_cmake_flake8 do not create the same environment as the CI tests (e.g. does not even bring in python3-flake8-docstrings)

That's because we install it from pip and debian packages cannot install dependencies from pip.

> If someone does happen to have python3-flake8-docstrings installed on their system, the tests fail in a multitude of ways (different version, different configuration?)

That's correct, it's an out of date version with bugs/false positives in it.

> Does it really need to lean on non-system versions?

Unfortunately, yes I believe it does. You could argue that we don't need all of these linter extensions in the first place, which I would tend to agree with a bit (thought obviously others on the team thought they were worthwhile), but if you really want to push against this, I'd try to find the point where these linters were added and comment there.

launch/launch/actions/execute_process.py
```diff
@@ -513,6 +519,23 @@ def __expand_substitutions(self, context):
...
+                ))
+        except KeyError:
+            emulate_tty = self.__emulate_tty
+        # guard against windows - tty emulation is not yet working
+        #     https://github.com/ros2/launch/issues/268
+        if emulate_tty and platform.system() == 'Windows':
+            self.__logger.warning(
+                'tty emulation not yet supported on windows, disabling'
+            )
+            emulate_tty = False
```
wjwwood: Is this needed? Doesn't setting `emulate_tty` to `True` on Windows gracefully degrade?

stonier: I wanted to ensure there was a logged warning that doesn't become contradictory if, for whatever reason, it started gracefully working. This made sure that it never provides an incorrect message.

I expect the better solution would be to explicitly test for the capability of tty emulation pre-execution, or to check for tty emulation post-execution and issuing a warning based on that. I'm not exactly certain how to implement that right now though.

wjwwood: That's what I'm saying, `osrf_pycommon` already tests this, e.g. if `import pty` fails it will automatically fallback to non-emulation:

https://github.com/osrf/osrf_pycommon/blob/8ba78744d9df7de70d824b4d76c11afe9e81ccd1/osrf_pycommon/process_utils/impl.py#L154-L156

So I think the warning is ok, though I'm not convinced we'll ever support it and it might just be annoying to Windows users who will otherwise have to conditionally set this setting. But either way, I don't think you need to set `emulate_tty` to `False` here.

The only reason it was failing before was that your test asserted that emulation was happening, and on Windows it wasn't. With the work around in the test, this check and logic and warning I think are unnecessary. Though we could keep the warning if we need to.

It will also gracefully degrade if you run out of pty's which can happen.

stonier:
> > Does it really need to lean on non-system versions?

> Unfortunately, yes I believe it does. You could argue that we don't need all of these linter extensions in the first place, which I would tend to agree with a bit (thought obviously others on the team thought they were worthwhile), but if you really want to push against this, I'd try to find the point where these linters were added and comment there.

It's not a big drama, but it does certainly complicate things and the ROS infrastructure promises that one has come to rely on start to fall away.

The implication here is that I can install ament debian packages for linting, but they have only been verified to work with hidden pip dependencies. There is no guarantee that they will even be functional out of the box without their pip dependencies.

Additionally, the `test_depends` elements in `package.xml` are no longer a true representation and subsequently a `rosdep install` will no longer get everything you need to build and run your tests.

wjwwood:
> There is no guarantee that they will even be functional out of the box without their pip dependencies.

I agree, though I also see the counter point of how constraining it is when we cannot use the tools we need to get the job done (not saying that's the case here, but it has been in similar cases in the past).

Perhaps @dirk-thomas can comment on this, since he spends a lot of time working on these tools and dependencies. (he's on vacation right now though, so it might be later)

> Additionally, the test_depends elements in package.xml are no longer a true representation and subsequently a rosdep install will no longer get everything you need to build and run your tests.

`rosdep` will still work, since we map the rosdep key's in the `package.xml` to pip not to apt in these cases. The only thing that does not work is that apt packages cannot not automatically install them. This has come up over and over again in ROS 1, e.g.: https://github.com/ros/catkin/issues/746

launch/launch/actions/execute_process.py
stonier: Sorry for the latency - dropped in the last commit. That there (osrf_pycommon) is certainly doing the guarding and if a warning is desired, probably the best place for it so not not going to make any decision here.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7584/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3715/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6208/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7410/)

wjwwood:
I had to rebase this because of how our CI works, but here's a new round:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7596/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3723/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6216/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7422/)

wjwwood:
Thanks @stonier for the contribution and for taking time to iterate with us!

wjwwood:
Unfortunately this broke a bunch of our tests last night, so I'll revert and fix it up.

dirk-thomas:
> Unfortunately this broke a bunch of our tests last night, so I'll revert and fix it up.

Just for the record: the tests failing to match their output seem to be related to this change http://build.ros2.org/view/Eci/job/Eci__nightly-fastrtps_ubuntu_bionic_amd64/21/testReport/

PR #269, opened 2019-07-09
Always accept posix paths in LaunchDescriptionSource

ivanpauno:
On Linux/Mac, `LaunchDescriptionSource` will work as before.
On Windows:
- `LaunchDescriptionSource` will still accept paths in the windows way(e.g.: `'path\\to\\file'`).
- Will also accept path in the posix style (e.g. `'path/to/file'`).

It will avoid unnecessary usage of `os.path.join` or `pathlib.Path` in many launch files.
It will also allow always using posix style in the new [launch xml and yaml formats](https://github.com/ros2/launch/pull/226). I can do another workaround for that, but this seems like a better change.

### DIFF
--- a/launch/launch/launch_description_source.py
+++ b/launch/launch/launch_description_source.py
@@ -14,6 +14,7 @@
 
 """Module for the LaunchDescriptionSource class."""
 
+from pathlib import Path
 import traceback
 from typing import Optional
 from typing import Text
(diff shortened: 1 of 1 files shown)


ivanpauno:
The three loaders used were actually accepting both kind of paths (python loading machinery, xml loader, yaml loader).
I will close it, as it's not actually needed.

PR #273, opened 2019-07-16
Add parsing method for dirname substitution

ivanpauno:
Also, corrected problem for parsing substitutions with no arguments.

### DIFF
--- a/launch/launch/frontend/expose.py
+++ b/launch/launch/frontend/expose.py
@@ -40,16 +40,13 @@ def instantiate_action(entity: 'Entity', parser: 'Parser') -> Action:
 
 def instantiate_substitution(
     type_name: Text,
-    args: Iterable[Iterable[SomeSubstitutionsType]]
+    args: Iterable[SomeSubstitutionsType] = []
 ) -> Substitution:
     """Call the registered substitution parsing method, according to `args`."""
     if type_name not in substitution_parse_methods:
         raise RuntimeError(
             'Unknown substitution: {}'.format(type_name))
-    if args:
-        subst_type, kwargs = substitution_parse_methods[type_name](*args)
-    else:
-        subst_type, kwargs = substitution_parse_methods[type_name]([])
+    subst_type, kwargs = substitution_parse_methods[type_name](args)
     return subst_type(**kwargs)
 
 
--- a/launch/launch/frontend/expose.py
+++ b/launch/launch/frontend/expose.py
@@ -16,6 +16,7 @@
 
 import inspect
 from typing import Iterable
+from typing import Optional
 from typing import Text
 
 from ..action import Action
@@ -40,12 +41,13 @@ def instantiate_action(entity: 'Entity', parser: 'Parser') -> Action:
 
 def instantiate_substitution(
     type_name: Text,
-    args: Iterable[SomeSubstitutionsType] = []
+    args: Optional[Iterable[SomeSubstitutionsType]] = None
 ) -> Substitution:
     """Call the registered substitution parsing method, according to `args`."""
     if type_name not in substitution_parse_methods:
         raise RuntimeError(
             'Unknown substitution: {}'.format(type_name))
+    args = [] if args is None else args
     subst_type, kwargs = substitution_parse_methods[type_name](args)
     return subst_type(**kwargs)
 
--- a/launch/launch/frontend/expose.py
+++ b/launch/launch/frontend/expose.py
@@ -40,13 +40,16 @@ def instantiate_action(entity: 'Entity', parser: 'Parser') -> Action:
 
 def instantiate_substitution(
     type_name: Text,
-    args: Iterable[SomeSubstitutionsType]
+    args: Iterable[Iterable[SomeSubstitutionsType]]
 ) -> Substitution:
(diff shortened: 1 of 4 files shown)


launch/launch/frontend/expose.py
```diff
@@ -40,13 +40,16 @@ def instantiate_action(entity: 'Entity', parser: 'Parser') -> Action:
...
 ) -> Substitution:
     """Call the registered substitution parsing method, according to `args`."""
     if type_name not in substitution_parse_methods:
         raise RuntimeError(
             'Unknown substitution: {}'.format(type_name))
-    subst_type, kwargs = substitution_parse_methods[type_name](*args)
+    if args:
+        subst_type, kwargs = substitution_parse_methods[type_name](*args)
+    else:
+        subst_type, kwargs = substitution_parse_methods[type_name]([])
```
hidmic: @ivanpauno hmm, `substitution_parse_methods[type_name](*args)` should be enough.

launch/launch/substitutions/this_launch_file_dir.py
```diff
@@ -14,20 +14,31 @@
...
+@expose_substitution('dirname')
 class ThisLaunchFileDir(Substitution):
     """Substitution that returns the absolute path to the current launch file."""
 
     def __init__(self) -> None:
         """Constructor."""
         super().__init__()
 
+    @classmethod
+    def parse(cls, data: Iterable[SomeSubstitutionsType]):
```
hidmic: @ivanpauno the actual problem is this method taking a `data` argument when it shouldn't

ivanpauno: All the methods have the same signature:
```python3
```
Regardless if data have 1, 2, 3, ..., 10 arguments.
Only deleting it for 0 arguments is strange. I preferred consistency.

launch/launch/frontend/expose.py
ivanpauno: check my other comment

hidmic: Taking a second look at this, something's odd about the star expansion. Why do we end up with a list that has a single list of normalized substitutions (lists themselves)?

launch/launch/substitutions/this_launch_file_dir.py
hidmic: Fair enough. Not sure if `[ ]` is better than `None`.

launch/launch/frontend/expose.py
ivanpauno: Yeap, check https://github.com/ros2/launch/pull/273/commits/816b94a71ff7e2e4eefac94ab0ffc55f79bb2684.

launch/launch/substitutions/this_launch_file_dir.py
ivanpauno: I prefer `[ ]`, so the parsing method only needs to check the len of the list, and not check for `None`.

launch/launch/substitutions/this_launch_file_dir.py
```diff
@@ -14,20 +14,31 @@
...
+        """Parse `EnviromentVariable` substitution."""
+        if len(data) > 1:
```
hidmic: @ivanpauno shouldn't it be
```suggestion
        if len(data) != 0:
```
?

ivanpauno: Corrected in https://github.com/ros2/launch/pull/273/commits/dc75c2375e7911624a16b08577e6aa3ec5d8050a.

ivanpauno:
CI (up to launch, launch_xml, launch_yaml, launch_ros, test_launch_ros, only fastrtps):
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/7578/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3710/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6203/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7405/)

launch/launch/frontend/expose.py
```diff
@@ -40,13 +40,13 @@ def instantiate_action(entity: 'Entity', parser: 'Parser') -> Action:
 
 def instantiate_substitution(
     type_name: Text,
-    args: Iterable[SomeSubstitutionsType]
+    args: Iterable[SomeSubstitutionsType] = []
```
wjwwood: This is a usually a bad idea, though in this case I think it's technically safe, it would be better to avoid it, see:

https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

ivanpauno: Thanks! I didn't know that.

ivanpauno:
CI (up to launch, launch_xml, launch_yaml, launch_ros, test_launch_ros, only fastrtps):
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/7581/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3712/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6205/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7407/)

PR #274, opened 2019-07-17
fix linter warnings

dirk-thomas:
Addresses the warnings reported in the [CI nightly build](http://build.ros2.org/view/Eci/job/Eci__nightly-fastrtps_ubuntu_bionic_amd64/20/testReport/launch.test/test_flake8/test_flake8/): [Build Status](http://build.ros2.org/job/Eci__overlay_ubuntu_bionic_amd64/2/) (CI overlay before)

With the fix the warnings are resolved: [Build Status](http://build.ros2.org/job/Eci__overlay_ubuntu_bionic_amd64/11/)

### DIFF
--- a/launch/launch/action.py
+++ b/launch/launch/action.py
@@ -24,8 +24,8 @@ from .launch_context import LaunchContext
 from .launch_description_entity import LaunchDescriptionEntity
 
 if False:
-    from .frontend import Entity
-    from .frontend import Parser
+    from .frontend import Entity  # noqa: F401
+    from .frontend import Parser  # noqa: F401
 
 
 class Action(LaunchDescriptionEntity):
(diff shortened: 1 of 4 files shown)


launch/launch/action.py
```diff
@@ -24,8 +24,8 @@
 from .launch_description_entity import LaunchDescriptionEntity
 
 if False:
-    from .frontend import Entity
-    from .frontend import Parser
+    from .frontend import Entity  # noqa: F401
```
sloretz: I haven't seen this style of type annotations before. Is there a thread with more context? The annotations [are strings](https://github.com/ros2/launch/pull/274/files#diff-87194eeec55a75ef4357257623e817e5R52), but I would have expected them to be the class.

```
def parse(entity: Entity, parser: Parser):
```

 It's legal python since the [annotation can be any expression](https://www.python.org/dev/peps/pep-3107/#parameters), but I haven't seen any examples like this before.

sloretz: Nevermind, found an example. It's for avoiding import cycles

https://mypy.readthedocs.io/en/latest/common_issues.html#import-cycles

PR #277, opened 2019-07-18
Revert "Revert "[execute_process] emulate_tty configurable and defaults to true""

wjwwood:
This is a refactor of https://github.com/ros2/launch/pull/265 so that it doesn't break our other tests.

Detecting it in the first place was my fault because I narrowed our CI testing too much and so didn't exercise the failing tests.

Reverts ros2/launch#276

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -38,6 +38,8 @@ import launch.logging
 from osrf_pycommon.process_utils import async_execute_process
 from osrf_pycommon.process_utils import AsyncSubprocessProtocol
 
+import yaml
+
 from .emit_event import EmitEvent
 from .opaque_function import OpaqueFunction
 from .timer_action import TimerAction
@@ -95,6 +97,7 @@ class ExecuteProcess(Action):
             'sigterm_timeout', default=5),
         sigkill_timeout: SomeSubstitutionsType = LaunchConfiguration(
             'sigkill_timeout', default=5),
+        emulate_tty: bool = True,
         prefix: Optional[SomeSubstitutionsType] = None,
         output: Text = 'log',
         output_format: Text = '[{this.name}] {line}',
@@ -173,6 +176,8 @@ class ExecuteProcess(Action):
             as a string or a list of strings and Substitutions to be resolved
             at runtime, defaults to the LaunchConfiguration called
             'sigkill_timeout'
+        :param: emulate_tty emulate a tty (terminal), defaults to
+            the LaunchConfiguration called 'emulate_tty'
         :param: prefix a set of commands/arguments to preceed the cmd, used for
             things like gdb/valgrind and defaults to the LaunchConfiguration
             called 'launch-prefix'
@@ -211,6 +216,7 @@ class ExecuteProcess(Action):
         self.__shell = shell
         self.__sigterm_timeout = normalize_to_list_of_substitutions(sigterm_timeout)
         self.__sigkill_timeout = normalize_to_list_of_substitutions(sigkill_timeout)
+        self.__emulate_tty = emulate_tty
         self.__prefix = normalize_to_list_of_substitutions(
             LaunchConfiguration('launch-prefix', default='') if prefix is None else prefix
         )
@@ -578,6 +584,16 @@ class ExecuteProcess(Action):
                 ', '.join(cmd), cwd, 'True' if env is not None else 'False'
             ))
         try:
+            emulate_tty = yaml.safe_load(
+                context.launch_configurations['emulate_tty']
+            )
+            if type(emulate_tty) is not bool:
+                raise TypeError('emulate_tty is not boolean [{}]'.format(
+                    type(emulate_tty)
+                ))
+        except KeyError:
+            emulate_tty = self.__emulate_tty
+        try:
             transport, self._subprocess_protocol = await async_execute_process(
                 lambda **kwargs: self.__ProcessProtocol(
                     self, context, process_event_args, **kwargs
@@ -586,7 +602,7 @@ class ExecuteProcess(Action):
                 cwd=cwd,
                 env=env,
                 shell=self.__shell,
-                emulate_tty=False,
+                emulate_tty=emulate_tty,
                 stderr_to_stdout=False,
             )
         except Exception:
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -43,7 +43,7 @@ from .opaque_function import OpaqueFunction
 from .timer_action import TimerAction
 
 from ..action import Action
-from ..condition import evaluate_condition_expression
+from ..conditions import evaluate_condition_expression
 from ..event import Event
 from ..event_handler import EventHandler
 from ..event_handlers import OnProcessExit
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -96,7 +96,7 @@ class ExecuteProcess(Action):
             'sigterm_timeout', default=5),
         sigkill_timeout: SomeSubstitutionsType = LaunchConfiguration(
             'sigkill_timeout', default=5),
-        emulate_tty: bool = True,
+        emulate_tty: bool = False,
         prefix: Optional[SomeSubstitutionsType] = None,
         output: Text = 'log',
         output_format: Text = '[{this.name}] {line}',
@@ -175,7 +175,7 @@ class ExecuteProcess(Action):
             as a string or a list of strings and Substitutions to be resolved
             at runtime, defaults to the LaunchConfiguration called
             'sigkill_timeout'
-        :param: emulate_tty emulate a tty (terminal), defaults to True, but can
+        :param: emulate_tty emulate a tty (terminal), defaults to False, but can
             be overridden with the LaunchConfiguration called 'emulate_tty',
             the value of which is evaluated as true or false according to
             :py:func:`evaluate_condition_expression`.
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -38,13 +38,12 @@ import launch.logging
 from osrf_pycommon.process_utils import async_execute_process
 from osrf_pycommon.process_utils import AsyncSubprocessProtocol
 
-import yaml
-
 from .emit_event import EmitEvent
 from .opaque_function import OpaqueFunction
 from .timer_action import TimerAction
 
 from ..action import Action
+from ..condition import evaluate_condition_expression
 from ..event import Event
 from ..event_handler import EventHandler
 from ..event_handlers import OnProcessExit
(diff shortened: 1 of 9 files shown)


dirk-thomas:
If want a fast CI response for this change you can try the new CI overlay job. It uses all unchanged packages from last nights build and only builds the specified changes: see http://build.ros2.org/view/Eci/job/Eci__overlay_ubuntu_bionic_amd64/21/parameters/ which finished in less than 15 min.

wjwwood:
Ok, I found the issue, it was the ANSI escape sequences (printed by the console logger to reset terminal colors) were being captured when emulate tty was enabled, which broke output comparison. I updated launch testing to strip these by default (with an option to not strip them in case someone needed that feature later).

@pbaughman can you look at 12c0563? I moved this around because I noticed it and thought it might be a race where the event handlers could miss some events from processes included in the launch description being tested. I don't think it was the issue, but while debugging I considered it might be an issue. I can revert it if needed, but I think the new code is more correct.

wjwwood:
Here's full CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7626/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3742/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6235/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7473/)

dirk-thomas:
A CI build on build.ros2.org shows the same test failures as previously: http://build.ros2.org/view/Eci/job/Eci__overlay_ubuntu_bionic_amd64/25/testReport/

wjwwood:
Configuration looks ok, not sure. It definitely fixed it in my local vm and I pushed everything. The job used the right commit of ros2/launch (afaict).

The failing test names and captured output are kind of not very useful though.

wjwwood:
Is it possible that the overlaid `demo_nodes_cpp` is using the `launch` from the underlay?

dirk-thomas:
> Is it possible that the overlaid `demo_nodes_cpp` is using the `launch` from the underlay?

It shouldn't. But I wouldn't go as far as to say that it is impossible :wink:

wjwwood:
It's possible this fix isn't perfect, let's see what ci.ros2.org says, and I'll investigate with CI jobs containing print statements in launch if needed.

wjwwood:
Looks like it is failing on ci.ros2.org too (just reading the output so far), I guess I'll have to have another look. It might be the `\n` versus `\r\n`, but I don't know why my VM works though.

wjwwood:
Ok, all the tests in `demo_nodes_cpp` were fixed, but none of those used the regex feature :|. So now I changed how newline replacement works because when emulating a tty, `\r\n` is used as the line ending rather than just `\n` or `os.linesep` as it is in files. I updated the logic to handle this. We do the same in `osrf_pycommon`: https://github.com/osrf/osrf_pycommon/blob/8ba78744d9df7de70d824b4d76c11afe9e81ccd1/osrf_pycommon/process_utils/execute_process_pty.py#L80-L81

But it's still failing on the build.ros2.org ci: http://build.ros2.org/view/Eci/job/Eci__overlay_ubuntu_bionic_amd64/26/console

I'm not sure what's happening just yet.

wjwwood:
Running a smaller set of CI on ci.ros2.org, because I think it's possible that nothing is fixed on build.ros.org (`demo_nodes_cpp` nor `intra_process_demo`) but that before `demo_nodes_cpp` was fixed on ci.ros2.org and only `intra_process_demo` and other regex based tests were still failing:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7629/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3744/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6237/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7476/)

wjwwood:
Yeah, looks like the build.ros2.org CI job is not catching the fix some how, the `demo_nodes_cpp` issues are fixed in https://ci.ros2.org/job/ci_linux/7626/testReport/ (from the first set of CI I started on ci.ros2.org).

pbaughman:
> @pbaughman can you look at 12c0563? I moved this around because I noticed it and thought it might be a race where the event handlers could miss some events from processes included in the launch description being tested. I don't think it was the issue, but while debugging I considered it might be an issue. I can revert it if needed, but I think the new code is more correct.

@wjwwood I never considered that the order of event handlers and execute processes could matter but yeah - you're right.  In a perfect world, I would hope it wouldn't matter but in the real world, yours is almost certainly more correct.

wjwwood:
I need to still look at this failure (last one):

https://ci.ros2.org/job/ci_linux/7629/testReport/launch.test.launch.actions/test_include_launch_description/test_include_launch_description_launch_arguments/

But it's failing on nightly too:

https://ci.ros2.org/view/nightly/job/nightly_linux_release/1245/testReport/launch.test.launch.actions/test_include_launch_description/test_include_launch_description_launch_arguments/

So, I'd say this is up for review.

launch/launch/actions/execute_process.py
```diff
@@ -577,6 +583,16 @@ def __expand_substitutions(self, context):
...
+        try:
+            emulate_tty = yaml.safe_load(
+                context.launch_configurations['emulate_tty']
+            )
+            if type(emulate_tty) is not bool:
+                raise TypeError('emulate_tty is not boolean [{}]'.format(
+                    type(emulate_tty)
+                ))
+        except KeyError:
+            emulate_tty = self.__emulate_tty
```
hidmic: I'm usually OK with Python's style of simply letting things throw but in this case I think that checking for `'emulate_tty' in context.launch_configurations` instead would aid readability.

launch_testing/launch_testing/asserts/assert_output.py
```diff
@@ -16,6 +16,19 @@
...
+
+
+def normalize_lineseps(lines):
+    """
+    Normalize and then return the given lines to all use `\n`.
+    """
+    lines = lines.replace(os.linesep, '\n')
+    # This happens (even on Linux and macOS) when capturing I/O from an
+    # emulated tty.
+    lines = lines.replace('\r\n', '\n')
```
hidmic: @wjwwood hmm, funky... Do we know why?

launch/test/launch/actions/test_emulate_tty.py
```diff
@@ -0,0 +1,70 @@
...
+
+@pytest.mark.parametrize('test_input,expected', [
+    # use the default defined by ExecuteProcess
+    (None, tty_expected_unless_windows()),
+    # redundantly override the default via LaunchConfiguration
+    ('true', tty_expected_unless_windows()),
+    # override the default via LaunchConfiguration
+    ('false', 0)
+])
+def test_emulate_tty(test_input, expected):
```
hidmic: @wjwwood it'd be nice to test that passing `emulate_tty` upon `ExecuteProcess` construction works too.

launch/launch/actions/execute_process.py
```diff
@@ -577,6 +583,16 @@ def __expand_substitutions(self, context):
             self.__logger.info("process details: cmd=[{}], cwd='{}', custom_env?={}".format(
                 ', '.join(cmd), cwd, 'True' if env is not None else 'False'
             ))
+        try:
+            emulate_tty = yaml.safe_load(
+                context.launch_configurations['emulate_tty']
+            )
```
hidmic: @wjwwood meta: I think `yaml` parsing here is fine for now, but I also think we need to get this straight in `launch`. We're already applying YAML parsing under the hood for parameters, maybe we should standardize it (or not)?

launch_testing/launch_testing/asserts/assert_output.py
wjwwood: Yeah, it uses `\r` to move the carriage return back to the beginning of the line. Most print functions emulate this behavior for you when you're connected to a tty and print a lone `\n`.

It's a quirk of these kinds of things, which I've mostly learned from experience. There's scarcely any documentation about these behaviors across operating systems and different configurations. So yeah, funky.

launch/launch/actions/execute_process.py
wjwwood: In the original pull request I was going to mention this, but I didn't just to keep things moving. We have a function for this kind of parsing which we use in the condition evaluations:

https://github.com/ros2/launch/blob/master/launch/launch/conditions/evaluate_condition_expression_impl.py

It's a subset of YAML. I think we can/should stabilize on something. I don't mind using yaml, though we would need to depend on pyyaml. Which brings up another issue which is that launch doesn't declare a dependency on yaml after this pr. So that's not right. I think this is why I made the evaluation function in the first place. I think I will reuse that and drop the use of YAML for now. Allowing `true`(any cap)/`1`/`false`/`0` is documented in other parts of the code too.

launch/launch/actions/execute_process.py
wjwwood: Have a look at 1b84e42.

launch/launch/actions/execute_process.py
wjwwood: See 1b84e42

launch/test/launch/actions/test_emulate_tty.py
wjwwood: Check 9b413c1.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7635/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3749/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6242/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7482/)

wjwwood:
@stonier I refactored two parts of the original pull request, just FYI in case you care to have a look.

wjwwood:
9b413c1 also fixes the test failure that is also on master for the nightly's.

stonier:
Looks good to me - nicer way of checking the boolean.

wjwwood:
I'm looking into https://ci.ros2.org/job/ci_linux/7635/testReport/(root)/projectroot/test_logging_output_format/ now, which also seems to be related to this change.

wjwwood:
So this last issue has become a bit crazy. I'm still working on it, but I'm going to change the default of `emulate_tty` to be `False` (current behavior), and make a new pull request to change it to be `True` by default along with any fixes needed to make the last kind of failing test pass.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7654/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3765/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6261/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7498/)

wjwwood:
@stonier FYI I had to disable emulate_tty by default right now, so that I can get some fixes in and at least have the feature available. I'll keep poking at the issue and try to get it so that we can re-enable it by default.

wjwwood:
CI test failures are from new overnight linter warnings.

stonier:
> @stonier FYI I had to disable emulate_tty by default right now, so that I can get some fixes in and at least have the feature available. I'll keep poking at the issue and try to get it so that we can re-enable it by default.

Understood, thanks.

PR #278, opened 2019-07-18
Add assertWaitForStartup method to match assertWaitForShutdown

pbaughman:
Previously, a test could easily wait for a process to shut down, but there was no way for a test to wait for a process to start.  When testing complicated launch files, it's as important to be able to assert on started processes as it is to assert on stopped processes.

Before, a process was only added into the proc_info object on shutdown.  After this change, processes are added to the proc_info object when the start, and then again when they shutdown, overwriting the previous entry.  That means if a process is in proc_info, it's started up.  We need to check if an exit_code is present to know that a process shut down.

This PR also adds two tests - one for the assertWaitForShutdown method which was previously not covered by unit tests - and one for the new method.

Finally, over in 'launch' it looks like an OnProcessStart EventHandler was added, but it wasn't added to the event handler module's '__init__.py' file so it was difficult to import.

### DIFF
--- a/launch_testing/launch_testing/proc_info_handler.py
+++ b/launch_testing/launch_testing/proc_info_handler.py
@@ -14,6 +14,7 @@
 
 import threading
 from launch.actions import ExecuteProcess  # noqa
+from launch.events.process import ProcessExited
 
 from .util import NoMatchingProcessException
 from .util import resolveProcesses
@@ -118,7 +119,7 @@ class ActiveProcInfoHandler(ProcInfoHandler):
                 )[0]
 
                 process_event = self._proc_info_handler[resolved_process]
-                return hasattr(process_event, 'returncode')
+                return isinstance(process_event, ProcessExited)
             except NoMatchingProcessException:
                 return False
 
--- a/launch_testing/launch_testing/proc_info_handler.py
+++ b/launch_testing/launch_testing/proc_info_handler.py
@@ -129,3 +129,30 @@ class ActiveProcInfoHandler(ProcInfoHandler):
             )
 
         assert success, "Timed out waiting for process '{}' to finish".format(process)
+
+    def assertWaitForStartup(self,
+                             process,
+                             cmd_args=None,
+                             *,
+                             timeout=10):
+        success = False
+
+        def proc_is_started():
+            try:
+                resolveProcesses(
+                    info_obj=self._proc_info_handler,
+                    process=process,
+                    cmd_args=cmd_args,
+                    strict_proc_matching=True
+                )
+                return True
+            except NoMatchingProcessException:
+                return False
+
+        with self._sync_lock:
+            success = self._sync_lock.wait_for(
+                proc_is_started,
+                timeout=timeout
+            )
+
+        assert success, "Timed out waiting for process '{}' to start".format(process)
--- a/launch_testing/launch_testing/proc_info_handler.py
+++ b/launch_testing/launch_testing/proc_info_handler.py
@@ -110,13 +110,15 @@ class ActiveProcInfoHandler(ProcInfoHandler):
 
         def proc_is_shutdown():
             try:
-                resolveProcesses(
+                resolved_process = resolveProcesses(
                     info_obj=self._proc_info_handler,
                     process=process,
                     cmd_args=cmd_args,
                     strict_proc_matching=True
-                )
-                return True
+                )[0]
+
+                process_event = self._proc_info_handler[resolved_process]
+                return hasattr(process_event, 'returncode')
             except NoMatchingProcessException:
                 return False
 
--- a/launch_testing/launch_testing/test_runner.py
+++ b/launch_testing/launch_testing/test_runner.py
@@ -20,8 +20,9 @@ import launch
 from launch import LaunchDescription
 from launch import LaunchService
 from launch.actions import RegisterEventHandler
-from launch.event_handlers import OnProcessExit, OnProcessStart
+from launch.event_handlers import OnProcessExit
 from launch.event_handlers import OnProcessIO
+from launch.event_handlers import OnProcessStart
 
 from .io_handler import ActiveIoHandler
 from .parse_arguments import parse_launch_arguments
--- a/launch_testing/launch_testing/test_runner.py
+++ b/launch_testing/launch_testing/test_runner.py
@@ -20,7 +20,7 @@ import launch
 from launch import LaunchDescription
 from launch import LaunchService
 from launch.actions import RegisterEventHandler
-from launch.event_handlers import OnProcessExit
+from launch.event_handlers import OnProcessExit, OnProcessStart
 from launch.event_handlers import OnProcessIO
 
 from .io_handler import ActiveIoHandler
@@ -125,6 +125,9 @@ class _RunnerWorker():
             launch.actions.IncludeLaunchDescription(
                 launch.LaunchDescriptionSource(launch_description=test_ld),
                 launch_arguments=parsed_launch_arguments
+            ),
+            RegisterEventHandler(
+                OnProcessStart(on_start=lambda info, unused: proc_info.append(info))
             ),
             RegisterEventHandler(
                 OnProcessExit(on_exit=lambda info, unused: proc_info.append(info))
--- a/launch_testing/test/conftest.py
+++ b/launch_testing/test/conftest.py
@@ -31,14 +31,14 @@ def _source_test_loader(generate_test_description_fn,
     and modules.
     Rather than refactoring launch_testing and python unittest, we leverage what we already have.
     This method stuffs test methods into unittest.TestCase classes, and then stuffs those
-    classes into modules before loading them as a TestRun
+    classes into modules before loading them as a TestRun.
 
     :param: generate_test_description_fn A python function with a signature matching
-        generate_test_description.
+        `generate_test_description`'s.
     :param: pre_shutdown_tests A list of test methods to run concurrently with the processes
-        described by the generate_test_description function
+        described by `generate_test_description`.
     :param: post_shutdown_tests A list of test methods to run after the launched processes
-        have shut down
+        have shut down.
     """
     test_module = types.ModuleType('test_module')
     test_module.generate_test_description = generate_test_description_fn
--- a/launch_testing/test/conftest.py
+++ b/launch_testing/test/conftest.py
@@ -1,0 +1,69 @@
+# Copyright 2019 Apex.AI, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import types
+import unittest
+
+from launch_testing import post_shutdown_test
+from launch_testing.loader import LoadTestsFromPythonModule
+import pytest
+
+
+def _source_test_loader(generate_test_description_fn,
+                        pre_shutdown_tests=[],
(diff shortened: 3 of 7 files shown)


launch_testing/launch_testing/proc_info_handler.py
```diff
@@ -109,6 +109,35 @@ def assertWaitForShutdown(self,
...
+            try:
+                resolved_process = resolveProcesses(
+                    info_obj=self._proc_info_handler,
+                    process=process,
+                    cmd_args=cmd_args,
+                    strict_proc_matching=True
+                )[0]
+
+                process_event = self._proc_info_handler[resolved_process]
+                return hasattr(process_event, 'returncode')
```
hidmic: @pbaughman why not checking explicitly for process exit events?

launch_testing/test/launch_testing/test_proc_info_assertions.py
```diff
@@ -0,0 +1,137 @@
...
+    # This is kind of a weird test-within-a-test, but it's the easiest way to get
+    # all of the proc_info handlers hooked up correctly without digging deep into
+    # the guts of the runner
+    class ProcInfoTests(unittest.TestCase):
+
+        def test_01_check_running_process(self, proc_info, good_process):
+            try:
+                proc_info.assertWaitForShutdown(good_process, timeout=3)  # Should raise
+                assert False, "assertWaitForShutdown did not raise"
+            except AssertionError as e:
```
hidmic: @pbaughman there's [`self.assertRaises()`](https://docs.python.org/3.6/library/unittest.html#unittest.TestCase.assertRaises) too.

launch_testing/test/launch_testing/test_proc_info_assertions.py
```diff
@@ -0,0 +1,137 @@
...
+
+        def test_02_check_when_process_exits(self, proc_info, good_process):
+            proc_info.assertWaitForShutdown(good_process, timeout=15)
+
+    test_module = types.ModuleType('test_module')
+    test_module.generate_test_description = generate_test_description
+    test_module.ProcInfoTests = ProcInfoTests
+
+    runner = LaunchTestRunner(
+        LoadTestsFromPythonModule(test_module)
```
hidmic: @pbaughman meta: can't we feed test cases in other, more direct way?

launch_testing/launch_testing/proc_info_handler.py
pbaughman: Because when I write Python, I forget that types are a thing!  I like your idea - it's more explicit.

launch_testing/test/launch_testing/test_proc_info_assertions.py
pbaughman: Yeah, this is just a brain-fart.  I'll change it.

launch_testing/test/launch_testing/test_proc_info_assertions.py
pbaughman: @hidmic Are you thinking a factory method that the tests use, or are you thinking something else?  I agree this is getting a little out of hand.  The following test files do this or something similar:
  * test_proc_info_assertions.py
  * test_xml_output.py (doesn't actually run the tests though)
  * test_resolve_process.py
  * test_runner_results.py

I want to use the real 'LaunchTestRunner' from the tests, which needs a `TestRun` object which is somewhat tightly coupled to the `LoadTestsFromPythonModule` function that is normally used to create them.

I could create a different factory function that makes TestRun objects without so much boilerplate.  Would that be a test-only thing, or would it be worth making that part of the real API?  Again - I agree that this could be better - open to input

pbaughman:
@hidmic I addressed your first two comments.  Let me puzzle about the 3rd one a bit.  I'll probably make a function like "LoadTestsFromPythonModule" that doesn't need everything to be packed into a module first.

launch_testing/test/launch_testing/test_proc_info_assertions.py
hidmic: > I could create a different factory function that makes TestRun objects without so much boilerplate.

That'd be ideal :)

pbaughman: I think no matter what we do, we're going to have to stuff things into a module so we don't have to replicate the discovery logic in https://github.com/python/cpython/blob/master/Lib/unittest/loader.py#L98-L136

At the end of the day, the unittest loader expects a module.

I'm going to make this function available to the tests.  We can decide later if the function should be part of the regular API instead.  I worry that right now I don't understand the edge cases well enough to support this as part of the regular API.

pbaughman:
@hidmic Ok, please tell me what you think of the latest commit.  It gets some boilerplate out of the tests and into a common conftest.py file

I was not able to update test_resolve_process.py because I couldn't figure out how to use a pytest fixture from a python unittest setUpClass function, so it still packs things into a module itself.

launch_testing/launch_testing/test_runner.py
```diff
@@ -20,7 +20,7 @@
 from launch import LaunchDescription
 from launch import LaunchService
 from launch.actions import RegisterEventHandler
-from launch.event_handlers import OnProcessExit
+from launch.event_handlers import OnProcessExit, OnProcessStart
```
hidmic: @pbaughman nit: consider one import per line.

launch_testing/test/conftest.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def _source_test_loader(generate_test_description_fn,
+                        pre_shutdown_tests=[],
+                        post_shutdown_tests=[]):
+
+    # The launch_testing API knows how to load tests from python modules.  Rather than
+    # refactoring launch_testing and python unittest, we'll leverage what we already have
+    # by stuffing test cases into a module first, then using the existing infrastructure to
+    # load it
```
hidmic: @pbaughman nit: consider making this comments part of the function docstring

launch_testing/test/conftest.py
```diff
@@ -0,0 +1,69 @@
...
+            setattr(PostShutdownTestClass, test_func.__name__, test_func)
+
+        setattr(test_module, 'PostShutdownTests', PostShutdownTestClass)
+
+    return LoadTestsFromPythonModule(test_module)
+
+
+@pytest.fixture
+def source_test_loader():
+
```
hidmic: @pbaughman nit: unnecessary line?

launch_testing/test/conftest.py
```diff
@@ -0,0 +1,69 @@
...
+    return _source_test_loader
+
+
+@pytest.fixture(scope='class')
+def source_test_loader_class_fixture(request):
+
+    def wrapper(self, *args, **kwargs):
+        return _source_test_loader(*args, **kwargs)
+
+    request.cls.source_test_loader = wrapper
```
hidmic: @pbaughman is this here to support `unittest` style tests under `pytest`?

pbaughman: Yeah, some of the tests are part of test classes, and some are pytest style tests.  The fixture above is for 'pytest' tests and this one is for classes that inherit from unittest.TestCase

launch_testing/test/conftest.py
pbaughman: Do you mean the blank line?

hidmic: Yes!

launch_testing/test/conftest.py
hidmic: I see. Would you consider documenting it?

launch_testing/test/launch_testing/test_proc_info_assertions.py
```diff
@@ -0,0 +1,128 @@
...
+            pre_shutdown_tests=[
+                test_01_check_for_non_running_process,
+                test_02_check_when_process_runs
+            ]
+        )
+    )
+    results = runner.run()
+
+    for result in results.values():
+        assert result.wasSuccessful()
```
hidmic: Hmm, on a second look this seems like a launch test fit into a `pytest`. It's fine for now but we should probably standardize into something less terse and/or use launch test where appropriate.

hidmic:
CI going all the way up to `test_communication`:

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/7878/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3934/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6427/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7735/)

launch_testing/test/launch_testing/test_proc_info_assertions.py
pbaughman: @hidmic You're not wrong - line 58/105 explains what's going on.  I want to use the runner here because plumbing the proc_info handler to the launch is a bit complicated and would be prone to errors if I did it in the test.

If you think it would be more clear to make these into launch_tests, I can try to do it that way.  Making fake Launch events is a bit difficult - I felt it was less error prone to use 'launch' to make real events

hidmic: No, fake `launch` events doesn't sound great either. This likely falls in the bucket of things we need to iterate on as a whole to provide a better API to our users.

hidmic:
@pbaughman friendly ping

pbaughman:
@hidmic Thanks - got backed up with stuff - starting work on this right now!

pbaughman:
@hidmic Ok!  Give that a try

launch_testing/test/conftest.py
```diff
@@ -23,12 +23,23 @@
...
+
+    The launch_testing API knows how to load tests from files on disk and from python modules.
+    When we want to create a TestRun from a unit test, we usually have functions, not files
+    and modules.
+    Rather than refactoring launch_testing and python unittest, we leverage what we already have.
+    This method stuffs test methods into unittest.TestCase classes, and then stuffs those
+    classes into modules before loading them as a TestRun
+
+    :param: generate_test_description_fn A python function with a signature matching
+        generate_test_description.
```
hidmic: @pbaughman consider \`generate_test_description\`'s instead (plus backticks)

launch_testing/test/conftest.py
```diff
@@ -23,12 +23,23 @@
...
-
+    """
+    Create a TestRun from python functions.
+
+    The launch_testing API knows how to load tests from files on disk and from python modules.
+    When we want to create a TestRun from a unit test, we usually have functions, not files
+    and modules.
+    Rather than refactoring launch_testing and python unittest, we leverage what we already have.
+    This method stuffs test methods into unittest.TestCase classes, and then stuffs those
+    classes into modules before loading them as a TestRun
```
hidmic: @pbaughman nit: missing trailing period.

launch_testing/test/conftest.py
```diff
@@ -23,12 +23,23 @@
...
+
+    :param: generate_test_description_fn A python function with a signature matching
+        generate_test_description.
+    :param: pre_shutdown_tests A list of test methods to run concurrently with the processes
+        described by the generate_test_description function
+    :param: post_shutdown_tests A list of test methods to run after the launched processes
+        have shut down
```
hidmic: @pbaughman nit: missing trailing periods in both sentences.

launch_testing/test/conftest.py
```diff
@@ -23,12 +23,23 @@
...
+    When we want to create a TestRun from a unit test, we usually have functions, not files
+    and modules.
+    Rather than refactoring launch_testing and python unittest, we leverage what we already have.
+    This method stuffs test methods into unittest.TestCase classes, and then stuffs those
+    classes into modules before loading them as a TestRun
+
+    :param: generate_test_description_fn A python function with a signature matching
+        generate_test_description.
+    :param: pre_shutdown_tests A list of test methods to run concurrently with the processes
+        described by the generate_test_description function
```
hidmic: @pbaughman consider \`generate_test_description\` instead (plus backticks)

pbaughman:
@hidmic I'm good for one more round of changes - hang on a sec. . .

hidmic:
Running CI anyways, as those are small documentation nits. Thanks for the contribution Pete!

hidmic:
Ok, CI is happy, going in. Thanks for the contribution @pbaughman !

PR #279, opened 2019-07-19
Support CLI commands testing

hidmic:
This pull request encompasses all additions to `launch`, and mainly `launch_testing`, to allow for CLI commands testing. 

Connected to https://github.com/ros2/ros2cli/pull/304.


launch_testing/launch_testing/proc_info_handler.py
```diff
@@ -100,14 +100,11 @@ def __getitem__(self, key):
...
-                              *,
-                              timeout=10):
-
-        success = False
-
+    def waitForShutdown(self,
+                        process,
+                        cmd_args=None,
+                        *,
+                        timeout=10):
```
dirk-thomas: I see that the code was using alignment before but our style guide explicitly avoids that. Please wrap after the open parenthesis and indent the following lines one level.

hidmic: Good catch. Done in 60f12a0d2144e7dc88ded827f6ac3390e80f7643.

hidmic:
This fixes issues in `launch_testing` and that's regardless of CLI testing. I'll run CI on this patch only so we can get it in.

hidmic:
Same as in https://github.com/ros2/ros2cli/pull/304, https://github.com/ros2/launch/pull/279/commits/9891ee09d5738862e8946d68d6b1d13817d5d5a5 introduces significant changes and improvements.

launch_testing/launch_testing/tools/output.py
```diff
@@ -77,3 +82,49 @@ def expected_output_from_file(path):
...
+    """
+    if (text is not None) == (lines is not None):
+        raise ValueError('Either lines or text, but not both, must be specified')
+
+    if (expected_text is not None) == (expected_lines is not None):
+        raise ValueError('Either expected lines or text, but not both, must be specified')
+
+    if expected_text is not None:
+        if text is None:
+            text = '\n'.join(lines)
```
ivanpauno: IMO, documentation is not clear enough about this.
I thought you could use both `lines` and `expected_lines`, or `text` and `expected_text`. But I didn't get than combinations like `lines` and `expected_text` were valid.

launch_testing/launch_testing/tools/output.py
```diff
@@ -77,3 +82,49 @@ def expected_output_from_file(path):
             return [re.compile(regex) for regex in f.read().splitlines()]
 
     raise RuntimeError('could not find output check file: {}'.format(path))
+
+
+def expect_output(
+    text=None,
```
ivanpauno: Maybe, it's clearer if you only use keyworded arguments.

launch_testing/launch_testing/markers.py
```diff
@@ -0,0 +1,44 @@
...
+        def _wrapper(*args, **kwargs):
+            n = times
+            while n > 1:
+                try:
+                    return func(*args, **kwargs)
+                except AssertionError:
+                    n -= 1
+            return func(*args, **kwargs)
+        # Keep signature for argument binding
+        _wrapper.__signature__ = inspect.signature(func)
```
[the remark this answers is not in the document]
hidmic: > and `functools.wraps` preserves the signature

Not for the `inspect` module. Try [`inspect.getfullargspec()`](https://docs.python.org/3.6/library/inspect.html#inspect.getfullargspec) on a decorated function that has no `__signature__`.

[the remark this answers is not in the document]
hidmic: After an f2f with @ivanpauno and some additional digging into `inspect`'s documentation, `inspect.signature()` provides an alternative to `inspect.getfullargspec()` but while the former abides to the `__wrapped__` attribute that `functools.wraps()` gives the latter doesn't. `inspect .getfullargspec()` was actually deprecated in favor of `inspect.signature` in Python 3.5 and then brought back in 3.6 (see [here](https://docs.python.org/3.6/library/inspect.html#inspect.getfullargspec)).

Conclusion: we're moving away from `inspect.getfullargspec()`.

launch_testing/launch_testing/tools/output.py
hidmic: That was the case originally, but not requiring `text` to be a keyword plays its role when `expect_output` is used as a condition for `ProcessProxy.wait_for_output` (or we force a boilerplate lambda). See [here](https://github.com/ros2/launch/pull/279/files#diff-e6ac5ee5445a9a82cfbe47983f5e4529R66-R84).

launch_testing/launch_testing/tools/output.py
hidmic: Done in d3982e54e9449d63d353bf0a00753b3489b75e7d

launch_testing/launch_testing/markers.py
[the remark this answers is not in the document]
hidmic: Done in e007f549299d7c56c5e0ca978af6f28fd7eda257.

hidmic:
CI (full, against Fast-RTPS, Connext and Opensplice)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8398/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4313/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6837/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8297/)

launch_testing/launch_testing/tools/output.py
ivanpauno: Makes sense

PR #280, opened 2019-07-19
fix encoding handling when writing to stdout and log files

dirk-thomas:
Related to ros2/rosidl#388.

### DIFF
--- a/launch/launch/logging/__init__.py
+++ b/launch/launch/logging/__init__.py
@@ -334,8 +334,9 @@ def get_screen_handler():
     """
     if launch_config.screen_handler is None:
         stream = codecs.StreamWriter(sys.stdout, errors='replace')
-        stream.encode = lambda msg, errors='strict': (
-            msg.encode(locale.getpreferredencoding(False), errors).decode(),
+        stream.encode = lambda msg, errors='replace': (
+            msg.encode(locale.getpreferredencoding(False), errors).decode(
+                errors=errors),
             msg)
         launch_config.screen_handler = handlers.StreamHandler(stream)
         launch_config.screen_handler.setFormatter(launch_config.screen_formatter)
--- a/launch/launch/logging/__init__.py
+++ b/launch/launch/logging/__init__.py
@@ -335,8 +335,7 @@ def get_screen_handler():
     if launch_config.screen_handler is None:
         stream = codecs.StreamWriter(sys.stdout, errors='replace')
         stream.encode = lambda msg, errors='replace': (
-            msg.encode(locale.getpreferredencoding(False), errors).decode(
-                errors=errors),
+            msg.encode(locale.getpreferredencoding(False), errors).decode(),
             msg)
         launch_config.screen_handler = handlers.StreamHandler(stream)
         launch_config.screen_handler.setFormatter(launch_config.screen_formatter)
--- a/launch/launch/logging/__init__.py
+++ b/launch/launch/logging/__init__.py
@@ -335,7 +335,8 @@ def get_screen_handler():
     if launch_config.screen_handler is None:
         stream = codecs.StreamWriter(sys.stdout, errors='replace')
         stream.encode = lambda msg, errors='replace': (
-            msg.encode(locale.getpreferredencoding(False), errors).decode(),
+            msg.encode(locale.getpreferredencoding(False), errors).decode(
+                locale.getpreferredencoding(False), errors=errors),
             msg)
         launch_config.screen_handler = handlers.StreamHandler(stream)
         launch_config.screen_handler.setFormatter(launch_config.screen_formatter)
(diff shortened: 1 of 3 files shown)


dirk-thomas:
With these fixes in launch the build succeeds but fails the pub/sub tests as well as results in invalid tokens in the xUnit result file: [Build Status](https://ci.ros2.org/job/ci_windows/7456/)

launch/launch/logging/__init__.py
```diff
@@ -331,7 +333,12 @@ def get_screen_handler():
     See launch_config() documentation for screen logging configuration.
     """
     if launch_config.screen_handler is None:
-        launch_config.screen_handler = handlers.StreamHandler(sys.stdout)
+        stream = codecs.StreamWriter(sys.stdout, errors='replace')
+        stream.encode = lambda msg, errors='replace': (
+            msg.encode(locale.getpreferredencoding(False), errors).decode(
+                errors=errors),
+            msg)
+        launch_config.screen_handler = handlers.StreamHandler(stream)
```
wjwwood: This building a custom `StreamWriter` seems like a departure from normal use, is there a SO or example somewhere that you're pulling from?

dirk-thomas: Since the `StreamHandler` itself doesn't support an encoding the stream has to apply the encoding: see https://www.reddit.com/r/learnpython/comments/95uzy2/logging_how_to_define_encoding_of/

dirk-thomas:
The latest state of this patch passed the test execution without encode/decode exceptions: [Build Status](https://ci.ros2.org/job/ci_windows/7462/) (though some tests are still failing for unrelated reasons).

launch/launch/logging/__init__.py
```diff
@@ -331,7 +333,12 @@ def get_screen_handler():
     See launch_config() documentation for screen logging configuration.
     """
     if launch_config.screen_handler is None:
-        launch_config.screen_handler = handlers.StreamHandler(sys.stdout)
+        stream = codecs.StreamWriter(sys.stdout, errors='replace')
+        stream.encode = lambda msg, errors='replace': (
+            msg.encode(locale.getpreferredencoding(False), errors).decode(
+                locale.getpreferredencoding(False), errors=errors),
```
hidmic: @dirk-thomas do we need to `decode()`? I think it'll be encoded again when it reaches `sys.stdout`.

launch/launch/logging/__init__.py
```diff
@@ -331,7 +333,12 @@ def get_screen_handler():
     See launch_config() documentation for screen logging configuration.
     """
     if launch_config.screen_handler is None:
-        launch_config.screen_handler = handlers.StreamHandler(sys.stdout)
+        stream = codecs.StreamWriter(sys.stdout, errors='replace')
+        stream.encode = lambda msg, errors='replace': (
```
hidmic: @dirk-thomas hmm can't we use `codecs.getwriter(locale.getpreferredencoding(False))` to get the appropriate `StreamWriter` subclass instead of patching an instance?

launch/launch/logging/__init__.py
```diff
@@ -358,7 +365,8 @@ def get_log_file_handler(file_name='launch.log'):
             else:
                 launch_config.log_handler_factory = handlers.FileHandler
         file_path = get_log_file_path(file_name)
-        file_handler = launch_config.log_handler_factory(file_path)
+        file_handler = launch_config.log_handler_factory(
+            file_path, encoding='utf-8')
```
hidmic: @dirk-thomas why do we force `UTF-8` for files and use `locale.getpreferredencoding()` for standard streams only?

launch/launch/logging/__init__.py
dirk-thomas: > do we need to decode()?

Yes, we need `decode()` since the API for writing only accepts `str` - not `bytes`.

> I think it'll be encoded again when it reaches `sys.stdout`.

Exactly that is why it needs to be decoded into a `str`.

launch/launch/logging/__init__.py
dirk-thomas: We usually write all files in UTF-8. The standard streams have their own encoding which we have to use as is.

launch/launch/logging/__init__.py
dirk-thomas: I am not sure how you can use that API to replace the existing logic. Can you please post the snippet you have in mind which sets the `launch_config.screen_handler`.

launch/launch/logging/__init__.py
hidmic: Hmm, I see the intent now. We _could_ use the binary raw stream e.g. `sys.stdout.buffer`, but if a client library replaces `sys.stdout` with another object it'll fall apart. I guess we'll have to wait for [`reconfigure()`](https://docs.python.org/3/library/io.html#io.TextIOWrapper.reconfigure).

launch/launch/logging/__init__.py
hidmic: Nevermind, this suggestion only applies if `stream.encode` actually does encode.

launch/launch/logging/__init__.py
hidmic: Fair enough.

PR #286, opened 2019-07-25
reenable emulate_tty by default

wjwwood:
This is a follow up of https://github.com/ros2/launch/pull/277

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -96,7 +96,7 @@ class ExecuteProcess(Action):
             'sigterm_timeout', default=5),
         sigkill_timeout: SomeSubstitutionsType = LaunchConfiguration(
             'sigkill_timeout', default=5),
-        emulate_tty: bool = False,
+        emulate_tty: bool = True,
         prefix: Optional[SomeSubstitutionsType] = None,
         output: Text = 'log',
         output_format: Text = '[{this.name}] {line}',
@@ -175,7 +175,7 @@ class ExecuteProcess(Action):
             as a string or a list of strings and Substitutions to be resolved
             at runtime, defaults to the LaunchConfiguration called
             'sigkill_timeout'
-        :param: emulate_tty emulate a tty (terminal), defaults to False, but can
+        :param: emulate_tty emulate a tty (terminal), defaults to True, but can
             be overridden with the LaunchConfiguration called 'emulate_tty',
             the value of which is evaluated as true or false according to
             :py:func:`evaluate_condition_expression`.
@@ -345,6 +345,8 @@ class ExecuteProcess(Action):
             raise RuntimeError('Signal event received before execution.')
         if self._subprocess_transport is None:
             raise RuntimeError('Signal event received before subprocess transport available.')
+        if self._subprocess_protocol is None:
+            raise RuntimeError('Signal event received before subprocess protocol available.')
         if self._subprocess_protocol.complete.done():
             # the process is done or is cleaning up, no need to signal
             self.__logger.debug(
@@ -390,6 +392,8 @@ class ExecuteProcess(Action):
     def __on_process_stdout(
         self, event: ProcessIO
     ) -> Optional[SomeActionsType]:
+        if self.__stdout_buffer.closed:
+            raise RuntimeError('OnProcessIO (stdout) received after stdout buffer closed.')
         self.__stdout_buffer.write(event.text.decode(errors='replace'))
         self.__stdout_buffer.seek(0)
         last_line = None
@@ -409,6 +413,8 @@ class ExecuteProcess(Action):
     def __on_process_stderr(
         self, event: ProcessIO
     ) -> Optional[SomeActionsType]:
+        if self.__stderr_buffer.closed:
+            raise RuntimeError('OnProcessIO (stderr) received after stderr buffer closed.')
         self.__stderr_buffer.write(event.text.decode(errors='replace'))
         self.__stderr_buffer.seek(0)
         last_line = None
--- a/launch_testing/launch_testing/test_runner.py
+++ b/launch_testing/launch_testing/test_runner.py
@@ -59,7 +59,6 @@ class _RunnerWorker():
         self._test_tr = threading.Thread(
             target=self._run_test,
             name='test_runner_thread',
-            daemon=True
         )
 
     def run(self):
@@ -159,6 +158,9 @@ class _RunnerWorker():
         ).run(self._test_run.post_shutdown_tests)
 
         self._results.append(inactive_results)
+
+        # Join the test thread before returning to avoid it being daemonized.
+        self._test_tr.join()
 
         return self._results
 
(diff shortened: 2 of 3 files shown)


launch/launch/actions/execute_process.py
```diff
@@ -390,6 +392,8 @@ def __on_process_stdin(
     def __on_process_stdout(
         self, event: ProcessIO
     ) -> Optional[SomeActionsType]:
+        if self.__stdout_buffer.closed:
+            raise RuntimeError('OnProcessIO (stdout) received after stdout buffer closed.')
```
wjwwood: This this check, I get tracebacks like this:

```
  File "/home/william/ros2_ws/build/launch/launch/launch_service.py", line 350, in run
    self.__loop_from_run_thread.run_until_complete(run_loop_task)
  File "/usr/lib/python3.6/asyncio/base_events.py", line 484, in run_until_complete
    return future.result()
  File "/home/william/ros2_ws/build/launch/launch/launch_service.py", line 240, in __run_loop
    await process_one_event_task
  File "/usr/lib/python3.6/asyncio/coroutines.py", line 126, in send
    return self.gen.send(value)
  File "/home/william/ros2_ws/build/launch/launch/launch_service.py", line 177, in _process_one_event
  File "/usr/lib/python3.6/asyncio/coroutines.py", line 110, in __next__
    return self.gen.send(None)
  File "/home/william/ros2_ws/build/launch/launch/launch_service.py", line 186, in __process_event
    entities = event_handler.handle(event, self.__context)
  File "/home/william/ros2_ws/build/launch/launch/event_handlers/on_process_io.py", line 73, in handle
    return self.__on_stdout(event)
  File "/home/william/ros2_ws/build/launch/launch/actions/execute_process.py", line 396, in __on_process_stdout
    raise RuntimeError('OnProcessIO (stdout) received after stdout buffer closed.')
RuntimeError: OnProcessIO (stdout) received after stdout buffer closed.

[ERROR] [launch]: Caught exception in launch (see debug for traceback): OnProcessIO (stdout) received after stdout buffer closed.
```

I believe this because this line buffering logic was not designed in a way to allow for ProcessIO events to occur after the ProcessExit event comes. @ivanpauno do you have any ideas about how we could change it to allow for that race condition?

launch_testing/launch_testing/test_runner.py
```diff
@@ -59,7 +59,6 @@ def __init__(self,
         self._test_tr = threading.Thread(
             target=self._run_test,
             name='test_runner_thread',
-            daemon=True
```
wjwwood: @pbaughman do you know why this was a daemon thread?

As a daemon thread I sporadically get crashes on exit because this daemon thread tries to write to stdout which is already closed by the main thread (the main thread exiting will close the global file handles but not join daemon threads). See:

https://stackoverflow.com/questions/45267439/fatal-python-error-and-bufferedwriter

Also, see:

https://docs.python.org/3.7/library/threading.html#thread-objects (part about daemonic threads)

It seems to work ok, but I wanted to know if there was a reason for it before.

wjwwood: Example of the error I was getting before I changed this:

```
----------------------------------------------------------------------
Ran 1 test in 0.050s

Fatal Python error: could not acquire lock for <_io.BufferedWriter name='<stderr>'> at interpreter shutdown, possibly due to daemon threads

Thread 0x00007fe5f2ddd700 (most recent call first):
  File "/usr/lib/python3.6/unittest/runner.py", line 220 in run
  File "/home/william/ros2_ws/build/launch_testing/launch_testing/test_runner.py", line 180 in _run_test
  File "/usr/lib/python3.6/threading.py", line 864 in run
  File "/usr/lib/python3.6/threading.py", line 916 in _bootstrap_inner
  File "/usr/lib/python3.6/threading.py", line 884 in _bootstrap

Current thread 0x00007fe5f7392740 (most recent call first):
[1]    21192 abort (core dumped)  python3 -m launch_testing.launch_test  --verbose
```

pbaughman: @wjwwood It's a daemon thread because I didn't want it to accidentally prevent the program from exiting if you ctrl+c and the worker thread got hung and I had had problems with this in a past life, but that was for a program where it didn't matter if there was a nasty stack trace when shutting down.  It doesn't necessarily need to be.

launch/launch/actions/execute_process.py
ivanpauno: I'm flushing the buffers after a `ProcessExit` event.
https://github.com/ros2/launch/blob/f82c364dd1f5ea67beeaec9ecaa03a4ac22aa043/launch/launch/actions/execute_process.py#L675-L678.
The method for flushing buffers leaves them in a wrong state for handling another `ProcessIO` event:
https://github.com/ros2/launch/blob/f82c364dd1f5ea67beeaec9ecaa03a4ac22aa043/launch/launch/actions/execute_process.py#L434-L446

One easy change, is to restore the buffers to a correct state after handling `ProcessExit` event.
It will, at least, don't raise an error with a new `ProcessIO` event.
The only bad thing: I think it's impossible to ensure the buffer will be completely flushed after the process exited.

Another option: Ignore `ProcessIO` events after flushing the buffers, instead of raising an error.

ivanpauno: Raising an error here, as you proposed is also fine. If the daemon node of `launch_testing` is causing this race condition, I think the best solution is stop using a daemon node.

hidmic: > Ignore ProcessIO events after flushing the buffers, instead of raising an error.

Hmm, if we cannot guarantee order, neither of those options will do. I don't know how many or how often we get this sort of out-of-band `ProcessIO` events, but we _could_ stop buffering after `ProcessExit`.

ivanpauno: > Hmm, if we cannot guarantee order, neither of those options will do. I don't know how many or how often we get this sort of out-of-band `ProcessIO` events, but we _could_ stop buffering after `ProcessExit`.

Sounds reasonable.

wjwwood: > If the daemon node of launch_testing is causing this race condition, I think the best solution is stop using a daemon node.

That's not the cause, the cause is using a pty for stdout and stderr of the subprocess, which means there is different timing (apparently) for when data becomes available on those pipes and the process exiting. Perhaps instead we should not flush until the stdout and stderr are closed (I think you can setup a protocol event for when that happens).

wjwwood: Specifically this one: https://docs.python.org/3/library/asyncio-protocol.html#asyncio.SubprocessProtocol.pipe_connection_lost

I'll need to think about the lifetime of these objects and make sure things stick around long enough.

tonynajjar:
Hello, the PR has been opened for 5 years, could we know if you still want to go through with it? If not can the PR be closed?
I'm asking because that's a "wide reach" feature, in the sense that so many ROS users have this flag enabled for all their nodes; it would be nice to get a conclusion on whether this will become the default or not. Thank you

Ryanf55:
People are still getting hung up with it online.
https://robotics.stackexchange.com/questions/104557/ros2-launch-no-way-to-get-log-displayed-in-console

Mixing different loggers in large apps causes some logs to be dropped without this.

PR #288, opened 2019-07-25
Add support for not optional environment variable substitution

ivanpauno:
Based on comment https://github.com/ros2/ros2_documentation/pull/302#discussion_r306315723.

This is not exactly what I first commented, but mimics what we where doing in `LaunchConfiguration`.
I think it makes more sense.
https://github.com/ros2/launch/blob/0f81d987039420036a7153c622b075c993ff1558/launch/launch/substitutions/launch_configuration.py#L32-L40
https://github.com/ros2/launch/blob/0f81d987039420036a7153c622b075c993ff1558/launch/launch/substitutions/launch_configuration.py#L67-L76

This breaks API. If someone was relying on `default_value` being `''`, not it will raise a `SubstitutionError`.
The fix is passing `default_value=''` explicitly.

### DIFF
--- a/launch/launch/substitutions/environment_variable.py
+++ b/launch/launch/substitutions/environment_variable.py
@@ -17,8 +17,10 @@
 import os
 from typing import Iterable
 from typing import List
+from typing import Optional
 from typing import Text
 
+from .substitution_failure import SubstitutionFailure
 from ..frontend.expose import expose_substitution
 from ..launch_context import LaunchContext
 from ..some_substitutions_type import SomeSubstitutionsType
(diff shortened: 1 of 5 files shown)


launch/launch/substitutions/environment_variable.py
```diff
@@ -37,23 +39,38 @@ def __init__(
...
         *,
-        default_value: SomeSubstitutionsType = ''
+        default_value: Optional[SomeSubstitutionsType] = None
     ) -> None:
-        """Constructor."""
+        """
+        Construct an enviroment variable substitution.
+
+        The enviroment variable could be optional or not:
+            - When `default_value` is `None`, the environment variable is non optional.
```
wjwwood:
```suggestion
            - When `default_value` is `None`, the environment variable is not optional.
```

launch/launch/substitutions/environment_variable.py
```diff
@@ -37,23 +39,38 @@ def __init__(
...
 
     @classmethod
     def parse(cls, data: Iterable[SomeSubstitutionsType]):
         """Parse `EnviromentVariable` substitution."""
         if len(data) < 1 or len(data) > 2:
             raise TypeError('env substitution expects 1 or 2 arguments')
         kwargs = {'name': data[0]}
-        if len(data) == 2:
-            kwargs['default_value'] = data[1]
+        kwargs['default_value'] = data[1] if len(data) == 2 else None
```
wjwwood: What happens if `len(data)` > 2?

ivanpauno: I forgot undoing this change (it doesn't change behavior, but the old style is better).
L70 -71 handles `len(data) > 2`, it raises an error.

ivanpauno:
Rebased with master, since there were merge conflicts.

Full CI (only fastrtps):
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/7671/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3779/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6278/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7516/)

ivanpauno:
CI failures are all unrelated. Same three failures in windows [today nightly job](https://ci.ros2.org/view/nightly/job/nightly_win_rel/1260/#showFailuresLink).

PR #291, opened 2019-07-26
Add deprecated argument to LaunchDescription, add ThisLaunchFile substitution

ivanpauno:
Similar to ROS 1.
It allows deprecating a launch description.

### DIFF
--- a/launch/launch/actions/include_launch_description.py
+++ b/launch/launch/actions/include_launch_description.py
@@ -104,8 +104,18 @@ class IncludeLaunchDescription(Action):
         else:
             return self.__launch_arguments
 
-    def _get_launch_file_location(self):
+    def _get_launch_file(self):
         return os.path.abspath(self.__launch_description_source.location)
+
+    def _get_launch_file_directory(self):
+        launch_file_location = self._get_launch_file()
+        if os.path.exists(launch_file_location):
+            launch_file_location = os.path.dirname(launch_file_location)
+        else:
+            # If the location does not exist, then it's likely set to '<script>' or something
+            # so just pass it along.
+            launch_file_location = self.__launch_description_source.location
+        return launch_file_location
 
     def describe_sub_entities(self) -> List[LaunchDescriptionEntity]:
         """Override describe_sub_entities from LaunchDescriptionEntity to return sub entities."""
@@ -115,15 +125,13 @@ class IncludeLaunchDescription(Action):
     def visit(self, context: LaunchContext) -> List[LaunchDescriptionEntity]:
         """Override visit to return an Entity rather than an action."""
         launch_description = self.__launch_description_source.get_launch_description(context)
-        location = self._get_launch_file_location()
-        if os.path.exists(location):
-            # If the location does not exist, then it's likely set to '<script>' or something.
-            context.extend_locals({
-                'current_launch_file_path': location,
-            })
-            context.extend_locals({
-                'current_launch_file_directory': os.path.dirname(location),
-            })
+        # If the location does not exist, then it's likely set to '<script>' or something.
+        context.extend_locals({
+            'current_launch_file_path': self._get_launch_file(),
+        })
+        context.extend_locals({
+            'current_launch_file_directory': self._get_launch_file_directory(),
+        })
 
         # Do best effort checking to see if non-optional, non-default declared arguments
         # are being satisfied.
--- a/launch/launch/actions/include_launch_description.py
+++ b/launch/launch/actions/include_launch_description.py
@@ -105,14 +105,7 @@ class IncludeLaunchDescription(Action):
             return self.__launch_arguments
 
     def _get_launch_file_location(self):
-        launch_file_location = os.path.abspath(self.__launch_description_source.location)
-        if os.path.exists(launch_file_location):
-            launch_file_location = os.path.dirname(launch_file_location)
-        else:
-            # If the location does not exist, then it's likely set to '<script>' or something
-            # so just pass it along.
-            launch_file_location = self.__launch_description_source.location
-        return launch_file_location
+        return os.path.abspath(self.__launch_description_source.location)
 
     def describe_sub_entities(self) -> List[LaunchDescriptionEntity]:
         """Override describe_sub_entities from LaunchDescriptionEntity to return sub entities."""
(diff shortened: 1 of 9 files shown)


launch/launch/launch_description.py
```diff
@@ -45,12 +48,19 @@ class LaunchDescription(LaunchDescriptionEntity):
...
     ) -> None:
         """Constructor."""
         self.__entities = list(initial_entities) if initial_entities is not None else []
+        self.__deprecated = deprecated
 
     def visit(self, context: LaunchContext) -> Optional[List[LaunchDescriptionEntity]]:
         """Override visit from LaunchDescriptionEntity to visit contained entities."""
+        if self.__deprecated is not None:
+            launch.logging.get_logger().warning(
+                'Deprecated launch file: {}'.format(self.__deprecated)
```
wjwwood: It would be more useful if this included where or which launch description was deprecated. I'm try to remember if this information was already piped some where, as it could be captured as `<string>` (if done in an interpreter), or `some_launch_file_launch.py`, or `some_launch_file.launch.xml`, etc... It should be possible (technically) to ensure the launch description knows where it came from in most cases (when being included by another launch file or being run directly from `ros2 launch`, etc...).

ivanpauno:
@wjwwood While addressing your comment, I added a `ThisLaunchFile` substitution and refactored `ThisLaunchFileDir` code a little bit.

launch/launch/actions/include_launch_description.py
```diff
@@ -122,9 +115,15 @@ def describe_sub_entities(self) -> List[LaunchDescriptionEntity]:
...
-        })
+        location = self._get_launch_file_location()
+        if os.path.exists(location):
+            # If the location does not exist, then it's likely set to '<script>' or something.
+            context.extend_locals({
+                'current_launch_file_path': location,
+            })
+            context.extend_locals({
+                'current_launch_file_directory': os.path.dirname(location),
+            })
```
wjwwood: Why move this logic here? Also as currently written when it is `<string>` the `current_launch_file_directory` is never set, whereas before it was.

I think it should always be set, just like `__file__` is always set in python even if you're in an interpreter (in which case it uses `<string>` as well).

launch/launch/launch_description.py
```diff
@@ -45,12 +48,23 @@ class LaunchDescription(LaunchDescriptionEntity):
...
         """Constructor."""
         self.__entities = list(initial_entities) if initial_entities is not None else []
+        self.__deprecated = deprecated
 
     def visit(self, context: LaunchContext) -> Optional[List[LaunchDescriptionEntity]]:
         """Override visit from LaunchDescriptionEntity to visit contained entities."""
+        if 'current_launch_file_path' in context.get_locals_as_dict():
+            if self.__deprecated is not None:
+                launch.logging.get_logger().warning(
+                    ' launch file [{}] is deprecated: {}'.format(
```
wjwwood: Do you need a leading space when doing a warning message?

launch/launch/launch_description.py
```diff
@@ -124,3 +138,8 @@ def add_entity(self, entity: LaunchDescriptionEntity) -> None:
     def add_action(self, action: Action) -> None:
         """Add an action to the LaunchDescription."""
         self.add_entity(action)
+
+    @property
+    def deprecated(self) -> Optional[Text]:
+        """Getter for deprecated."""
```
wjwwood: This is a bit odd for me, as I'd expect `ld.deprecated` to return a boolean, and `ld.deprecation_reason` to return `None` if no deprecated or a string if deprecated.

If we leave it as-is, I think you should put into the docstring that `None` means it is not deprecated, and that a string means it is deprecated (and also what the string is, i.e. the description of why it is deprecated).

launch/launch/substitutions/this_launch_file.py
```diff
@@ -0,0 +1,58 @@
...
+        Perform the substitution by returning the path to the current launch file.
+
+        If there is no current launch file, i.e. if run from a script, then an
+        error is raised.
+
+        :raises: SubstitutionFailure if not in a launch file
+        """
+        if 'current_launch_file_path' not in context.get_locals_as_dict():
+            raise SubstitutionFailure(
+                'ThisLaunchFile used outside of a launch file (in a script)')
```
wjwwood: I would prefer this to be an "exceptional" exception, like `'current_launch_file_path' unexpectedly not set` and the logic elsewhere to produce a value of `<string>` when it is a script.

I understand wanting the substitution to fail when it's not a file, but in that case I'd just check here for `os.path.exists` instead of in the logic to generate the `current_launch_file_path`.

I could imagine `$(filename)` being used with `$(file-exists $(filename))` or having a no-throw version like `$(filename-no-throw)`.

launch/launch/launch_description.py
```diff
@@ -45,12 +48,23 @@ class LaunchDescription(LaunchDescriptionEntity):
...
+        *,
+        deprecated_reason: Optional[Text] = None
     ) -> None:
         """Constructor."""
         self.__entities = list(initial_entities) if initial_entities is not None else []
+        self.__deprecated_reason = deprecated_reason
 
     def visit(self, context: LaunchContext) -> Optional[List[LaunchDescriptionEntity]]:
         """Override visit from LaunchDescriptionEntity to visit contained entities."""
+        if 'current_launch_file_path' in context.get_locals_as_dict():
```
wjwwood: What about the else case here? Maybe just a runtime error? I think it should never happen.

ivanpauno: I think it happens if you create a `LaunchDescription` in a script and the you manually create a  `LaunchService` with it, since `current_launch_file_path` local is added by `IncludeLaunchDescription` action (if you don't use `IncludeLaunchDescription`, that local doesn't exist).

Maybe, logging an std warning without the file path would be a good idea.
I will do that.

ivanpauno:
CI up to launch, launch_xml, launch_yaml, launch_ros, test_launch_ros (only fastrtps):
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/7718/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3821/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6319/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7573/)

PR #296, opened 2019-08-05
Fix launch_testing output filtering

hidmic:
Precisely what the title says. A spin off from #279, as it's orthogonal to CLI testing.

### DIFF
--- a/launch_testing/launch_testing/tools/output.py
+++ b/launch_testing/launch_testing/tools/output.py
@@ -43,10 +43,10 @@ def basic_output_filter(
         filtered_prefixes = get_default_filtered_prefixes()
     if filtered_patterns is None:
         filtered_patterns = get_default_filtered_patterns()
-    filtered_patterns = map(re.compile, filtered_patterns)
+    filtered_patterns = list(map(re.compile, filtered_patterns))
 
     def _filter(output):
-        filtered_output = []
+        filtered_output_lines = []
         for line in output.splitlines():
             # Filter out stdout that comes from underlying DDS implementation
             # Note: we do not currently support matching filters across multiple stdout lines.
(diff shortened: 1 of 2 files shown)


hidmic:
CI (up to `test_communication`, `test_rclcpp`, `launch_testing` and `launch_testing_ros`):

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/7814/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3890/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6378/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7646/)

launch_testing/launch_testing/tools/output.py
```diff
@@ -43,21 +43,22 @@ def basic_output_filter(
...
                 continue
             if any(pattern.match(line) for pattern in filtered_patterns):
                 continue
-            filtered_output.append(line)
-        if output.endswith(os.linesep):
-            filtered_output.append(os.linesep)
-        return os.linesep.join(filtered_output)
+            filtered_output_lines.append(line)
+        filtered_output = os.linesep.join(filtered_output_lines)
+        if filtered_output and output.endswith('\n'):
```
ivanpauno: Why `/n` and not `os.linesep`? Is `os.linesep` replaced in other place?

hidmic: Ok, this is more complicated than it appears at first. If universal newline support is available (it is by default on Python 3+), `'\n'`, `'\r'` and `'\r\n'` are all converted to `'\n'`. Thus, internally, Python simply uses `'\n'` e.g. for multi-line strings, and then makes up for differences on IO. See [PEP 278](https://www.python.org/dev/peps/pep-0278/) and [PEP 3116](https://www.python.org/dev/peps/pep-3116/). 

However, it looks like [`str.splitlines()`](https://docs.python.org/3.6/library/stdtypes.html#str.splitlines) is way more comprehensive in terms of line boundaries, and Python could've been compiled without universal newline support.

Bottom line, see c9f852512e9b7c78b2521b85b437ba4cbc35ccad.

hidmic:
Alright, going in !

PR #297, opened 2019-08-06
Avoid closing event loop in atexit hook on windows

ivanpauno:
Fixes https://github.com/ros2/demos/issues/354.
We could just avoid registering it in all the platforms, as the original problem seems to be solved in python 3.6.

Note: I haven't found a python issue related to this problem (i.e.: closing loop hangs when called from atexit hook).

### DIFF
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -18,6 +18,7 @@ import asyncio
 import atexit
 import collections.abc
 import logging
+import os
 import signal
 import threading
 import traceback
@@ -54,12 +55,15 @@ _g_loops_used = set()  # type: Set
 # This is only really required pre-3.6, see:
 #   https://github.com/ros2/launch/issues/84
 #   https://bugs.python.org/issue23548
-@atexit.register
-def close_loop():
-    global _g_loops_used
-    for loop in _g_loops_used:
-        if not loop.is_closed():
-            loop.close()
+# Don't add this on Windows. It causes:
+#   https://github.com/ros2/demos/issues/354.
+if os.name != 'nt':
+    @atexit.register
+    def close_loop():
+        global _g_loops_used
+        for loop in _g_loops_used:
+            if not loop.is_closed():
+                loop.close()
 
 
 class LaunchService:


launch/launch/launch_service.py
```diff
@@ -54,12 +55,15 @@
...
-@atexit.register
-def close_loop():
-    global _g_loops_used
-    for loop in _g_loops_used:
-        if not loop.is_closed():
-            loop.close()
+# Don't add this on Windows. It causes:
+#   https://github.com/ros2/demos/issues/354.
+if os.name != 'nt':
+    @atexit.register
```
hidmic: @ivanpauno you're right about not needing this anymore in Python 3.6. @wjwwood @dirk-thomas is there any intention to (even partially) support Python 3.5? If not, we can simply drop this.

dirk-thomas: We are waiting for @nuclearsandwich to resolve this question. At the moment we claim Dashing supports Python 3 (see [REP 2000](https://github.com/ros-infrastructure/rep/blame/7cbc7ed3c822bd1acae60f43b1d7470d7de52863/rep-2000.rst#L382)) but in fact we use 3.6+ syntax in some Python code (see ros2/ros2#735). Depending on which way we go this might not be necessary.

ivanpauno: I think we can go ahead with this fix, and delete the `atexit` hook in a follow-up PR if we decide not to support Pyhton 3.5 more.

wjwwood: I agree that we can continue with this patch as-is and not change anything substantial w.r.t. Python 3.5 support.

pbaughman: Just adding a little info - I think 'launch' is already using some stuff from the 'Types' module that isn't part of python 3.5.  Maybe this was changed, but I think the horse has left the barn on 3.5

pbaughman: Specifically, [this, I think](https://github.com/ros2/launch/blob/master/launch/launch/actions/opaque_coroutine.py#L20)
Sorry I said 'types' above but I mean t'typing'

```
$ python3 --version
Python 3.5.2
$ python3 -c 'from typing import Coroutine'
  File "<string>", line 1, in <module>
ImportError: cannot import name 'Coroutine'
```

ivanpauno:
CI up to launch, launch_xml, launch_yaml, launch_ros, test_launch_ros, demo_nodes_cpp (only fastrtps):
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/7813/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3889/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6377/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7645/)

hidmic:
@ivanpauno @wjwwood backport material?

wjwwood:
The atexit thing is needed for Python 3.5 I believe.

hidmic:
Are we still supporting Python 3.5? If so, [this](https://github.com/ros2/launch/blob/dashing/launch/launch/actions/opaque_coroutine.py) should not be there.

ivanpauno:
> The atexit thing is needed for Python 3.5 I believe.

I believe `atexit` thing was needed in python 3.5 on Linux, but I'm not sure if it was needed on Windows too (the bug why atexit was added was reported on Linux, and probably atexit hook wasn't doing anything on Windows).

wjwwood:
I thought that was the thinking for Dashing, @nuclearsandwich isn't it needed in Dashing for Debian or something?

nuclearsandwich:
> I thought that was the thinking for Dashing, @nuclearsandwich isn't it needed in Dashing for Debian or something?

Please maintain support for Python 3.5 in Dashing. I was trying to find the time this week to fix the outstanding regressions and restore full Python 3.5 support but it didn't happen.

PR #299, opened 2019-08-08
Don't create a log directory every time the launch logger is imported.

### LINKED ISSUE #293: Launch creates a directory on every `ros2` cli invocation
## Bug report

Running *any* `ros2` CLI commands creates a new directory under `~/.ros/log`.  This is basically because `ros2cli` loads every entry point when it starts up, including the `ros2launch` entry point.  `ros2launch` ends up importing `launch`, which imports `logging` (from https://github.com/ros2/launch/blob/master/launch/launch/logging/__init__.py), which creates a directory every time it is imported.

**Required Info:**

- Operating System:
  - Ubuntu Bionic (though this is a generic problem)
- Installation type:
  - Source
- Version or commit hash:
  - 7961b80b5d517228f35ac0b88b6832a2cf0f21f4
- DDS implementation:
  - Fast-RTPS
  - rclpy

#### Steps to reproduce issue
```
rm -rf ~/.ros/log/*
ros2 topic list
ls ~/.ros/log
```

#### Expected behavior
No new directory is created from running a `ros2 topic` command.

#### Actual behavior
A new directory is created every time, of the form "2019-07-31-19-55-52-357234-ros2-bionic-5413"

#### Additional information
The problem is here: https://github.com/ros2/launch/blob/master/launch/launch/logging/__init__.py#L410 .  Since that is not "lazy", it creates a new directory every time that module is imported.

Talking to @hidmic about it, his suggestion is that we make this lazy.  He suggests replacing the setting of attributes on a function (that is currently done at https://github.com/ros2/launch/blob/master/launch/launch/logging/__init__.py#L52), which making a `LaunchConfig` object.  Once we have that object, we can create an `@property` for the `log_dir`, which will then only create the directory (and the name) when it is accessed.

clalancette:
To do this, we revamp the launch_config() function into a class
with member variables and methods.  We then make the 'log_dir'
a property, and if it is None when it is accessed, we create
the directory then.  This ensures that we will not create the
directory until someone actually wants to use it.

The design of this class makes it so that we can no longer call `configure` to do partial configuration.  However, this isn't something we currently use, and if we want to do partial configuration in the future we can add more methods to the `LaunchConfig` class to do so.

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7842/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3910/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6399/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7680/)

Fixes #293

### DIFF
--- a/launch/launch/logging/__init__.py
+++ b/launch/launch/logging/__init__.py
@@ -41,56 +41,15 @@ class LaunchConfig:
     """Launch Logging Configuration class."""
 
     def __init__(self):
-        self._log_dir = None
-        self.file_handlers = {}
-        self.screen_handler = None
-        self.screen_formatter = None
-        self._log_handler_factory = None
+        self.reset()
 
     def reset(self):
         self._log_dir = None
         self.file_handlers = {}
         self.screen_handler = None
         self.screen_formatter = None
+        self.file_formatter = None
         self._log_handler_factory = None
-
-    def configure(
-        self,
-        *,
-        log_format=None,
-        log_style=None,
-    ):
-        """
-        Set up launch logging.
-
-        This function allows you to:
-
-          - Configure log file formats.
-
-        Setup only has side effects for the arguments provided.
-        The setup process is idempotent.
-
-        :param log_format: the format used when logging to the main launch log file,
-            as expected by the `logging.Formatter` constructor.
-            Alternatively, the 'default' alias can be given to log verbosity level,
-            logger name and logged message.
-        :param log_style: the log style used if no alias is given for log_format.
-            No style can be provided if a format alias is given.
-        """
-        if log_format is not None:
-            if log_format == 'default':
-                log_format = '{created:.7f} [{levelname}] [{name}]: {msg}'
-                if log_style is not None:
-                    raise ValueError(
-                        'Cannot set a custom format style for the "default" log format.'
-                    )
-            if log_style is None:
-                log_style = '{'
-            self.file_formatter = logging.Formatter(
-                log_format, style=log_style
-            )
-            for handler in self.file_handlers.values():
-                handler.setFormatter(self.file_formatter)
 
     def set_level(self, new_level):
         """
@@ -209,6 +168,34 @@ class LaunchConfig:
             self.screen_handler = handlers.StreamHandler(stream)
             self.screen_handler.setFormatter(self.screen_formatter)
         return self.screen_handler
+
+    def set_log_format(self, log_format, log_style):
+        """
+        Set up launch log file format.
+
+        :param log_format: the format used when logging to the main launch log file,
+            as expected by the `logging.Formatter` constructor.
+            Alternatively, the 'default' alias can be given to log verbosity level,
+            logger name and logged message.
+        :param log_style: the log style used if no alias is given for log_format.
+            No style can be provided if a format alias is given.
+        """
+        if log_format is not None:
+            if log_format == 'default':
+                log_format = '{created:.7f} [{levelname}] [{name}]: {msg}'
+                if log_style is not None:
+                    raise ValueError(
+                        'Cannot set a custom format style for the "default" log format.'
+                    )
+            if log_style is None:
+                log_style = '{'
+            self.file_formatter = logging.Formatter(
+                log_format, style=log_style
+            )
+            for handler in self.file_handlers.values():
+                handler.setFormatter(self.file_formatter)
+        else:
+            self.file_formatter = None
 
     def get_log_file_path(self, file_name='launch.log'):
         """
@@ -461,9 +448,9 @@ def reset():
         logger.setLevel(logging.NOTSET)
         del logger.handlers[:]
     launch_config.reset()
-    launch_config.configure(log_format='default')
     launch_config.level = logging.INFO
     launch_config.set_screen_format('default', None)
+    launch_config.set_log_format('default', None)
     logging.setLoggerClass(LaunchLogger)
 
 
--- a/launch/launch/logging/__init__.py
+++ b/launch/launch/logging/__init__.py
@@ -55,7 +55,6 @@ class LaunchConfig:
     def configure(
         self,
         *,
-        log_dir=None,
         screen_format=None,
         screen_style=None,
         log_format=None,
@@ -67,7 +66,6 @@ class LaunchConfig:
 
         This function allows you to:
 
-          - Configure the location of log files on disk.
           - Configure screen and log file formats.
 
         Setup only has side effects for the arguments provided.
@@ -78,7 +76,6 @@ class LaunchConfig:
           - 'default' to log verbosity level, logger name and logged message
           - 'default_with_timestamp' to add timestamps to the 'default' format
 
-        :param log_dir: used as base path for all log file collections.
         :param screen_format: format specification used when logging to the screen,
             as expected by the `logging.Formatter` constructor.
             Alternatively, aliases for common formats are available, see above.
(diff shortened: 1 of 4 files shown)


launch/launch/logging/__init__.py
```diff
@@ -407,25 +425,15 @@ def __init__(self, *args, **kwargs):
...
-    # Back to default logging setup
-    launch_config.log_dir = None
-    launch_config.file_handlers = {}
-    launch_config.screen_handler = None
-    launch_config(
-        level=logging.INFO, log_dir=default_log_dir,
-        log_format='default', screen_format='default'
-    )
+    launch_config.reset()
+    launch_config.configure(level=logging.INFO, log_dir=None,
```
hidmic: @clalancette `log_dir=None` has no effect here.

hidmic: Also, I wonder if `launch_config.reset()` shouldn't result in this configuration already.

hidmic: Hmm, and what would happen if someone calls `launch_config.reset()` instead of `launch.logging.reset()`?

clalancette: Good point.  I'm thinking that we should just remove the `configure` method completely, and just pass those parameters to `reset`.  That would alleviate both of your concerns above.  What do you think about that?

clalancette: Hm, I tried it out, and it got ugly.  So scratch that.

Instead, what about something like https://github.com/ros2/launch/blob/launch-logger-config-getters-setters/launch/launch/logging/__init__.py , where we switch to everything being a getter/setter?  Do you think that is better?

hidmic: That looks much better!

clalancette: All right, I'm going to merge that branch to this PR, and we can continue the review with that.

launch/launch/logging/__init__.py
```diff
@@ -30,123 +30,203 @@
...
+        if self._log_handler_factory is None:
+            if os.name != 'nt':
+                self._log_handler_factory = handlers.WatchedFileHandler
+            else:
+                self._log_handler_factory = handlers.FileHandler
+        return self._log_handler_factory
+
+    log_handler_factory = property(get_log_handler_factory, set_log_handler_factory)
+
+    def set_screen_format(self, screen_format, screen_style):
```
clalancette: I'll point out that both for `set_screen_format` and `set_log_format`, I got rid of the keyword-only args in favor of traditional args.  I also got rid of default values (essentially `None`) for the arguments.  I slightly prefer this style for small numbers of arguments like this, but I'm willing to change either (or both) decisions.  Let me know what you think.

launch/launch/logging/__init__.py
```diff
@@ -30,123 +30,203 @@
...
+
+    def set_level(self, new_level):
+        """
+        Set up launch logging verbosity level for all loggers.
+
+        :param new_level: the default log level used for all loggers.
+        """
+        logging.root.setLevel(new_level)
+
+    level = property(None, set_level)
```
hidmic: @clalancette nit: you may use decorators instead.

launch/launch/logging/__init__.py
```diff
@@ -407,25 +441,16 @@ def __init__(self, *args, **kwargs):
...
-    launch_config.screen_handler = None
-    launch_config(
-        level=logging.INFO, log_dir=default_log_dir,
-        log_format='default', screen_format='default'
-    )
+    launch_config.reset()
+    launch_config.level = logging.INFO
+    launch_config.set_screen_format('default', None)
+    launch_config.set_log_format('default', None)
     logging.setLoggerClass(LaunchLogger)
```
hidmic: @clalancette hmm I wonder if `launch_config.reset()` shouldn't already set reasonable defaults.

launch/launch/logging/__init__.py
```diff
@@ -326,52 +406,6 @@ def get_output_loggers(process_name, output_config):
...
-                launch_config.log_handler_factory = handlers.FileHandler
-        file_path = get_log_file_path(file_name)
-        file_handler = launch_config.log_handler_factory(
-            file_path, encoding='utf-8')
-        file_handler.setFormatter(launch_config.file_formatter)
-        launch_config.file_handlers[file_name] = file_handler
-    return launch_config.file_handlers[file_name]
-
-
 def _make_unique_log_dir(*, base_path):
```
hidmic: @clalancette move upwards maybe?

launch/launch/logging/__init__.py
hidmic: >  I got rid of the keyword-only args in favor of traditional args

I _personally_ think that makes sense for the first, necessary argument but not for the rest e.g. `set_screen_format(self, screen_format, *, screen_style='}')` would be better IMO.

launch/launch/logging/__init__.py
```diff
@@ -30,123 +30,203 @@
...
+                screen_style = '{'
+            self.screen_formatter = logging.Formatter(
+                screen_format, style=screen_style
+            )
+            if self.screen_handler is not None:
+                self.screen_handler.setFormatter(self.screen_formatter)
+        else:
+            self.screen_formatter = None
+
+    def get_screen_handler(self):
```
hidmic: @clalancette I wonder if `logging` handlers' getters should module-level functions instead of part of the configuration. These _use_ the configuration but are not part of it IMO.

launch/launch/logging/__init__.py
clalancette: Actually, not for this particular one.  You can only use decorators if you implement a getter, apparently.  I learned that while doing this :).  So if you just want a setter, you have to do it this way.  Alternatively, you could argue that we want a getter for `level`, in which case I could switch to decorators (here and elsewhere).

launch/launch/logging/__init__.py
clalancette: Yeah, that's a good point.  I'll switch to that, I like it better.

launch/launch/logging/__init__.py
clalancette: Yeah, I wasn't sure where to have this live.  I can move it above the `LaunchConfig` class now.

launch/launch/logging/__init__.py
clalancette: The thing that breaks it is this: https://github.com/ros2/launch/pull/299/files#diff-01b1f102bf9f7e90ec6fd88bf1ac3780R178 .  There, we *do* hold onto the screen_handler so we can later set a screen_formatter on it.  At least as it stands, I think it needs to be part of the class.  Unless you have another idea here?

launch/launch/logging/__init__.py
clalancette: I ended up adding a getter too, so now I've switch to properties.

launch/launch/logging/__init__.py
hidmic: Hmm that's true. Nevermind.

clalancette:
New CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7858/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3921/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6412/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7718/)

clalancette:
All of the failures listed here were the same from the nightlies on the same day.  And @hidmic seems to be happy with this, so I'm going to merge.  Thanks!

PR #302, opened 2019-08-13
Use renamed remove_ansi_escape_sequences.

clalancette:
This should only go in after https://github.com/osrf/osrf_pycommon/pull/53


clalancette:
CI is in https://github.com/osrf/osrf_pycommon/pull/53#issuecomment-520954345

PR #306, opened 2019-08-14
Use signal.CTRL_C_EVENT on windows instead of signal.SIGINT

ivanpauno:
We were using `signal.SIGTERM` on windows instead of `signal.SIGINT`.
I think that `signal.CTRL_C_EVENT` is the correct signal to be send in replacement of `signal.SIGINT`, which is not supported in [SubprocessTransport.send_signal](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.send_signal).

Related with https://github.com/ros2/ros2cli/pull/315.
See https://github.com/pypa/setuptools/issues/1818.

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -354,12 +354,8 @@ class ExecuteProcess(Action):
             return None
         if platform.system() == 'Windows' and typed_event.signal_name == 'SIGINT':
             # TODO(wjwwood): remove this when/if SIGINT is fixed on Windows
-            self.__logger.warning(
-                "'SIGINT' sent to process[{}] not supported on Windows, escalating to 'SIGTERM'"
-                .format(self.process_details['name']),
-            )
             typed_event = SignalProcess(
-                signal_number=signal.SIGTERM,
+                signal_number=signal.CTRL_C_EVENT,
                 process_matcher=lambda process: True)
         self.__logger.info("sending signal '{}' to process[{}]".format(
             typed_event.signal_name, self.process_details['name']


ivanpauno:
> I presume you tried this and it works. If so, I wonder why [Python documentation](https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.asyncio.subprocess.Process.send_signal) states we need custom subprocess creation flags to make it work.

I've tried and it works well.

I've also read that part of the documentation, and I forgot about it after seeing that it was working :smile:.
I will double check if `osrf_pycommon` is already adding that flag, or if we can add it.

----

This PR doesn't solve the following problem: a Python program catching `KeyboardInterrupt` and ignoring it (ctrl-c event also raise a `KeyboardInterrupt` in python). After `sigterm_timeout`, `ExecuteProcess` will send a `SIGTERM` signal, that will have the same problem with executables installed with `entry_points` `console_scripts`. That's why I opened an issue in `setuptools` repo.

ivanpauno:
> I presume you tried this and it works. If so, I wonder why [Python documentation](https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.asyncio.subprocess.Process.send_signal) states we need custom subprocess creation flags to make it work.

After further testing, I realized that when I send the [ctrl_c_event](https://docs.microsoft.com/en-us/windows/console/console-process-groups), the parent process was also being finished :smile: (it wasn't a problem in the example I tried).
And the funny thing, is that if I use `creationflags=subprocess.CREATE_NEW_PROCESS_GROUP`, the child process stop receiving signals (something broken on Python?).

I will just close this, and wait for an answer in https://github.com/pypa/setuptools/issues/1818.

clalancette:
For what it is worth, this change does solve the "zombie" Python processes I reported over in https://github.com/ros2/demos/pull/107#issuecomment-510505166 .  I guess it is still not the correct solution, but at least for that use case it is definitely an improvement.

ivanpauno:
> For what it is worth, this change does solve the "zombie" Python processes I reported over in [ros2/demos#107 (comment)](https://github.com/ros2/demos/pull/107#issuecomment-510505166) . I guess it is still not the correct solution, but at least for that use case it is definitely an improvement.

The problem is that it isn't just killing the child process, is killing the launch service itself (it's in the same process group, and ctrl_c_event work in that way on windows).
I also tried running the child process in a new process group, but it stopped receiving signals at all (a bug in python?).

The "zombie" python process is generated because the installed executable wrapper of the python program (by `setuptools`), is not handling `TerminateProcess` correctly (a Windows API equivalent to SIGTERM).

I haven't found a workaround for this combination on unfortunate bugs on Windows (sending SIGINT to asincio subprocesses is broken, creating subprocesses with CREATE_NEW_PROCESS_GROUP creationflag seems to be broken, setuptools installed console_scripts handles `TerminateProcess`  incorrectly).

Note: Here is a branch of `osrf_pycommon` adding support to `creationflags` https://github.com/ivanpauno/osrf_pycommon/tree/ivanpauno/add-creationflags.

ivanpauno:
I haven't tried escalating to `kill`, but it seems to be the same than `term` on Windows:
https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.asyncio.subprocess.Process.kill

PR #307, opened 2019-08-15
Fix test_expose_decorators failures

ivanpauno:
I never realized, but `test_expose_decorators` is failing in some nightly jobs (since I merged launch frontend stuff):

- [nightly_win_rep](https://ci.ros2.org/view/nightly/job/nightly_win_rep/1565/#showFailuresLink)
- [nightly_linux_repeated](https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/1557/testReport/junit/launch.test.launch.frontend/test_expose_decorators/test_expose_decorators_2_11_/)

I guess this will solve the issue, but I'm not sure.
Also, I don't understand the difference between this nightly jobs and manual jobs or [nightly_linux_debug](https://ci.ros2.org/view/nightly/job/nightly_linux_debug/1283/#showFailuresLink) where this test is passing.

### DIFF
--- a/launch/launch/frontend/expose.py
+++ b/launch/launch/frontend/expose.py
@@ -91,7 +91,7 @@ def __expose_impl(name: Text, parse_methods_map: dict, exposed_type: Text):
                 'Exposed {} parser for {} is not a callable or a class'
                 ' containg a parse method'.format(exposed_type, name)
             )
-        if name in parse_methods_map and found_parse_method is not parse_methods_map[name]:
+        if name in parse_methods_map and found_parse_method != parse_methods_map[name]:
             raise RuntimeError(
                 'Two {} parsing methods exposed with same name.'.format(exposed_type)
             )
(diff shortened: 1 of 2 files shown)


ivanpauno:
@nuclearsandwich @dirk-thomas What could be the difference between `nightly_linux_repeated` and `ci_linux`, and between `nightly_win_rep` and `ci_windows`, to make the test fail in one of them and not in the other?

dirk-thomas:
The `ci_*` are by default triggered without a specific CMake build type (which is `None`) which on non-Windows is neither the same as `Debug` nor `Release`. On Windows it does default to `Release`.

For non-`repeated` jobs we try to run tests multiple in case they fail to kind of look over flaky tests. If they pass at least once in N runs we consider them passing. The `repeated` jobs do the opposite if a test passes it re-runs the test N times to check if it might be flaky. If the test fails even once it is considered a `failure`.

nuclearsandwich:
Nothing to add to @dirk-thomas's answer except that you can set the CMake build type as a parameter when launching CI jobs to check whether any potential fixes address the issue.

ivanpauno:
Thanks @dirk-thomas @nuclearsandwich !
Now I realize that the only jobs where this test is failing are of the `repeated` kind, and it isn't failing in the first try, but in the repetitions.

CI, up to launch, using `--retest-until-fail 10`:
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/7918/)

PR #308, opened 2019-08-15
 Support LaunchService injection into pre-shutdown tests.

### LINKED ISSUE #295: [launch testing] Support repeated execution of executables under test
#### Feature description

As it stems from https://github.com/ros2/ros2cli/pull/304, specifically [this discussion](https://github.com/ros2/ros2cli/pull/304#discussion_r307924218), in order to minimize the fragility  of some test cases, repeated execution of a given executable has to be supported by `launch_testing`.

One way this can be achieved is through [injection](https://github.com/ros2/launch/blob/7961b80b5d517228f35ac0b88b6832a2cf0f21f4/launch_testing/launch_testing/test_runner.py#L86) of the [`LaunchService` instance](https://github.com/ros2/launch/blob/7961b80b5d517228f35ac0b88b6832a2cf0f21f4/launch_testing/launch_testing/test_runner.py#L45) into pre shutdown tests.

hidmic:
Precisely what the title says. Closes #295.

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -332,6 +332,10 @@ class ExecuteProcess(Action):
         self,
         context: LaunchContext
     ) -> Optional[LaunchDescription]:
+        typed_event = cast(ShutdownProcess, context.locals.event)
+        if not typed_event.process_matcher(self):
+            # this event whas not intended for this process
+            return None
         return self._shutdown_process(context, send_sigint=True)
 
     def __on_signal_process_event(
--- a/launch_testing/launch_testing/test_runner.py
+++ b/launch_testing/launch_testing/test_runner.py
@@ -87,6 +87,7 @@ class _RunnerWorker():
         self._test_run.bind(
             self._test_run.pre_shutdown_tests,
             injected_attributes={
+                'launch_service': self._launch_service,
                 'proc_info': proc_info,
                 'proc_output': proc_output,
                 'test_args': test_args,
@@ -95,6 +96,7 @@ class _RunnerWorker():
                 full_context,
                 # Add a few more things to the args dictionary:
                 **{
+                    'launch_service': self._launch_service,
                     'proc_info': proc_info,
                     'proc_output': proc_output,
                     'test_args': test_args
--- a/launch_testing/example_processes/terminating_proc
+++ b/launch_testing/example_processes/terminating_proc
@@ -41,4 +41,4 @@ if __name__ == "__main__":
 
     print("Shutting Down")
 
-    sys.exit(0) 
+    sys.exit(0)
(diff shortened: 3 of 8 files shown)


hidmic:
> Is the idea just to loop using launch_process or to add some extra functionality to make repeated tests looks nicer?

I was thinking about simply looping, yes. We could also instrument a decorator for our tests that repeats it until it passes or timeout is reached, akin to `pytest-repeat`.

hidmic:
CI up to `launch_testing` and `test_communication`:

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/7924/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3958/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6451/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7775/)

hidmic:
Alright, all green. Going in! @dirk-thomas feel free to comment async.

launch_testing/launch_testing/test_runner.py
```diff
@@ -87,6 +87,7 @@ def run(self):
         self._test_run.bind(
             self._test_run.pre_shutdown_tests,
             injected_attributes={
+                'launch_service': self._launch_service,
```
pbaughman: @hidmic Not directly related to this PR, but do you have any guidance for a process for deprecating 'injected_attributes' from launch_testing?  Something like 'put a warning in the next release when someone uses an injected attribute'  then in the release after that remove injected attributes?

I think injected_args are better and less confusing in just about every way.  They still require the user to know ahead of time what 'magic words' can be added as arguments to a test function, but beyond that they're way less mysterious

hidmic: Hmm, you could wrap them using a `property` that can generate a `warning` once if these are used.

pbaughman: @hidmic Yeah - I know how to do the implementation side - my question is more about the administrative side.  Do we need a full release of "WARNING THIS WILL BE DEPRECATED" or is it sufficient to add the warning, them actually deprecate once all of OSRF's stuff is updated?

launch/launch/actions/execute_process.py
```diff
@@ -332,6 +332,10 @@ def __on_shutdown_process_event(
         self,
         context: LaunchContext
     ) -> Optional[LaunchDescription]:
+        typed_event = cast(ShutdownProcess, context.locals.event)
+        if not typed_event.process_matcher(self):
+            # this event whas not intended for this process
```
dirk-thomas: Spelling.

dirk-thomas: Also that `this` is used twice but references different entities is a bit hard to read.

PR #312, opened 2019-08-16
 Enable launch test discovery in pytest

### LINKED ISSUE #237: Hook up launch tests in pure Python packages
#### Feature description

As a follow-up of https://github.com/ros2/launch_ros/pull/15#issuecomment-488157564, we currently lack a way to register launch tests to be run in pure Python packages. We need a way to do so.

As of https://github.com/ros2/launch/pull/236, `launch_test` can now be run from Python code. We'd just be lacking a `pytest` plugin to deal with launch test files properly.

hidmic:
Closes #237. This pull request allows for launch tests discovery and execution via `pytest`. It also sets up the scene for downstream packages i.e. `launch_testing_ros` to customize this.

Note this is not a tight integration between `launch_testing` and `pytest` but simply a mechanism for the latter to be able to execute launch tests.

Depends on #308.


hidmic:
CI up to `launch_testing`, `launch_testing_ros` and `test_communication` for extra safety:

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/7943/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3974/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6467/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7790/)

launch_testing/test/launch_testing/examples/README.md
```diff
@@ -1,12 +1,12 @@
 # Examples
 
-## `good_proc.test.py`
```
ivanpauno: I understand why moving the examples, but I think that the README should stay at the top level directory of the package (in `launch_testing`). It'll be more visible there.

launch_testing/test/launch_testing/test_examples.py
```diff
@@ -1,45 +0,0 @@
-# Copyright 2019 Apex.AI, Inc.
```
ivanpauno: Why was this file deleted?

ivanpauno:
Extra comment, instead of `rostest` files [needing a pytest mark](https://github.com/ros2/launch_ros/blob/68f7d32e66a9808a876fc838a8a61c7ef6d07575/launch_testing_ros/test/examples/talker_listener_launch_test.py#L29-L30) and `launchtest` files just needing a `generate_test_description` function in the file, why not requiring a mark in them too? Something like, `@pytest.mark.launchtest`.
I prefer the difference being more explicit (than one explicit and the other implicit).

Also, aren't files with `@pytest.mark.rostest` mark being collected by both hooks now?

launch_testing/launch_testing/pytest/hooks.py
```diff
@@ -0,0 +1,124 @@
...
+            test_runs=LoadTestsFromPythonModule(
+                module, name=module.__name__
+            )
+        )
+
+
+def find_launch_test_entrypoint(path):
+    try:
+        return getattr(path.pyimport(), 'generate_test_description', None)
+    except SyntaxError:
```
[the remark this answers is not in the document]
hidmic: A `SyntaxError` is the only error I'm expecting. I'd rather not catch (and hide) unexpected errors e.g. some exception as a result of code being executed at the module level.

launch_testing/test/launch_testing/examples/README.md
hidmic: Sure, we can move it there. See 8014b4c94a5cded069bca06d4f031398f6c76a4b.

launch_testing/test/launch_testing/test_examples.py
hidmic: Because it was `pytest` boilerplate running a `launch_test` over the examples. Now that `pytest` can do that on its own, there's no need to preserve it.

hidmic:
> Extra comment, instead of rostest files needing a pytest mark and launchtest files just needing a generate_test_description function in the file, why not requiring a mark in them too? Something like, @pytest.mark.launchtest.
I prefer the difference being more explicit (than one explicit and the other implicit).

Right now, any module in the search path that has a `generate_test_description()` is taken as a launch test and thus any mark was redundant in the pure launch test case. But I see the value in being explicit in both cases. Check 44917596b72ea7d105147e6d3f57af8f10d01446.

> Also, aren't files with `@pytest.mark.rostest` mark being collected by both hooks now?

Nope, [this hook](https://github.com/ros2/launch/pull/312/files#diff-4b19f7cb02f9cd8fcafc7b387b3cb854R102) collects them  appropriately once.

hidmic:
I had to fix other failing tests, due to examples not being installed and `pytest` compulsively re-writing `asserts` within `launch_testing`. @ivanpauno PTAL.

launch_testing/launch_testing/parametrize.py
```diff
@@ -35,22 +35,16 @@ def parametrize(argnames, argvalues):
...
                 partial = functools.partial(func, **partial_args)
                 functools.update_wrapper(partial, func)
-                self.__calls.append(
-                    (partial, partial_args)
-                )
-
-        def __iter__(self):
-            return iter(self.__calls)
+                yield partial, partial_args
+        setattr(_wrapped, '__parametrized__', True)
```
ivanpauno: Couldn't be:
```python
_wrapped.__parametrized__ = True
```
?

launch_testing/launch_testing/asserts/assert_exit_codes.py
```diff
@@ -12,6 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""
+A module providing exit code assertions.
+
+PYTEST_DONT_REWRITE
```
ivanpauno: Maybe, add a comment/link explaining why this is needed (e.g.: https://docs.pytest.org/en/latest/assert.html#assertion-introspection-details).

launch_testing/launch_testing/pytest/hooks.py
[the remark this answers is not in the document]
ivanpauno: This is old code, but just as reference:
https://github.com/ros2/launch_ros/blob/bf3d800a53ea53d9465da9298c104998a7461dcf/ros2launch/ros2launch/command/launch.py#L125-L141

Couldn't value error happen here too?

launch_testing/launch_testing/asserts/assert_exit_codes.py
hidmic: Sure, done in c7e6f8043a9643ac2f3a6f2206c7186011dff32e.

launch_testing/launch_testing/parametrize.py
hidmic: Yes, it can be. Done in 0705777c90b01c4b253ba5f03c464ef09ff220bb.

hidmic:
Alright, all green! Going in.

launch_testing/setup.py
```diff
@@ -11,10 +11,11 @@
     data_files=[
         ('share/ament_index/resource_index/packages', ['resource/launch_testing']),
         ('lib/launch_testing', glob.glob('example_processes/**')),
-        ('share/launch_testing/examples', glob.glob('examples/[!_]**')),
+        ('share/launch_testing/examples', glob.glob('test/launch_testing/examples/[!_]**')),
     ],
     entry_points={
-        'console_scripts': ['launch_test=launch_testing.launch_test:main']
+        'console_scripts': ['launch_test=launch_testing.launch_test:main'],
+        'pytest11': ['launch = launch_testing.pytest.hooks'],
```
dirk-thomas: This registers the marker `launch` but the tests use the marker `launch_test` everywhere?

hidmic: Launch tests that are run using `pytest` do, or otherwise it wouldn't work. We have many launch tests in other repositories that still run outside `pytest`. I haven't changed those.

PR #304, opened 2019-08-14
Add support for conditions in IncludeLaunchDescription actions

hidmic:
Fixes https://github.com/ros2/launch/issues/303. I couldn't find a rationale anywhere as to why an `Action` `execute(...)` should return a `List[Action]` as opposed to the more general `List[LaunchDescriptionEntity]`, and we have clear use case at hand here where that is not true i.e. for `IncludeLaunchDescription`.

PR #317, opened 2019-08-22
[backport] Add support for conditions in IncludeLaunchDescription actions (#304)

chapulina:
* Make Action.execute() return LaunchDescriptionEntity instances.

Signed-off-by: Michel Hidalgo <michel@ekumenlabs.com>

* Make sure IncludeLaunchDescription propagates kwargs.

Signed-off-by: Michel Hidalgo <michel@ekumenlabs.com>

### DIFF
--- a/launch/launch/action.py
+++ b/launch/launch/action.py
@@ -14,7 +14,6 @@
 
 """Module for Action class."""
 
-from typing import cast
 from typing import List
 from typing import Optional
 from typing import Text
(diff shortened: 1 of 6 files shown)


hidmic:
@wjwwood mind to take a look too? You were out when I made this conceptual change, that to me is sound but I'd like to cross check with you.

chapulina:
I also don't know enough Python to tell if this could bring backwards compatibility issues. Are we ok on that front, @hidmic ?

hidmic:
It's not bringing anything that wasn't there already in `dashing` and I don't see how it could break downstream packages either.

PR #322, opened 2019-08-28
Optionally remove ready fn arg from generate_test_description

### LINKED ISSUE #305: Change how launch_testing waits for the launched system to be ready
## Feature request
Change the ready_fn in generate_test_description into a launch action

#### Feature description

The original intent of the ready_fn was to avoid an issue frequently encountered with ros1's rostest where each test would have to contain logic to wait for the launched system to be ready to be interacted with.  The purpose of the ready_fn is to force the test author to clearly indicate when it's safe for the tests to start to run.

What I'd like to do is change the ready_fn from a function passed into generate_test_description to something that's included in the launch description itself.  This would allow launch_testing to introspect the launch description and make sure that the action was included, as well as give a better error message if the element is omitted.  

Starting with the [good_proc_test example](https://github.com/ros2/launch/blob/master/launch_testing/examples/good_proc.test.py) I'm proposing that after this change, it would be possible to do this instead:

```
def generate_test_description():
        dut_process,

        # Start tests right away - no need to wait for anything
        launch_testing.actions.ReadyToTest(),  # <--- I'm not married to this name
    ])
```

This should simplify the launch_testing API a bit, and maybe make it easier to write a pytest plugin in the future instead of needing a special runner.

I can keep support for the old way pretty easily, so we won't break anything.

I would have the test_runner call generate_test_description to get the launch description and the test context, just like it does now.  Then introspect the launch description for a ReadyToTest action and plumb that to the python event it's currently using to delay the start of the tests

pbaughman:
Resolves #305

This change lets test launch descriptions include a launch_testing.actions.ReadyToTest action to signal when it's safe for tests to start.

This is intended to be used instead of passing a ready_fn into the generate_test_description function which then has to be plumbed into the launch description with an launch.actions.OpaqueFunction

Longer term, I would like to deprecate the use of the ready_fn in favor of the ReadyToTest action.  The former is annoying to plumb especially in tests where the test launch description is generated by many nested functions.  I feel the latter is easier to understand and will also be easier to improve the behavior of in the future - for example we could support multiple ReadyToTest actions instead of using the somewhat silly [ready aggregator](https://github.com/ros2/launch/blob/master/launch_testing/launch_testing/ready_aggregator.py)

### DIFF
--- a/launch_testing/README.md
+++ b/launch_testing/README.md
@@ -23,6 +23,27 @@ launch_test launch_testing/examples/good_proc.test.py
 
 #### The Launch Description
 
+```python
+def generate_test_description():
+
+    return launch.LaunchDescription([
+        launch.actions.ExecuteProcess(
+            cmd=[path_to_process],
+        ),
+
+        # Start tests right away - no need to wait for anything in this example.
+        # In a more complicated launch description, we might want this action happen
+        # once some process starts or once some other event happens
+        launch_testing.actions.ReadyToTest()
+    ])
+```
+The `generate_test_description` function should return a `launch.LaunchDescription` object that launches the system to be tested.
+
+The launch description needs to include a ReadyToTest action to signal to the test framework that it's safe to start the active tests.
+
+In the above example, there is no need to delay the start of the tests so the ReadyToTest action is a peer to the process under test and will signal to the framework that it's safe to start around the same time the ExecuteProcess action is run.
+
+In older style tests, a function called `ready_fn` is declared as an argument to `generate_test_description` and must be plumbed into the launch description with an `OpaqueFunction`.
 
 ```python
 def generate_test_description(ready_fn):
@@ -36,9 +57,6 @@ def generate_test_description(ready_fn):
         launch.actions.OpaqueFunction(function=lambda context: ready_fn()),
     ])
 ```
-
-The `generate_test_description` function should return a `launch.LaunchDescription` object that launches the system to be tested.
-It should also call the `ready_fn` that is passed in to signal when the tests should start.  In the `good_proc.test.py` example, there is no need to delay the start of the tests so the `ready_fn` is called concurrently when the launching of the process under test.
 
 #### Active Tests
 
--- a/launch_testing/README.md
+++ b/launch_testing/README.md
@@ -42,7 +42,7 @@ The `generate_test_description` function should return a `launch.LaunchDescripti
 
 The launch description needs to include a `ReadyToTest` action to signal to the test framework that it's safe to start the active tests.
 
-In the above example, there is no need to delay the start of the tests so the ReadyToTest action is a peer to the process under test and will signal to the framework that it's safe to start around the same time the ExecuteProcess action is run.
+In the above example, there is no need to delay the start of the tests so the `ReadyToTest` action is a peer to the process under test and will signal to the framework that it's safe to start around the same time the `ExecuteProcess` action is run.
 
 In older style tests, a function called `ready_fn` is declared as an argument to `generate_test_description` and must be plumbed into the launch description with an `OpaqueFunction`.
 
--- a/launch_testing/README.md
+++ b/launch_testing/README.md
@@ -40,7 +40,7 @@ def generate_test_description():
 
 The `generate_test_description` function should return a `launch.LaunchDescription` object that launches the system to be tested.
 
-The launch description needs to include a ReadyToTest action to signal to the test framework that it's safe to start the active tests.
+The launch description needs to include a `ReadyToTest` action to signal to the test framework that it's safe to start the active tests.
 
 In the above example, there is no need to delay the start of the tests so the ReadyToTest action is a peer to the process under test and will signal to the framework that it's safe to start around the same time the ExecuteProcess action is run.
 
--- a/launch_testing/README.md
+++ b/launch_testing/README.md
@@ -37,6 +37,7 @@ def generate_test_description():
         launch_testing.actions.ReadyToTest()
     ])
 ```
+
 The `generate_test_description` function should return a `launch.LaunchDescription` object that launches the system to be tested.
 
 The launch description needs to include a ReadyToTest action to signal to the test framework that it's safe to start the active tests.
--- a/launch_testing/launch_testing/loader.py
+++ b/launch_testing/launch_testing/loader.py
@@ -17,17 +17,36 @@ import inspect
 import itertools
 import unittest
 
+from .actions import ReadyToTest
+
 
 def _normalize_ld(launch_description_fn):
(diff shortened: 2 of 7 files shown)


pbaughman:
@hidmic @wjwwood FYI^^  I think this is cleaner and more consistent with the rest of 'launch' and also solves a problem we have internally where we have a bunch of nested functions that generate our launch descriptions for test and it's a pain in the butt to figure out who needs to pass the 'ready-fn' and who needs to add the OpaqueFunction wrapper.

No rush on this - open to feedback.

launch_testing/README.md
```diff
@@ -23,6 +23,27 @@ launch_test launch_testing/examples/good_proc.test.py
...
+            cmd=[path_to_process],
+        ),
+
+        # Start tests right away - no need to wait for anything in this example.
+        # In a more complicated launch description, we might want this action happen
+        # once some process starts or once some other event happens
+        launch_testing.actions.ReadyToTest()
+    ])
+```
+The `generate_test_description` function should return a `launch.LaunchDescription` object that launches the system to be tested.
```
wjwwood:
```suggestion

The `generate_test_description` function should return a `launch.LaunchDescription` object that launches the system to be tested.
```

launch_testing/README.md
```diff
@@ -23,6 +23,27 @@ launch_test launch_testing/examples/good_proc.test.py
...
+
+The launch description needs to include a ReadyToTest action to signal to the test framework that it's safe to start the active tests.
```
wjwwood:
```suggestion
The launch description needs to include a `ReadyToTest` action to signal to the test framework that it's safe to start the active tests.
```

launch_testing/README.md
```diff
@@ -23,6 +23,27 @@ launch_test launch_testing/examples/good_proc.test.py
...
+
+In the above example, there is no need to delay the start of the tests so the ReadyToTest action is a peer to the process under test and will signal to the framework that it's safe to start around the same time the ExecuteProcess action is run.
```
wjwwood:
```suggestion
In the above example, there is no need to delay the start of the tests so the `ReadyToTest` action is a peer to the process under test and will signal to the framework that it's safe to start around the same time the `ExecuteProcess` action is run.
```

pbaughman:
@wjwwood Unless someone chimes in with an objection, I'll do a follow-up PR that puts a deprecation warning on the ready_fn, and on the implicit `self.proc_info` and `self.proc_output` fields that get injected into the TestCase classes.

wjwwood:
@hidmic @ivanpauno any comments here before I run CI and merge?

launch_testing/launch_testing/loader.py
```diff
@@ -17,18 +17,57 @@
...
+                """Recursively search LaunchDescription entities for all ReadyToTest actions."""
+                for entity in entities:
+                    if isinstance(entity, ReadyToTest):
+                        yield entity
+                    yield from iterate_ready_to_test_actions(
+                        entity.describe_sub_entities()
+                    )
+                    for conditional_sub_entity in entity.describe_conditional_sub_entities():
+                        yield from iterate_ready_to_test_actions(
+                            conditional_sub_entity[1]
```
ivanpauno: shouldn't be:
```python3
for conditional_sub_entity in entity.describe_conditional_sub_entities():
    yield from iterate_ready_to_test_actions(
        [conditional_sub_entity[1]]
    )
```

or better:
```python3
yield from iterate_ready_to_test_actions(
    conditional_sub_entity[1] for conditional_sub_entity in entity.describe_conditional_sub_entities()
)
```

pbaughman: @ivanpauno 
I don't think the first one is correct.  [describe_conditional_sub_entities](https://github.com/ros2/launch/blob/master/launch/launch/action.py#L95) returns a a tuple of `(text description, iterable(entity))` so conditional_sub_entity[1] should already be iterable

I've got a test case that hits this path [here](https://github.com/ros2/launch/pull/322/files#diff-6b08971ae572019599efe076a73b7b94R119-R135) so I think it's working correctly, but I may be misunderstanding your concern.

I also added a [new test case](https://github.com/ros2/launch/pull/322/files#diff-6b08971ae572019599efe076a73b7b94R137-R163) that has multiple conditional actions, and requires recusing farther to find the ReadyAction

ivanpauno: yes, you're right, I read the return type incorrectly.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8160/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4142/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6645/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8039/)

wjwwood:
Looks like Linux has this error:

```
ros2: error: argument Call `ros2 <command> -h` for more detailed usage.: invalid choice: 'test' (choose from 'action', 'daemon', 'extension_points', 'extensions', 'msg')
```

And Windows has lots and lots of test failures.

pbaughman:
@wjwwood Thanks - I'll investigate

pbaughman:
@wjwwood I think maybe this just needs to be rebased.  Most of the failures are failure to look up the 'launch' package with ament_package_index.

Looking at the git history, I see
```
Author:     Dirk Thomas <dirk-thomas@users.noreply.github.com>       
AuthorDate: Fri Aug 30 09:42:05 2019 -0700                           
Commit:     GitHub <noreply@github.com>                              
CommitDate: Fri Aug 30 09:42:05 2019 -0700                           

    install package marker and manifest (#323)
    
    Signed-off-by: Dirk Thomas <dirk-thomas@users.noreply.github.com>
```

. . . was created 3 commits after I branched off.

I rebased this onto origin/master - can you try the CI again please?

wjwwood:
Sure! CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8161/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4143/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6646/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8040/)

pbaughman:
Those latest results look a lot better.  I checked the Linux and Windows results - it all looks like racy tests that fail for reasons unrelated to launch_testing

pbaughman:
@wjwwood Did I correctly interpret the CI results, or are these failures unique to my branch?

wjwwood:
Yeah, it appears to be on the nightly's from around that time: https://ci.ros2.org/view/nightly/job/nightly_linux_release/1303/testReport/

PR #324, opened 2019-08-30
[Dashing] Restore support for Python 3.5

nuclearsandwich:
This pull request fixes python 3.5 compatibility by removing unsupported type hints.

Some of the type hints introduced in Dashing are not compatible with
Python 3.5. Where possible I tried to convert the hints to type comments
but for the type hints within function argument lists that did not
appear to be possible.

This is going to make backporting future changes more challenging since
we will have to:
1. Review the backported code for incompatible type hints
2. Reconcile merge conflicts caused by the removal of type hints on
the Dashing development branches.

To address flake8 failures this PR also includes 06b0b01 which is a backport of #300.

There were two cases where type comments were usable but exceeded the 99 char line limit. Without knowledge of how to preserve both the type comment and the line length requirement I removed the type comments in 3914f85.

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -190,7 +190,7 @@ class ExecuteProcess(Action):
         self.__cmd = [normalize_to_list_of_substitutions(x) for x in cmd]
         self.__name = name if name is None else normalize_to_list_of_substitutions(name)
         self.__cwd = cwd if cwd is None else normalize_to_list_of_substitutions(cwd)
-        self.__env = None   # type: Optional[List[Tuple[List[Substitution], List[Substitution]]]]
+        self.__env = None  # type: Optional[List[Tuple[List[Substitution], List[Substitution]]]]
         if env is not None:
             self.__env = []
             for key, value in env.items():
@@ -216,12 +216,12 @@ class ExecuteProcess(Action):
         self.__log_cmd = log_cmd
         self.__on_exit = on_exit
 
-        self.__process_event_args = None   # type: Optional[Dict[Text, Any]]
-        self._subprocess_protocol = None   # type: Optional[Any]
+        self.__process_event_args = None  # type: Optional[Dict[Text, Any]]
+        self._subprocess_protocol = None  # type: Optional[Any]
         self._subprocess_transport = None
-        self.__completed_future = None   # type: Optional[asyncio.Future]
-        self.__sigterm_timer = None   # type: Optional[TimerAction]
-        self.__sigkill_timer = None   # type: Optional[TimerAction]
+        self.__completed_future = None  # type: Optional[asyncio.Future]
+        self.__sigterm_timer = None  # type: Optional[TimerAction]
+        self.__sigkill_timer = None  # type: Optional[TimerAction]
         self.__shutdown_received = False
         self.__stdout_buffer = io.StringIO()
         self.__stderr_buffer = io.StringIO()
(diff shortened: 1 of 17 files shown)


launch/launch/actions/execute_process.py
```diff
@@ -216,12 +216,12 @@ def __init__(
         self.__log_cmd = log_cmd
         self.__on_exit = on_exit
 
-        self.__process_event_args = None  # type: Optional[Dict[Text, Any]]
-        self._subprocess_protocol = None  # type: Optional[Any]
+        self.__process_event_args = None   # type: Optional[Dict[Text, Any]]
```
jacobperron: Is this space-related change necessary or can it be reverted (same for other instances)?

launch/launch/actions/execute_process.py
```diff
@@ -190,14 +190,14 @@ def __init__(
...
-        self.__env = None  # type: Optional[List[Tuple[List[Substitution], List[Substitution]]]]
+        self.__env = None   # type: Optional[List[Tuple[List[Substitution], List[Substitution]]]]
         if env is not None:
             self.__env = []
             for key, value in env.items():
                 self.__env.append((
                     normalize_to_list_of_substitutions(key),
                     normalize_to_list_of_substitutions(value)))
-        self.__additional_env: Optional[List[Tuple[List[Substitution], List[Substitution]]]] = None
+        self.__additional_env = None
```
jacobperron: You can explicitly ignore the line length constraint for the line by appending `# noqa: E501`:

```suggestion
        self.__additional_env = None  # type: Optional[List[Tuple[List[Substitution], List[Substitution]]]]  # noqa: E501
```

launch/launch/actions/execute_process.py
nuclearsandwich: Ah no these were the result of an overzealous sed. I'll fix that up.

launch/launch/actions/execute_process.py
nuclearsandwich: Is having a noqa'd line preferable to ditching the type comment? I wasn't sure.

jacobperron: Since we're not running mypy tests it's probably okay to ditch the comment. I don't have a preference.

launch/launch/logging/__init__.py
```diff
@@ -56,7 +56,7 @@ def launch_config(
     screen_style=None,
     log_format=None,
     log_style=None,
-    log_handler_factory=None,
+    log_handler_factory=None
```
nuclearsandwich:
```suggestion
    log_handler_factory=None,
```

nuclearsandwich:
I'll run Dashing CI on this once I get its rclpy sibling going.

nuclearsandwich:
It appears that some type hints in function parameters may be supported in spite of the errors I thought they were causing. I'm going to try and reduce the diff here before seeking further review.

launch/launch/actions/execute_process.py
nuclearsandwich: I'll prefer to leave it as is then.

nuclearsandwich:
> It appears that some type hints in function parameters may be supported in spite of the errors I thought they were causing. I'm going to try and reduce the diff here before seeking further review.

There was nothing unnecessarily removed here. So this is good for :eyes:.

nuclearsandwich:
CI in is :green_book: https://github.com/ros2/launch_ros/pull/65#issuecomment-527895849

PR #326, opened 2019-09-03
Unindent setup.cfg options

hidmic:
Precisely what the title says. This was causing Python 2.7-based `bloom-release` to fail on `launch_testing`.

### DIFF
--- a/launch_testing/setup.cfg
+++ b/launch_testing/setup.cfg
@@ -1,4 +1,4 @@
 [coverage:run]
-    # This will let coverage find files with 0% coverage (not hit by tests at all)
-    source = .
-    omit = setup.py
+# This will let coverage find files with 0% coverage (not hit by tests at all)
+source = .
+omit = setup.py


dirk-thomas:
The same fix might be necessary here: https://github.com/ros2/launch_ros/blob/8f99605a38cf2757ccbc3b2fd54e521569ceba2b/launch_testing_ros/setup.cfg

pbaughman:
> The same fix might be necessary here: https://github.com/ros2/launch_ros/blob/8f99605a38cf2757ccbc3b2fd54e521569ceba2b/launch_testing_ros/setup.cfg

Whoops - that also has a dangling reference to 'apex_launchtest' that should probably get changed to launch_testing.   I suspect that line isn't doing anything, though, otherwise we would've seen some linter failures.  We might just want to delete the part that says:

```
[flake8]
    # For the purpose of import-order style checker, consider apex_launchtest as a local package
    import-order-style = "apex_launchtest"
```

pbaughman:
I commented too soon.  I see that's exactly what was done here: https://github.com/ros2/launch_ros/pull/66/files  Sorry for the noise.

Issue #321, opened 2019-08-28
[regression] Improve error message when failing to launch


jacobperron:
## Bug report

https://github.com/ros2/launch/pull/315 has regressed.

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - source
- Version or commit hash:
  - latest

#### Steps to reproduce issue

```
from launch import LaunchDescription
from launch_ros.actions import Node

raise ValueError('oops')

def generate_launch_description():
    return LaunchDescription([
            Node(package='foo', node_executable='foo')
        ])
```

#### Expected behavior

Reports the error with a reasonable description.

#### Actual behavior

Outputs:

```
> ros2 launch foo.launch.py
[INFO] [launch]: All log files can be found below /home/jacob/.ros/log/2019-08-27-17-26-48-584167-warner-23796
[INFO] [launch]: Default logging verbosity is set to INFO
[ERROR] [launch]: Caught exception in launch (see debug for traceback): Caught exception when trying to load file of format [{}]:
```

#### Additional information

I'm not sure what caused the regression, but I think it's valuable to add a regression test this time ;)

jacobperron:
Resolved by #333

PR #315, opened 2019-08-20
Improve error message when a failing to include launch file

ivanpauno:
Fixes https://github.com/ros2/launch/issues/314.

- The error message was better only when a `.xml` or `.yaml` file failed to be included, because an error in the logic of `any_launch_file_utilities.py`.
- I improved the message further, by using `__cause__` attribute. See https://docs.python.org/3/library/exceptions.html, https://stackoverflow.com/questions/24752395/python-raise-from-usage.

PR #333, opened 2019-09-19
Fix InvalidLaunchFileError error message.

hidmic:
Precisely what the title says.

### DIFF
--- a/launch/launch/invalid_launch_file_error.py
+++ b/launch/launch/invalid_launch_file_error.py
@@ -26,11 +26,9 @@ class InvalidLaunchFileError(Exception):
             self._error_message = 'The launch file may have a syntax error'
             ', or its format is unknown'
         else:
-            self._error_message = 'Caught exception when trying to load file of format [{}]:'
-            ' {}'.format(
-                self._extension,
-                self._likely_errors[0]
-            )
+            self._error_message = (
+                'Caught exception when trying to load file of format [{}]: {}'
+            ).format(self._extension, self._likely_errors[0])
             self.__cause__ = self._likely_errors[0]
 
     def __str__(self):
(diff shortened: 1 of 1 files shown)


launch/launch/invalid_launch_file_error.py
```diff
@@ -26,11 +26,9 @@ def __init__(self, extension='', *, likely_errors=None):
...
             ', or its format is unknown'
         else:
-            self._error_message = 'Caught exception when trying to load file of format [{}]:'
-            ' {}'.format(
-                self._extension,
-                self._likely_errors[0]
-            )
+            self._error_message = (
+                'Caught exception when trying to load file of format [{}]: {}'
+            ).format(self._extension, self._likely_errors[0])
```
ivanpauno: Oh, I was just formatting the second line, right? Didn't know that.

jacobperron:
I think this resolves #321 ?

ivanpauno:
> I think this resolves #321 ?

Was the only problem in #321 the bad message formatting? The issue is not clear to me.
If that's the case, this fixes #321.

jacobperron:
> Was the only problem in #321 the bad message formatting? The issue is not clear to me.
If that's the case, this fixes #321.

Yes. Error messages were improved in #315, but then "regressed" when I started seeing uninformative error messages:

```
[ERROR] [launch]: Caught exception in launch (see debug for traceback): Caught exception when trying to load file of format [{}]:
```

I've confirmed that this PR resolves my issue.

launch/launch/invalid_launch_file_error.py
```diff
@@ -26,11 +26,9 @@ def __init__(self, extension='', *, likely_errors=None):
             self._error_message = 'The launch file may have a syntax error'
             ', or its format is unknown'
```
ivanpauno: Can you correct this too?
```python3
self._error_message = \
    'The launch file may have a syntax error, or its format is unknown'
```

hidmic:
Running CI for `launch`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8195/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4171/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6673/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8069/)

PR #310, opened 2019-08-15
Fix launch argument listing/checking issues

ivanpauno:
- Fixes `describe_sub_entities` `describe_conditional_sub_entities`.
- Declared launch arguments in `GroupAction` actions were ignored. Fixed that.

Both items affect the logic of checking/getting launch arguments:
https://github.com/ros2/launch/blob/89de94182947ed8fa61fbe7554866e871d8a5121/launch/launch/launch_description.py#L101-L128
https://github.com/ros2/launch/blob/89de94182947ed8fa61fbe7554866e871d8a5121/launch/launch/actions/include_launch_description.py#L139-L152

PR #334, opened 2019-09-19
Fix default Action describe_conditional_sub_entities() implementation.

hidmic:
Precisely what the title says. Current implementation does not seem to match the documentation in `LaunchDescriptionEntity` nor the mock used to test `get_launch_arguments`. It also fails on conditional `GroupAction`s.

### DIFF
--- a/launch/launch/action.py
+++ b/launch/launch/action.py
@@ -97,13 +97,7 @@ class Action(LaunchDescriptionEntity):
         Iterable[LaunchDescriptionEntity],  # list of conditional sub-entities
     ]]:
         """Override describe_conditional_sub_entities from LaunchDescriptionEntity."""
-        return [
-            (
-                'Conditionally included by {}'.format(self.describe()),
-                entity,
-            )
-            for entity in self.get_sub_entities()
-        ] if self.condition is not None else []
+        return [('Conditionally included by {}'.format(self.describe()), self.get_sub_entities())]
 
     def visit(self, context: LaunchContext) -> Optional[List[LaunchDescriptionEntity]]:
         """Override visit from LaunchDescriptionEntity so that it executes."""
--- a/launch/launch/action.py
+++ b/launch/launch/action.py
@@ -97,7 +97,7 @@ class Action(LaunchDescriptionEntity):
         Iterable[LaunchDescriptionEntity],  # list of conditional sub-entities
     ]]:
         """Override describe_conditional_sub_entities from LaunchDescriptionEntity."""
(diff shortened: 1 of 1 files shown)


ivanpauno:
After reading [this](https://github.com/ros2/launch/blob/c65a315856fdcb20701db004fccd85d2a0dd53d6/launch/launch/actions/timer_action.py#L95-L100), I think we should revert the [changes I did to action API](https://github.com/ros2/launch/pull/310), and instead of that, check if the entity has a condition directly here:
https://github.com/ros2/launch/blob/c65a315856fdcb20701db004fccd85d2a0dd53d6/launch/launch/launch_description.py#L119-L120
doing
```python3
_conditional_inclusion = entity.condition is not None
```

hidmic:
@ivanpauno launch argument discovery is orthogonal to this patch, though I think the code as-is is fine and in its most general form. 

In any case, you made me realize this patch was missing something. PTAL at https://github.com/ros2/launch/pull/334/commits/eba43223851b3137b34649b17e555ee942686a21.

hidmic:
Running CI for `launch`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8196/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4172/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6674/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8070/)

ivanpauno:
> @ivanpauno launch argument discovery is orthogonal to this patch, though I think the code as-is is fine and in its most general form.

I think that we could handle conditional inclusion of an action independently of if sub-actions are conditional or not, though the current implementation (with this fix) is ok.

launch/launch/action.py
```diff
@@ -98,11 +98,7 @@ def describe_conditional_sub_entities(self) -> List[Tuple[
     ]]:
         """Override describe_conditional_sub_entities from LaunchDescriptionEntity."""
         return [
-            (
-                'Conditionally included by {}'.format(self.describe()),
-                entity,
-            )
-            for entity in self.get_sub_entities()
+            ('Conditionally included by {}'.format(self.describe()), self.get_sub_entities())
```
wjwwood: I think the reason that this was structured as a `[(str, entity), ...]` before is that each entity may have a different reason for being included. As you have it here, matches what I had documented before, and should also work, `[(str, [entity]), ...]`, also allowing for groups to have the same condition reason. But I just wanted to reiterate clearly what the change in structure here was about.

PR #336, opened 2019-09-19
Add support to ignore fields when parsing ExecuteProcess.

hidmic:
Precisely what the title says. The previous mechanism to ignore `cmd` (so that `ExecuteProcess.parse` can be reused by derived classes) was not properly implemented. This is a fix for it and an improvement towards `parse` functions reuse, but it doesn't scale that well either. But IMO that's a separate issue.

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -241,61 +241,87 @@ class ExecuteProcess(Action):
         self.__stderr_buffer = io.StringIO()
 
     @classmethod
-    def parse(
+    def _parse_cmdline(
         cls,
-        entity: Entity,
-        parser: Parser,
-        cmd_arg_name: str = 'cmd'
+        cmd: Text,
+        parser: Parser
     ):
         """
-        Return the `ExecuteProcess` action and kwargs for constructing it.
+        Parse text apt for command line execution.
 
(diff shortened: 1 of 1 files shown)


hidmic:
Running CI for `launch`, `launch_ros` and `test_launch_ros`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8199/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4174/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6677/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8073/)

jacobperron:
This change appears to have broken some of my launch files using `<executable>` tags. For example, attempting to launch the following results in an error:

    <launch>
      <executable cmd="echo foo" output="screen" />
    </launch>

Error:

    [ERROR] [launch]: Caught exception in launch (see debug for traceback): Caught exception when trying to load file of format [xml]: _parse_cmdline() missing 1 required positional argument: 'parser'

ivanpauno:
> Running CI for launch, launch_ros and test_launch_ros:

I ignore this part of your comment, but you should also run tests of: launch_xml, launch_yaml

hidmic:
Yes, my bad. I completely forgot about `launch_xml` and `launch_yaml`.

PR #346, opened 2019-10-16
Warn that old-style ready_fn and test attributes will be deprecated

pbaughman:
I'd like to get rid of the `ready_fn`  and the 'magic' creation of `self.proc_info` and `self.proc_output`, and other `self.whatever` attributes on test classes in launch_testing

This PR issues a warning when you use the 'old' way.

If your `generate_test_description` function [looks like this](https://github.com/ros2/launch/blob/9c6ae70e5c57dc4f32545108fdb6ffac78d4155f/launch_testing/test/launch_testing/examples/good_proc_launch_test.py#L47) instead [of like this](https://github.com/ros2/launch/blob/9c6ae70e5c57dc4f32545108fdb6ffac78d4155f/launch_testing/test/launch_testing/examples/ready_action_test.py#L52) You will get the following warning when you run the test:

```
root@93575d1ee932:/launch# ros2 test launch_testing/test/launch_testing/examples/good_proc_launch_test.py 
Running with ROS_DOMAIN_ID 1
ROS_DOMAIN_ID 1
[WARNING] [launch_testing.loader]: Passing ready_fn as an argument to
generate_test_description will be deprecated in a future release.  Include a
launch_testing.actions.ReadyToTest action in the LaunchDescription instead.
```

If you access proc_output or proc_info via `self` [like this](https://github.com/ros2/launch/blob/9c6ae70e5c57dc4f32545108fdb6ffac78d4155f/launch_testing/test/launch_testing/examples/good_proc_launch_test.py#L76) instead of [like this](https://github.com/ros2/launch/blob/9c6ae70e5c57dc4f32545108fdb6ffac78d4155f/launch_testing/test/launch_testing/examples/terminating_proc_launch_test.py#L55-L65) you will get the following warning when you run the test:

```
test_exit_code (good_proc_launch_test.TestProcessOutput) ... 
[WARNING] [launch_testing.loader]: Automatically adding attributes like self.proc_info and
self.proc_output to the test class will be deprecated in a future release.  Instead, add
proc_info or proc_output to the test method argument list to access the test object you need
```

### DIFF
--- a/launch_testing/launch_testing/loader.py
+++ b/launch_testing/launch_testing/loader.py
@@ -253,41 +253,17 @@ def _partially_bind_matching_args(unbound_function, arg_candidates):
 
 def _give_attribute_to_tests(data, attr_name, test_suite):
 
-    # Accessing the proc_info, proc_output, etc. . . objects via self.proc_info
-    # will be removed in a future release.  Wrap these objects in a way that produces a warning
-    # when they are used
-    class _warn_wrapper:
-
-        def __init__(self, wrapped):
-            self.__warned = False
-            self.__wrapped = wrapped
-
-        def __getattr__(self, attr):
-            self._warn_once()
-            return getattr(self.__wrapped, attr)
-
-        def __getitem__(self, key):
-            self._warn_once()
-            return self.__wrapped[key]
-
-        def __iter__(self):
-            self._warn_once()
-            return self.__wrapped.__iter__()
-
-        def _warn_once(self):
-
-            if self.__warned:
-                return
-
+    def _warn_getter(self):
+        if not hasattr(self, '__warned'):
             _logger.warning(
-                'Automatically adding attributes like self.proc_info and self.proc_output '
+                'Automatically adding attributes like self.{0} '
                 'to the test class will be deprecated in a future release.  '
-                'Instead, add proc_info or proc_output to the test method argument list to '
-                'access the test object you need'
+                'Instead, add {0} to the test method argument list to '
+                'access the test object you need'.format(attr_name)
             )
-            self.__warned = True
+            setattr(self, '__warned', True)
 
-    data = _warn_wrapper(data)
+        return data
 
     # Test suites can contain other test suites which will eventually contain
     # the actual test classes to run.  This function will recursively drill down until
@@ -296,7 +272,7 @@ def _give_attribute_to_tests(data, attr_name, test_suite):
     # The effect of this is that every test will have `self.attr_name` available to it so that
     # it can interact with ROS2 or the process exit coes, or IO or whatever data we want
     for test in _iterate_tests_in_test_suite(test_suite):
-        setattr(test, attr_name, data)
+        setattr(test.__class__, attr_name, property(fget=_warn_getter))
 
 
 def _iterate_test_classes_in_test_suite(test_suite):
--- a/launch_testing/launch_testing/loader.py
+++ b/launch_testing/launch_testing/loader.py
@@ -17,7 +17,12 @@ import inspect
 import itertools
 import unittest
 
+import launch.logging
+
 from .actions import ReadyToTest
+
+
+_logger = launch.logging.get_logger(__name__)
 
 
 def _normalize_ld(launch_description_fn):
@@ -35,8 +40,15 @@ def _normalize_ld(launch_description_fn):
         fn_args = inspect.getfullargspec(launch_description_fn)
 
         if 'ready_fn' in fn_args.args + fn_args.kwonlyargs:
-            # This is an old-style launch_description function which epects ready_fn to be passed
+            # This is an old-style launch_description function which expects ready_fn to be passed
             # in to the function
+            # This type of launch description will be deprecated in the future.  Warn about it
+            # here
+            _logger.warning(
+                'Passing ready_fn as an argument to generate_test_description will '
+                'be deprecated in a future release.  Include a launch_testing.actions.ReadyToTest '
+                'action in the LaunchDescription instead.'
+            )
             return normalize(launch_description_fn(**kwargs))
         else:
             # This is a new-style launch_description which should contain a ReadyToTest action
@@ -240,6 +252,43 @@ def _partially_bind_matching_args(unbound_function, arg_candidates):
 
 
 def _give_attribute_to_tests(data, attr_name, test_suite):
+
+    # Accessing the proc_info, proc_output, etc. . . objects via self.proc_info
+    # will be removed in a future release.  Wrap these objects in a way that produces a warning
+    # when they are used
+    class _warn_wrapper:
+
+        def __init__(self, wrapped):
+            self.__warned = False
+            self.__wrapped = wrapped
+
+        def __getattr__(self, attr):
+            self._warn_once()
+            return getattr(self.__wrapped, attr)
+
+        def __getitem__(self, key):
+            self._warn_once()
+            return self.__wrapped[key]
+
+        def __iter__(self):
+            self._warn_once()
+            return self.__wrapped.__iter__()
+
+        def _warn_once(self):
+
+            if self.__warned:
+                return
+
+            _logger.warning(
+                'Automatically adding attributes like self.proc_info and self.proc_output '
+                'to the test class will be deprecated in a future release.  '
+                'Instead, add proc_info or proc_output to the test method argument list to '
+                'access the test object you need'
+            )
+            self.__warned = True
+
+    data = _warn_wrapper(data)
+
     # Test suites can contain other test suites which will eventually contain
     # the actual test classes to run.  This function will recursively drill down until
     # we find the actual tests and give the tests a reference to the process
--- a/launch_testing/launch_testing/loader.py
+++ b/launch_testing/launch_testing/loader.py
@@ -265,14 +265,10 @@ def _give_attribute_to_tests(data, attr_name, test_suite):
 
         return data
 
-    # Test suites can contain other test suites which will eventually contain
-    # the actual test classes to run.  This function will recursively drill down until
-    # we find the actual tests and give the tests a reference to the process
-
     # The effect of this is that every test will have `self.attr_name` available to it so that
     # it can interact with ROS2 or the process exit coes, or IO or whatever data we want
-    for test in _iterate_tests_in_test_suite(test_suite):
-        setattr(test.__class__, attr_name, property(fget=_warn_getter))
+    for cls in _iterate_test_classes_in_test_suite(test_suite):
+        setattr(cls, attr_name, property(fget=_warn_getter))
 
 
 def _iterate_test_classes_in_test_suite(test_suite):
--- a/launch_testing/launch_testing/loader.py
+++ b/launch_testing/launch_testing/loader.py
@@ -15,14 +15,20 @@
 import functools
 import inspect
 import itertools
+import os
 import unittest
-
-import launch.logging
+import warnings
 
 from .actions import ReadyToTest
 
 
-_logger = launch.logging.get_logger(__name__)
+# Patch up the warnings module to streamline the warning messages.  See
+# https://docs.python.org/3/library/warnings.html#warnings.showwarning
+def slim_formatwarning(msg, *args, **kwargs):
+    return 'Warning: ' + str(msg) + os.linesep
+
+
+warnings.formatwarning = slim_formatwarning
 
 
 def _normalize_ld(launch_description_fn):
@@ -44,9 +50,9 @@ def _normalize_ld(launch_description_fn):
             # in to the function
             # This type of launch description will be deprecated in the future.  Warn about it
             # here
-            _logger.warning(
+            warnings.warn(
                 'Passing ready_fn as an argument to generate_test_description will '
-                'be deprecated in a future release.  Include a launch_testing.actions.ReadyToTest '
+                'be removed in a future release.  Include a launch_testing.actions.ReadyToTest '
                 'action in the LaunchDescription instead.'
             )
             return normalize(launch_description_fn(**kwargs))
@@ -255,7 +261,7 @@ def _give_attribute_to_tests(data, attr_name, test_suite):
 
     def _warn_getter(self):
         if not hasattr(self, '__warned'):
-            _logger.warning(
+            warnings.warn(
                 'Automatically adding attributes like self.{0} '
                 'to the test class will be deprecated in a future release.  '
                 'Instead, add {0} to the test method argument list to '
--- a/launch_testing/test/launch_testing/examples/good_proc_launch_test.py
+++ b/launch_testing/test/launch_testing/examples/good_proc_launch_test.py
@@ -22,6 +22,7 @@ import launch
 import launch.actions
 
 import launch_testing
+import launch_testing.actions
 from launch_testing.asserts import assertSequentialStdout
 
 import pytest
@@ -44,13 +45,13 @@ dut_process = launch.actions.ExecuteProcess(
 
 
 @pytest.mark.launch_test
-def generate_test_description(ready_fn):
+def generate_test_description():
 
     return launch.LaunchDescription([
         dut_process,
 
         # Start tests right away - no need to wait for anything
-        launch.actions.OpaqueFunction(function=lambda context: ready_fn()),
+        launch_testing.actions.ReadyToTest(),
     ])
 
 
--- a/launch_testing/test/launch_testing/test_proc_info_assertions.py
+++ b/launch_testing/test/launch_testing/test_proc_info_assertions.py
@@ -21,12 +21,13 @@ import launch
 import launch.actions
 import launch.events.process
 from launch_testing.test_runner import LaunchTestRunner
+import launch_testing.actions
 import launch_testing.util
 
 
 def test_wait_for_shutdown(source_test_loader):
 
-    def generate_test_description(ready_fn):
+    def generate_test_description():
         TEST_PROC_PATH = os.path.join(
             ament_index_python.get_package_prefix('launch_testing'),
             'lib/launch_testing',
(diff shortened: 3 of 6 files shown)


pbaughman:
@hidmic ready_fn and self.proc_whatever arguments are very very old and have both been replaced by something better and more similar to regular `pytest` tests.  I'm hoping we can issue a warning for a while, and then remove these all-together in a future release.  I'll also prepare a branch with these features removed so we can run CI and figure out what ROS packages need to be updated to work the 'new' way.

launch_testing/launch_testing/loader.py
```diff
@@ -231,6 +243,43 @@ def _partially_bind_matching_args(unbound_function, arg_candidates):
...
+
+            _logger.warning(
+                'Automatically adding attributes like self.proc_info and self.proc_output '
+                'to the test class will be deprecated in a future release.  '
+                'Instead, add proc_info or proc_output to the test method argument list to '
+                'access the test object you need'
+            )
+            self.__warned = True
+
+    data = _warn_wrapper(data)
```
hidmic: @pbaughman meta: wouldn't it be simpler to turn inject attributes into properties that warn on access? Instead of trying to cover every possible way these attribute may be used I mean.

launch_testing/launch_testing/loader.py
```diff
@@ -231,6 +243,43 @@ def _partially_bind_matching_args(unbound_function, arg_candidates):
...
+        def _warn_once(self):
+
+            if self.__warned:
+                return
+
+            _logger.warning(
+                'Automatically adding attributes like self.proc_info and self.proc_output '
+                'to the test class will be deprecated in a future release.  '
+                'Instead, add proc_info or proc_output to the test method argument list to '
+                'access the test object you need'
```
hidmic: @pbaughman I think it'd be nice to use `attr_name` in the warning to give the user more precise directions. WDYT?

launch_testing/launch_testing/loader.py
```diff
@@ -17,9 +17,14 @@
 import itertools
 import unittest
 
+import launch.logging
+
 from .actions import ReadyToTest
 
 
+_logger = launch.logging.get_logger(__name__)
```
hidmic: @pbaughman meta: I wonder if we shouldn't use the `warnings` module instead.

launch_testing/launch_testing/loader.py
pbaughman: @hidmic I will try - I'm not sure what it takes to dynamically add a property to a class but I'm sure I can figure it out

launch_testing/launch_testing/loader.py
pbaughman: That's a good idea

launch_testing/launch_testing/loader.py
pbaughman: I'm not sure I've used that module before, so I can't speak to the pros and cons.  You're referring [to this](https://docs.python.org/3/library/warnings.html) right?

hidmic: Right.

launch_testing/launch_testing/loader.py
hidmic: See [here](https://docs.python.org/3.6/howto/descriptor.html#properties).

pbaughman:
@hidmic Here is as checklist of packages that use launch_testing and do things the "old" way.  I will check them off as they're updated:

- [ ] rcl
- [ ] ros1_bridge
- [ ] demos
- [ ] rcutils
- [ ] system_tests
- [ ] launch_ros
- [ ] ros2cli

This list was generated by by doing:

```
wget https://raw.githubusercontent.com/ros2/ros2/master/ros2.repos
vcs import src < ros2.repos
grep -R "ready_fn"
``` 

And then ignoring stuff in launch_testing

hidmic:
I think that's a pretty accurate recollection. Watch out, some new `ros2cli` tests are hopefully coming along with https://github.com/ros2/ros2cli/pull/304.

launch_testing/launch_testing/loader.py
pbaughman: @hidmic Using the warnings module changes

```
test_count_to_four (good_proc_launch_test.TestGoodProcess) ... [WARNING] [launch_testing.loader]: Automatically adding attributes like self.proc_outputto the test class will be deprecated in a future release.  Instead, add proc_output to the test method argument list to access the test object you need
```

To 

```
test_count_to_four (good_proc_launch_test.TestGoodProcess) ... /launch/install/launch_testing/lib/python3.6/site-packages/launch_testing/loader.py:251: FutureWarning: Automatically adding attribut
es like self.proc_output to the test class will be deprecated in a future release.  Instead, add proc_output to the test method argument list to access the test object you need
  FutureWarning,
```

It tries to print 1 line of source code at the end - that's why it says "FutureWarning"  I'm not sure how to get it to stop doing that, but I also didn't look very hard.

It doesn't seem like using the 'warnings' module is a clear winner here

hidmic: I don't know how you're using the module. I think that using the `stacklevel` parameter appropriately you may get to refer to the exact line that triggered the warning (which is cool IMO).

launch_testing/launch_testing/loader.py
```diff
@@ -231,14 +243,27 @@ def _partially_bind_matching_args(unbound_function, arg_candidates):
...
+
     # Test suites can contain other test suites which will eventually contain
     # the actual test classes to run.  This function will recursively drill down until
     # we find the actual tests and give the tests a reference to the process
 
     # The effect of this is that every test will have `self.attr_name` available to it so that
     # it can interact with ROS2 or the process exit coes, or IO or whatever data we want
     for test in _iterate_tests_in_test_suite(test_suite):
-        setattr(test, attr_name, data)
+        setattr(test.__class__, attr_name, property(fget=_warn_getter))
```
hidmic: @pbaughman careful with this, as the property applies to all tests that use that same class. Since attributes are always available, I guess it's fine in this case, though you may want to iterate over test classes instead.

launch_testing/launch_testing/loader.py
pbaughman: @hidmic It's already pointing to the exact line that triggered the warning.  The source code is:

```
warnings.warn(                                                                         
    'Passing ready_fn as an argument to generate_test_description will '               
    'be deprecated in a future release.  Include a launch_testing.actions.ReadyToTest '
    'action in the LaunchDescription instead.',                                        
    FutureWarning,                                                                     
)                                                                                      
```

That's where the somewhat useless "FutureWarning," comes from in the printed message.

I can play with the stacklevel bit.  When warning about 'self.proc_info' and 'self.proc_output' I can get it to point to the test file that the user wrote:

```
test_full_output (good_proc_launch_test.TestProcessOutput) ... /launch/launch_testing/test/launch_testing/examples/good_proc_launch_test.py:81: FutureWarning: Automatically adding attributes like self.proc_output to the test class will be deprecated in a future release.  Instead, add proc_output to the test method argument list to access the test object you need
  with assertSequentialStdout(self.proc_output, dut_process) as cm:
```

When warning about using the ready_fn I doubt I can get it to show part of the generate_test_description function because it's long-gone from the stack by the time raise this warning.

Then options as I see them are:
1) Use warnings.warn for both warnings, but the one about the ready_fn has a useless extra "FutureWarning," printed at the end which is confusing
2) Use warnings.warn for one warning, but not the other - now we have two different formatted warnings, but at least one of them prints the file name and the line that it came from
3) Use logger warnings for both.  They're formatted the same, and work the same in the code but we don't see what line a warning for self.proc_info or self.proc_output comes from

launch_testing/launch_testing/loader.py
hidmic: Yeah, basically. Do you think it's that risky?

launch_testing/launch_testing/loader.py
hidmic: > When warning about using the ready_fn I doubt I can get it to show part of the generate_test_description function because it's long-gone from the stack by the time raise this warning.

True that. Hmm, what if we warn them as soon as we see that `generate_test_description` takes a `ready_fn` argument? To make it extra robust, we should also either ban a user-provided `ready_fn` parameter or stay silent and just use the user's if one's actually there.

launch_testing/launch_testing/loader.py
pbaughman: I think it's riskier than doing it this way, and also this whole method goes away once we stop injecting attributes into test classes, so the sub-optimal way of doing it is relatively short-lived.

I suppose I could do 
```
for test in _iterate_tests_in_test_suite(test_suite):
    if not hasattr(test.__class__, attr_name):
        setattr(test.__class__, attr_name, property(fget=_warn_getter))
```
That doesn't seem a whole lot better though, unless there's some huge performance penalty doing `setattr` or constructing a property object

pbaughman:
Note, this PR is on-hold until I fix https://github.com/ros2/launch/issues/347

Otherwise the existing ros2cli tests can't be updated to use the ReadyToTest action because it's not discoverable by launch_testing

launch_testing/launch_testing/loader.py
hidmic: Why not using [`_iterate_test_classes_in_test_suite()`](https://github.com/ros2/launch/blob/f73bc56caa9c7b3c53b264475728392d0e1ce9cf/launch_testing/launch_testing/loader.py#L253-L258)?

pbaughman: Sorry, yeah - because I forgot it was a thing.  Thank you for being a good reviewer

wjwwood:
What's the status of this pull request? Seems like the issue referenced in the most recent comment is closed.

pbaughman:
Hey @wjwwood .  This PR causes warnings in other packages, and I haven't been able to resolve them all.

[This PR in ros2/system_tests](https://github.com/ros2/system_tests/pull/391) has test failures when I make the change
[This PR in ros2/ros2cli](https://github.com/ros2/ros2cli/pull/376) is even more bizarre - the test times out waiting for the processes to start up

The situation is slightly complicated because I'm currently not doing any new work for Apex right now.  I can tackle these with @dejanpan 's approval, or maybe someone else from Apex will take over this PR.  Hopefully he can chime in.

wjwwood:
Ok, sounds good and thanks for the update. I'll leave it as-is for now.

ivanpauno:
@pbaughman @dejanpan I've added this to the backlog, I can remove the label if you're planning to work on it in the near future.

pbaughman:
@ivanpauno I've reached out to Apex again to see if they want me to tackle this.  Forgive my ignorance of your process but what does the "backlog" label mean, functionally, and what effect does it have on me if it's added or removed?

mjcarroll:
It means it's not actively being looked at or worked on, unless someone
wants to revive it.

In our weekly triage process, it essentially filters it from our queue of
issues to look at.

pbaughman:
@ivanpauno I get to finish this after all!  Unless I find some awful underlying problem I should have this ready to go by Monday or Tuesday next week.

pbaughman:
Just a quick update that I'm still working on this.  I got the system_tests PR merged when we determined the test failures were also on the nightlies.

the ros2cli test failures appear to be a real race condition in the test that is taking some effort to resolve.  Progress is being made here: https://github.com/ros2/ros2cli/pull/454 and here: https://github.com/ros2/ros2cli/pull/376

pbaughman:
It looks like one other package had some tests added that use the ready_fn
  - [ ] geometry2: https://github.com/ros2/geometry2/pull/243/files

All of the other packages have been updated

jacobperron:
Full CI to see where we stand:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9710/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5366/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7915/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9604/)

pbaughman:
@jacobperron Thank you - I will take a look at the CI results as soon as I get up tomorrow and rebase or fix any issues.  I appreciate you sticking with this one

pbaughman:
The failure in `launch_testing.test.launch_testing.test_flake8.test_flake8` was the real deal - I've fixed the import order in test_proc_info_assertions.py and test_resolve_process.py.  Continuing to investigate the other failures to see if they're related to this change.

pbaughman:
The other test failures don't look related to this change, but I don't know where you keep your list of problematic tests to ensure these tests are all on it.

jacobperron:
@pbaughman What I typically do is look at the nightly jobs ("*_repeated") to see if failures are an existing issue: https://ci.ros2.org/view/nightly/

The other failures (besides launch_testing flake8) look like there already occurring nightly to me.

jacobperron:
Confirming flake8 failure is resolved:
Linux [Build Status](https://ci.ros2.org/job/ci_linux/9721/)

launch_testing/launch_testing/loader.py
jacobperron: If we use `warnings`, then we see the warning messages if the tests happen to be run with `pytest` (instead of `ros2 test`). I tried locally and do not see launch logs if I use `pytest`.

pbaughman: Is this because of the way the pytest wrapper for launch tests captures all stdout and only shows it on failure?  IE, if you run with `-s` do you see the warning output?

jacobperron: I do see the warning with `-s`, albeit buried in the rest of the test output. If we use the `warnings` module then we also see warning messages in the pytest summary. Using the `warnings` module also lets users to handle the warnings differently if they want, for example ignoring them or treating them as errors. Perhaps this is also configurable with the launch logger; I don't know.

pbaughman: Yeah, I really dislike how the pytest wrapper combines as many tests as there are in a file into one mega-test.  It's quite difficult to interpret the output.  I can switch this to the 'warnings' module as long as we don't go around and around about the message that's printed.  I didn't take that route originally because I couldn't figure out how to get it to display a warning without a bunch of meaningless (to the end user) extra stuff

pbaughman:
@jacobperron Is this good to merge then?  Not sure if you're waiting on an action from me or not.

launch_testing/launch_testing/loader.py
jacobperron: I think if you use the warnings module with the default warning (UserWarning), then I don't think it displays "extra stuff". Even if it is ugly, I think it gets the message across. IMO, the warnings module is more visible for users compared to the launch logger warning. If someone else wants to weigh in, I'm not going to die on this hill. @hidmic ?

jacobperron:
@pbaughman see https://github.com/ros2/launch/pull/346#discussion_r397588149

launch_testing/launch_testing/loader.py
hidmic: Had to parse the discussion again, it's been a while.

To let the user know a feature is deprecated, I'd certainly stick to the `warnings` module. Not only for `launch`, but for every other ROS 2 core Python package out there too. @pbaughman I still think you may be able to get better output for `ready_fn` deprecation warnings if you do so when you bind them to the test function instead.

hidmic: And about `pytest` aggregating all output in a rather confusing way, I agree. That's because there's a clear boundary between `launch_testing` and `pytest` (see [here](https://github.com/ros2/launch/blob/a5ec2a46595e7b711a1ffc70fe2b2dfd966aac9c/launch_testing/launch_testing/pytest/hooks.py#L42-L61)), and thus an entire launch test (with all its test cases) is a single `pytest`. To get fine grained output we need deeper integration. Turning `launch_testing` into an extension of `pytest` has been and is a path to achieve that.

pbaughman:
@jacobperron 
Ok, here's what I can make the output from the example test `good_proc.test.py` look like:

```
root@4841d43f6eb4:/launch# launch_test launch_testing/test/launch_testing/examples/good_proc_launch_test.py                                                                                      
Warning: Passing ready_fn as an argument to generate_test_description will be removed in a future release.  Include a launch_testing.actions.ReadyToTest action in the LaunchDescription instead.
[INFO] [launch]: All log files can be found below /root/.ros/log/2020-03-26-17-11-32-354584-4841d43f6eb4-1027
test_count_to_four (good_proc_launch_test.TestGoodProcess) ... Warning: Automatically adding attributes like self.proc_output to the test class will be deprecated in a future release.  Instead,
add proc_output to the test method argument list to access the test object you need
[INFO] [python3-1]: process started with pid [1030]               
[python3-1] Starting Up                                           
[python3-1] Loop 0                                                    
```

Hopefully this is satisfactory.  @hidmic I'm sorry, I just don't understand what you're getting at when you say
> I still think you may be able to get better output for ready_fn deprecation warnings if you do so when you bind them to the test function instead.
Unless I somehow inject code into the `generate_test_description` function, I can't make the stack trace point to the inside the `generate_test_description` function

Hopefully the latest commit produces an output that works for everybody tough.

hidmic:
> Unless I somehow inject code into the `generate_test_description` function, I can't make the stack trace point to the inside the `generate_test_description` function

Well, there's [`warnings.warn_explicit()`](https://docs.python.org/3/library/warnings.html#warnings.warn_explicit) with which you can refer to the test source file, but consider that an FYI.

> Hopefully the latest commit produces an output that works for everybody tough.

What you have is good enough I guess. @jacobperron ?

pbaughman:
Not sure how I ended up with a flake8 error in `launch_context.py` - that's not part of the changes in this PR

```
AssertionError: Found 1 code style errors / warnings:   ./launch/launch_context.py:124:5: A003 "locals" is a python builtin, consider renaming the class attribute assert 1 == 0
```

Maybe a new version of flake8 that added this error?

jacobperron:
We've switched our builds over to Ubuntu Focal, which recently had an update to flake8 I believe. The same error exists on master and shouldn't block this PR.

jacobperron:
Things looks fine compared to the nightlies. I'll just hold to retrigger the Windows build once the fix is merged.

pbaughman:
@jacobperron @wjwwood I'll put together a branch that actually removes these features, and I guess you can hold onto it until after Foxy is done and out the door.  Does that work for you guys?

jacobperron:
Sounds good to me. Maybe you can open it as a draft PR so it doesn't get accidentally merged before Foxy?

pbaughman:
@jacobperron No problem - How do I make something a 'draft PR' ?

PR #348, opened 2019-10-23
Support launch test reruns when using pytest

hidmic:
Precisely what the title says. Without this patch, test failures coupled with a `--reruns` option result in:

```bash
AttributeError: 'NoneType' object has no attribute 'launch_service'
```

as it happens [here](https://ci.ros2.org/job/ci_windows/8434/testReport/junit/ros2topic.test/test_echo_pub/test_echo_pub/).

### DIFF
--- a/launch_testing/launch_testing/pytest/hooks.py
+++ b/launch_testing/launch_testing/pytest/hooks.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import copy
 from unittest import TestCase
 
 import pytest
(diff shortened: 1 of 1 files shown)


hidmic:
CI (up to `launch_testing`, sanity check only)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8531/)

hidmic:
> Is it pytest that is calling runtest and overwriting test_runs?

It's the combination of `rerunfailures` plugin calling `runtest()` multiple times and `unittest.TextTestRunner` eventually calling `unittest.TestSuite.run`, which [drops tests and replaces them with `None`](https://github.com/python/cpython/blob/0f221d09cad46bee38d1b7a7822772df66c53028/Lib/unittest/suite.py#L60-L81).

hidmic:
@mjcarroll PTAL

PR #352, opened 2019-10-25
Make launch_testing.markers.retry_on_failure decorator more robust.

hidmic:
Precisely what the title says, to help deal with `unittest.TestCase.subTest` usage correctly.

### DIFF
--- a/launch_testing/launch_testing/markers.py
+++ b/launch_testing/launch_testing/markers.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 import functools
+import inspect
+import unittest
 
 
 def keep_alive(test_description):
@@ -28,14 +30,18 @@ def retry_on_failure(*, times):
     assert times > 0
 
     def _decorator(func):
+        assert 'self' == list(inspect.signature(func).parameters)[0]
         @functools.wraps(func)
-        def _wrapper(*args, **kwargs):
+        def _wrapper(self, *args, **kwargs):
             n = times
             while n > 1:
                 try:
-                    return func(*args, **kwargs)
+                    ret = func(self, *args, **kwargs)
+                    if isinstance(self, unittest.TestCase):
+                        assert self._outcome.success
+                    return ret
                 except AssertionError:
                     n -= 1
-            return func(*args, **kwargs)
+            return func(self, *args, **kwargs)
         return _wrapper
     return _decorator
--- a/launch_testing/launch_testing/markers.py
+++ b/launch_testing/launch_testing/markers.py
@@ -41,6 +41,8 @@ def retry_on_failure(*, times):
                         assert self._outcome.success
                     return ret
                 except AssertionError:
+                    self._outcome.errors.clear()
+                    self._outcome.success = True
                     n -= 1
             return func(self, *args, **kwargs)
         return _wrapper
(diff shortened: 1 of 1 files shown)


hidmic:
CI (up to `ros2topic`, `ros2action`, `ros2service`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8571/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4445/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6993/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8489)

launch_testing/launch_testing/markers.py
```diff
@@ -28,14 +30,18 @@ def retry_on_failure(*, times):
...
         @functools.wraps(func)
-        def _wrapper(*args, **kwargs):
+        def _wrapper(self, *args, **kwargs):
             n = times
             while n > 1:
                 try:
-                    return func(*args, **kwargs)
+                    ret = func(self, *args, **kwargs)
+                    if isinstance(self, unittest.TestCase):
+                        assert self._outcome.success
```
ivanpauno: Why is this necessary?

hidmic: [`unittest.TestCase.subTest`](https://github.com/python/cpython/blob/master/Lib/unittest/case.py#L547) catches all `AssertionError` exceptions thrown and simply updates the test case result. That allows it to continue going over other sub tests, but breaks this decorator (which expects an `AssertionError` if there's a failure). This extra assertion makes sure that those updates do not pass by unnoticed.

hidmic:
Well, @BMarchi and I just found out this isn't enough. Errors persist within `self._outcome`. Also, this `retry_on_failure` decorator does not `setUp()` and `tearDown()` for every attempt, which may be counter-intuitive. I think we can fix the first issue by restoring `self._outcome` state, but the latter would require re-writing the base `TestCase.run()` implementation. 

@pbaughman @dirk-thomas @wjwwood I think this is yet another reason why we may want to move towards `pytest`, I don't see value in re-implementing more and more of `unittest` or `launch_testing` t to get something that's already available in `pytest`.

pbaughman:
@hidmic You won't get an argument from me.  The reason this is not a pytest plugin already is because I bill my clients hourly, so it was cheaper to get it working this way.

pbaughman:
Side-topic:  Is the intended use case to re-launch the launched processes and re-run the test cases on failure, or just to re-run a test case with the same launched processe?  The latter case seems like it's a magnet for counter-intuitive edge cases, like when there are side-effects in the test case.

hidmic:
> Side-topic: Is the intended use case to re-launch the launched processes and re-run the test cases on failure, or just to re-run a test case with the same launched processe? 

I see value in both. The latter is helpful to cope with discovery latencies.

> The latter case seems like it's a magnet for counter-intuitive edge cases, like when there are side-effects in the test case.

True, it should be clear though what to expect from your test.

BMarchi:
CI (up to `ros2topic`, `ros2action`, `ros2service`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8612/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4466/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7032/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8530)

hidmic:
@mjcarroll let's get this in once CI is fully green.

hidmic:
Ok, going in!

PR #358, opened 2019-11-18
Fix junit XML when launch dies early

pbaughman:
- old junit XML showed 0 cases, 0 failures when a test stopped early
  - Colcon test-result would not spot this as a failure
  - New method uses same logic as 'skip' when there's a skip decorator on the launch description

I'm not sure how this slipped through for so long, but I just started noticing this with tests that shutdown early.  If a test [stops early](https://github.com/ros2/launch/blob/2f32b9fa94a5cd441271a9b23850db18b9c7d28e/launch_testing/launch_testing/test_runner.py#L162-L169) either because of ctrl+c or because all of the processes in the launch description stop before the active tests complete, we generate a [fail result](https://github.com/ros2/launch/blob/2f32b9fa94a5cd441271a9b23850db18b9c7d28e/launch_testing/launch_testing/test_runner.py#L263-L266)

This fail result correctly sets the exit code for launch_testing, but the generated junit XML looks something like this

```
<testsuites errors="0" failures="0" name="rcutils.test_logging_long_messages" tests="0" time="0">
  <testsuite errors="0" failures="0" name="rcutils.test_logging_long_messages.launch_tests" skipped="0" tests="0" time="0" />
</testsuites>   
```

When 'failures' and 'errors' is set to 0, colcon test-result won't pick this up as a failing result.

After this PR, the generated junit XML looks a bit like this:

```
<?xml version='1.0' encoding='utf-8'?>                                                                                                                                                     
<testsuites errors="0" failures="1" name="rcutils.test_logging_long_messages" tests="1" time="0">
    <testsuite errors="0" failures="1" name="rcutils.test_logging_long_messages.launch_tests" skipped="0" tests="1" time="0">
        <testcase classname="rcutils.TestLoggingLongMessages" name="test_logging_output" time="0"> 
             <failure message="Exception: Launch stopped before the active tests finished.&#10;" />
        </testcase>
    </testsuite>
</testsuites>         
```

Now that failures is set to a non-zero value, colcon test-result will correctly detect this failure.

The logic I'm using for the FailResult is almost exactly the same as the skip result down below.  We loop through all of the test cases in the test run and call "AddFailure" to log the failure.

### DIFF
--- a/launch_testing/launch_testing/test_result.py
+++ b/launch_testing/launch_testing/test_result.py
@@ -20,14 +20,20 @@ import unittest
 class FailResult(unittest.TestResult):
     """For test runs that fail when the DUT dies unexpectedly."""
 
+    def __init__(self, test_run, message):
+        super().__init__()
+        for case in test_run.all_cases():
+            self.addFailure(case, (Exception, Exception(message), None))
+            self.testsRun += 1
+
     @property
     def testCases(self):
-        return []
+        return [failure[0] for failure in self.failures]
 
     @property
     def testTimes(self):
         """Get a dict of {test_case: elapsed_time}."""
-        return {}
+        return {failure[0]: 0 for failure in self.failures}
 
     def wasSuccessful(self):
         return False
--- a/launch_testing/launch_testing/test_runner.py
+++ b/launch_testing/launch_testing/test_runner.py
@@ -264,7 +264,7 @@ class LaunchTestRunner(object):
                 # The most likely cause was ctrl+c, so we'll abort the test run
                 results[run] = FailResult(
                     test_run=run,
-                    message="Launch stopped before the active tests finished."
+                    message='Launch stopped before the active tests finished.'
                 )
                 break
 
--- a/launch_testing/launch_testing/test_runner.py
+++ b/launch_testing/launch_testing/test_runner.py
@@ -262,7 +262,10 @@ class LaunchTestRunner(object):
                 continue
             except _LaunchDiedException:
                 # The most likely cause was ctrl+c, so we'll abort the test run
-                results[run] = FailResult()
+                results[run] = FailResult(
+                    test_run=run,
+                    message="Launch stopped before the active tests finished."
+                )
                 break
 
         return results
--- a/launch_testing/test/launch_testing/test_xml_output.py
+++ b/launch_testing/test/launch_testing/test_xml_output.py
@@ -98,16 +98,43 @@ class TestXmlFunctions(unittest.TestCase):
             return runner.run(cases)
 
     def test_fail_results_serialize(self):
+
+        def generate_test_description(ready_fn):
+            raise Exception('This should never be invoked')  # pragma: no cover
+
+        def test_fail_always(self):
+            assert False  # pragma: no cover
+
+        def test_pass_always(self):
+            pass  # pragma: no cover
+
+        test_runs = self.source_test_loader(
+            generate_test_description,
+            pre_shutdown_tests=[
+                test_fail_always,
+            ],
+            post_shutdown_tests=[
+                test_fail_always,
+                test_pass_always
+            ]
+        )
+
+        self.assertEqual(1, len(test_runs))  # Not a parametrized launch, so only 1 run
+
         xml_tree = unittestResultsToXml(
             name='fail_xml',
             test_results={
-                'active_tests': FailResult()
+                'active_tests': FailResult(test_run=test_runs[0])
             }
         )
 
         # Simple sanity check - see that there's a child element called active_tests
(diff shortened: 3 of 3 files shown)


pbaughman:
@hidmic I just started noticing this with rcutil tests.  I don't know why I didn't spot this before - it seems like a pretty obvious failure/problem.

pbaughman:
@hidmic poke.  Thoughts on this?

dirk-thomas:
I would suggest to reconsider when these result files get generated. In `ament_cmake` when adding a test the "failed" result file is being generated at the very beginning of the test execution. And only after the test has finished the result file is being overwritten with the actual result. That makes the approach resilient to almost any kind of failure of the test execution and makes sure that there is always a result file present - even when the test execution is killed half way through.

hidmic:
> I would suggest to reconsider when these result files get generated. In ament_cmake when adding a test the "failed" result file is being generated at the very beginning of the test execution.

@dirk-thomas I don't think we're going to be able to pull that one off. We might be able to write a bare result file somewhat closer to the beginning of the process execution, but that's it. On a side note, `pytest` doesn't even try to (see the `pytest_sessionfinish()` hook in the [`pytest.junitxml`](https://doc.pytest.org/en/3.6.2/_modules/_pytest/junitxml.html) module).

dirk-thomas:
> I don't think we're going to be able to pull that one off.

Do you have a reason why not? That should be fairly simple and purely additional to whatever else is done.

hidmic:
> Do you have a reason why not?

Why can't we have the same resiliency that `ament_cmake` achieves? We'd need another process to put those bare result files in place. `launch_testing_ament_cmake` can take care of that, but for pure Python packages `pytest` won't. We *can* easily write a file ASAP, but that isn't quite the same. 

I personally think that having some sort of process monitor running test processes would be beneficial as it could collect much more information than just a bare result file (like exit codes or core dumps).

pbaughman:
> @pbaughman it looks like a fine patch to me. Having said that, it doesn't really make the failure mode any less obscure (processes died, but why? before which test?). Having `None` in the failure message isn't super helpful either.

It's not really a matter of 'before which test.'  It's 'we expected the launch_description to still be running while the active tests were running, but it stopped on its own before all of the active tests were done.  Please fix your launch description'

You're right about the failure message in the description though - sorry - I copy/pasted that from earlier in the development process.  There's actually a message attached [here](https://github.com/ros2/launch/blob/1f8a5690e1fb35f2b373bb49aacdc3b54795839a/launch_testing/launch_testing/test_result.py#L26)

It should read

```
<?xml version='1.0' encoding='utf-8'?>                                                                                                                                                     
<testsuites errors="0" failures="1" name="rcutils.test_logging_long_messages" tests="1" time="0">
    <testsuite errors="0" failures="1" name="rcutils.test_logging_long_messages.launch_tests" skipped="0" tests="1" time="0">
        <testcase classname="rcutils.TestLoggingLongMessages" name="test_logging_output" time="0"> 
             <failure message="Exception: Launch stopped before the active tests finished.&#10;" />
        </testcase>
    </testsuite>
</testsuites>                                                                                                                                   
```

The message itself comes from [here](https://github.com/ros2/launch/blob/1f8a5690e1fb35f2b373bb49aacdc3b54795839a/launch_testing/launch_testing/test_runner.py#L267)

pbaughman:
@dirk-thomas 

> I would suggest to reconsider when these result files get generated. In `ament_cmake` when adding a test the "failed" result file is being generated at the very beginning of the test execution. And only after the test has finished the result file is being overwritten with the actual result. That makes the approach resilient to almost any kind of failure of the test execution and makes sure that there is always a result file present - even when the test execution is killed half way through.

"When" wouldn't fix this problem.  The issue is with the generated results themselves.  Even if I generated them before running any tests, they would still be wrong and not detected as a failure by `colcon test-result`  In fact, that's what's happening here when running with `colcon test`.  ament_cmake generated some fine "this test didn't generate any results and is therefore a fail" XML, and then launch_testing overwrote them with XML that's intended to convey "Failure" but were actually passing results indicating that no tests were run.

pbaughman:
Can we move forward with this?

hidmic:
@pbaughman thanks for the bump. 

I'll start CI, but I'd like @dirk-thomas to confirm he's good with this patch.


* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8749/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4578/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7151/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8666/)

hidmic:
@dirk-thomas any other thoughts? I'll merge otherwise.

PR #360, opened 2019-12-09
Fix for module import issue

### LINKED ISSUE #357: launch_testing test file import can cause confusing module import issues
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - source
- Version or commit hash:
  - Dashing
- DDS implementation:
  - Doesn't matter, this is a pure-python thing
  - N/A

#### Steps to reproduce issue

We have a package called `lidar_integration.`  This package installs a python module also called lidar_integration and it has some tests that run with `ros2 test.`  The test file is called `foo_test.py`

When we run the test with `colcon test` or with `ros2 test foo_test.py --package_name lidar_integration` we get a bizarre import error that was difficult to debug, but easy to understand once we figured out what's going on.  The test tries to do

```
from lidar_integration import useful_tool
```

But then fails with someting like (sorry, error message is from memory)

```
useful_tool not a member of lidar_integration
```

Because of the way the test file was imported by launch_testing [here](https://github.com/ros2/launch/blob/2f32b9fa94a5cd441271a9b23850db18b9c7d28e/launch_testing/launch_testing/launch_test.py#L78-L80), when you import lidar_integration from the test, instead of getting the lidar_integration module which contains `useful_tool`, you get test file which was named 'lidar_integration' when it was imported because `lidar_integration` is the package name and that was used to name the module when the SourceFileLoader loaded it

#### Expected behavior

I'm not sure, exactly.  I wanted to get @hidmic 's opinion on this.  If you run the test with ros2 test foo_test.py, there's no problem becaues when we import the test file we name it `foo_test`.  `foo_test` doesn't collide with any modules names so there's no problem.

I suspect we use the package name as the module name because this information makes it into the junit XML generated by launch_testing (I'm speculating based on the git blame history of this line).  I have not yet investigated what happens to the junit XML if I change the name of the imported test file.

#### Additional information

I want your take on how to solve this problem.  My ideas thus far are

  * Always use the launch_test_file_basename when calling _load_python_file_as_module
    * Pros:  Solves this case
    * Cons:  Could still collide in other cases (like if the test was named lidar_integration.py).  Maybe messes up the junit XML results (I haven't checked yet)
  * Move this test into another package called `test_lidar_integration`
    * Pros:  Solves the problem.  Not totally inconsistent with packages like test_launch_testing
    * Cons: This solves our problem, but it doesn't help anybody else whose confused by the same issue
  * Maybe when we're importing the module, try to import that name and issue a warning if the import succeeds
    * Pros: Would make it easier to understand what's going on
    * Cons: Does not actually solve the problem.  You'd still need to do one of the other things above

pbaughman:
- Try a 'real' import of the test file name before we load the module from source.
    This prevents the source file loader from hiding the real module (if it exists)
    in the sys.modules cache

Closes #357

### DIFF
--- a/launch_testing/launch_testing/launch_test.py
+++ b/launch_testing/launch_testing/launch_test.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import argparse
-from importlib.machinery import SourceFileLoader
+import importlib.util
 import logging
 import os
 import sys
@@ -28,9 +28,10 @@ _logger_ = logging.getLogger(__name__)
 
 def _load_python_file_as_module(test_module_name, python_file_path):
     """Load a given Python launch file (by path) as a Python module."""
-    # Taken from launch_testing to not introduce a weird dependency thing
-    loader = SourceFileLoader(test_module_name, python_file_path)
-    return loader.load_module()
+    spec = importlib.util.spec_from_file_location(test_module_name, python_file_path)
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
 
 
 def add_arguments(parser):


pbaughman:
@hidmic Here's the fix.  It's more explanation than code

pbaughman:
@wjwwood launch gives the file a fixed name "python_launch_file" so unless someone makes a module with the same name, launch will not encounter the same problem.

I don't remember why 'launch' and 'launch_testing' don't use the same load_python_launch_file_as_module method.

An alternative fix would be to have launch_testing use the same method as launch to import the test file - maybe import it as `python_launch_test` instead of `python_launch_file`, but as mentioned in the issue that will change the names of the tests that get generated.

hidmic:
> Seems like a hack, but maybe it will help someone avoid a rough edge, though I wouldn't expect launch to do this for you, so it may be surprising either way.

It is a bit of a hack, and I agree that it might silently incur in unwanted side effects. I didn't know of a superior alternative though, for the reasons that @pbaughman already mentioned in #357.

Having said that, after digging a bit further into `importlib` documentation, there might be a way to import a module while preventing the interpreter from caching it in `sys.modules`. @pbaughman mind to try [this](https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly)?

pbaughman:
@hidmic I tried doing this:

```
def _load_python_file_as_module(test_module_name, python_file_path):                 
    """Load a given Python launch file (by path) as a Python module."""              

    spec = importlib.util.spec_from_file_location(test_module_name, python_file_path)
    return importlib.util.module_from_spec(spec)                                     
```

But now things are blowing up here: https://github.com/ros2/launch/blob/master/launch_testing/launch_testing/loader.py#L154

with an error "module lidar_integration has no attribute 'generate_test_description'

I think I've just reversed the problem.  Before, it was looking in the test file for things that should be found in the real module.  Now I think it's looking in the real module for things that should be found in the test file.

pbaughman:
Maybe the correct fix here is "don't name your test files the same thing as a real module" and a better warning/error message when you do. . .

pbaughman:
@hidmic Ah, no I was wrong. . . This does it:

```
def _load_python_file_as_module(test_module_name, python_file_path):                 
    """Load a given Python launch file (by path) as a Python module."""              
    # Taken from launch_testing to not introduce a weird dependency thing            
    spec = importlib.util.spec_from_file_location(test_module_name, python_file_path)
    module = importlib.util.module_from_spec(spec)                                   
    spec.loader.exec_module(module)                                                  
    return module                                                                    
```

hidmic:
CI (up to `launch_testing`, `launch_testing_ros`, `test_communication`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8941/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4710/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7311/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8847)

hidmic:
Alright, all green. Going in!

PR #363, opened 2019-12-12
Update reference to example launch test file

jacobperron:


jacobperron:
@hidmic I've updated some other references and fixed some formatting. PTAL.

PR #365, opened 2019-12-13
Handle case where output buffer is closed during shutdown

### LINKED ISSUE #364: I/O operation on closed file in __on_process_stderr
## Bug report

**Required Info:**

- Operating System:
  - 18.04
- Installation type:
  - source
- Version or commit hash:
  - c644dfa8e1f1433e200b76eb52ba22f392838f48
- DDS implementation:
  - RTI Connect Micro, but I don't think this part matters
  - N/A

#### Steps to reproduce issue

I can get this to happen with the rcl test "test_rmw_impl_id_check" if also running stress on the system

```
terminal 1> stress -m 6 -c 6
terminal 2> launch_test build/rcl/test/test/test_rmw_impl_id_check__rmw_apex_dds_RelWithDebInfo.py --verbose
. . .
. . .
  File "/home/pete.baughman/gc/apex_ws/build/launch/launch/launch_service.py", line 377, in run_async
    await process_one_event_task
  File "/usr/lib/python3.6/asyncio/coroutines.py", line 126, in send
    return self.gen.send(value)
  File "/home/pete.baughman/gc/apex_ws/build/launch/launch/launch_service.py", line 287, in _process_one_event
  File "/usr/lib/python3.6/asyncio/coroutines.py", line 110, in __next__
    return self.gen.send(None)
  File "/home/pete.baughman/gc/apex_ws/build/launch/launch/launch_service.py", line 296, in __process_event
    entities = event_handler.handle(event, self.__context)
  File "/home/pete.baughman/gc/apex_ws/build/launch/launch/event_handlers/on_process_io.py", line 75, in handle
    return self.__on_stderr(event)
  File "/home/pete.baughman/gc/apex_ws/build/launch/launch/actions/execute_process.py", line 414, in __on_process_stderr
    self.__stderr_buffer.write(event.text.decode(errors='replace'))
ValueError: I/O operation on closed file
```

This happens when the execute_process object is trying to print 
```
[rcl]: Expected RMW implementation identifier of 'garbage' but instead found 'rmw_apex_dds', exiting with 102
```

#### Expected behavior
Launch Service should not die when trying to print to a buffer during shutdown

#### Actual behavior
It seems like this stderr message is racing with the shutdown of the launch service.

#### Additional information
  - I don't see where the __stderr_buffer gets closed yet - continuing to investigate
  - I'm also having trouble reproducing this in the nightly docker because I can't get rcl to build in there.  Continuing to investigate that too

pbaughman:
- Prevent crash during launch shutdown when a process IO event happens
    after the buffers have been closed

  - Use unbuffered output in that case so IO still has a chance of being seen

Fixes #364 

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -427,7 +427,13 @@ class ExecuteProcess(Action):
         self, event: ProcessIO
     ) -> Optional[SomeActionsType]:
         to_write = event.text.decode(errors='replace')
-        try:
+        if self.__stdout_buffer.closed:
+            # __stdout_buffer was probably closed by __flush_buffers on shutdown.  Output without
+            # buffering.
+            self.__stdout_logger.info(
+                self.__output_format.format(line=to_write, this=self)
+            )
+        else:
             self.__stdout_buffer.write(to_write)
             self.__stdout_buffer.seek(0)
             last_line = None
@@ -443,19 +449,18 @@ class ExecuteProcess(Action):
             self.__stdout_buffer.truncate(0)
             if last_line is not None:
                 self.__stdout_buffer.write(last_line)
-        except ValueError:
-            # __stdout_buffer was probably closed by __flush_buffers on shutdown.  Output without
-            # buffering.
-            self.__stdout_logger.info(
-                self.__output_format.format(line=to_write, this=self)
-            )
-
 
     def __on_process_stderr(
         self, event: ProcessIO
     ) -> Optional[SomeActionsType]:
         to_write = event.text.decode(errors='replace')
-        try:
+        if self.__stderr_buffer.closed:
+            # __stderr buffer was probably closed by __flush_buffers on shutdown.  Output without
+            # buffering.
+            self.__stderr_logger.info(
+                self.__output_format.format(line=to_write, this=self)
+            )
+        else:
             self.__stderr_buffer.write(to_write)
             self.__stderr_buffer.seek(0)
             last_line = None
@@ -471,13 +476,6 @@ class ExecuteProcess(Action):
             self.__stderr_buffer.truncate(0)
             if last_line is not None:
                 self.__stderr_buffer.write(last_line)
-        except ValueError:
-            # __stderr buffer was probably closed by __flush_buffers on shutdown.  Output without
-            # buffering.
-            self.__stderr_logger.info(
-                self.__output_format.format(line=to_write, this=self)
-            )
-
 
     def __flush_buffers(self, event, context):
         with self.__stdout_buffer as buf:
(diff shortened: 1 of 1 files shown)


pbaughman:
@ivanpauno Some code duplication, but fixes the crash on shutdown in the nightly docker image

dejanpan:
@wjwwood fyi and ask if we can get this in a little faster.

launch/launch/actions/execute_process.py
```diff
@@ -426,40 +426,58 @@ def __on_process_stdin(
...
+                else:
+                    last_line = line
+                    break
+            self.__stdout_buffer.seek(0)
+            self.__stdout_buffer.truncate(0)
+            if last_line is not None:
+                self.__stdout_buffer.write(last_line)
+        except ValueError:
+            # __stdout_buffer was probably closed by __flush_buffers on shutdown.  Output without
+            # buffering.
```
hidmic: @pbaughman why not checking `if self.__stdout_buffer.closed` instead of hoping `ValueError` was raised for the reasons you expect?

launch/launch/actions/execute_process.py
```diff
@@ -426,40 +426,58 @@ def __on_process_stdin(
...
+                else:
+                    last_line = line
+                    break
+            self.__stderr_buffer.seek(0)
+            self.__stderr_buffer.truncate(0)
+            if last_line is not None:
+                self.__stderr_buffer.write(last_line)
+        except ValueError:
+            # __stderr buffer was probably closed by __flush_buffers on shutdown.  Output without
+            # buffering.
```
hidmic: @pbaughman same as above.

pbaughman: @hidmic Yeah, I was coming at this from the approach of "this is a race condition" but this method is synchronized with the one that closes the buffers, right?  They're not going to get closed out from under us while the method is running?

pbaughman:
@hidmic I've addressed your comments

launch/launch/actions/execute_process.py
hidmic: It's `asyncio`, so no, that won't happen. There're no guarantees regarding execution order, and that's what probably caused this issue in the first place.

hidmic:
CI (up to `launch`, `test_launch_ros`, `test_communication`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8902/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4686/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7287/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8820)

pbaughman:
@hidmic The windows failure looks like a problem with building, not testing.

hidmic:
Alright, going in!

Issue #372, opened 2020-01-29
Error passing None target action to OnProcess* event handlers


jacobperron:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - Binaries or from source
- Version or commit hash:
  - Eloquent or 0822c9711e7d6543a8ba3361e3b5a8db41268c44
- DDS implementation:
  - N/A
- Client library (if applicable):
  - N/A

#### Steps to reproduce issue

Launch the following launch file with `ros2 launch`:
```python
import launch
import launch_ros

def generate_launch_description():
    talker_node = launch_ros.actions.Node(
        name='talker', package='demo_nodes_cpp', node_executable='talker', output='screen'
    )   
    listener_node = launch_ros.actions.Node(
        name='listener', package='demo_nodes_cpp', node_executable='listener', output='screen'
    )   

    return launch.LaunchDescription([
        # Register an event that starts the listener_node when a process starts
        launch.actions.RegisterEventHandler(
            launch.event_handlers.OnProcessStart(
                target_action=None,
                on_start=listener_node
            )   
        ),  
        # Launch the talker_node
        talker_node
    ])  
```

#### Expected behavior

The talker and listener nodes are started without any errors.

#### Actual behavior

We see the error:

```
AttributeError: 'str' object has no attribute 'perform'
```

Output with --debug option:

<details>
<summary>Debug</summary>
<pre>
[DEBUG] [launch.launch_context]: emitting event synchronously: 'launch.events.IncludeLaunchDescription'
[DEBUG] [launch.launch_context]: emitting event synchronously: 'launch.events.IncludeLaunchDescription'
[INFO] [launch]: All log files can be found below /home/jacob/.ros/log/2020-01-29-09-44-25-398369-warner-32334
[INFO] [launch]: Default logging verbosity is set to DEBUG
[DEBUG] [launch]: processing event: '<launch.events.include_launch_description.IncludeLaunchDescription object at 0x7f913b9fb3c8>'
[DEBUG] [launch]: processing event: '<launch.events.include_launch_description.IncludeLaunchDescription object at 0x7f913b9fb3c8>' ✓ '<launch.event_handlers.on_include_launch_description.OnIncludeLaunchDescription object at 0x7f9131659e48>'
[DEBUG] [launch]: processing event: '<launch.events.include_launch_description.IncludeLaunchDescription object at 0x7f912d5ba128>'
[...]

PR #375, opened 2020-02-03
Only allow ExecuteProcess actions to execute once

jacobperron:
It doesn't make sense to execute the same ExecuteProcess instance more than once and doing so
leads to uninformative errors. This change explicitly checks if the action has already been
executed and raises a runtime error with a more explicit error message.

Connected to #372

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -240,6 +240,8 @@ class ExecuteProcess(Action):
         self.__stdout_buffer = io.StringIO()
         self.__stderr_buffer = io.StringIO()
 
+        self.__executed = False
+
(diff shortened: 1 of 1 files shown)


jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9229/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4918/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7536/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9149/)
* Windows-container [Build Status](http://ci.ros2.org/job/ci_windows-container/30/)

PR #378, opened 2020-02-10
Add the ability to assert in stdout or stderr.

clalancette:
Since we are switching the underlying logging to be stderr,
this is going to be important for tests.  This commit
refactors assertInStdout() to be assertInStream(), which
takes in the stream to be asserted on.  It then recreates
assertInStdout() on top of that, and adds in assertInStderr().

It also changes assertWaitFor() to wait on the stderr stream
by default.

This goes along with https://github.com/ros2/rcutils/pull/196

### DIFF
--- a/launch_testing/launch_testing/asserts/assert_output.py
+++ b/launch_testing/launch_testing/asserts/assert_output.py
@@ -26,6 +26,84 @@ from ..tools.text import build_text_match
 from ..util import resolveProcesses
 
 
+def assertInStream(proc_output,
+                   expected_output,
+                   process,
+                   cmd_args=None,
+                   *,
+                   output_filter=None,
+                   strict_proc_matching=True,
+                   strip_ansi_escape_sequences=True,
+                   stream='stderr'):
+    """
+    Assert that 'output' was found in a stream of a process.
+
+    :param proc_output: The process output captured by launch_test.  This is usually injected
+    into test cases as self._proc_output
+    :type proc_output: An launch_testing.IoHandler
+
+    :param expected_output: The output to search for
+    :type expected_output: string or regex pattern or a list of the aforementioned types
+
+    :param process: The process whose output will be searched
+    :type process: A string (search by process name) or a launch.actions.ExecuteProcess object
+
+    :param cmd_args: Optional.  If 'process' is a string, cmd_args will be used to disambiguate
+    processes with the same name.  Pass launch_testing.asserts.NO_CMD_ARGS to match a proc without
+    command arguments
+    :type cmd_args: string
+
+    :param output_filter: Optional. A function to filter output before attempting any assertion.
+    :type output_filter: callable
+
+    :param strict_proc_matching: Optional (default True), If proc is a string and the combination
+    of proc and cmd_args matches multiple processes, then strict_proc_matching=True will raise
+    an error.
+    :type strict_proc_matching: bool
+
+    :param strip_ansi_escape_sequences: If True (default), strip ansi escape
+    sequences from actual output before comparing with the output filter or
+    expected output.
+    :type strip_ansi_escape_sequences: bool
+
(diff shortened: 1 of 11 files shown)


launch_testing/launch_testing/asserts/assert_output.py
```diff
@@ -77,9 +81,17 @@ def assertInStdout(proc_output,
...
+        if stream == 'stdout':
+            full_output = ''.join(
+                output.text.decode() for output in proc_output[proc] if output.from_stdout
+            )
+        elif stream == 'stderr':
+            full_output = ''.join(
+                output.text.decode() for output in proc_output[proc] if output.from_stderr
+            )
+        else:
+            raise ValueError("Invalid value for stream; must be 'stout' or 'stderr'")
```
hidmic: @clalancette 
```suggestion
            raise ValueError("Invalid value for stream; must be 'stdout' or 'stderr'")
```

launch_testing/launch_testing/asserts/assert_output.py
```diff
@@ -91,3 +103,91 @@ def assertInStdout(proc_output,
...
+def assertInStdout(proc_output,
+                   expected_output,
+                   process,
+                   cmd_args=None,
+                   *,
+                   output_filter=None,
+                   strict_proc_matching=True,
+                   strip_ansi_escape_sequences=True):
+    """
+    Assert that 'output' was found in the standard output of a process.
```
hidmic: @clalancette meta: I wonder if we should refer to the base function to avoid replicating docs.

launch_testing/launch_testing/asserts/assert_output.py
```diff
@@ -64,6 +65,9 @@ def assertInStdout(proc_output,
     sequences from actual output before comparing with the output filter or
     expected output.
     :type strip_ansi_escape_sequences: bool
+
+    :param stream: Which stream to examine.  This must be one of 'stderr' or 'stdout'.
```
hidmic: @clalancette meta: it's not currently implemented, but what if one wants to check both streams' output, in chronological order? Maybe a `'both'` option may help that use case, which is the supported one by process proxies (see [here](https://github.com/ros2/launch/blob/b99eb3811b80fe9c685151543b5dbb690a52674b/launch_testing/launch_testing/tools/process.py#L64-L82)).

launch_testing/launch_testing/asserts/assert_output.py
clalancette: Oops, will fix.

launch_testing/launch_testing/asserts/assert_output.py
clalancette: Yeah, I wasn't quite sure how to go about this.  I think I'll do what you suggest and just point the documentation to `assertInStream`.

launch_testing/launch_testing/asserts/assert_output.py
clalancette: It's certainly functionality that we can add, though I'd prefer not to add it until we run across a use-case that needs it.

hidmic: I'd generally agree with that, but the change is small and IMO it'll reduce the likelihood of users hacking their way forward if they ever want to do that. 

Having said that, I'm fine to move forward if you strongly prefer not to allow this patch to get bigger.

clalancette: Yeah, for now I'm going to err on the side of keeping this smaller (I already have to add one more thing here based on discussion from a couple of days ago).  If (probably when) a user opens an issue, it should be an easy addition.

wjwwood: The problem with "chronological" is that in my experience, since we'd be implementing this by capturing both and interleaving them (AFAIK), the order of messages is not deterministic.

Issue #263, opened 2019-06-13
Change `cmd` parameter in `ExecuteProcess`


ivanpauno:
#### Feature description
The current format of `cmd` is a little confusing. e.g.:
```python3
ExecuteProcess(['ls -las'])  # fails
ExecuteProcess(['ls -las'], shell=True)  # Ok
ExecuteProcess(['ls', '-las'])  # Ok
ExecuteProcess(['ls', '-las'], shell=True)  # Ok
```

That is maybe not of much interest, but the following case:
```python3
ExecuteProcess(['command', EnvironmentVariable('USER_ARGUMENTS')])
```
It's not way of specifying more than one argument in that environment variable, e.g.:
```bash
export USER_ARGUMENTS="-opt1 -opt2 -opt3"
```
That will make the program crash. The same problem applies for any substitution.
I don't see a possible workaround for that (without forcing shell=True).

My idea is to change the `cmd` type from:
```python3
cmd: Iterable[SomeSubstitutionsType]
```
to:
```python3
cmd: SomeSubstitutionsType
```

And use `shlex.split` after the substitutions are performed.

For backwards compatibility, we could accept both:
```python3
cmd: Union[SomeSubstitutionsType, Iterable[SomeSubstitutionsType]]
```

I realize about the problem while working in the launch frontend: https://github.com/ros2/launch/pull/226#discussion_r293036243.

ivanpauno:
@hidmic FYI

wjwwood:
That solution seems reasonable to me.

hidmic:
There's a bit of an issue with this change. We're essentially forcing shell-like parsing on otherwise plain command line arguments. And thus, something like `cmd=['ls', 'my/subdir/with spaces/']` would stop working the way it currently does. Same goes for prefixes, where we already are `shlex.split`'ing them. We probably need a flag for the user to explicitly opt-in (or opt-out) this parsing behavior.

ivanpauno:
> There's a bit of an issue with this change. We're essentially forcing shell-like parsing on otherwise plain command line arguments. And thus, something like `cmd=['ls', 'my/subdir/with spaces/']` would stop working the way it currently does. Same goes for prefixes, where we already are `shlex.split`'ing them. We probably need a flag for the user to explicitly opt-in (or opt-out) this parsing behavior.

I see. So, though it's true that it's not 100% backwards compatible, I think that the change is still reasonable and won't affect much users (and the ones affected should be able to fix the problem easily).

[...]

PR #379, opened 2020-02-14
Fix parsing of cmd line arguments in XML and yaml file

### LINKED ISSUE #337: [launch_frontend] FindPackage substitution not concatenated with surrounding text
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - source
- Version or commit hash:
  - 897cae8f6bf5602e250f9e521497b9eb70e14a85
- DDS implementation:
  - N/A
  - N/A

#### Steps to reproduce issue

Create a launch file with the following content and launch it.
```
<launch>
  <executable cmd="echo $(find-pkg-share rclcpp)/sub/path/foo" output="screen"/>
</launch>
```

#### Expected behavior

I would expect to see output like the following:

    [echo-2] /home/jacob/ros2_ws/install/rclcpp/share/rclcpp/sub/path/foo

#### Actual behavior

The substituted string for the share directory and the subsequent text, `/sub/path/foo`, are separated with a space:

    [echo-2] /home/jacob/ros2_ws/install/rclcpp/share/rclcpp /sub/path/foo

#### Additional information

There's the same behavior with `find-pkg-prefix` as well.

I've noticed that if we do the substitution in an `<arg>` tag, then it works as expected.
I'm not sure if tags other than `<executable>` are exhibiting the same bug.

#### Workaround

First do the substitution in an `<arg>` tag, for example:

```
<launch>
  <arg name="rclcpp_sub_dir" default="$(find-pkg-share rclcpp)/sub/path/foo" />
  <executable cmd="echo $(var rclcpp_sub_dir)" output="screen"/>
</launch>
```

---

I'm not sure where to look to resolve this issue. It's possible it should be fixed in `launch` (not `launch_ros`).
Maybe someone can point me in the right direction.

ivanpauno:
Fixes https://github.com/ros2/launch/issues/337.

The rules are a bit tricky, but I added a bunch of test cases. I will add any other suggested test case if needed.

I first though about solving the problem by addressing https://github.com/ros2/launch/issues/263.
But I later realized that:
- That change breaks API and it's not backportable. This fixes #337 in a backportable way.
- Using a string (`SomeSubstitutionsType`) for specifying commands, instead of a list (`List[SomeSubstitutionsType]`), makes concatenation of arguments more complex (e.g.: when passing args from `Node` to `ExecuteProcess`).

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -256,32 +256,52 @@ class ExecuteProcess(Action):
            list again as a `TextSubstitution`.
         :returns: a list of command line arguments.
         """
-        args = []
+        result_args = []
         arg = []
+
+        def _append_arg():
+            nonlocal arg
+            result_args.append(arg)
+            arg= []
         for sub in parser.parse_substitution(cmd):
             if isinstance(sub, TextSubstitution):
-                some_args = shlex.split(sub.text)
-                if not some_args:
+                tokens = shlex.split(sub.text)
+                if not tokens:
+                    # Sting with just spaces.
+                    # Appending args allow splitting two substitutions
+                    # separated by a space.
+                    # e.g.: `$(subst1 asd) $(subst2 bsd)` will be two separate arguments.
+                    _append_arg()
                     continue
                 if sub.text[0].isspace():
+                    # Needed for splitting from the previous argument
+                    # e.g.: `$(find-exec bsd) asd`
+                    # It splits `asd` from the path of `bsd` executable.
                     if len(arg) != 0:
-                        args.append(arg)
-                        arg = []
-                arg.append(TextSubstitution(text=some_args[0]))
-                if len(some_args) > 1:
-                    args.append(arg)
-                    arg = []
-                    arg.append(TextSubstitution(text=some_args[-1]))
-                if len(some_args) > 2:
-                    args.extend([TextSubstitution(text=x)] for x in some_args[1:-1])
+                        _append_arg()
+                arg.append(TextSubstitution(text=tokens[0]))
+                if len(tokens) > 1:
+                    # Needed to split the first argument when more than one token.
+                    # e.g. `$(find-pkg-prefix csd)/asd bsd`
+                    # will split `$(find-pkg-prefix csd)/asd` from `bsd`.
+                    _append_arg()
+                    arg.append(TextSubstitution(text=tokens[-1]))
+                if len(tokens) > 2:
+                    # If there are more than two tokens, just add all the middle tokens to
+                    # `result_args`.
+                    # e.g. `$(find-pkg-prefix csd)/asd bsd dsd xsd`
+                    # 'bsd' 'dsd' will be added.
+                    result_args.extend([TextSubstitution(text=x)] for x in tokens[1:-1])
                 if sub.text[-1].isspace():
-                    args.append(arg)
-                    arg = []
+                    # Allows splitting from next argument.
+                    # e.g. `exec $(find-some-file)`
+                    # Will split `exec` argument from the result of `find-some-file` substitution.
+                    _append_arg()
             else:
                 arg.append(sub)
         if arg:
-            args.append(arg)
-        return args
+            result_args.append(arg)
+        return result_args
 
     @classmethod
     def parse(
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -262,7 +262,7 @@ class ExecuteProcess(Action):
         def _append_arg():
             nonlocal arg
             result_args.append(arg)
-            arg= []
+            arg = []
         for sub in parser.parse_substitution(cmd):
             if isinstance(sub, TextSubstitution):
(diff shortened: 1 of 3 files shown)


launch/launch/actions/execute_process.py
```diff
@@ -257,14 +257,30 @@ def _parse_cmdline(
         :returns: a list of command line arguments.
         """
         args = []
```
jacobperron: It might help with readability if the name of this variable was more distinct from `arg`. Maybe `result_args`?

launch/launch/actions/execute_process.py
```diff
@@ -257,14 +257,30 @@ def _parse_cmdline(
...
-                text = shlex.split(text)
-                text = [TextSubstitution(text=item) for item in text]
-                args.extend(text)
+        arg = []
+        for sub in parser.parse_substitution(cmd):
+            if isinstance(sub, TextSubstitution):
+                some_args = shlex.split(sub.text)
+                if not some_args:
+                    continue
+                if sub.text[0].isspace():
```
jacobperron: I think a comment describing this if-block would be good. Same for the other blocks below.

launch/test/launch/frontend/test_substitutions.py
```diff
@@ -172,3 +190,48 @@ def test_eval_subst():
     expr = subst[0]
     assert isinstance(expr, PythonExpression)
     assert 'asdbsd' == expr.perform(LaunchContext())
+
+
+def expand_cmd_subs(cmd_subs: List[SomeSubstitutionsType]):
+    return [perform_substitutions_without_context(x) for x in cmd_subs]
+
+
+class MockedParser:
```
jacobperron: Just a suggestion,
```suggestion
class MockParser:
```

launch/test/launch/frontend/test_substitutions.py
```diff
@@ -172,3 +190,48 @@ def test_eval_subst():
...
+
+    def parse_substitution(self, value: Text) -> SomeSubstitutionsType:
+        return parse_substitution(value)
+
+
+def test_execute_process_parse_cmd_line():
+    """Test ExecuteProcess._parse_cmd_line."""
+    parser = MockedParser()
+
+    cmd_text: Text = '$(test path)/a/b/c asd csd $(test asd)/bsd/csd'
```
jacobperron: I suggest adding a test with two substitutions making up a single word, e.g.

```py
'$(test foo)$(test bar)'
```

launch/launch/actions/execute_process.py
```diff
@@ -257,14 +257,30 @@ def _parse_cmdline(
...
-                text = arg.text
-                text = shlex.split(text)
-                text = [TextSubstitution(text=item) for item in text]
-                args.extend(text)
+        arg = []
+        for sub in parser.parse_substitution(cmd):
+            if isinstance(sub, TextSubstitution):
+                some_args = shlex.split(sub.text)
+                if not some_args:
+                    continue
```
hidmic: @ivanpauno what if you have multiple substitutions separated with a space e.g. `$(find that) $(find this)`?

jacobperron: Sounds like a good unit test.

launch/launch/actions/execute_process.py
ivanpauno: I renamed variables, hope it's easier to read now.

launch/launch/actions/execute_process.py
ivanpauno: I added comments in all the conditions. I'm not sure if they clarify something, or if they make the code more obscure :rofl:.

launch/test/launch/frontend/test_substitutions.py
ivanpauno: Done

launch/test/launch/frontend/test_substitutions.py
ivanpauno: Added, thanks for the suggestion!

launch/launch/actions/execute_process.py
ivanpauno: Good catch! Added test case too.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9367/)

ivanpauno:
Fixed linter failures in https://github.com/ros2/launch/pull/379/commits/6e74c919eb8ca7679b4ca44894061da56800aaa4, the other failures are also happening in [nightlies](https://ci.ros2.org/view/nightly/job/nightly_linux_release/1447/#showFailuresLink).

> The code is a bit complex, but I don't have suggestions for making it better.

It would be possible to tweak the [grammar](https://github.com/ros2/launch/blob/master/launch/share/launch/frontend/grammar.lark) and add a different [Transformer](https://github.com/ros2/launch/blob/b99eb3811b80fe9c685151543b5dbb690a52674b/launch/launch/frontend/parse_substitution.py#L36) to get directly a list of arguments, instead of manually converting one thing to the other.

PR #390, opened 2020-02-25
Add delay parameter to retry_on_failure decorator

jacobperron:
Also added documentation for the parameters to the docstring.

Motivated by flaky tests in ros2action: https://github.com/ros2/ros2cli/pull/454

### DIFF
--- a/launch_testing/launch_testing/markers.py
+++ b/launch_testing/launch_testing/markers.py
@@ -26,12 +26,13 @@ def keep_alive(test_description):
     return test_description
 
 
-def retry_on_failure(*, times, delay=0.0):
+def retry_on_failure(*, times, delay=None):
     """
     Mark a test case to be retried up to `times` on AssertionError.
 
     :param times: The number of times to rety the test.
     :param delay: The time to wait between retries, in seconds.
+      A value of None will result in zero delay.
     """
     assert times > 0
 
@@ -51,7 +52,8 @@ def retry_on_failure(*, times, delay=0.0):
                     self._outcome.errors.clear()
                     self._outcome.success = True
                     n -= 1
-                time.sleep(delay)
+                if delay is not None:
+                    time.sleep(delay)
             return func(self, *args, **kwargs)
         return _wrapper
     return _decorator
--- a/launch_testing/launch_testing/markers.py
+++ b/launch_testing/launch_testing/markers.py
@@ -26,7 +26,7 @@ def keep_alive(test_description):
     return test_description
 
 
-def retry_on_failure(*, times, delay=0.1):
+def retry_on_failure(*, times, delay=0.0):
     """
     Mark a test case to be retried up to `times` on AssertionError.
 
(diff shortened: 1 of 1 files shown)


launch_testing/launch_testing/markers.py
```diff
@@ -25,8 +26,13 @@ def keep_alive(test_description):
     return test_description
 
 
-def retry_on_failure(*, times):
-    """Mark a test case to be retried up to `times` on AssertionError."""
+def retry_on_failure(*, times, delay=0.1):
```
wjwwood: Having `delay=0.1` changes the default behavior for this decorator, can you just use `0.0` or maybe `None` and do something like `if delay is not None: ...` instead? That makes approving this change easier for me.

jacobperron: Done 8ed2390

launch_testing/launch_testing/markers.py
```diff
@@ -45,6 +51,7 @@ def _wrapper(self, *args, **kwargs):
                     self._outcome.errors.clear()
                     self._outcome.success = True
                     n -= 1
+                time.sleep(delay)
```
dirk-thomas: A sleep of `0` might still change behavior since it might suspend the calling thread (not sure how that is implemented in Python).

jacobperron: True, if you think it's better we can default to `None` and avoid sleeping altogether in this case.

jacobperron: See 1c4be5f

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9440/)
* Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/5133/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7720/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9367/)
* Windows-container [Build Status](http://ci.ros2.org/job/ci_windows-container/212/)

jacobperron:
I'm almost certain that the compiler warning is a false positive coming from the git log of cyclonedds:

```
15:25:50     Fix warning -Wimplicit-int-float-conversion
15:25:50     
15:25:50     ```
15:25:50     /opt/ros/master/src/eclipse-cyclonedds/cyclonedds/src/tools/pubsub/common.c:586:28: warning: implicit conversion from 'long' to 'double' changes value from 9223372036854775807 to 9223372036854775808 [-Wimplicit-int-float-conversion]
15:25:50         if(nanosec > nextafter(INT64_MAX, 0)) {
15:25:50                      ~~~~~~~~~ ^~~~~~~~~
15:25:50     /usr/include/stdint.h:134:22: note: expanded from macro 'INT64_MAX'
15:25:50                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
15:25:50     /usr/include/stdint.h:116:24: note: expanded from macro '__INT64_C'
15:25:50                             ^~~~~~
15:25:50     <scratch space>:345:1: note: expanded from here
15:25:50     9223372036854775807L
15:25:50     ^~~~~~~~~~~~~~~~~~~~
15:25:50     1 warning generated.
15:25:50     ```
```

:smile:

Issue #400, opened 2020-04-11
[Launch Testing] failure messages are generic, unuseful


rotu:
In ros2cli, there are tests that command line tools produce expected output. e.g. https://github.com/ros2/ros2cli/blob/96cf5b949d1aa870111bba0664eca1c359d3307c/ros2lifecycle/test/test_cli.py#L363-L374

notably this command which uses `expect_output`:
```
        assert launch_testing.tools.expect_output(
            expected_lines=['unconfigured [1]'],
            text=lifecycle_command.output,
            strict=True
        )
```

When these tests fail, they are reported in CI with an extremely generic error message, stating only the test case and its parameter:

https://ci.ros2.org/view/nightly/job/nightly_osx_debug/1588/testReport/ros2lifecycle.src.ros2.ros2cli.ros2lifecycle.test/test_cli/test_cli/

> ## Error Message
> some test cases have failed: test_cli[rmw_cyclonedds_cpp] failed at TestROS2LifecycleCLI.test_list_all_lifecycle_hidden_node_transitions_with_hidden_flag[rmw_cyclonedds_...
> test_cli[rmw_cyclonedds_cpp] failed at TestROS2LifecycleCLI.test_list_all_lifecycle_node_transitions[rmw_cyclonedds_cpp]
> 
> ## Stacktrace
> some test cases have failed:
> test_cli[rmw_cyclonedds_cpp] failed at TestROS2LifecycleCLI.test_list_all_lifecycle_hidden_node_transitions_with_hidden_flag[rmw_cyclonedds_cpp]
> ...
> test_cli[rmw_cyclonedds_cpp] failed at TestROS2LifecycleCLI.test_list_all_lifecycle_node_transitions[rmw_cyclonedds_cpp]

rotu:
@dirk-thomas I notice more verbose error output is present in the full text log:
```
----------------------------- Captured stderr call -----------------------------
test_count_all_lifecycle_nodes[rmw_cyclonedds_cpp] (test_cli.TestROS2LifecycleCLI) ... ok
test_count_lifecycle_nodes[rmw_cyclonedds_cpp] (test_cli.TestROS2LifecycleCLI) ... ok
test_get_hidden_lifecycle_node_state[rmw_cyclonedds_cpp] (test_cli.TestROS2LifecycleCLI) ... FAIL
test_get_lifecycle_node_state[rmw_cyclonedds_cpp] (test_cli.TestROS2LifecycleCLI) ... FAIL
[...]

PR #404, opened 2020-04-17
Improve jUnit output for launch tests when run with py.test

hidmic:
Addresses #400. I'll start by saying that while I'm not super happy with this patch -- it relies a bit too much on `pytest` internals --, it's the one I could come up with without undergoing major surgery.

CI (up to `launch_testing` and `ros2topic`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10291/)

### DIFF
--- a/launch_testing/launch_testing/pytest/hooks.py
+++ b/launch_testing/launch_testing/pytest/hooks.py
@@ -15,6 +15,8 @@
 import copy
 from unittest import TestCase
 
+from _pytest._code.code import ReprFileLocation
+
 import pytest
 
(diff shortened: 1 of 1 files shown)


hidmic:
> This should really be revisited - so adding a TODO / follow up ticket would be good.

I agree. I tried hard to make this better, but I was unable to without turning `launch_testing` into a `pytest` plugin. I'll open an issue to track future shim development and/or integration efforts.

hidmic:
See https://github.com/ros2/launch/issues/405. I'll merge this now.

Issue #387, opened 2020-02-19
DeprecationWarning: The loop argument is deprecated since Python 3.8


dirk-thomas:
And scheduled for removal in Python 3.10.

E.g. https://github.com/ros2/launch/blob/f243fc6b3bcd276c194151e6933cda90761b1944/launch/launch/actions/timer_action.py#L84

dirk-thomas:
Instead of creating separate tickets for each deprecation warning I will add more here:

* `launch_testing/test/launch_testing/test_launch_test_runner_validation.py:15`:

  > DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses

zmk5:
I submitted a fix for the second deprecation warning. I'm still trying to figure out what the deprecation in the first one is.

zmk5:
Just checked. It actually didn't work for me. I'll try and fix the issue.

zmk5:
Hey @dirk-thomas I have resolved all deprecation warnings associated with this issue. The team has also reviewed and merged the changes. You can close this issue if you would like 😄

ivanpauno:
Thanks for your contribution @zmk5 !!!

PR #409, opened 2020-04-28
removed deprecated loop parameter call (#387)

zmk5:
According to the [documentation](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait). Two things are deprecated by Python 3.10: the use of coroutines and the `loop` parameter. Since the first parameter is an `asyncio.Future`, we don't have to worry about the first one. For the second, I just removed the use of the `loop` parameter and it passed all the `pytest` unit tests. If merged, I believe this should close issue #387 .


zmk5:
Apologies I misread your original issue. You want **all** the deprecated uses of `loop` to be removed. I'll fix those and add them to the rest.

PR #407, opened 2020-04-27
fixed depcrecation warning of imp to importlib (issue #387)

zmk5:
This is a fix for the second deprecation warning that dirk found in issue #387 . I'm still trying to figure out the first.

PR #410, opened 2020-04-28
removed deprecated loop parameter call (#387)

zmk5:
According to the [documentation](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait). Two things are deprecated by Python 3.10: the use of coroutines and the `loop` parameter. Since the first parameter is an `asyncio.Future`, we don't have to worry about the first one. For the second, I just removed the use of the `loop` parameter.

The other deprecation warning raised during testing comes from the use of `asyncio.Queue()`. The [docs](https://docs.python.org/3/library/asyncio-queue.html#asyncio.Queue) state that the loop parameter is deprecated, so it can be removed without issue. Finally, the last deprecation warning came from the use of `asyncio.wait()` that tried to use `loop=` parameter and was simply removed.

Removing these still had `launch` pass all unit testing with `pytest`. If merged, I believe this should close issue #387 .

### DIFF
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -174,7 +174,7 @@ class LaunchService:
                 # Set the asyncio loop for the context.
                 self.__context._set_asyncio_loop(this_loop)
                 # Recreate the event queue to ensure the same event loop is being used.
-                new_queue = asyncio.Queue(loop=this_loop)
+                new_queue = asyncio.Queue()
                 while True:
                     try:
                         new_queue.put_nowait(self.__context._event_queue.get_nowait())
@@ -351,7 +351,6 @@ class LaunchService:
                         while not done:
                             done, pending = await asyncio.wait(
                                 entity_futures,
-                                loop=this_loop,
                                 timeout=1.0,
                                 return_when=asyncio.FIRST_COMPLETED
                             )
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -158,7 +158,7 @@ class LaunchService:
         return number_of_entity_future_pairs == 0 and self.__context._event_queue.empty()
 
     @contextlib.contextmanager
-    def _prepare_run_loop(self, loop):
+    def _prepare_run_loop(self):
         try:
             # Acquire the lock and initialize the loop.
             with self.__loop_from_run_thread_lock:
@@ -166,7 +166,7 @@ class LaunchService:
                     raise RuntimeError(
                         'LaunchService cannot be run multiple times concurrently.'
                     )
-                this_loop = asyncio.get_event_loop() if loop is None else loop
+                this_loop = asyncio.get_event_loop()
 
                 if self.__debug:
                     this_loop.set_debug(True)
@@ -299,7 +299,7 @@ class LaunchService:
                 #     'launch.LaunchService',
                 #     "processing event: '{}' x '{}'".format(event, event_handler))
 
-    async def run_async(self, *, shutdown_when_idle=True, loop=None) -> int:
+    async def run_async(self, *, shutdown_when_idle=True) -> int:
         """
         Visit all entities of all included LaunchDescription instances asynchronously.
 
@@ -318,7 +318,7 @@ class LaunchService:
             )
 
         return_code = 0
-        with self._prepare_run_loop(loop) as (this_loop, this_task):
+        with self._prepare_run_loop() as (this_loop, this_task):
             # Log logging configuration details.
             launch.logging.log_launch_config(logger=self.__logger)
 
@@ -388,7 +388,7 @@ class LaunchService:
         """
         loop = osrf_pycommon.process_utils.get_loop()
         run_async_task = loop.create_task(self.run_async(
-            shutdown_when_idle=shutdown_when_idle, loop=loop
+            shutdown_when_idle=shutdown_when_idle
         ))
         loop.run_until_complete(run_async_task)
         return run_async_task.result()
--- a/launch_testing/test/launch_testing/test_launch_test_runner_validation.py
+++ b/launch_testing/test/launch_testing/test_launch_test_runner_validation.py
@@ -24,7 +24,8 @@ from launch_testing.test_runner import LaunchTestRunner
 
 
 def make_test_run_for_dut(generate_test_description_function):
-    module = importlib.util.module_from_spec('test_module')
+    test_spec = importlib.util.spec_from_loader('test_module', loader=None)
+    module = importlib.util.module_from_spec(test_spec)
     module.generate_test_description = generate_test_description_function
     return LoadTestsFromPythonModule(module)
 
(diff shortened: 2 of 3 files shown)


launch_testing/test/launch_testing/test_launch_test_runner_validation.py
```diff
@@ -24,7 +24,8 @@
 
 
 def make_test_run_for_dut(generate_test_description_function):
-    module = imp.new_module('test_module')
+    test_spec = importlib.util.spec_from_loader('test_module', loader=None)
+    module = importlib.util.module_from_spec(test_spec)
```
dirk-thomas, line 28: This change has already been merged in #407 and should therefore not be part of this PR anymore.

zmk5: Didn't even notice this was still in there. I will remove this fix from the PR.

hidmic:
> We should delete the loop parameter from our API too, e.g. `run_async` and `_prepare_run_loop`.

I agree.

---

On a bit of a broader note, skimming through the code I see `asyncio` event loops being used and passed around as if these were local instances, while `asyncio` API in general, and this deprecation in particular, suggest event loops are bound to OS threads. I can only guess `launch` is how it is because we intended to isolate `launch` from client code using `asyncio` and/or to furnish `launch` API with some degree of thread-safety (I can see a few locks here and there). The fact that it only partially achieves either of those goals makes me think this is somewhat brittle. And I wonder if being more transparent about `launch` leveraging (non thread-safe) `asyncio` and letting multi-threaded code use [standard API](https://docs.python.org/3/library/asyncio-task.html#asyncio.run_coroutine_threadsafe) to interact with the loop wouldn't be less prone to errors and make for an easier integration of `launch` in an `async` ecosystem. @wjwwood for thoughts.

ivanpauno:
> On a bit of a broader note, skimming through the code I see `asyncio` event loops being used and passed around as if these were local instances, while `asyncio` API in general, and this deprecation in particular, suggest event loops are global (actually bound to OS threads).

I think we still need to keep track of the loop being used, because for some part of the API it's needed.
Particularly, it's needed when interacting with a loop running in another `thread` (e.g.: here https://github.com/ros2/launch/blob/fb2549ca343f7250741498ab32381b6d0a6a5d60/launch/launch/launch_service.py#L115-L117).
We shouldn't provide a `loop` parameter in the API though.

> I can only guess launch is how it is because we intended to isolate launch from client code using asyncio and/or to furnish launch API with some degree of thread-safety (I can see a few locks here and there). The fact that it only partially achieves either of those goals makes me think this is somewhat brittle. And I wonder if being more transparent about launch leveraging (non thread-safe) asyncio and letting multi-threaded code use standard API to interact with the loop wouldn't be less prone to errors and make for an easier integration of launch in an async ecosystem. @wjwwood for thoughts.

I'm not super sure how the result of this proposal would look, so I don't have an opinion.

hidmic:
> Particularly, it's needed when interacting with a loop running in another thread

Correct, but in a single place. And there's specific `asyncio` API to deal with that case. IMHO wrapping it so that a small fraction of the API can be thread-safe (while none of the rest nor most of its implementation is) results in false expectations and fun surprises (see #126 or dig up older discussions about `launch_testing`).

hidmic:
Anyways, @zmk5 once @ivanpauno last request is fulfilled, I think this will be good to go too.

zmk5:
Sure, I can remove the loop parameter from the function calls where I removed the use of `loop`. I will have to check that these functions are not being called elsewhere in package for tests or other scripts, etc.

zmk5:
Removed the remaining uses of `loop` within the `launch_service.py` API functions. `pytest` returned 100% success after I made these changes locally.

wjwwood:
I think the loop argument was also because on Windows the standard global loop was not compatible with the subprocess part of asyncio, so you need to make sure you used that loop instead. See:

https://github.com/osrf/osrf_pycommon/blob/ae04fa12f02e1ae84d68fc7c0ad4ee28b49efc93/osrf_pycommon/process_utils/get_loop_impl.py#L21-L36

I don't know if these changes break any of that or if that is no longer needed on Windows. But someone should investigate that before we remove their use...

wjwwood:
I guess that `asyncio.set_event_loop()` takes care of that, but we should make sure.

zmk5:
Hey William, I'll try and figure out a way to test this on Windows and get back to you with the results from `pytest`. I just need to figure out how to get a windows license 🤣

wjwwood:
I can run our CI which has Windows as part of it, but I only hope any issues show up from the testing. I guess they would, but I'm not 100% sure, as it might only be an issue in corner cases we don't test properly.

zmk5:
Sounds good! I'll look into the link you posted and verify if the standard global loop in windows is fixed.

wjwwood:
Full CI with Fast-RTPS only and the test argument `--packages-above launch`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10569/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6021/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8608/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10467/)

zmk5:
@wjwwood I ran my changes on both Ubuntu 20.04 and macOS Catalina. Both passed the `pytest` unit tests provided by the package.

zmk5:
These seemed to pass the CI tests. It seems, the instability in the windows build is do to something with `tf`.

launch/launch/launch_service.py
```diff
@@ -389,7 +388,7 @@ def run(self, *, shutdown_when_idle=True) -> int:
         """
         loop = osrf_pycommon.process_utils.get_loop()
         run_async_task = loop.create_task(self.run_async(
```
ivanpauno, line 390: Shouldn't the loop be set here? with `asyncio.set_event_loop()`

ivanpauno:
> I think the loop argument was also because on Windows the standard global loop was not compatible with the subprocess part of asyncio, so you need to make sure you used that loop instead. See:
> 
> https://github.com/osrf/osrf_pycommon/blob/ae04fa12f02e1ae84d68fc7c0ad4ee28b49efc93/osrf_pycommon/process_utils/get_loop_impl.py#L21-L36
> 
> I don't know if these changes break any of that or if that is no longer needed on Windows. But someone should investigate that before we remove their use...

It seems that `ProactorEventLoop` is being used by [default on Windows](https://docs.python.org/3/library/asyncio-eventloop.html#event-loop-implementations), and that's why the missing `set_event_loop` didn't affect behavior.

Continue using `osrf_pycommon.process_utils.get_loop()` or not is a discussion for another PR, for now, we should make sure we're setting that loop.

ivanpauno:
Failures are unrelated, merging!

wjwwood:
CI looks fine, those issues are known already.

PR #414, opened 2020-05-06
Fix up parser.py

rotu:
1. *always* pass a file object to parser implementation. Passing a string introduces no extra expressiveness and the file is not guaranteed to be properly closed.
2. Use more Pythonic type detection for whether passed object is a string or a file.
3. Clean up type declarations
4. Clean up parser ordering logic
5. Make parser ordering deterministic
6. Remove use of Ellipsis. It makes the declaration look like a stub.
Signed-off-by: Dan Rose <dan@digilabs.io>

### DIFF
--- a/launch/launch/frontend/parser.py
+++ b/launch/launch/frontend/parser.py
@@ -1,4 +1,5 @@
 # Copyright 2019 Open Source Robotics Foundation, Inc.
+# Copyright 2020 Rover Robotics, c/o Dan Rose
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -14,15 +15,16 @@
 
 """Module for Parser class and parsing methods."""
 
-import io
-import os
+import os.path
 from typing import Any
 from typing import List
 from typing import Optional
 from typing import Text
+from typing import TextIO
 from typing import Tuple
 from typing import Type
 from typing import Union
+from typing import TYPE_CHECKING
 
 from pkg_resources import iter_entry_points
 
@@ -33,21 +35,21 @@ from .parse_substitution import replace_escaped_characters
 from ..action import Action
 from ..invalid_launch_file_error import InvalidLaunchFileError
 from ..some_substitutions_type import SomeSubstitutionsType
-from ..utilities import is_a
 
 interpolation_fuctions = {
     entry_point.name: entry_point.load()
     for entry_point in iter_entry_points('launch.frontend.interpolate_substitution_method')
 }
 
-if False:
-    from ..launch_description import LaunchDescription  # noqa: F401
+
+if TYPE_CHECKING:
+    from ..launch_description import LaunchDescription
 
 
 class InvalidFrontendLaunchFileError(InvalidLaunchFileError):
     """Exception raised when the given frontend launch file is not valid."""
 
-    ...
+    pass
 
 
 class Parser:
@@ -132,7 +134,7 @@ class Parser:
     @classmethod
     def load(
         cls,
-        file: Union[str, io.TextIOBase],
+        file: Union[Text, TextIO],
     ) -> (Entity, 'Parser'):
         """
         Parse an Entity from a markup language-based launch file.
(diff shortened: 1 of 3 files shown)


rotu:
If this should be split into multiple PRs, please suggest how to split it up.

launch/launch/utilities/file_path.py
```diff
@@ -0,0 +1,19 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import typing
+
+# Type of a filesystem path
+FilePath = typing.Union[str, bytes, os.PathLike]
```
ivanpauno, line 19: I would rename this as `FilePathType`.

ivanpauno:
> I presume this includes #415 

I think that improvements in #415 are orthogonal to the ones here.

hidmic:
> I think that improvements in #415 are orthogonal to the ones here.

Same pattern, different packages. You're right, my bad. On a second look, my comment over there was addressed here.

launch/launch/frontend/parser.py
```diff
@@ -1,4 +1,5 @@
 # Copyright 2019 Open Source Robotics Foundation, Inc.
+# Copyright 2020 Rover Robotics, c/o Dan Rose
```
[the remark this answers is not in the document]
rotu: I mean we don't even need a copyright statement to claim a copyright. The only purpose is to track down the copyright holder, and for authorship (vanity). The c/o serves both purposes.

There definitely are copyrightable elements to this change. We presumptively assume copyrightable elements in *every* PR (hence the DCO). I don't think we should have a per-file copyright notice, but if we have one, it makes sense to keep it as complete as possible.

I asked this question on answers but @tfoote closed it as off topic so... who knows? https://answers.ros.org/question/349677/when-should-i-add-a-copyright-at-the-top-of-a-source-file/

launch/launch/utilities/file_path.py
rotu: I don't like that because none of the type annotation types have `Type` in them, to keep them succinct. I did make it a type var to better match e.g. `typing.AnyStr`.

launch/launch/frontend/parser.py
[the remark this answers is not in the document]
ivanpauno: > I mean we don't even need a copyright statement to claim a copyright. The only purpose is to track down the copyright holder, and for authorship

The propose of the copyright note is to track down copyright holders, not authors.
Please delete the author from the note.

> I don't think we should have a per-file copyright notice, but if we have one, it makes sense to keep it as complete as possible.

Adding Rover Rovotics copyright note is ok.

launch/launch/utilities/file_path.py
ivanpauno: > I don't like that because none of the type annotation types have Type in them, to keep them succinct. I did make it a type var to better match e.g. typing.AnyStr.

I think that in the case of `typing` is clear, because that module only has type annotation types.
Now, the file has `_type` suffix in the name, so it's clear that it only contains a type definition.

If you also would like to add `FilePath` to `utilities` `__all__` dictionary, please rename it in that case as `FilePathType` (or re-export a renamed version).

----

You can omit this last suggestion if you want.

launch/launch/frontend/parser.py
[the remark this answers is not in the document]
dirk-thomas: > Newer copyright notes should be added on top of old ones.

Sorry for the confusion but the opposite is the case. Please move the new copyright line back below the existing one.

launch/launch/utilities/typing_file_path.py
```diff
@@ -0,0 +1,19 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import typing
+
+# Type of a filesystem path
+FilePath = typing.TypeVar('FilePath', str, bytes, os.PathLike)
```
dirk-thomas, line 19: A separate file for a single type which also only used in a single file once seems unnecessary to me.

launch/launch/frontend/parser.py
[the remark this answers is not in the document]
ivanpauno: Sorry, I got confused on how the rule was.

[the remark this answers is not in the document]
rotu: That should be documented somewhere...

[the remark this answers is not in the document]
dirk-thomas: > That should be documented somewhere...

Please consider to contribute to the developer guide: https://github.com/ros2/ros2_documentation/blob/master/source/Contributing/Developer-Guide.rst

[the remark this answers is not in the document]
rotu: Copyright notices in open source are so often unmaintained that they can't generally be trusted. For this reason, I'm not comfortable documenting them further in the dev guide.

sloretz:
@dirk-thomas @ivanpauno where is this PR at? Is it ready for CI? Is there feedback that still needs to be addressed?

rotu:
I need to fix the copyrights to say “Open Avatar Inc.” instead of “Rover Robotics ℅ Dan Rose” since Rover Robotics is a dba. I’ll move it out of draft again when I get around to that

ivanpauno:
@jacobperron ok to merge this before Foxy relase? should be merge after the release?

jacobperron:
@ivanpauno Unless you think it's crucial, I'd rather hold until after the rosdistro freeze (May 27th).

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11040/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6365/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8991/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10963/)

launch/launch/frontend/parser.py
```diff
@@ -14,15 +15,16 @@
...
+import os.path
 from typing import Any
 from typing import List
 from typing import Optional
 from typing import Text
+from typing import TextIO
 from typing import Tuple
 from typing import Type
 from typing import Union
+from typing import TYPE_CHECKING
```
ivanpauno, line 27: flake8 isn't passing in some of the platforms because this is not correctly ordered (and because of other unrelated errors).
Can you reorder? Thanks!

launch/launch/utilities/typing_file_path.py
dirk-thomas: This comment is still pending.

rotu: Files are cheap and there is not a better place to put it. This code is written as intended.

launch/launch/frontend/parser.py
rotu: Done. I didn't see this because https://github.com/ament/ament_lint/issues/254

rotu:
Blech. I struggle with Git sometimes, but I think this is now ready for merge.

ivanpauno:
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6444/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9100/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11075/)

ivanpauno:
Failing tests are unrelated, merging!
Thanks for the fix @rotu

PR #415, opened 2020-05-06
Close YAML file when we're done.

rotu:
Signed-off-by: Dan Rose <dan@digilabs.io>


launch_yaml/launch_yaml/parser.py
```diff
@@ -30,13 +31,22 @@ class Parser(frontend.Parser):
...
-        if len(tree) != 1:
-            raise RuntimeError('Expected only one root')
-        type_name = list(tree.keys())[0]
-        return (Entity(tree[type_name], type_name), cls())
+        try:
+            fileobj = open(file, 'r')
+            didopen = True
+        except TypeError:
+            fileobj = file
+            didopen = False
```
hidmic, line 42: @rotu I'm not very fond of using exceptions for flow control. Why not checking?

rotu: Using exceptions for expected circumstances may not be natural in C++ but it's [idiomatic in Python](https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/) so long as exceptions are specific and try-blocks are tight.

The reason here is because file can be a path-like object. Sure we're only expecting a `str` but it would be sensible to pass a `pathlib.Path` or `os.PathLike` and this code does the correct thing in all those cases.

hidmic: > The reason here is because file can be a path-like object. Sure we're only expecting a str but it would be sensible to pass a pathlib.Path or os.PathLike and this code does the correct thing in all those cases.

Consider updating type annotations if so.

hidmic:
@rotu ping.

rotu:
converting to draft to wait until #414 is done

ivanpauno:
@rotu friendly ping

rotu:
Thanks, @ivanpauno

launch_yaml/launch_yaml/parser.py
```diff
@@ -29,14 +31,23 @@ class Parser(frontend.Parser):
 
     @classmethod
     def load(
-        cls,
-        stream: Union[str, io.TextIOBase],
+            cls,
+            file: Union[FilePath, TextIO],
```
ivanpauno, line 35: this seems to be over-indented

rotu: oops

ivanpauno:
Rci job is passing, merging!
Thanks for the fix @rotu!

PR #421, opened 2020-05-13
Fix launch warnings

clalancette:
This PR fixes a bunch of warnings in launch.

1.  pylint on Focal complains that:
```
./launch_testing/legacy/output.py:46:31: E741 ambiguous variable name 'l'
    return [str.encode(l) for l in rmw_output_filter.splitlines()]
```
2.  pytest on Focal and Bionic complains that:
```
  Warning: direct construction of LaunchTestModule has been deprecated, please use LaunchTestModule.from_parent
```
3.  Launch itself complains that:
```
  Warning: Automatically adding attributes like self.proc_info to the test class will be deprecated in a future release.  Instead, add proc_info to the test method argument list to access the test object you need
```
(this is fixed in several tests)
4.  pytest on Bionic complains that:
```
  Warning: The 'junit_family' default value will change to 'xunit2' in pytest 6.0.
  Add 'junit_family=xunit1' to your pytest.ini file to keep the current format in future versions of pytest and silence this warning.
```
(I'm less sure on this one whether our infrastructure wants xunit1 or xunit2, to be honest).

With all of these fixes in place, I get green CI again.  Note that the first one overlaps with #420 , so I'll just rebase after that lands.


dirk-thomas:
This overlaps with #420.

clalancette:
> This overlaps with #420.

Yep, pointed out in the initial description.  Only the first one does.

launch_testing/pytest.ini
```diff
@@ -3,3 +3,4 @@
 testpaths = test
 # Add arguments for launch tests
 addopts = --launch-args dut_arg:=test
+junit_family=xunit1
```
dirk-thomas, line 6: This is not compatible with ci.ros2.org: https://github.com/ros2/ci/blob/a7c41fbce3f1acee2a98a0e2865a8c9a0fb0db8a/ros2_batch_job/__main__.py#L381

clalancette: Ah, thanks for the feeback.  Will set to xunit2.

dirk-thomas: CI already sets the flag automatically. So why do you want to set it explicitly (which would be the only package which does that)?

clalancette: Because otherwise there's a warning from CI:

```
  Warning: The 'junit_family' default value will change to 'xunit2' in pytest 6.0.
  Add 'junit_family=xunit1' to your pytest.ini file to keep the current format in future versions of pytest and silence this warning.
```

I *think* this needs a custom pytest.ini to add options, but I really don't know.  @ivanpauno @hidmic any thoughts here?

ivanpauno: I think that if the warning is silenced by explicitly specifying the `junit_familiy` here and CI passes, it's ok this approach (besides that CI already sets the junit family).

hidmic: > I think this needs a custom pytest.ini to add options

Correct. 

> Because otherwise there's a warning from CI:

So you think the existence of `pytest.ini` itself triggers the warning?

dirk-thomas: Can you please reference the build where you get this warning. On ci.ros2.org this should happened because of ros2/ci#406.

dirk-thomas: > I think that if the warning is silenced by explicitly specifying the `junit_familiy` here and CI passes, it's ok this approach (besides that CI already sets the junit family).

The question is *why* it is necessary since the CI logic tries to workaround this already. And if it is necessary I would assume it is the same for *all* packages and then should also be addressed for all.

clalancette: Oh, right.  This doesn't happen in CI because of the override.  I only saw it *locally*, which is why we need it.

dirk-thomas: >  I only saw it locally, which is why we need it.

Then please apply it in a separate pull request to *all* packages.

clalancette: > Then please apply it in a separate pull request to _all_ packages.

I really don't see why.  This improves it for this package; adding it to other packages to fix it there is orthogonal.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10755/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6150/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8757/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10645/)

clalancette:
All right, green CI and approval on both this and ros2/launch_ros#150.  I'm going to merge these; we can follow up with more pytest warning fixes in a follow-up.

Issue #287, opened 2019-07-25
Support for respawn in `ExecuteProcess`


ivanpauno:
## Feature request

Add support for `respawn` and `respawn_delay`.

#### Feature description

ROS 1 supported those attributes in `node` tag. We could have something similar in `ExecuteProcess` action. The behavior could be emulated with event handlers, but having an argument in the constructor for that may be easier to use.

There are some point of discussions:
- We could just implement it for the frontend implementations (XML, YAML), and avoid adding an argument to `ExecuteProcess` constructor.
- Should event handlers targeting the original process target the new one?

Previous discussion: https://github.com/ros2/ros2_documentation/pull/302#discussion_r306309349

ivanpauno:
@wjwwood @hidmic FYI.

> We could just implement it for the frontend implementations (XML, YAML), and avoid adding an argument to ExecuteProcess constructor.

I prefer implementing it in `ExecuteProcess`.

> Should event handlers targeting the original process target the new one?

I think they should. I don't see an use case were that isn't desired.

mkhansen-intel:
+1 for getting this feature, is it planned for anytime soon?

wjwwood:
> +1 for getting this feature, is it planned for anytime soon?

I think the best assumption is no. Maybe @hidmic or @ivanpauno were planning on doing it, but I'm full booked up until well after ROSCon.

wjwwood:
> It would also be nice if we could signal somehow that the node has been respawned to the lifecycle_manager node so that it could re-activate any node that crashes. Not to digress too far, but is there a mechanism for doing something like that?

Assuming it knows the node's name, it could look for the node to enter the unconfigured state (or whatever the first lifecycle event published is) on the event topic for that node. Launch does the same thing in order to produce launch events for each state transition.

ivanpauno:
> +1 for getting this feature, is it planned for anytime soon?

I'm not planning to work on this in a near future.
If someone wants to work on it, I can help doing reviews.

mkhansen-intel:
> Assuming it knows the node's name, it could look for the node to enter the unconfigured state (or whatever the first lifecycle event published is) on the event topic for that node. Launch does the same thing in order to produce launch events for each state transition.

That makes sense. We'll discuss that as a potential way of doing this. It won't matter until the 'respawn' capability is available though.

Jconn:
has anything changed here? This is a nice to have for us

iuhilnehc-ynos:
[...]

PR #426, opened 2020-06-05
Add respawn and respawn_delay support

iuhilnehc-ynos:
This PR is related to https://github.com/ros2/launch/issues/287

Signed-off-by: Chen Lihui <lihui.chen@sony.com>

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -406,7 +406,8 @@ class ExecuteProcess(Action):
             # Do not handle shutdown more than once.
             return None
         self.__shutdown_received = True
-        self.__shutdown_future.set_result(None)
+        if self.__shutdown_future:
+            self.__shutdown_future.set_result(None)
         if self.__completed_future is None:
             # Execution not started so nothing to do, but self.__shutdown_received should prevent
             # execution from starting in the future.
@@ -750,9 +751,10 @@ class ExecuteProcess(Action):
             self.__logger.error("process has died [pid {}, exit code {}, cmd '{}'].".format(
                 pid, returncode, ' '.join(cmd)
             ))
+        await context.emit_event(ProcessExited(returncode=returncode, **process_event_args))
+        if returncode != 0:
             # respawn the process that abnormally died
             if not self.__shutdown_received and not context.is_shutdown and self.__respawn:
-                respawn_flag = True
                 if self.__respawn_delay > 0.0:
                     # wait for a timeout(`self.__respawn_delay`) to respawn the process
                     # and handle shutdown event with future(`self.__shutdown_future`)
@@ -762,16 +764,9 @@ class ExecuteProcess(Action):
                         loop=context.asyncio_loop,
                         return_when=asyncio.FIRST_COMPLETED
                     )
-                    if self.__shutdown_future.done():
-                        respawn_flag = False
-                if respawn_flag:
-                    await context.emit_event(
-                        ProcessExited(returncode=returncode, **process_event_args)
-                    )
+                if not self.__shutdown_future.done():
                     context.asyncio_loop.create_task(self.__execute_process(context))
                     return
-
-        await context.emit_event(ProcessExited(returncode=returncode, **process_event_args))
         self.__cleanup()
 
     def execute(self, context: LaunchContext) -> Optional[List[LaunchDescriptionEntity]]:
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -107,7 +107,7 @@ class ExecuteProcess(Action):
             Callable[[ProcessExited, LaunchContext], Optional[SomeActionsType]]
         ]] = None,
         respawn: bool = False,
-        respawn_delay: float = 0.0,
+        respawn_delay: Optional[float] = None,
         **kwargs
     ) -> None:
         """
@@ -244,7 +244,6 @@ class ExecuteProcess(Action):
         self.__shutdown_future = None  # type: Optional[asyncio.Future]
         self.__sigterm_timer = None  # type: Optional[TimerAction]
         self.__sigkill_timer = None  # type: Optional[TimerAction]
-        self.__shutdown_received = False
         self.__stdout_buffer = io.StringIO()
         self.__stderr_buffer = io.StringIO()
 
@@ -402,14 +401,12 @@ class ExecuteProcess(Action):
         return []
 
     def _shutdown_process(self, context, *, send_sigint):
-        if self.__shutdown_received:
-            # Do not handle shutdown more than once.
+        if self.__shutdown_future is None or self.__shutdown_future.done():
+            # Execution not started or already done, nothing to do.
             return None
-        self.__shutdown_received = True
-        if self.__shutdown_future:
-            self.__shutdown_future.set_result(None)
+        self.__shutdown_future.set_result(None)
         if self.__completed_future is None:
-            # Execution not started so nothing to do, but self.__shutdown_received should prevent
+            # Execution not started so nothing to do, but self.__shutdown_future should prevent
             # execution from starting in the future.
             return None
         if self.__completed_future.done():
@@ -555,7 +552,7 @@ class ExecuteProcess(Action):
 
         # the respawned process needs to reuse these StringIO resources,
         # close them only after receiving the shutdown
-        if self.__shutdown_received:
+        if self.__shutdown_future is None or self.__shutdown_future.done():
             self.__stdout_buffer.close()
             self.__stderr_buffer.close()
         else:
@@ -752,21 +749,20 @@ class ExecuteProcess(Action):
                 pid, returncode, ' '.join(cmd)
             ))
         await context.emit_event(ProcessExited(returncode=returncode, **process_event_args))
-        if returncode != 0:
-            # respawn the process that abnormally died
-            if not self.__shutdown_received and not context.is_shutdown and self.__respawn:
-                if self.__respawn_delay > 0.0:
-                    # wait for a timeout(`self.__respawn_delay`) to respawn the process
-                    # and handle shutdown event with future(`self.__shutdown_future`)
-                    # to make sure `ros2 launch` exit in time
-                    await asyncio.wait(
-                        [asyncio.sleep(self.__respawn_delay), self.__shutdown_future],
-                        loop=context.asyncio_loop,
-                        return_when=asyncio.FIRST_COMPLETED
-                    )
-                if not self.__shutdown_future.done():
-                    context.asyncio_loop.create_task(self.__execute_process(context))
-                    return
+        # respawn the process if necessary
+        if not context.is_shutdown and not self.__shutdown_future.done() and self.__respawn:
+            if self.__respawn_delay is not None and self.__respawn_delay > 0.0:
+                # wait for a timeout(`self.__respawn_delay`) to respawn the process
+                # and handle shutdown event with future(`self.__shutdown_future`)
+                # to make sure `ros2 launch` exit in time
+                await asyncio.wait(
+                    [asyncio.sleep(self.__respawn_delay), self.__shutdown_future],
+                    loop=context.asyncio_loop,
+                    return_when=asyncio.FIRST_COMPLETED
+                )
+            if not self.__shutdown_future.done():
+                context.asyncio_loop.create_task(self.__execute_process(context))
+                return
         self.__cleanup()
 
     def execute(self, context: LaunchContext) -> Optional[List[LaunchDescriptionEntity]]:
@@ -786,7 +782,7 @@ class ExecuteProcess(Action):
             )
         self.__executed = True
 
-        if self.__shutdown_received:
+        if context.is_shutdown:
             # If shutdown starts before execution can start, don't start execution.
             return None
 
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -199,7 +199,9 @@ class ExecuteProcess(Action):
             involved.
         :param: on_exit list of actions to execute upon process exit.
         """
-        super().__init__(**kwargs)
+        super().__init__()
+        self.__respawn = kwargs['respawn'] if 'respawn' in kwargs else None
+        self.__respawn_delay = kwargs['respawn_delay'] if 'respawn_delay' in kwargs else 0
         self.__cmd = [normalize_to_list_of_substitutions(x) for x in cmd]
         self.__name = name if name is None else normalize_to_list_of_substitutions(name)
         self.__cwd = cwd if cwd is None else normalize_to_list_of_substitutions(cwd)
@@ -234,6 +236,7 @@ class ExecuteProcess(Action):
         self._subprocess_protocol = None  # type: Optional[Any]
         self._subprocess_transport = None
         self.__completed_future = None  # type: Optional[asyncio.Future]
+        self.__shutdown_future = None  # type: Optional[asyncio.Future]
         self.__sigterm_timer = None  # type: Optional[TimerAction]
         self.__sigkill_timer = None  # type: Optional[TimerAction]
         self.__shutdown_received = False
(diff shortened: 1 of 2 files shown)


launch/launch/actions/execute_process.py
```diff
@@ -199,7 +199,9 @@ def __init__(
             involved.
         :param: on_exit list of actions to execute upon process exit.
         """
-        super().__init__(**kwargs)
+        super().__init__()
+        self.__respawn = kwargs['respawn'] if 'respawn' in kwargs else None
+        self.__respawn_delay = kwargs['respawn_delay'] if 'respawn_delay' in kwargs else 0
```
hidmic, line 204: @iuhilnehc-ynos why not adding `respawn` and `respawn_delay` explictly, with type annotations and documentation?

launch/launch/actions/execute_process.py
```diff
@@ -716,6 +735,31 @@ async def __execute_process(self, context: LaunchContext) -> None:
...
+                    except Exception:
+                        self.__logger.error(
+                            'exception occurred while executing asyncio.wait_for:\n'
+                            '{}'.format(traceback.format_exc())
+                        )
+                if self.__respawn_delay == 0 or respawn_flag:
+                    context.asyncio_loop.create_task(self.__execute_process(context))
+                    return
+
         await context.emit_event(ProcessExited(returncode=returncode, **process_event_args))
```
hidmic, line 763: @iuhilnehc-ynos I think a `ProcessExited` event has to be emitted regardless of whether the process gets respawned or not.

launch/launch/actions/execute_process.py
```diff
@@ -716,6 +735,31 @@ async def __execute_process(self, context: LaunchContext) -> None:
             self.__logger.error("process has died [pid {}, exit code {}, cmd '{}'].".format(
                 pid, returncode, ' '.join(cmd)
             ))
+            # respawn the process that abnormally died
+            if not context.is_shutdown and self.__respawn:
```
hidmic, line 739: @iuhilnehc-ynos hmm, I wonder if we should respawn if a shutdown event was sent explicitly (which is not the same as a full `launch` shutdown.)

launch/launch/actions/execute_process.py
```diff
@@ -716,6 +735,31 @@ async def __execute_process(self, context: LaunchContext) -> None:
...
+                        # and handle shutdown event with future(`self.__shutdown_future`)
+                        # to make sure `ros2 launch` exit in time
+                        await asyncio.wait_for(
+                            self.__shutdown_future,
+                            timeout=self.__respawn_delay,
+                            loop=context.asyncio_loop
+                        )
+                    except asyncio.exceptions.TimeoutError:
+                        respawn_flag = True
+                        self.__shutdown_future = create_future(context.asyncio_loop)
```
hidmic, line 753: @iuhilnehc-ynos I don't follow why this future has to be recreated.

launch/launch/actions/execute_process.py
```diff
@@ -716,6 +735,31 @@ async def __execute_process(self, context: LaunchContext) -> None:
...
+                    except Exception:
+                        self.__logger.error(
+                            'exception occurred while executing asyncio.wait_for:\n'
+                            '{}'.format(traceback.format_exc())
+                        )
+                if self.__respawn_delay == 0 or respawn_flag:
```
hidmic, line 759: @iuhilnehc-ynos nit: instead of checking for equality, set the `respawn_flag` to true in an `else` clause if `respawn_delay` is not greater than 0.

launch/launch/actions/execute_process.py
iuhilnehc-ynos: @hidmic 
Sorry for the late response.
Maybe I just want to avoid adding an argument to ExecuteProcess constructor and minimize the updating source code.

launch/launch/actions/execute_process.py
iuhilnehc-ynos: @hidmic 
I totally agree with you. I'll update it.

launch/launch/actions/execute_process.py
iuhilnehc-ynos: @hidmic 
Not very clear about this comment. Do you mean I need to add extra checking about **not self.__shutdown_received**?

launch/launch/actions/execute_process.py
iuhilnehc-ynos: @hidmic 
I think it's because I misuse the **asyncio.wait_for** that will cancel the future inside if timeout, so it needs to create_futurn again. Anyway, I'll update it by using **asyncio.wait** with **[asyncio.sleep, self.__shutdown_future]**.

launch/launch/actions/execute_process.py
iuhilnehc-ynos: @hidmic 
Thanks for your guide. I'll update it.

launch/launch/actions/execute_process.py
hidmic: Correct. I'd think the `respawn` feature should apply to executables that die due to reasons external to `launch`.

launch/launch/actions/execute_process.py
```diff
@@ -716,6 +750,27 @@ async def __execute_process(self, context: LaunchContext) -> None:
...
+                    # and handle shutdown event with future(`self.__shutdown_future`)
+                    # to make sure `ros2 launch` exit in time
+                    await asyncio.wait(
+                        [asyncio.sleep(self.__respawn_delay), self.__shutdown_future],
+                        loop=context.asyncio_loop,
+                        return_when=asyncio.FIRST_COMPLETED
+                    )
+                    if self.__shutdown_future.done():
+                        respawn_flag = False
+                if respawn_flag:
```
hidmic, line 767: @iuhilnehc-ynos hmm, do we need `respawn_flag`? why not

```python
if not self.__shutdown_received and not context.is_shutdown and self.__respawn:
    if self.__respawn_delay > 0.0:
        # ...
    if not self.__shutdown_future.done():
        context.asyncio_loop.create_task(self.__execute_process(context))
        return
```

which somewhat emphasizes the redundancy between `shutdown_received` and `shutdown_future`.

launch/launch/actions/execute_process.py
```diff
@@ -106,6 +106,8 @@ def __init__(
             SomeActionsType,
             Callable[[ProcessExited, LaunchContext], Optional[SomeActionsType]]
         ]] = None,
+        respawn: bool = False,
+        respawn_delay: float = 0.0,
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Yes, they're optional.

launch/launch/actions/execute_process.py
```diff
@@ -229,11 +234,14 @@ def __init__(
 
         self.__log_cmd = log_cmd
         self.__on_exit = on_exit
+        self.__respawn = respawn
+        self.__respawn_delay = respawn_delay
```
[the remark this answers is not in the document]
iuhilnehc-ynos: @hidmic just keep the same behavior with ros1, so the answer is no.

launch/launch/actions/execute_process.py
```diff
@@ -383,6 +406,7 @@ def _shutdown_process(self, context, *, send_sigint):
             # Do not handle shutdown more than once.
             return None
         self.__shutdown_received = True
+        self.__shutdown_future.set_result(None)
```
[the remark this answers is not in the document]
iuhilnehc-ynos: @hidmic 
No, here we need to add a mechanism to make sure 'ros launch' exit in time while calling 'asyncio.wait'.

launch/launch/actions/execute_process.py
iuhilnehc-ynos: @hidmic 
Yes, it seems it can work without an extra flag. And I notice that I can also move the 'create_task' behind.

launch/launch/actions/execute_process.py
```diff
@@ -716,6 +750,27 @@ async def __execute_process(self, context: LaunchContext) -> None:
...
+                        [asyncio.sleep(self.__respawn_delay), self.__shutdown_future],
+                        loop=context.asyncio_loop,
+                        return_when=asyncio.FIRST_COMPLETED
+                    )
+                    if self.__shutdown_future.done():
+                        respawn_flag = False
+                if respawn_flag:
+                    await context.emit_event(
+                        ProcessExited(returncode=returncode, **process_event_args)
+                    )
```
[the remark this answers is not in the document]
iuhilnehc-ynos: @hidmic I agree that emitting this event should not be affected by checking respawn

launch/launch/actions/execute_process.py
```diff
@@ -717,6 +752,21 @@ async def __execute_process(self, context: LaunchContext) -> None:
                 pid, returncode, ' '.join(cmd)
             ))
         await context.emit_event(ProcessExited(returncode=returncode, **process_event_args))
+        if returncode != 0:
```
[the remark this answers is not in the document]
iuhilnehc-ynos: @ivanpauno 
OK, I'll update it.

launch/launch/actions/execute_process.py
```diff
@@ -383,6 +406,8 @@ def _shutdown_process(self, context, *, send_sigint):
             # Do not handle shutdown more than once.
             return None
         self.__shutdown_received = True
```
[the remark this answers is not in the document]
iuhilnehc-ynos: @ivanpauno 
I just wonder if the following code can be replaced?
https://github.com/ros2/launch/blob/f243fc6b3bcd276c194151e6933cda90761b1944/launch/launch/actions/execute_process.py#L739-L741

Because **self.__shutdown_future = create_future(context.asyncio_loop)** is created after it.

launch/launch/actions/execute_process.py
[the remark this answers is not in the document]
iuhilnehc-ynos: @hidmic 
@ivanpauno also suggests it, I'll discuss it with him if I can only use this shutdown_future

launch/launch/actions/execute_process.py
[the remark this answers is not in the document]
hongbin-liao: @ivanpauno sorry, I don't see the point to relaunch the process if it exits successfully. Normally we want to respawn the process it's because the process dies unexpectedly. If the process exits successfully, it shouldn't be launched again (at least in my case). Or maybe we can have another flag to specify in which situation we want to respawn the process.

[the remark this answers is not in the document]
hidmic: > Normally we want to respawn the process it's because the process dies unexpectedly. If the process exits successfully, it shouldn't be launched again (at least in my case).

In that case, @hongbin-liao, feel free to submit a patch for consideration.

FWIW ROS 1 `roslaunch` respawn feature [did not pay any attention to return codes](https://github.com/ros/ros_comm/blob/12eaf946597739e833b2300ef28bb079f248a9e1/tools/roslaunch/src/roslaunch/pmon.py#L539).

launch/launch/actions/execute_process.py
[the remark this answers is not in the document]
ivanpauno: Good catch, that check doesn't have any sense :joy:.
The shutdown event handler is installed **after** that check, so `self.__shutdown_received` will always be true in that place.

That should be replaced with `context.is_shutdown()`.

launch/launch/actions/execute_process.py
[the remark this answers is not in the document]
ivanpauno: Sorry if I wasn't clear, but my last comment was in favor of **not** adding a warning, as it will triggered easily from a launch XML file.

launch/launch/actions/execute_process.py
```diff
@@ -555,7 +552,7 @@ def __flush_buffers(self, event, context):
 
         # the respawned process needs to reuse these StringIO resources,
         # close them only after receiving the shutdown
-        if self.__shutdown_received:
+        if self.__shutdown_future is None or self.__shutdown_future.done():
```
ivanpauno, line 555: why `or`? shouldn't be `and` here?

launch/launch/actions/execute_process.py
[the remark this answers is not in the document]
ivanpauno: Maybe, there can be another optional argument to specify that.
Sounds like an improvement for another PR though

launch/launch/actions/execute_process.py
iuhilnehc-ynos: @ivanpauno 
I think you mean **(self.__shutdown_future is not None and self.__shutdown_future.done())** here.
**self.__shutdown_future is None or self.__shutdown_future.done():** contains
**self.__shutdown_future is None** or **(self.__shutdown_future is not None and self.__shutdown_future.done())**.
Just for __flush_buffers() this interface, I think it needs to check if **self.__shutdown_future** is created or not.
So I use **or** here.

ivanpauno: After checking the code again, this looks correct to me.

launch/launch/actions/execute_process.py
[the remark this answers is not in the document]
hidmic: Meta: now that `respawn_delay` is optional, I think it'd make sense to raise if it is given a value and `respawn` evaluates to `True`.

launch/launch/actions/execute_process.py
hidmic: This is a peculiar way of writing this condition. `self.__shutdown_future is None` is a tautology in normal circumstances, as no buffers will ever be flushed if the process hasn't even started.

launch/launch/actions/execute_process.py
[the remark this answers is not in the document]
ivanpauno: How will [this case](https://github.com/ros2/launch/pull/426#discussion_r438123968) be handled?

launch/launch/actions/execute_process.py
ivanpauno: We create the buffers right away in `__init__`:

https://github.com/ros2/launch/blob/0472f9a3ce6a156dfa93ccd0261b9727cadacc47/launch/launch/actions/execute_process.py#L247-L248

thus I think this check is correct

ivanpauno:
@iuhilnehc-ynos can you add a test for the feature?

launch/launch/actions/execute_process.py
hidmic: Sure, my point is that if the process never started then `__flush_buffers()` will never be called.

launch/launch/actions/execute_process.py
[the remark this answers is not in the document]
hidmic: Aha! Very good point. Oh well, fine to leave as-is.

launch/launch/actions/execute_process.py
ivanpauno: That's a good point.
We should probably close buffers in `__cleanup` (no related to this PR).

iuhilnehc-ynos:
> @iuhilnehc-ynos can you add a test for the feature?

@ivanpauno 
OK.
I have added a test, please review it again.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11182/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6461/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9132/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11096/)

launch/test/launch/test_execute_process.py
```diff
@@ -84,3 +86,35 @@ def on_exit_function(context):
...
     assert on_exit_function.called
+
+
+def test_execute_process_with_respawn():
+    """Test launching a process with a respawn and respawn_delay attribute."""
+    def on_exit_callback(event, context):
+        on_exit_callback.called_count = on_exit_callback.called_count + 1
+    on_exit_callback.called_count = 0
+
+    respawn_delay = 2
```
hidmic, line 97: @iuhilnehc-ynos nit^N: 
```suggestion
    respawn_delay = 2.0
```

ivanpauno: Done

PR #427, opened 2020-06-08
Defer shutdown if already running

Blast545:
@ivanpauno Trying to address #418 with your proposed comments in #424. 

EDIT: This PR now skips running a process if the shutdown process started. And defers the shutdown process for the cases where the process is starting.

Signed-off-by: Jorge Perez <jjperez@ekumenlabs.com>

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -408,16 +408,6 @@ class ExecuteProcess(Action):
             return None
 
         self.__shutdown_future.set_result(None)
-
-        if (self.__completed_future is not None and
-                (self.process_details is None or self._subprocess_transport is None)):
-            # Defer shut down if the process is scheduled to be started
-            context.register_event_handler(
-                OnProcessStart(
-                    on_start=lambda event, context:
-                    self._shutdown_process(context, send_sigint=send_sigint)))
-            return None
-
         self.__shutdown_received = True
 
         if self.__completed_future is None:
@@ -427,6 +417,16 @@ class ExecuteProcess(Action):
         if self.__completed_future.done():
             # If already done, then nothing to do.
             return None
+
+        # Defer shut down if the process is scheduled to be started
+        if (self.process_details is None or self._subprocess_transport is None):
+            self.__shutdown_received = False
+            context.register_event_handler(
+                OnProcessStart(
+                    on_start=lambda event, context:
+                    self._shutdown_process(context, send_sigint=send_sigint)))
+            return None
+
         # Otherwise process is still running, start the shutdown procedures.
         context.extend_locals({'process_name': self.process_details['name']})
         actions_to_return = self.__get_shutdown_timer_actions()
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -754,6 +754,7 @@ class ExecuteProcess(Action):
             return
 
         pid = transport.get_pid()
+        # self.__action._subprocess_transport = transport
 
         await context.emit_event(ProcessStarted(**process_event_args))
 
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -409,8 +409,8 @@ class ExecuteProcess(Action):
 
         self.__shutdown_future.set_result(None)
 
-        if self.__completed_future is not None and \
-        (self.process_details is None or self._subprocess_transport is None):
+        if (self.__completed_future is not None and
+            (self.process_details is None or self._subprocess_transport is None)):
             # Defer shut down if the process is scheduled to be started
             context.register_event_handler(
                 OnProcessStart(
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -420,10 +420,6 @@ class ExecuteProcess(Action):
 
         self.__shutdown_received = True
 
-        # if not self.__running:
-        #     # Not running, skip shutdown
-        #     return None
-
         if self.__completed_future is None:
             # Execution not started so nothing to do, but self.__shutdown_future should prevent
             # execution from starting in the future.
@@ -720,12 +716,6 @@ class ExecuteProcess(Action):
         process_event_args = self.__process_event_args
         if process_event_args is None:
             raise RuntimeError('process_event_args unexpectedly None')
-
-        # if self.__shutdown_received:
-        #     # Shutdown received, do not run process
-        #     self.__cleanup()
-        #     return
-        # self.__running = True
 
         cmd = process_event_args['cmd']
         cwd = process_event_args['cwd']
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -406,25 +406,26 @@ class ExecuteProcess(Action):
             # Execution not started or already done, nothing to do.
             return None
 
-        self.__shutdown_future.set_result(None)
-        self.__shutdown_received = True
-
         if self.__completed_future is None:
             # Execution not started so nothing to do, but self.__shutdown_future should prevent
             # execution from starting in the future.
+            self.__shutdown_future.set_result(None)
             return None
         if self.__completed_future.done():
             # If already done, then nothing to do.
+            self.__shutdown_future.set_result(None)
             return None
 
         # Defer shut down if the process is scheduled to be started
         if (self.process_details is None or self._subprocess_transport is None):
-            self.__shutdown_received = False
+            # Do not set shutdown result, as event is postponed
             context.register_event_handler(
                 OnProcessStart(
                     on_start=lambda event, context:
                     self._shutdown_process(context, send_sigint=send_sigint)))
             return None
+
+        self.__shutdown_future.set_result(None)
 
         # Otherwise process is still running, start the shutdown procedures.
         context.extend_locals({'process_name': self.process_details['name']})
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -420,7 +420,7 @@ class ExecuteProcess(Action):
         self.__shutdown_received = True
 
         if not self.__running:
-            #"Not running, skip shutdown"
+            # Not running, skip shutdown
             return None
 
         if self.__completed_future is None:
@@ -737,7 +737,7 @@ class ExecuteProcess(Action):
             )
 
         if self.__shutdown_received:
-            #"Shutdown received, do not run process"
+            # Shutdown received, do not run process
             return
         self.__running = True
 
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -249,7 +249,6 @@ class ExecuteProcess(Action):
         self.__stderr_buffer = io.StringIO()
 
         self.__executed = False
-        # self.__running = False
 
     @classmethod
     def _parse_cmdline(
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -667,7 +667,6 @@ class ExecuteProcess(Action):
                 'process started with pid [{}]'.format(transport.get_pid()),
             )
             super().connection_made(transport)
-            self.__process_event_args['pid'] = transport.get_pid()
 
         def on_stdout_received(self, data: bytes) -> None:
             self.__context.emit_event_sync(ProcessStdout(text=data, **self.__process_event_args))
@@ -752,6 +751,7 @@ class ExecuteProcess(Action):
             return
 
         pid = transport.get_pid()
+        self.__process_event_args['pid'] = pid
         self._subprocess_transport = transport
 
         await context.emit_event(ProcessStarted(**process_event_args))
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -48,6 +48,7 @@ from ..event import Event
 from ..event_handler import EventHandler
 from ..event_handlers import OnProcessExit
 from ..event_handlers import OnProcessIO
+from ..event_handlers import OnProcessStart
 from ..event_handlers import OnShutdown
 from ..events import matches_action
 from ..events import Shutdown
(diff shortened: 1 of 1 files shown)


Blast545:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11061/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6378/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9005/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10977/)

Blast545:
CI of the opening issue:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10978/)

Blast545:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11066/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10980/)

launch/launch/actions/execute_process.py
```diff
@@ -686,6 +702,11 @@ async def __execute_process(self, context: LaunchContext) -> None:
                 ),
             )
 
+        if self.__shutdown_received:
+            # Shutdown received, do not run process
+            return
+        self.__running = True
```
hidmic, line 708: @Blast545 why not moving this past the try-block, where it's actually true?

ivanpauno, line 708: I would add this at the beginning of the method.

launch/launch/actions/execute_process.py
```diff
@@ -382,7 +384,21 @@ def _shutdown_process(self, context, *, send_sigint):
...
+                OnProcessStart(
+                    on_start=lambda event, context:
+                    self._shutdown_process(context, send_sigint=send_sigint)))
+            return None
+
         self.__shutdown_received = True
+
+        if not self.__running:
+            # Not running, skip shutdown
+            return None
```
ivanpauno, line 400: The logic can be more clear if it's reordered:

```python3
        if self.__shutdown_received:
            return None
        if not self.__running:
            return
        if self.process_details is None or self._subprocess_transport is None:
            # Defer shut down if the process is not ready
            context.register_event_handler(
                OnProcessStart(
                    on_start=lambda event, context:
                    self._shutdown_process(context, send_sigint=send_sigint)))
            return None
        ...
```

launch/launch/actions/execute_process.py
```diff
@@ -382,7 +384,21 @@ def _shutdown_process(self, context, *, send_sigint):
...
             # Do not handle shutdown more than once.
             return None
+
+        if self.__running and (self.process_details is None or self._subprocess_transport is None):
+            # Defer shut down if the process is not ready
+            context.register_event_handler(
+                OnProcessStart(
+                    on_start=lambda event, context:
+                    self._shutdown_process(context, send_sigint=send_sigint)))
+            return None
```
hidmic, line 394: @Blast545 I'd think this should happen once we're past checking for completion (line 409) and iff it's not `running` already (assuming `running` means precisely that without further considerations).

launch/launch/actions/execute_process.py
ivanpauno: It wouldn't be of help in that case ...
It would be true when the condition
```
self.process_details is None or self._subprocess_transport is None
```
is false.

The idea is to check if you are already awaiting `async_execute_process` (or about to await it).

ivanpauno: Disregard this.

launch/launch/actions/execute_process.py
Blast545: If that block happens after `self.__shutdown_received = True`, then the deferring of the shutdown does not occur because of the first `if self.__shutdown_received:`. I didn't find any other way to arrange it.

launch/launch/actions/execute_process.py
Blast545: Why? Actually @hidmic is suggesting me to change it to after the try block, and I think that would be a better idea, to avoid having the shutdown process to something that's not actually running.

launch/launch/actions/execute_process.py
Blast545: That was my original intention, but the block has to be called before the `self.__shutdown_received = True` to be able of deferring the shutdown process.

launch/launch/actions/execute_process.py
hidmic: If you're past the try-block, you're done awaiting for `async_execute_process` and transport is already not `None`. 

But it's true that there's no guarantee as of whether the protocol has established the connection or not. Which makes wonder. Why does the protocol inject the transport into an action "private" attribute? `ExecuteProcess.__execute_process()` could equally do `self._subprocess_transport = transport` and fill in the PID in process details once it's past the `await`. And set `self._running` accordingly.

ivanpauno: mmm ... I don't think that works.
Imaging the following scenario:
- __execute_process is called, and it's awaiting for async_execute_process to finish.
- The shutdown event is triggered, as `self.__running` is false it won't do nothing.
- At some point `async_execute_process` will finish, and we're not killing the launched process.

Writing that gave me an idea ... `__execute_process` should check `self.__shutdown` after `async_execute_process` returns!
That solves the problem and is cleaner than the running flag!

hidmic: > The shutdown event is triggered, as self.__running is false it won't do nothing.

If self.__running is false and the completion future is not None, that means we're half-way through an async exec and thus the shutdown has to be deferred. It's not a no-op.

> `__execute_process` should check self.__shutdown after async_execute_process returns!

We could do that, yeah, but we still need to check in `_shutdown_process` whether we can rely on `__execute_process` to shutdown or not. And `__execute_process` would have to call `_shutdown_process()` or a subset of it to get the same behavior (timed signalling and stuff).

ivanpauno: >  completion future is not None, that means we're half-way through an async exec and thus the shutdown has to be deferred

What completion future is indicating that?

> We could do that, yeah, but we still need to check in _shutdown_process whether we can rely on __execute_process to shutdown or not. And __execute_process would have to call _shutdown_process() or a subset of it to get the same behavior (timed signalling and stuff).

Yes

launch/launch/actions/execute_process.py
Blast545: `shutdown_received` is set if already started or finished, you mean setting that variable in every one of those scenarios?

Blast545:
After reading on the discussion here, isn't this check enough to check that the process is scheduled to be executed?
https://github.com/ros2/launch/blob/b79e8edd531162bedac89aa82ca25a2ea8def093/launch/launch/actions/execute_process.py#L402

I think that variable can avoid the use of the self.__running one added. Using `self.__completed_future not None` is enough to see if process is being prepared to be executed and if `self._subprocess_transport` is available, the process is running.

If that helds truth, bab39b7 should solve the problem faced. I think the checks for shutdown already in `execute` should deal with not starting processes after shutdown.

launch/launch/actions/execute_process.py
hidmic: > What completion future is indicating that?

[This one](https://github.com/ros2/launch/pull/427/files#diff-61622adb9c2241797a64d96e6065c414R798). Of course, that's up to where you put early returns, if any.

Personally, I'd rather not plague the implementation with early returns and instead have an early return on `execute()`, a flag to tell the process has started execution (can be completion future existence) and a flag to tell if it's actually running (can be transport availability). On shutdown, if the process didn't start you just don't, if the process is on its way you defer, and if the process is running you kill it. There's little gain in catching it mid-flight after `execute()` completed but `async_execute_process()` hasn't been submitted yet.

Blast545:
CI:
* Linux to check lint: [Build Status](http://ci.ros2.org/job/ci_linux/11079/)

launch/launch/actions/execute_process.py
hidmic: Yes. As it stands, if you'd got a `_shutdown_process()` on a finished action, you'd try to kill it.

Blast545: With the current implementation if an action finished normally then the `shutdown_received` variable is set and the `_shutdown_process` does nothing. I think I'm not following your idea.

hidmic:
@Blast545 one thing that is missing here is to set `_subprocess_transport` in `__execute_process()` instead of in the protocol subclass before emitting the `ProcessStart`, or otherwise you may defer a shutdown and miss it.

hidmic:
> It seems  `async_execute_process` is awaited, isn't this a guarantee the `connection_made` method will be called before emitting the `ProcessStarted` signal?

Not necessarily. The coroutine in which you await for `async_execute_process` may be scheduled by the event loop before the `connection_made` callback (callback is scheduled [here](https://github.com/python/cpython/blob/a3d6d23a923d62a713b9e4536d5f7ce7313d99c7/Lib/asyncio/base_subprocess.py#L181), future gets resolved [here](https://github.com/python/cpython/blob/a3d6d23a923d62a713b9e4536d5f7ce7313d99c7/Lib/asyncio/base_subprocess.py#L192)). The most likely scenario is that it is resolved first, but better to err on the safe side.

Blast545:
@hidmic with the commented part of f4e3b85 I tested locally setting the transport before emitting the ProcessStarted signal, but this caused the tests to hang. Am I missing something?

Blast545:
Last PR I think it addressed all your comments @hidmic 

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11135/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6430/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9083/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11053/)

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11054/)

launch/launch/actions/execute_process.py
```diff
@@ -241,6 +242,7 @@ def __init__(
         self.__stderr_buffer = io.StringIO()
 
         self.__executed = False
+        # self.__running = False
```
ivanpauno, line 245: please, delete this comment before merging

hidmic: +1 to that.

launch/launch/actions/execute_process.py
```diff
@@ -622,7 +636,6 @@ def connection_made(self, transport):
             )
             super().connection_made(transport)
             self.__process_event_args['pid'] = transport.get_pid()
```
hidmic, line 638: @Blast545 I think we should update this in the same place in which we set the `_subprocess_transport`, no?

hidmic: Also, I think the protocol no longer needs to know about the `__action`.

Blast545: Addressed both comments with 0516975 and c20dab4 respectively

Blast545:
Rebasing, and refactoring to match changes made on master.

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11197/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6474/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9158/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11115/)

* Windows up to composition to check usage: [Build Status](http://ci.ros2.org/job/ci_windows/11116/)

Blast545:
@ros_pull_request_builder

ivanpauno:
@ros-pull-request-builder retest this please

Blast545:
This is causing a regression in `launch_xml`

Master CI: [Build Status](http://ci.ros2.org/job/ci_linux/11201/)

This branch CI: [Build Status](http://ci.ros2.org/job/ci_linux/11202/)

ivanpauno:
@Blast545 the failing test is:

https://github.com/ros2/launch/blob/8d639818da9d458b9be4a8203a775bf2220ef417/launch_xml/test/launch_xml/test_include.py#L27

Uou can try it out locally to see what's going on.

Blast545:
Thanks for the tip @ivanpauno, I'm already checking, seems to be that moving the pid assignment in the `connection_made` of `__ProcessProtocol` causes the problem.

hidmic:
> seems to be that moving the pid assignment in the connection_made of __ProcessProtocol causes the problem.

That is surprising.

Blast545:
@hidmic I get this traceback output:

```
[1.533s] Traceback (most recent call last):
[1.533s]   File "/usr/lib/python3.6/asyncio/events.py", line 145, in _run
[1.533s]     self._callback(*self._args)
[1.533s]   File "/home/misato/ros_ws/build/osrf_pycommon/osrf_pycommon/process_utils/async_execute_process.py", line 263, in pipe_data_received
[1.533s]     self.on_stdout_received(data)
[1.533s]   File "/home/misato/ros_ws/build/launch/launch/actions/execute_process.py", line 673, in on_stdout_received
[1.533s]     self.__context.emit_event_sync(ProcessStdout(text=data, **self.__process_event_args))
[1.533s]   File "/home/misato/ros_ws/build/launch/launch/events/process/process_stdout.py", line 34, in __init__
[1.533s]     super().__init__(text=text, fd=1, **kwargs)
[1.533s]   File "/home/misato/ros_ws/build/launch/launch/events/process/process_io.py", line 35, in __init__
[1.533s]     super().__init__(**kwargs)
[1.533s] TypeError: __init__() missing 1 required keyword-only argument: 'pid'
```

It seems other handlers require the pid variable set there intermediately after the connection_made.

hidmic:
@Blast545 weren't we populating the same dictionary in `__execute_process()` ? Anyways, this is fine.

Blast545:
CI: [Build Status](http://ci.ros2.org/job/ci_linux/11209/)

Blast545:
@hidmic Yeah, it's the same structure. Something was interrupting the process in the middle of the `connection_made` method and the section I added.

Green CI up to `launch_xml` + `launch` and automatic build job without errors. Will merge as is.

PR #429, opened 2020-06-19
Remove the asyncio.wait loop parameter.

clalancette:
This has been deprecated for quite a long time, but Python 3.8
is now showing warnings about it.  Just remove it here.

The warning that this gets rid of looks like:

```
  /home/ubuntu/ros2_ws/src/ros2/launch/launch/launch/actions/execute_process.py:758: DeprecationWarning: The loop argument is deprecated since Python 3.8, and scheduled for removal in Python 3.10.
    await asyncio.wait(

```

One thing I'll point out is that I don't know if this is *exactly* the same as what was there before.  Looking into the Python code, if the loop parameter is None (the default), then it just gets the currently running loop and adds this `wait` to that.  But its not entirely clear to me that the currently running loop is `context.asyncio_loop`.  The tests pass, but I'm not sure if that is enough.  Looking for review from people with more expertise in this to say for sure.

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -772,7 +772,6 @@ class ExecuteProcess(Action):
                 # to make sure `ros2 launch` exit in time
                 await asyncio.wait(
                     [asyncio.sleep(self.__respawn_delay), self.__shutdown_future],
-                    loop=context.asyncio_loop,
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Seems to be a left over of #387.

clalancette:
All of the failing tests are also failing in nightlies, so this is good to go.  Thanks for the review.

PR #430, opened 2020-06-22
Remove pkg_resources.

clalancette:
Replace it with the use of the more modern importlib*
libraries.  There are a couple of reasons to do this:

1.  pkg_resources is quite slow to import; on my machine,
just firing up the python interpreter takes ~35ms, while
firing up the python interpreter and importing pkg_resources
takes ~175ms.  Firing up the python interpreter and importing
importlib_metadata takes ~70ms.  Removing 100ms per invocation
of the command-line both makes it speedier for users, and
will speed up our tests (which call out to the command-line
quite a lot).

2.  pkg_resources is somewhat deprecated and being replaced
by importlib.  https://importlib-metadata.readthedocs.io/en/latest/using.html
describes some of it

Note: By itself, this change is not enough to completely remove our
dependence on pkg_resources.  We'll also have to do something about
the console_scripts that setup.py generates.  That will be
a separate effort.

See ros2/ros2#919 for a lot more details on why we want to do this.

### DIFF
--- a/launch/launch/frontend/parser.py
+++ b/launch/launch/frontend/parser.py
@@ -26,7 +26,7 @@ from typing import Type
 from typing import TYPE_CHECKING
 from typing import Union
 
-from pkg_resources import iter_entry_points
+import importlib_metadata
 
 from .entity import Entity
 from .expose import instantiate_action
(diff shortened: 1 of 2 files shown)


jacobperron:
Should this be backported?

clalancette:
> Should this be backported?

Same response as https://github.com/ament/ament_lint/pull/260#issuecomment-648324681 .

PR #432, opened 2020-06-24
avoid side effect, defer until needed

dirk-thomas:
This avoids looking for and parsing the grammar on module load time. Instead it is deferred until when the parser is needed the first time.

As a side effect (pun intended) this fixes ros-visualization/rqt#230: https://ci.ros2.org/view/nightly/job/nightly_win_rel/1591/testReport/(root)/projectroot/rqt_py_common/

CI builds testing `launch` and above:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11285/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6534/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9235/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11190/) (unrelated test failures)

### DIFF
--- a/launch/launch/frontend/parse_substitution.py
+++ b/launch/launch/frontend/parse_substitution.py
@@ -107,7 +107,8 @@ def parse_substitution(string_value):
         # Grammar cannot deal with zero-width expressions.
         return [TextSubstitution(text=string_value)]
     if _parser is None:
-        _parser = Lark.open(get_grammar_path(), start='template')
+        with open(get_grammar_path(), 'r') as h:
+            _parser = Lark(h, start='template')
     tree = _parser.parse(string_value)
     transformer = ExtractSubstitution()
     return transformer.transform(tree)
--- a/launch/launch/frontend/parse_substitution.py
+++ b/launch/launch/frontend/parse_substitution.py
@@ -93,14 +93,21 @@ class ExtractSubstitution(Transformer):
     double_quoted_template = template
 
 
-grammar_file = os.path.join(get_package_share_directory('launch'), 'frontend', 'grammar.lark')
-parser = Lark.open(grammar_file, start='template')
-transformer = ExtractSubstitution()
+def get_grammar_path():
+    return os.path.join(
+        get_package_share_directory('launch'), 'frontend', 'grammar.lark')
+
+
+_parser = None
 
 
 def parse_substitution(string_value):
+    global _parser
     if not string_value:
         # Grammar cannot deal with zero-width expressions.
         return [TextSubstitution(text=string_value)]
-    tree = parser.parse(string_value)
+    if _parser is None:
+        _parser = Lark.open(get_grammar_path(), start='template')
+    tree = _parser.parse(string_value)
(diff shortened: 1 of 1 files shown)


launch/launch/frontend/parse_substitution.py
```diff
@@ -93,14 +93,21 @@ def template(self, fragments):
...
 
-grammar_file = os.path.join(get_package_share_directory('launch'), 'frontend', 'grammar.lark')
-parser = Lark.open(grammar_file, start='template')
-transformer = ExtractSubstitution()
+def get_grammar_path():
+    return os.path.join(
+        get_package_share_directory('launch'), 'frontend', 'grammar.lark')
+
+
+_parser = None
```
clalancette, line 101: In general, thumbs up to removing stuff from import time.  I guess my question is: does this have to be global at all?  Can't we just instantiate the Lark parser during `parse_substitution` below when we need it?

ivanpauno: It's possible, but we would initialize it many times, which is not needed.
Using a global looks good to me.

ivanpauno: The problem with the global is that we're never closing the file.
Maybe using a local isn't bad.

dirk-thomas: Parsing a grammar is a pretty heavy process. So doing it more than one isn't a good idea imo. That is why delaying the steps until it is needed is also beneficial to not affect the module load time.

launch/launch/frontend/parse_substitution.py
```diff
@@ -93,14 +93,21 @@ def template(self, fragments):
     double_quoted_template = template
 
 
-grammar_file = os.path.join(get_package_share_directory('launch'), 'frontend', 'grammar.lark')
-parser = Lark.open(grammar_file, start='template')
```
dirk-thomas, line 97: Does anyone know why `Load.open()` was chosen here instead of the more intuitive constructor `Lark()`?

ivanpauno: It was a completely random choice. I'm not sure which one is more intuitive.

dirk-thomas: The constructor is imo the preferred way since it allows arbitrary options to be passed. In newer versions that includes using a cached state of the parsed grammar.

ivanpauno: If you want to make that change here, it sounds good to me too.

dirk-thomas: Done in fc22b48347e3e6b8e575b2f5ec3a6a932abccbaf.

dirk-thomas: The constructor takes the actual grammar string rather than a path. I modified it in 1210d0ac93aca05977a8faaa3605fe9a99d1431a which also nicely shows that the file is being closed after parsing the grammar.

PR #435, opened 2020-07-02
Use unittest.TestCase.id() to put together jUnit XML output.

hidmic:
Precisely what the title says. This is particularly important if using `unittest.TestCase.subTest()` or otherwise a generic name is all you get (see [an example](https://ci.ros2.org/view/nightly/job/nightly_linux_debug/1602/testReport/junit/ros2topic.ros2topic.test/test_echo_pub/test_echo_pub/)).

### DIFF
--- a/launch_testing/launch_testing/junitxml.py
+++ b/launch_testing/launch_testing/junitxml.py
@@ -83,9 +83,10 @@ def unittestCaseToXml(test_result, test_case):
     class needs to be a launch_testing.TestResult class
     """
     case_xml = ET.Element('testcase')
-    full_classname = type(test_case).__module__ + '.' + type(test_case).__name__
+    full_methodname, _, qualifiers = test_case.id().partition(' ')
+    full_classname, _, methodname = full_methodname.rpartition('.')
     case_xml.set('classname', full_classname)
-    case_xml.set('name', test_case._testMethodName)
+    case_xml.set('name', (methodname + ' ' + qualifiers).strip())
     case_xml.set('time', str(round(test_result.testTimes[test_case], 3)))
 
     for failure in test_result.failures:
--- a/launch_testing/test/launch_testing/test_xml_output.py
+++ b/launch_testing/test/launch_testing/test_xml_output.py
@@ -86,6 +86,8 @@ class TestXmlFunctions(unittest.TestCase):
         class TestHost(unittest.TestCase):
             pass
 
+        TestHost.__qualname__ = 'TestHost'  # as if it was defined in global scope
+
         for n, test_case in enumerate(test_case_list):
             setattr(TestHost, 'test_{}'.format(n), test_case)
 


hidmic:
Full CI to make sure nothing breaks:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11372/)

launch_testing/launch_testing/junitxml.py
```diff
@@ -83,9 +83,10 @@ def unittestCaseToXml(test_result, test_case):
     class needs to be a launch_testing.TestResult class
     """
     case_xml = ET.Element('testcase')
-    full_classname = type(test_case).__module__ + '.' + type(test_case).__name__
+    full_methodname, _, qualifiers = test_case.id().partition(' ')
```
ivanpauno, line 86: can you give an example of how a full method name with qualifiers look?

launch_testing/test/launch_testing/test_xml_output.py
```diff
@@ -86,6 +86,8 @@ def unit_test_result_factory(self, test_case_list):
         class TestHost(unittest.TestCase):
             pass
 
+        TestHost.__qualname__ = 'TestHost'  # as if it was defined in global scope
```
ivanpauno, line 89: why is this needed?

launch_testing/launch_testing/junitxml.py
hidmic: Qualifiers apply to the test case, not the method. When using the [`subTest` feature](https://docs.python.org/3/library/unittest.html#unittest.TestCase.subTest), `unittest` appends message and parameters that I need to account for.

launch_testing/test/launch_testing/test_xml_output.py
hidmic: With the change in how we generate jUnit XML, tests that make use of this local class would have to be changed to account for its real qualified class name i.e. it's not `test_xml_output.TestHost`, more like `test_xml_output.unit_test_result_factory.<locals>.TestHost`. And I didn't think it was worth cluttering tests with that detail, therefore this line.

launch_testing/launch_testing/junitxml.py
ivanpauno: thanks for clarifying, sounds good

hidmic:
Alright, full Linux CI's green. This is trivial Python code, so I don't think it's worth running full CI on other platforms.

dirk-thomas:
Maybe I am missing something but how does the what improvement look like for the example above? The latest case with this patch seems to have the exact same output: https://ci.ros2.org/view/nightly/job/nightly_linux_debug/1606/testReport/ros2topic.ros2topic.test/test_echo_pub/test_echo_pub/

hidmic:
Uh, that's... unexpected.

PR #436, opened 2020-07-06
Use unittest.TestCase.id() for pytest failure reprs.

hidmic:
Follow-up after #435. This fixes the same problem but for `pytest` XML output.

### DIFF
--- a/launch_testing/launch_testing/pytest/hooks.py
+++ b/launch_testing/launch_testing/pytest/hooks.py
@@ -110,7 +110,7 @@ class LaunchTestItem(pytest.Item):
     def repr_failure(self, excinfo):
         if isinstance(excinfo.value, LaunchTestFailure):
             return LaunchTestFailureRepr(failures=[
-                (f'{type(test_case).__name__}.{test_case._testMethodName}', formatted_error)
+                (test_case.id(), formatted_error)
                 for test_run, test_result in excinfo.value.results.items()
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Can you please post an example test result (like the one from the original ticket) with this patch.

hidmic:
@dirk-thomas artificially forcing `test_echo_pub.py` in `ros2topic` to fail one gets:

```
FAIL: test_echo_pub.TestROS2TopicEchoPub.test_echo_basic (topic=&apos;/clitest/topic/echo_incompatible_qos&apos;, provide_qos=True, compatible_qos=False)
-----------------------------------------------------------------------------------------------------------------------------
  File &quot;/home/michel/Workspaces/focal_ros2_ws/src/ros2/ros2cli/ros2topic/test/test_echo_pub.py&quot;, line 227, in test_echo_basic
    assert False
AssertionError: assert False
```

in `pytest.xml`.

hidmic:
Considering that the rolling PR job already checks for style compliance, I don't think cross-platform CI will add any value.

dirk-thomas:
> Considering that the rolling PR job already checks for style compliance, I don't think cross-platform CI will add any value.

@hidmic Just keep in mind that the PR jobs only use linters available from Debian packages where as CI jobs on ci.ros2.org use several more / new ones.

hidmic:
Thanks for the FYI.

PR #438, opened 2020-07-13
Consolidate type_utils in a way that can be reused in substitution results that need to be coerced to a specific type

ivanpauno:
Follow up of https://github.com/ros2/launch_ros/pull/137.

This try to consolidate the `launch.frontend.type_utils` module, in a way it's usable from custom actions/descriptions that requires an user to specify the type.

To consolidate the coercion rules allowed I've dropped support of `Union[...]` and non uniform lists, we really don't need that.

I have to consolidate documentation (partially outdated), type annotations and add more test cases.

This could also be used in the future in places like:
- https://github.com/ros2/launch/blob/24ae1327b4fd6c358c40899569ceabbdb4bc57f4/launch/launch/actions/timer_action.py#L59

so we always use the same rules to convert result of substitutions to specific types.


launch/launch/utilities/type_utils.py
```diff
@@ -0,0 +1,451 @@
...
+    )
+    if types_in_list.issubset({str, Substitution}):
+        # list of mixed strings and substitutions
+        if data_type not in (None, str):
+            raise TypeError(err_msg.format(str))
+        return cast(List[Union[List[Substitution], str]], [
+            x if isinstance(x, str) else  # Don't convert strings to TextSubstitution
+            normalize_to_list_of_substitutions(cast(SomeSubstitutionsType, x)) for x in value
+        ])
+    if types_in_list.issubset({bool, Substitution}):
```
hidmic, line 380: @ivanpauno if `types_in_list == {float, int, bool, Substitution}`, this will not fail until the "typed" substitution is performed. Is that intentional?

launch/test/launch/utilities/test_type_utils.py
```diff
@@ -0,0 +1,149 @@
+# Copyright 2020 Open Source Robotics Foundation, Inc.
```
hidmic, line 1: @ivanpauno this is missing some tests e.g. for `normalize_typed_substitution()` and `perform_typed_substitution()`.

launch/launch/frontend/parser.py
```diff
@@ -83,11 +85,44 @@ def parse_action(self, entity: Entity) -> (Action, Tuple[Any]):
...
+    def parse_substitution(self, value: Text) -> List[Substitution]:
         """Parse a substitution."""
         return parse_substitution(value)
 
-    def escape_characters(self, value: Text) -> SomeSubstitutionsType:
+    def parse_if_substitutions(
+        self, value: StrSomeValueType
+    ) -> NormalizedValueType:
+        """
+        Parse substitutions in `value` if there are some.
```
hidmic, line 96: @ivanpauno nit: 
```suggestion
        Parse substitutions in `value` if there are any.
```

launch/launch/utilities/type_utils.py
```diff
@@ -0,0 +1,451 @@
...
+    """
+    output = [coerce_to_type(i, data_type, can_be_str) for i in value]
+    if not check_is_instance_of_valid_type(output, can_be_str):
+        raise ValueError(f'cannot convert value to {data_type}. Got value=`{value}`')
+    return cast(ListValueType, output)
+
+
+def get_typed_value(
+    value: Union[Text, List[Text]],
+    data_type: Optional[AllowedTypesType],
```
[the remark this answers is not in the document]
ivanpauno: no, `int` is also valid here for example.

launch/launch/utilities/type_utils.py
```diff
@@ -0,0 +1,451 @@
...
+            f"Got value='{value}', data_type='{data_type}'."
+        )
+
+    # Normalize each specific uniform list input
+    err_msg = (
+        "Got a list of '{}'"
+        f", expected a list of '{data_type}'. value='{value}'"
+    )
+    if types_in_list.issubset({str, Substitution}):
+        # list of mixed strings and substitutions
```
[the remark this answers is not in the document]
ivanpauno: That case is checking for a single substitution, this one is a list of substitutions and strings.

PS: This comment makes me think that I'm not handling a list of only substitutions correctly... I will solve that after adding tests.

launch/launch/utilities/type_utils.py
ivanpauno: Why?
It should fail right away.

launch/test/launch/utilities/test_type_utils.py
ivanpauno: this is missing a lot of tests :joy:

launch/launch/utilities/type_utils.py
```diff
@@ -0,0 +1,451 @@
...
+AllowedTypesTuple = ScalarTypesTuple + ListTypesTuple
+
+"""Allowed scalar types."""
+ScalarTypesType = Type[Union[int, float, bool, str]]
+"""Allowed uniform list types."""
+ListTypesType = Type[Union[
+    List[int], List[float], List[bool], List[str]
+]]
+"""Allowed types."""
+AllowedTypesType = Union[ScalarTypesType, ListTypesType]
```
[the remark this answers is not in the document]
ivanpauno: > I understand why the focus on uniform lists, but this is clearly at odds with Parser.get_attr() claiming it can handle non-uniform lists.

Non uniform lists add extra complexity, and I can't think of an use case for them.
I have to update old doc blocks.

launch/launch/utilities/type_utils.py
```diff
@@ -0,0 +1,451 @@
...
+    if type_obj in (int, float):
+        try:
+            return type_obj(value)
+        except ValueError:
+            if can_be_str:
+                return value
+            else:
+                raise
+
+    if bool != type_obj:
```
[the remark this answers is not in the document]
ivanpauno: Done in https://github.com/ros2/launch/pull/438/commits/df07ebd9359bb9ab87ec5893d398b66c6573e085

launch/launch/utilities/type_utils.py
```diff
@@ -0,0 +1,451 @@
...
+    return all(isinstance(x, type_obj) for x in value)
+
+
+def coerce_to_type(
+    value: Text,
+    data_type: Optional[AllowedTypesType] = None,
+    can_be_str: bool = False,
+) -> StrSomeValueType:
+    """
+    Coerce `value` to `type`.
```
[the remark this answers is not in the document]
ivanpauno: Done, see https://github.com/ros2/launch/pull/438/commits/df07ebd9359bb9ab87ec5893d398b66c6573e085

launch/launch/utilities/type_utils.py
```diff
@@ -0,0 +1,451 @@
...
+def check_type(
+    value: Any,
+    data_type: Optional[AllowedTypesType] = None,
+    can_be_str: bool = False,
+) -> bool:
+    """
+    Check if `value` is of `type`.
+
+    :param value: variable to check.
+    :param data_type: value will be enforced to be an instance of data_type.
```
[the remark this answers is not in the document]
ivanpauno: Yeap, see https://github.com/ros2/launch/pull/438/commits/df07ebd9359bb9ab87ec5893d398b66c6573e085.

ivanpauno:
@hidmic I'm still iterating on documentation and tests, but I've at least covered all the functions in `launch.utilities.type_utils`.
Can you take another look?

launch/launch/utilities/type_utils.py
ivanpauno: According to the tests I've added, that case is failing right away.

launch/test/launch/utilities/test_type_utils.py
ivanpauno: Ok, I think those functions are well covered now.

launch/launch/utilities/type_utils.py
```diff
@@ -0,0 +1,451 @@
...
+        if data_type not in (None, int):
+            raise TypeError(err_msg.format(int))
+        return cast(List[Union[List[Substitution], int]], [
+            normalize_to_list_of_substitutions(cast(SomeSubstitutionsType, x))
+            if is_substitution(x) else x for x in value
+        ])
+    if types_in_list.issubset({int, float, Substitution}):
+        # list of floats and substitutions
+        if data_type not in (None, float):
+            raise TypeError(err_msg.format(int))
```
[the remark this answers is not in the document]
ivanpauno: done

launch/launch/utilities/type_utils.py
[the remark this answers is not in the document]
ivanpauno: take a look to the new test cases

launch/launch/utilities/type_utils.py
```diff
@@ -0,0 +1,451 @@
...
+    types_in_list: Set[Optional[Type[Union[str, int, float, bool, Substitution]]]] = set()
+    for x in value:
+        if isinstance(x, ScalarTypesTuple):
+            types_in_list.add(type(x))
+        elif is_substitution(x):
+            types_in_list.add(Substitution)
+        else:
+            types_in_list.add(None)
+    if None in types_in_list:
+        raise TypeError(
```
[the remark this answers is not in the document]
ivanpauno: I think I've addressed this

ivanpauno:
Ok, this one is completely ready now.

launch/launch/utilities/type_utils.py
```diff
@@ -0,0 +1,565 @@
...
+        assert len(member_types) in (1, 2)
+        valid_types = tuple(member_types)
+        member_types.discard(str)
+        try:
+            member_type = member_types.pop()
+        except KeyError:
+            member_type = str
+        return (
+            all(isinstance(x, valid_types) for x in value[1:]) and
+            member_type in ScalarTypesTuple
```
hidmic, line 193: @ivanpauno meta, safe to ignore: wouldn't doing the check by construction be simpler? That is:

```python
if instance(value, list):
    if not value: 
       return True
    member_type = type(value[0])
    if member_type not in ScalarTypesTuple:
       return False
    valid_types = (member_type, str) if can_be_str else member_type
    return (all(isinstance(x, valid_types) for x in value[1:])
```

launch/launch/utilities/type_utils.py
```diff
@@ -0,0 +1,565 @@
...
+def is_instance_of(
+    value: Any,
+    data_type: Optional[AllowedTypesType] = None,
+    can_be_str: bool = False,
+) -> bool:
+    """
+    Check if `value` is instance of `data_type`.
+
+    :param value: variable to check.
+    :param data_type: class that `value` should be instance of.
```
hidmic, line 207: @ivanpauno nit:
```suggestion
    :param data_type: type that `value` should be an instance of.
```

launch/launch/utilities/type_utils.py
```diff
@@ -0,0 +1,565 @@
...
+        if not is_instance_of_valid_type(output, can_be_str):
+            raise ValueError(
+                f'{error_msg}: output type is not allowed, got {type(output)}'
+            )
+        return output
+
+    ensure_argument_type(value, str, 'value', 'coerce_to_type')
+    if data_type is None:
+        # No type specified, return best conversion.
+        return convert_as_yaml(value, f"Failed to convert '{value}' using yaml rules")
```
hidmic, line 260: @ivanpauno meta, safe to ignore: looking at this function, it seems unconditionally converting to `yaml` and doing type checks if `data_type` is not `None` may be simpler.

launch/launch/utilities/type_utils.py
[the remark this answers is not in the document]
hidmic: Oops, yeah, I misread the code initially.

launch/launch/utilities/type_utils.py
```diff
@@ -0,0 +1,565 @@
...
+    )
+
+
+def normalize_typed_substitution(
+    value: SomeValueType, data_type: Optional[AllowedTypesType]
+) -> NormalizedValueType:
+    """
+    Normalize a mix of substitution and values.
+
+    This function becomes handy when you need a substitution which result will be coerced to a
```
hidmic, line 378: @ivanpauno typo:
```suggestion
    This function becomes handy when you need a substitution whose result will be coerced to a
```

launch/launch/utilities/type_utils.py
```diff
@@ -0,0 +1,565 @@
...
+
+        def execute(self, context):
+            my_int = perform_typed_substitution(context, self.__my_int_normalized, int)
+            ...
+    ```
+
+    :param value: value to be normalized.
+    :param data_type: `value` can be either an instance of `data_type` or a substitution.
+        In the case of lists, `value` can be either a substitution or a list.
+        In the later case, its item should match the type specified by `data_type` or be a
```
hidmic, line 397: @ivanpauno typo: 
```suggestion
        In the later case, its items should match the type specified by `data_type` or be a
```

launch/launch/utilities/type_utils.py
```diff
@@ -0,0 +1,565 @@
...
+        except KeyError:
+            member_type = str
+        return (
+            all(isinstance(x, valid_types) for x in value[1:]) and
+            member_type in ScalarTypesTuple
+        )
+    return isinstance(value, ScalarTypesTuple)
+
+
+def is_instance_of(
```
hidmic, line 198: @ivanpauno meta, safe to ignore: this one's name is odd. I'd expect `is_instance_of` to be a variation of `isinstance` that supports a subset of `typing` generics and `is_instance_of_valid_type` to be sugar to call `is_instance_of` with `AllowedTypesTuple`.

launch/launch/utilities/type_utils.py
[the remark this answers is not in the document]
hidmic: I may be crazy, but `is_substitution()` implementation above appears to cover substitution lists that can be normalized:
```python
def is_substitution(x):
    return (
        isinstance(x, Substitution) or
        (
            isinstance(x, collections.abc.Iterable) and
            len(x) > 0 and
            all(isinstance(item, (Substitution, str)) for item in x) and
            any(isinstance(item, Substitution) for item in x)
        )
    )
```

launch/launch/utilities/type_utils.py
hidmic: Uh, I'm curious where and how it fails.

launch/launch/utilities/type_utils.py
```diff
@@ -0,0 +1,565 @@
...
+
+def perform_typed_substitution(
+    context: LaunchContext,
+    value: NormalizedValueType,
+    data_type: Optional[AllowedTypesType]
+) -> AllowedValueType:
+    """
+    Perform a normalized typed substitution.
+
+    This function becomes handy when you need a substitution which result will be coerced to a
```
hidmic, line 517: @ivanpauno typo:
```suggestion
    This function becomes handy when you need a substitution whose result will be coerced to a
```

launch/launch/utilities/type_utils.py
```diff
@@ -0,0 +1,565 @@
...
+                raise ValueError(
+                    'The input value is a list, cannot convert it to a scalar. '
+                    f"Got value='{value}', expected type {scalar_type}."
+                )
+        output = [
+            coerce_to_type(
+                perform_substitutions(context, cast(List[Substitution], x)), scalar_type)
+            if is_normalized_substitution(x) else x for x in value
+        ]
+        is_instance_of(output, data_type)
```
hidmic, line 559: @ivanpauno perhaps you wanted:
```suggestion
        assert is_instance_of(output, data_type)
```
instead? There's no side-effect to `is_instance_of` (nor should be IMO).

hidmic:
@ivanpauno I was thinking that we should deprecate and not remove what was in `launch.frontend.type_utils` i.e. redirect the call and warn that it has been relocated.

launch/launch/utilities/type_utils.py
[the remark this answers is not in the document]
ivanpauno: > Hmm, a mixed integer/float

The coercion functionality is flexible, and a substitution without a trailing `.` can be coerced to a `float`:
https://github.com/ros2/launch/blob/7cc161161eaff98b1f6cd3068148af9ed29fea00/launch/test/launch/utilities/test_type_utils.py#L133
But the checking functionality is more strict: https://github.com/ros2/launch/blob/7cc161161eaff98b1f6cd3068148af9ed29fea00/launch/test/launch/utilities/test_type_utils.py#L133

Python behaves in a similar way:
```python3
assert isinstance(3, float) == False
assert float('3') == 3.
```

If we want, we can make the system more flexible in the future.

launch/launch/frontend/entity.py
```diff
@@ -43,46 +45,40 @@ def get_attr(
...
-        List['Entity']
+        AllowedValueType,
+        List['Entity'],
     ]]:
         """
         Access an attribute of the entity.
 
         By default, it will try to return it as an string.
-        `types` states the expected types of the attribute. Type coercion or type checking is
+        `data_type` is the expected types of the attribute. Type coercion or type checking is
```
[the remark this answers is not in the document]
ivanpauno: I would say that the `set of types` part doesn't apply anymore.
I will use the singular:
```suggestion
        `data_type` is the expected type of the attribute. Type coercion or type checking is
```

launch/launch/frontend/entity.py
```diff
@@ -43,46 +45,40 @@ def get_attr(
...
         applied depending on the particular frontend.
 
-        The allowed types are:
-            - a scalar type i.e. `str`, `int`, `float`, `bool`;
-            - a uniform list i.e `List[str]`, `List[int]`, `List[float]`, `List[bool]`;
-            - a non-uniform list of known scalar types e.g. `List[Union[int, str]]`;
-            - a non-uniform list of any scalar type i.e. `list` or `List`;
-            - a `Union` of any of the above;
-            - `List[Entity]`, see below.
+        See :py:obj:`launch.utilities.AllowedTypesTuple` to see what literals `data_type` can be.
```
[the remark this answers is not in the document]
ivanpauno:
```suggestion
        See :py:obj:`launch.utilities.AllowedTypesTuple` to see what types are allowed.
```
I will apply the above suggestion.

launch/launch/frontend/parse_substitution.py
```diff
@@ -112,3 +114,37 @@ def parse_substitution(string_value):
...
+    """
+    Parse substitutions in `value`, if there are any, and return a normalized value type.
+
+    If `value` is a `str`, substitutions will be interpolated in it.
+    If `value` is a another type of scalar, it will be returned as-is.
+    If `value` is a list, the two rules above are applied to each item.
+    When interpolating substitutions in a string, `TextSubstitution` instances are resolved
+    and the original `str` is left.
+
+    :raise: `ValueError` if the result cannot be performed to a valid type.
```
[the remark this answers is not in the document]
ivanpauno: That sounds better.
```suggestion
    :raise: `ValueError` if the result cannot be parsed into a valid type.
```

launch/launch/utilities/type_utils.py
ivanpauno:
```python3
    member_type = type(value[0])
```
That line is a problem when `can_be_str == True`.
e.g.: `is_instance_of_valid_type(['$(var my_int)', 5], True)` will return `False`. 

The current logic is a bit hard to read, if you have a different way of improving the function let me know.

launch/launch/utilities/type_utils.py
ivanpauno: In some cases, the yaml conversion isn't used.
e.g.: for `int` or `float`

launch/launch/utilities/type_utils.py
ivanpauno: Hmm, I will ignore it for the moment, but we can improve that in the future.

ivanpauno:
> launch.frontend.type_utils

There's still a module called in that way, though I've moved most of the methods.
Also, a lots of the methods have been modified and renamed.

launch/launch/utilities/type_utils.py
ivanpauno: https://github.com/ros2/launch/pull/438/commits/cd3fafeca6102d69dc480c3a887ea74a2e659e8a

launch/launch/frontend/type_utils.py
```diff
@@ -12,241 +12,36 @@
...
+        issubclass(data_type.__args__[0], Entity)  # type: ignore
+
+
+def get_data_type_from_identifier(type_identifier: Text):
+    mapping = {
+        'str': str,
+        'bool': bool,
+        'float': float,
+        'int': int,
+        'strs_list': List[str],
```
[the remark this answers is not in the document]
ivanpauno: See https://github.com/ros2/launch/pull/438/commits/0e97e74eba5bb9a945820724830d543c8a0596b8.

launch/launch/utilities/type_utils.py
```diff
@@ -0,0 +1,565 @@
...
+        if not value:
+            return True  # Accept empty lists.
+        member_types = {type(x) for x in value}
+        if len(member_types) == 2:
+            if can_be_str is False or str not in member_types:
+                return False
+        elif len(member_types) > 2:
+            return False
+        else:
+            assert len(member_types) == 1
```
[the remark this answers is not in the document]
ivanpauno: Done https://github.com/ros2/launch/pull/438/commits/c2dacfcdee3fbbed3edd92b84741178b3f1f3e75.

launch/launch/utilities/type_utils.py
ivanpauno: Fixed in https://github.com/ros2/launch/pull/438/commits/cd3fafeca6102d69dc480c3a887ea74a2e659e8a.

launch/launch/utilities/type_utils.py
hidmic: Good point, `member_type = next((type(x) for x in value if type(x) is not str), str)` ?

launch/launch/utilities/type_utils.py
hidmic: Right, but it could be i.e. it'd have the same effect.

launch/launch/utilities/type_utils.py
ivanpauno: Thanks! That looks much nicer.
See https://github.com/ros2/launch/pull/438/commits/ab35da787620e63b4dbe05f59e62b0a08e920bf4.

launch/launch/utilities/type_utils.py
```diff
@@ -0,0 +1,567 @@
...
+    assert is_typing_list(typing.List[int])
+    assert not is_typing_list(int)
+    ```
+
+    """
+    return data_type is List or (
+        hasattr(data_type, '__origin__') and
+        hasattr(data_type, '__args__') and
+        data_type.__origin__ in  # type: ignore
+        (list, List) and  # On Linux/Mac is List, on Windows is list.
```
hidmic, line 122: Note to myself: this is crazy. I wonder why.

launch/launch/utilities/type_utils.py
```diff
@@ -0,0 +1,567 @@
...
+            ...
+    ```
+
+    List of substitutions coerced a list to strings can be confused with a single substitution
+    that is coerced to a list of strings.
+    e.g.: `['asd', TextSubstitution(text='bsd')]`
+    To avoid confusions, the passed value will always be interpreted as a single substitution
+    if possible, like in the above example (which will resolved to `'asdbsd'`).
+    To make it a list of substitutions:
+    `['asd', [TextSubstitution(text='bsd')]]`
```
hidmic, line 386: +1

launch/test/launch/utilities/test_type_utils.py
```diff
@@ -0,0 +1,578 @@
...
+    assert is_instance_of_valid_type_impl([1., 2.])
+    assert is_instance_of_valid_type_impl(['asd', 'bsd'])
+    assert is_instance_of_valid_type_impl([True, False])
+
+    assert not is_instance_of_valid_type_impl([1, '2'])
+    assert not is_instance_of_valid_type_impl(object)
+    assert not is_instance_of_valid_type_impl(test_is_instance_of_valid_type)
+    assert not is_instance_of_valid_type_impl({'key': 'value'})
+
+    assert is_instance_of_valid_type_impl(1, True)
```
hidmic, line 104: @ivanpauno nit: here and below, using a `can_be_str` kwarg helps readability.

launch/test/launch/utilities/test_type_utils.py
```diff
@@ -0,0 +1,578 @@
...
+    assert coerce_to_type_impl('tRuE', data_type=str) == 'tRuE'
+    assert coerce_to_type_impl("'1'", data_type=str) == "'1'"
+    assert coerce_to_type_impl("'off'", data_type=str) == "'off'"
+    assert coerce_to_type_impl("''1''", data_type=str) == "''1''"
+    assert coerce_to_type_impl('{1}', data_type=str) == '{1}'
+
+    assert coerce_to_type_impl('1', data_type=int) == 1
+    assert coerce_to_type_impl('1000', data_type=int) == 1000
+
+    assert coerce_to_type_impl('1', data_type=float) == 1.0
```
hidmic, line 226: @ivanpauno hmm, so coercion is flexible enough to accept an integer representation when `float` is explicitly given as the expected data type, but otherwise fails (even when a `List[float]` is specified)? I find it surprising at least.

launch/launch/utilities/type_utils.py
ivanpauno: This code was already there before this PR.
I think it wasn't platform dependent, but we where using at that time a different Python version on Windows.

The Python `typing` module is super new and still being developed, so it's not weird that this kind of detail changed.
I guess that the latest version is using `List`, and not `list`.

hidmic: > The Python typing module is super new and still being developed, so it's not weird that this kind of detail changed.

Yeah, that I know. To have a type change on a platform-basis is, nonetheless, curious. Looking at the `typing` module sources, it looks pretty platform-independent to me.

launch/test/launch/utilities/test_type_utils.py
ivanpauno: done.
See https://github.com/ros2/launch/pull/438/commits/150deca7d9c9b37f030cce350450dadd0c623b0f and https://github.com/ros2/launch/pull/438/commits/99cc84c1e1d89ad706f2cbc2ddd3d1acc021ab98.

launch/launch/utilities/type_utils.py
ivanpauno: > Yeah, that I know. To have a type change on a platform-basis is, nonetheless, curious. Looking at the typing module sources, it looks pretty platform-independent to me.

Yes, it's platform independent, but it changed between Python versions.
The comment is misleading, I will correct it.

ivanpauno: Done in https://github.com/ros2/launch/pull/438/commits/9ab3835a7d4f37b212858fd0744ca409bc424cb0

launch/test/launch/utilities/test_type_utils.py
ivanpauno: > @ivanpauno hmm, so coercion is flexible enough to accept an integer representation when float is explicitly given as the expected data type, but otherwise fails (even when a List[float] is specified)? I find it surprising at least.


I think that when `List[float]` is explicitly specified we could accept both ints and floats members.
IMO, we can make this more permissive in the future.

hidmic: Yeap. I'd be slightly inclined towards making it consistent (uniform) no, but I won't block on it.

launch/launch/utilities/type_utils.py
hidmic: > Yes, it's platform independent, but it changed between Python versions.

Aha! Now it makes sense.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11679/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6792/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9543/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11644/)

dirk-thomas:
It seems that this change resulted in a regression across all platforms: see e.g. https://ci.ros2.org/view/nightly/job/nightly_linux_debug/1641/testReport/test_launch_ros.test.test_launch_ros.frontend/test_node_frontend/test_launch_remapping_xml/

@ivanpauno Please address asap.

PR #440, opened 2020-07-16
fix max() with empty sequence

dirk-thomas:
The nightly Windows builds are failing during the tests of `ros2bag` with the following internal error in `pytest`:

```
INTERNALERROR> Traceback (most recent call last):
...
INTERNALERROR>   File "C:\ci\ws\install\Lib\site-packages\launch_testing\pytest\hooks.py", line 41, in __init__
INTERNALERROR>     max_length = max(
INTERNALERROR> ValueError: max() arg is an empty sequence
```

I can't say *why* the sequence is empty in this case, But this patch fixes the logic to handle that case.

Windows build: [Build Status](https://ci.ros2.org/job/ci_windows/11393/)

### DIFF
--- a/launch_testing/launch_testing/pytest/hooks.py
+++ b/launch_testing/launch_testing/pytest/hooks.py
@@ -38,11 +38,12 @@ class LaunchTestFailureRepr:
     """A `_pytest._code.code.ExceptionReprChain`-like object."""
 
     def __init__(self, failures):
-        max_length = max(
-            len(line)
+        lines = [
+            line
             for _, error_description in failures
             for line in error_description.splitlines()
-        )
+        ]
+        max_length = max(len(line) for line in lines) if lines else 3
         thick_sep_line = '=' * max_length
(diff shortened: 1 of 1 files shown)


ipa-hsd:
Will this fix be back-ported for `foxy`? I am using `launch_testing` for a [dummy package](https://github.com/ipa-hsd/ros2_tests/blob/main/pub_sub_tests/test/test_pub_sub_launch.py) and I get [ValueError: max() arg is an empty sequence](https://github.com/ipa-hsd/ros2_tests/runs/2094470484?check_suite_focus=true#step:4:839). When I changed to `rolling`, the behaviour is as [expected](https://github.com/ipa-hsd/ros2_tests/runs/2095249595?check_suite_focus=true#step:4:1205).

hidmic:
See https://github.com/ros2/launch/pull/491.

ipa-hsd:
Thanks for the quick response!

PR #443, opened 2020-07-30
Handle empty strings in type coercion.

hidmic:
Follow-up after #438.

### DIFF
--- a/launch/launch/utilities/type_utils.py
+++ b/launch/launch/utilities/type_utils.py
@@ -231,6 +231,10 @@ def coerce_to_type(
     :return: `value` coerced to `data_type`.
     """
     def convert_as_yaml(value, error_msg):
+        # Forward empty strings
+        if not value:
+            return value
+
         try:
             output = yaml.safe_load(value)
         except Exception as err:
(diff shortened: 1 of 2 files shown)


launch/launch/utilities/type_utils.py
```diff
@@ -231,6 +231,10 @@ def coerce_to_type(
     :return: `value` coerced to `data_type`.
     """
     def convert_as_yaml(value, error_msg):
+        # Forward empty strings
+        if not value:
```
dirk-thomas, line 235: I wondering if this need to match what the comment suggests. This condition will be true for various other kind of input, like `0`, `None`, an empty list/set/dict?

hidmic:
CI up to `launch` and  `test_launch_ros`:

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/11703/)
* Linux aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/6800/)
* MacOS [Build Status](https://ci.ros2.org/job/ci_osx/9554/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/11652/)

launch/launch/utilities/type_utils.py
dirk-thomas: Nvm, the function argument is supposed to be a string.

hidmic: If you look at line 251, it is ensured that value is a `str` object.

PR #445, opened 2020-08-04
Fix type_utils.extract_type() function.

hidmic:
Deal with `typing.List`'s without arguments.

CentOS CI [Build Status](https://ci.ros2.org/job/ci_linux-centos/74/)


hidmic:
Going in!

PR #449, opened 2020-08-11
Refactor launch service run_async loop to wait on futures and queued events

jacobperron:
Fixes https://github.com/ros2/launch_ros/issues/169

Otherwise, it's possible to run into a race condition between checking for 'idle' and processing events from the queue.

**Edit:**

As described in https://github.com/ros2/launch/pull/449#issuecomment-672209093, we can **not** be idle because we're waiting for futures to complete, but if there are no more events in the queue then we end up blocking forever.

To resolve this corner-case that results in deadlock, this PR refactors the run_async loop such that we concurrently wait on futures and events in the queue.

cc/ @wjwwood

### DIFF
--- a/launch/launch/launch_context.py
+++ b/launch/launch/launch_context.py
@@ -182,6 +182,7 @@ class LaunchContext:
 
     def emit_event_sync(self, event: Event) -> None:
         """Emit an event synchronously."""
+        self.__logger.debug("emitting event synchronously: '{}'".format(event.name))
         with self._event_queue_lock:
             self._event_queue.put_nowait(event)
 
--- a/launch/launch/launch_context.py
+++ b/launch/launch/launch_context.py
@@ -16,6 +16,7 @@
 
 import asyncio
 import collections
+from threading import Lock
 from typing import Any
 from typing import Dict
 from typing import Iterable
@@ -42,6 +43,7 @@ class LaunchContext:
         self.__argv = argv if argv is not None else []
 
         self._event_queue = asyncio.Queue()  # type: asyncio.Queue
+        self._event_queue_lock = Lock()
         self._event_handlers = collections.deque()  # type: collections.deque
         self._completion_futures = []  # type: List[asyncio.Future]
 
@@ -73,6 +75,15 @@ class LaunchContext:
 
     def _set_asyncio_loop(self, loop: asyncio.AbstractEventLoop) -> None:
         self.__asyncio_loop = loop
+
+    def _set_event_queue(self, new_queue: asyncio.Queue) -> None:
+        with self._event_queue_lock:
+            while True:
+                try:
+                    new_queue.put_nowait(self._event_queue.get_nowait())
+                except asyncio.QueueEmpty:
+                    break
+            self._event_queue = new_queue
 
     @property
     def asyncio_loop(self):
@@ -171,13 +182,24 @@ class LaunchContext:
 
     def emit_event_sync(self, event: Event) -> None:
         """Emit an event synchronously."""
-        self.__logger.debug("emitting event synchronously: '{}'".format(event.name))
-        self._event_queue.put_nowait(event)
+        with self._event_queue_lock:
+            self._event_queue.put_nowait(event)
 
     async def emit_event(self, event: Event) -> None:
         """Emit an event."""
-        self.__logger.debug("emitting event: '{}'".format(event.name))
-        await self._event_queue.put(event)
+        with self._event_queue_lock:
+            self.__logger.debug("emitting event: '{}'".format(event.name))
+            await self._event_queue.put(event)
+
+    def is_event_queue_empty(self) -> bool:
+        with self._event_queue_lock:
+            return self._event_queue.empty()
+
+    async def get_next_event(self):
+        with self._event_queue_lock:
+            if self._event_queue.empty():
+                return None
+            return await self._event_queue.get()
 
     def perform_substitution(self, substitution: Substitution) -> Text:
         """Perform substitution on given Substitution."""
--- a/launch/launch/launch_context.py
+++ b/launch/launch/launch_context.py
@@ -16,7 +16,6 @@
 
 import asyncio
 import collections
-from threading import Lock
 from typing import Any
 from typing import Dict
 from typing import Iterable
@@ -43,7 +42,6 @@ class LaunchContext:
         self.__argv = argv if argv is not None else []
 
         self._event_queue = asyncio.Queue()  # type: asyncio.Queue
-        self._event_queue_lock = Lock()
         self._event_handlers = collections.deque()  # type: collections.deque
         self._completion_futures = []  # type: List[asyncio.Future]
 
@@ -75,15 +73,6 @@ class LaunchContext:
 
     def _set_asyncio_loop(self, loop: asyncio.AbstractEventLoop) -> None:
         self.__asyncio_loop = loop
-
-    def _set_event_queue(self, new_queue: asyncio.Queue) -> None:
-        with self._event_queue_lock:
-            while True:
-                try:
-                    new_queue.put_nowait(self._event_queue.get_nowait())
-                except asyncio.QueueEmpty:
-                    break
-            self._event_queue = new_queue
 
     @property
     def asyncio_loop(self):
@@ -183,24 +172,12 @@ class LaunchContext:
     def emit_event_sync(self, event: Event) -> None:
         """Emit an event synchronously."""
         self.__logger.debug("emitting event synchronously: '{}'".format(event.name))
-        with self._event_queue_lock:
-            self._event_queue.put_nowait(event)
+        self._event_queue.put_nowait(event)
 
     async def emit_event(self, event: Event) -> None:
         """Emit an event."""
-        with self._event_queue_lock:
-            self.__logger.debug("emitting event: '{}'".format(event.name))
-            await self._event_queue.put(event)
-
-    def is_event_queue_empty(self) -> bool:
-        with self._event_queue_lock:
-            return self._event_queue.empty()
-
-    async def get_next_event(self):
-        with self._event_queue_lock:
-            if self._event_queue.empty():
-                return None
-            return await self._event_queue.get()
+        self.__logger.debug("emitting event: '{}'".format(event.name))
+        await self._event_queue.put(event)
 
     def perform_substitution(self, substitution: Substitution) -> Text:
         """Perform substitution on given Substitution."""
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -155,7 +155,7 @@ class LaunchService:
     def _is_idle(self):
         number_of_entity_future_pairs = self._prune_and_count_entity_future_pairs()
         number_of_entity_future_pairs += self._prune_and_count_context_completion_futures()
-        return number_of_entity_future_pairs == 0 and self.__context.is_event_queue_empty()
+        return number_of_entity_future_pairs == 0 and self.__context._event_queue.empty()
 
     @contextlib.contextmanager
     def _prepare_run_loop(self):
@@ -174,7 +174,13 @@ class LaunchService:
                 # Set the asyncio loop for the context.
                 self.__context._set_asyncio_loop(this_loop)
                 # Recreate the event queue to ensure the same event loop is being used.
-                self.__context._set_event_queue(asyncio.Queue())
+                new_queue = asyncio.Queue()
+                while True:
+                    try:
+                        new_queue.put_nowait(self.__context._event_queue.get_nowait())
+                    except asyncio.QueueEmpty:
+                        break
+                self.__context._event_queue = new_queue
 
                 self.__loop_from_run_thread = this_loop
 
@@ -263,9 +269,8 @@ class LaunchService:
                 self.__shutting_down = False
 
     async def _process_one_event(self) -> None:
-        next_event = await self.__context.get_next_event()
-        if next_event is not None:
-            await self.__process_event(next_event)
+        next_event = await self.__context._event_queue.get()
+        await self.__process_event(next_event)
 
     async def __process_event(self, event: Event) -> None:
         self.__logger.debug("processing event: '{}'".format(event))
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -155,7 +155,7 @@ class LaunchService:
     def _is_idle(self):
         number_of_entity_future_pairs = self._prune_and_count_entity_future_pairs()
         number_of_entity_future_pairs += self._prune_and_count_context_completion_futures()
-        return number_of_entity_future_pairs == 0 and self.__context._event_queue.empty()
+        return number_of_entity_future_pairs == 0 and self.__context.is_event_queue_empty()
 
     @contextlib.contextmanager
     def _prepare_run_loop(self):
@@ -174,13 +174,7 @@ class LaunchService:
                 # Set the asyncio loop for the context.
                 self.__context._set_asyncio_loop(this_loop)
                 # Recreate the event queue to ensure the same event loop is being used.
-                new_queue = asyncio.Queue()
-                while True:
-                    try:
-                        new_queue.put_nowait(self.__context._event_queue.get_nowait())
-                    except asyncio.QueueEmpty:
-                        break
-                self.__context._event_queue = new_queue
+                self.__context._set_event_queue(asyncio.Queue())
 
                 self.__loop_from_run_thread = this_loop
 
@@ -269,8 +263,9 @@ class LaunchService:
                 self.__shutting_down = False
 
     async def _process_one_event(self) -> None:
-        next_event = await self.__context._event_queue.get()
-        await self.__process_event(next_event)
+        next_event = await self.__context.get_next_event()
+        if next_event is not None:
+            await self.__process_event(next_event)
 
     async def __process_event(self, event: Event) -> None:
         self.__logger.debug("processing event: '{}'".format(event))
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -342,6 +342,11 @@ class LaunchService:
 
                     # Stop running if we're shutting down and there's no more work
                     if self.__shutting_down and is_idle:
+                        if (
+                            process_one_event_task is not None and
+                            not process_one_event_task.done()
+                        ):
+                            process_one_event_task.cancel()
                         break
 
                     # Collect futures to wait on
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -329,6 +329,7 @@ class LaunchService:
                 return loop.default_exception_handler(context)
             this_loop.set_exception_handler(_on_exception)
 
+            process_one_event_task = None
             while True:
                 try:
                     # Check if we're idle, i.e. no on-going entities (actions) or events in
@@ -347,9 +348,12 @@ class LaunchService:
                     entity_futures = [pair[1] for pair in self._entity_future_pairs]
                     entity_futures.extend(self.__context._completion_futures)
 
-                    # Create a task to process any events in the queue
-                    # and add it to the list of awaitables
-                    process_one_event_task = this_loop.create_task(self._process_one_event())
+                    # If the current task is done, create a new task to process any events
+                    # in the queue
+                    if process_one_event_task is None or process_one_event_task.done():
+                        process_one_event_task = this_loop.create_task(self._process_one_event())
+
+                    # Add the process event task to the list of awaitables
                     entity_futures.append(process_one_event_task)
 
                     # Wait on events and futures until an event is processed or a timeout occurs
@@ -362,7 +366,6 @@ class LaunchService:
                         if process_one_event_task in done:
                             break
                         elif len(done) == 0:
-                            process_one_event_task.cancel()
                             break
                         entity_futures = set(entity_futures).difference(done)
 
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -349,16 +349,23 @@ class LaunchService:
 
                     # If there is an event in the queue, create a task to process it
                     # and add it to the list of awaitables
+                    process_one_event_task = None
                     if not self.__context._event_queue.empty():
                         process_one_event_task = this_loop.create_task(self._process_one_event())
                         entity_futures.append(process_one_event_task)
 
                     if len(entity_futures) > 0:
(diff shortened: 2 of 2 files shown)


jacobperron:
See https://github.com/ros2/launch_ros/issues/169 for steps to reproduce the issue.

launch/launch/launch_context.py
```diff
@@ -171,13 +182,24 @@ def unregister_event_handler(self, event_handler: BaseEventHandler) -> None:
...
+            self.__logger.debug("emitting event: '{}'".format(event.name))
+            await self._event_queue.put(event)
+
+    def is_event_queue_empty(self) -> bool:
+        with self._event_queue_lock:
+            return self._event_queue.empty()
+
+    async def get_next_event(self):
+        with self._event_queue_lock:
+            if self._event_queue.empty():
```
ivanpauno, line 200: what if [shutdown_when_idle](https://github.com/ros2/launch/blob/657745c315ba6a61984b66f168f9c34b3a7e2108/launch/launch/launch_service.py#L302) was false?

This shouldn't be returning immediately in that case, but awaiting somebody to put a new event in the queue.
Maybe, you can add an extra argument to the method `return_when_empty` or something like that.

ivanpauno:
Actually, I'm not sure this is the right way to handle the problem.

You can currently emit an event thread safely if you want, e.g.:
https://github.com/ros2/launch/blob/657745c315ba6a61984b66f168f9c34b3a7e2108/launch/launch/launch_service.py#L115-L118
So, I'm not sure if the event Queue should have a thread safe protection.

launch/launch/launch_context.py
```diff
@@ -171,13 +182,24 @@ def unregister_event_handler(self, event_handler: BaseEventHandler) -> None:
 
     def emit_event_sync(self, event: Event) -> None:
         """Emit an event synchronously."""
-        self.__logger.debug("emitting event synchronously: '{}'".format(event.name))
```
wjwwood, line 174: Why remove this debug line?

launch/launch/launch_context.py
wjwwood: Regardless of that setting, it will cause a busy loop involving `LaunchService._process_one_event()`.

wjwwood:
Reading the code I'm not sure how the two lines you pointed out @jacobperron could ever be run concurrently, but maybe I'm not using my imagination properly :)

launch/launch/launch_context.py
jacobperron: Yeah, I was worried about a busy loop here. I think a better alternative is to somehow add an event to the queue if there are no more futures to wait on and the queue is empty.

jacobperron:
> Reading the code I'm not sure how the two lines you pointed out @jacobperron could ever be run concurrently, but maybe I'm not using my imagination properly :)

Looking again, I might be wrong about the reason for the hange.

It looks like `is_idle` evaluates to `False` on this line:

https://github.com/ros2/launch/blob/40f3be3c42ca9f6490813642aa75ffb1012c99a5/launch/launch/launch_service.py#L332

Which means either the queue is not empty or there are still futures to wait on. But when we get to processing events (from the queue) the queue is empty and we block forever on this line:

https://github.com/ros2/launch/blob/40f3be3c42ca9f6490813642aa75ffb1012c99a5/launch/launch/launch_service.py#L357

launch/launch/launch_context.py
jacobperron: It was not intentional. Added back 098b56a

jacobperron:
It looks like the future created here is never marked as done, even though logging suggests the function is run to completion:

https://github.com/ros2/launch_ros/blob/8c05ebb4be52dcd360e47694875643e4f53a010e/launch_ros/launch_ros/actions/load_composable_nodes.py#L173

ivanpauno:
> It looks like the future created here is never marked as done, even though logging suggests the function is run to completion:
> 
> https://github.com/ros2/launch_ros/blob/8c05ebb4be52dcd360e47694875643e4f53a010e/launch_ros/launch_ros/actions/load_composable_nodes.py#L173

mmmm, so [run_in_executor](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor) is using the default executor, which seems to actually be multi-threaded (reference needed here).
So, maybe that's the problem (?) ...

ivanpauno:
> reference needed here

Reference found: https://github.com/python/cpython/blob/374d998b507d34a6c0a3816a163926a8ba0c483f/Lib/asyncio/base_events.py#L808-L812.
IMO, that's the problem

jacobperron:
Yeah, the [docs say](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.set_default_executor) the default executor must be a [ThreadPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor).

ivanpauno:
What I think it's happening is the following:

- There's no more events to be processed
- We are waiting for some completion feature to be done
- the launch service awaits for a new event instead of waiting for the remaining completion futures to be done: https://github.com/ros2/launch/blob/657745c315ba6a61984b66f168f9c34b3a7e2108/launch/launch/launch_service.py#L341
https://github.com/ros2/launch/blob/657745c315ba6a61984b66f168f9c34b3a7e2108/launch/launch/launch_service.py#L271-L273
  Thus, it waits for ever.

Easy hack: Using [get](https://docs.python.org/3/library/queue.html#queue.Queue.get) with a timeout.
Not ideal, but I will need to think for real to figure out a better solution :joy:.

ivanpauno:
> Yeah, the [docs say](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.set_default_executor) the default executor must be a [ThreadPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor).

I like looking at the code :joy:

jacobperron:
> Easy hack: Using get with a timeout.

Unfortunately, [`asyncio.Queue.get()`](https://docs.python.org/3/library/asyncio-queue.html#asyncio.Queue.get) does not have a timeout argument..

wjwwood:
That was my initial theory too but the shutdown event didn’t clear the hang. 

A better solution would be to have a noop event like “check for idle” or “future completed” that would cause get to return and would be posted after the future is completed.

ivanpauno:
I think I have figured out a fix (it works well in my mind at least):

The issue:

[_is_iddle](https://github.com/ros2/launch/blob/657745c315ba6a61984b66f168f9c34b3a7e2108/launch/launch/launch_service.py#L158) is returning true if there is either a future available or an event has not be processed. But we are then blocking on getting a new event (see [1](https://github.com/ros2/launch/blob/657745c315ba6a61984b66f168f9c34b3a7e2108/launch/launch/launch_service.py#L362) and [2](https://github.com/ros2/launch/blob/657745c315ba6a61984b66f168f9c34b3a7e2108/launch/launch/launch_service.py#L272)).

We could rather do the following:
- Collect all futures like [here](https://github.com/ros2/launch/blob/657745c315ba6a61984b66f168f9c34b3a7e2108/launch/launch/launch_service.py#L347-L349).
- Wait on the first completed future like [here](https://github.com/ros2/launch/blob/657745c315ba6a61984b66f168f9c34b3a7e2108/launch/launch/launch_service.py#L352-L356).

If we always do that, not only after shutdown has happened, we should never run into the race (as we would check again `_is_idle` after any future has completed).

wjwwood:
@ivanpauno how do you wait on events in the queue and the incomplete futures at the same time?

ivanpauno:
> @ivanpauno how do you wait on events in the queue and the incomplete futures at the same time?

Using `asyncio.wait` with `asyncio.FIRST_COMPLETED` condition:

https://github.com/ros2/launch/blob/657745c315ba6a61984b66f168f9c34b3a7e2108/launch/launch/launch_service.py#L352-L356

IIUC, that allows you to wait on a collection of awaitables at the same time, and it will return when one of them was completed.
In that way, we can check when any future is completed if the launch service is idle or not.

We should take care of not processing another event if `wait` of another reason, and not because an event was processed. But that can be checked.

wjwwood:
Ok, so the set of things being waited on in each loop would be `await process_one_event()` + the context's futures + the action futures? Some of the logic might need to change, but it sounds like it should work.

jacobperron:
Sounds good to me. I'll give it a try.

jacobperron:
Almost seems to work; now I get stuck in the inner shutdown loop. Maybe I've hit a different bug related to shutdown, but I don't understand why there is a `continue` here: https://github.com/ros2/launch/blob/657745c315ba6a61984b66f168f9c34b3a7e2108/launch/launch/launch_service.py#L340

What I see is this condition evaluates to `True`:

https://github.com/ros2/launch/blob/657745c315ba6a61984b66f168f9c34b3a7e2108/launch/launch/launch_service.py#L337

Then we call `_shutdown` and `continue`. And the next iteration, `_is_idle()` evaluates to `False` (I guess because the shutdown causes another event to be added to the queue):

https://github.com/ros2/launch/blob/657745c315ba6a61984b66f168f9c34b3a7e2108/launch/launch/launch_service.py#L336

Then we get stuck waiting to get something from the queue and repeatedly see this debug:

https://github.com/ros2/launch/blob/657745c315ba6a61984b66f168f9c34b3a7e2108/launch/launch/launch_service.py#L359

I guess this logic is broken (I don't know under what circumstance this is True):

https://github.com/ros2/launch/blob/657745c315ba6a61984b66f168f9c34b3a7e2108/launch/launch/launch_service.py#L344-L346

If I remove the `continue`, my test works as expected, but maybe there are other unintended consequences.

wjwwood:
> Then we call _shutdown and continue. And the next iteration, _is_idle() evaluates to False (I guess because the shutdown causes another event to be added to the queue):

That continue ensure that `self.__shutting_down` is true:

https://github.com/ros2/launch/blob/657745c315ba6a61984b66f168f9c34b3a7e2108/launch/launch/launch_service.py#L342

At least I think that's the case. Maybe that's no longer the case.

ivanpauno:
@jacobperron can you put that code somewhere?

jacobperron:
I've reverted the original changes and instead refactored the `run_async` loop to wait on both queued events and futures (92bfb44091a2cebbe71c9f5a9775453939648668). I also added some logic to wait until the process_one_event task is done or a timeout occurs (19e0d68).

jacobperron:
This solves my original issue, but now the `test_launch_service.py` tests are hangning :upside_down_face:

jacobperron:
The test is emitting an event from a thread and it gets get stuck here:

https://github.com/ros2/launch/blob/19e0d6890e5c742fab5ed13c8a21dffe2a2c68f5/launch/test/launch/test_launch_service.py#L73

Launch service call stack ends up here:

https://github.com/ros2/launch/blob/19e0d6890e5c742fab5ed13c8a21dffe2a2c68f5/launch/launch/launch_service.py#L115-L117

and waiting forever for the result

https://github.com/ros2/launch/blob/19e0d6890e5c742fab5ed13c8a21dffe2a2c68f5/launch/launch/launch_service.py#L124-L125

jacobperron:
I guess the issue is that we nee to `await` always in the loop, in the case that there are no futures or events to wait on I'm running a busy loop.

jacobperron:
@ivanpauno @wjwwood Okay, I think I've got it now (1118883). PTAL at the latest changes.

launch/launch/launch_service.py
```diff
@@ -338,28 +338,34 @@ def _on_exception(loop, context):
...
+                    while True:
+                        done, pending = await asyncio.wait(
+                            entity_futures,
+                            timeout=1.0,
+                            return_when=asyncio.FIRST_COMPLETED
+                        )
+                        if process_one_event_task in done:
+                            break
+                        elif len(done) == 0:
                             process_one_event_task.cancel()
```
ivanpauno, line 365: why cancelling here? this was called only when shutdown happened

launch/launch/launch_service.py
```diff
@@ -338,28 +338,34 @@ def _on_exception(loop, context):
...
                             break
-                        entity_futures = [pair[1] for pair in self._entity_future_pairs]
-                        entity_futures.append(process_one_event_task)
-                        entity_futures.extend(self.__context._completion_futures)
-                        done = set()  # type: Set[asyncio.Future]
-                        while not done:
```
ivanpauno, line 351: before, we were waiting on all completion futures to be completed even when shutdown was called.
Now we aren't.

launch/launch/launch_service.py
jacobperron: It's possible that the `process_one_event_task` doesn't complete (timeout). Every time this happens, tasks were accumulating and during shutdown I would see all kinds of errors saying "task was destroyed but it is pending". Calling cancel here fixed those errors.

Before, we blocked forever until this task was completed so we didn't need to cancel.

launch/launch/launch_service.py
jacobperron: TBH, this condition was confusing to me. `while not done` will be true as soon as a single awaitable is done (the set is not empty), not all of them.

In this PR, we actually do wait for all futures, because `is_idle` will be `False` as long as there is still futures not done.

launch/launch/launch_service.py
ivanpauno: > It's possible that the process_one_event_task doesn't complete (timeout). Every time this happens, tasks were accumulating and during shutdown I would see all kinds of errors saying "task was destroyed but it is pending". Calling cancel here fixed those errors.

That is because `process_one_event_task` have to be cancelled if shutdown happened, but it shouldn't be cancelled just because nothing was done. We actually don't strictly need a timeout in the above wait.

launch/launch/launch_service.py
ivanpauno: > TBH, this condition was confusing to me. while not done will be true as soon as a single awaitable is done (the set is not empty), not all of them.

Yes, this is confusing because we're always waiting on the same future, it's not that we delete from the collection the ones that were already done ....

I really don't know what was the intention of the previous code

launch/launch/launch_service.py
jacobperron: `process_one_event_task` goes out of scope if we timeout, so we can't cancel it later during shutdown unless we hold a reference to it (maybe there's a way to get a reference from asynchio?). If I remove the timeout case:

```diff
- elif len(done) == 0:
-    process_one_event_task.cancel()
-    break
```

Then my tests block forever again. I think this is because all futures complete, but there are no events in the queue.

ivanpauno: > Then my tests block forever again. I think this is because all futures complete, but there are no events in the queue.

Yes, the idea is that you check if `is_idle` again in the loop, and `break` it if that happens.

jacobperron: We can check `is_idle` again in this loop, then we have to re-collect the entity futures. It seems a bit redundant. Instead, we could hold a reference to `process_one_event_task` and only create a new one if it is done. I guess the code would be a little bit simpler. But, I don't see the issue with canceling, unless you think it's an expensive operation.

jacobperron: We do want to avoid accidentally canceling the processing mid-event. See 9c4c641 for a fix.

jacobperron:
I did a minor refactor 41a31d2 and added an guard against a pending task during shutdown f33e68e. Let me know if there's anything else.

launch/launch/launch_service.py
```diff
@@ -338,28 +339,39 @@ def _on_exception(loop, context):
...
+                        process_one_event_task = this_loop.create_task(self._process_one_event())
+
+                    # Add the process event task to the list of awaitables
+                    entity_futures.append(process_one_event_task)
+
+                    # Wait on events and futures until an event is processed or a timeout occurs
+                    while True:
+                        done, pending = await asyncio.wait(
+                            entity_futures,
+                            timeout=1.0,
```
ivanpauno, line 368: is the timeout needed?

launch/launch/launch_service.py
```diff
@@ -338,28 +339,39 @@ def _on_exception(loop, context):
...
-                                timeout=1.0,
-                                return_when=asyncio.FIRST_COMPLETED
-                            )
-                            if not done:
-                                self.__logger.debug(
-                                    'still waiting on futures: {}'.format(entity_futures)
-                                )
-                    else:
-                        await process_one_event_task
+                        entity_futures = set(entity_futures).difference(done)
```
ivanpauno, line 373: can you use pending instead of doing this?
```suggestion
                        entity_futures = pending
```

launch/launch/launch_service.py
```diff
@@ -338,28 +339,39 @@ def _on_exception(loop, context):
...
-                        if is_idle:
+
+                    # Stop running if we're shutting down and there's no more work
+                    if self.__shutting_down and is_idle:
+                        if (
+                            process_one_event_task is not None and
+                            not process_one_event_task.done()
+                        ):
                             process_one_event_task.cancel()
+                        break
```
ivanpauno, line 350: are the shutdown event handlers going to be executed if we just break here?
should something be awaited after shutdown happens?

I have to think a bit more ... I haven't figured out what the previous code was doing.

launch/launch/launch_service.py
```diff
@@ -338,28 +339,39 @@ def _on_exception(loop, context):
...
+                    # If the current task is done, create a new task to process any events
+                    # in the queue
+                    if process_one_event_task is None or process_one_event_task.done():
+                        process_one_event_task = this_loop.create_task(self._process_one_event())
+
+                    # Add the process event task to the list of awaitables
+                    entity_futures.append(process_one_event_task)
+
+                    # Wait on events and futures until an event is processed or a timeout occurs
+                    while True:
```
ivanpauno, line 365: I think we don't need this inner loop.
We're not processing a new event if there's already one being processing (L358), so I think that the code would be simpler if we don't have an inner loop.

launch/launch/launch_service.py
jacobperron: I think this logic is equivalent to the code before:

https://github.com/ros2/launch/blob/f3f0d69d8b43eecbcf968ee06611960a33be6313/launch/launch/launch_service.py#L342-L346

If there are still events in the queue, then `is_idle` will be `False` and we'll continue to handle them.

launch/launch/launch_service.py
jacobperron: You're right, we can simplify. See b002b92

launch/launch/launch_service.py
jacobperron: Removed the inner loop (b002b92), so this does not apply anymore.

launch/launch/launch_service.py
jacobperron: Nope, removed in b002b92

launch/launch/launch_service.py
ivanpauno: you're right

launch/launch/launch_service.py
```diff
@@ -338,28 +339,34 @@ def _on_exception(loop, context):
...
-                                self.__logger.debug(
-                                    'still waiting on futures: {}'.format(entity_futures)
-                                )
-                    else:
-                        await process_one_event_task
+                        break
+
+                    # Collect futures to wait on
+                    entity_futures = [pair[1] for pair in self._entity_future_pairs]
+                    entity_futures.extend(self.__context._completion_futures)
```
ivanpauno, line 354: this is only another option, and not required:
```suggestion
                    entity_futures = []
                    if self.__context._event_queue.empty():
                        entity_futures = [pair[1] for pair in self._entity_future_pairs]
                        entity_futures.extend(self.__context._completion_futures)
```
this will avoid the checking in L358 about when creating a new task for processing an event, and will avoid the previous race.

Maybe it makes the code harder to understand, I don't mind strongly

jacobperron: I don't understand how it avoids the condition for creating a new task below. Or are you suggesting that makes it so we only have to wait for the task in the queue at L365, and it doesn't get spuriously woken up by another future completing?

hidmic:
Great finding! LGTM too with green CI.

launch/launch/launch_service.py
ivanpauno: > Or are you suggesting that makes it so we only have to wait for the task in the queue at L365, and it doesn't get spuriously woken up by another future completing?

That's exactly my idea, avoiding the spurious waits.
We only need those wakes when the queue is empty, if not we can be waiting for something to be put in the queue when nobody is going to do that.

> I don't understand how it avoids the condition for creating a new task below

Maybe, we still need that check.
Because I imagine that some of the completion futures can generate a new event when they are done.

---

The current code is fine. It avoids the hung, which is the important issue.

jacobperron: I've applied your suggestion in fc4e9ef

jacobperron:
Testing packages recursively depending on `launch`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11926/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6970/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9717/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11863/)

launch/launch/launch_service.py
```diff
@@ -338,28 +339,37 @@ def _on_exception(loop, context):
...
+                        process_one_event_task = this_loop.create_task(self._process_one_event())
+
+                    # Add the process event task to the list of awaitables
+                    entity_futures.append(process_one_event_task)
+
+                    # Wait on events and futures
+                    await asyncio.wait(
+                        entity_futures,
+                        return_when=asyncio.FIRST_COMPLETED
+                    )
```
wjwwood, line 371: I'm not sure taking the timeout loop out here is better. The debug message telling you that you're still waiting on some set of futures periodically was useful.

I agree the previous loop condition was odd, maybe it should be `while len(pending) > 0:`?

Something like:

```python
pending = list(entity_futures)
while pending:
    self.__logger.debug(...)
    done, pending = await asyncio.wait(entity_futures, return_when=asyncio.FIRST_COMPLETE)
```

wjwwood: Oh I see the shutdown loop has been removed completely and this is for normal runtime as well... nvm I guess.

jacobperron:
The Rpr job is failing due to an infrastructure issue. Since the failure is unrelated to this change and CI is green, I'll go ahead and merge.

Thanks for the reviews everyone!

jacobperron:
I've decided not to backport this to Eloquent or Dashing since there's been a lot of changes since they were last touched.

PR #450, opened 2020-08-13
Add pytest.ini so local tests don't display warning (#428)

jacobperron:
This should help resolve one of the outstanding failures in ros2/ci#503


jacobperron:
Technically, we only need this for the CMake package https://github.com/ros2/launch/pull/451

PR #453, opened 2020-08-14
Add new conditions for checking launch configuration values

### LINKED ISSUE #452: Condition for checking launch configuration value
#### Feature description

I've come across the need to condition launch actions based on the value of a launch configuration.

For example, in my particular case I want to only launch a node container if the name of a container is not provided (otherwise, launch an action to load nodes into the existing container):

```python
    arg_container_name = DeclareLaunchArgument(
        name='container', default_value=None,
        description=(
            'Name of an existing node container to load launched nodes into. '
            'Otherwise, a new container will be created.'
        )
    )   

    # If an existing container is not provided, start container and load nodes into it
    image_processing_container = ComposableNodeContainer(
        condition=LaunchConfigurationEquals('container', None),  # We need a condition here
        name='image_proc_container',
        namespace='',
        package='rclcpp_components',
        executable='component_container',
    )
```

In my example, I only care if the value is `None` or not, but I can imagine being able to condition on any value would be useful.

I was able to achieve the behavior I want by wrapping the check in a `PythonExpression`:

```python
    # Sentinel value for unset 'container' argument
    container_name_not_set = '<container name not set>'

    arg_container_name = DeclareLaunchArgument(
        name='container', default_value=container_name_not_set,
        description=(
            'Name of an existing node container to load launched nodes into. '
            'Otherwise, a new container will be created.'
        )
    )   

    is_container_set = PythonExpression(["'", LaunchConfiguration('container'), f"' != '{container_name_not_set}'"])

    # If an existing container is not provided, start container and load nodes into it
    image_processing_container = ComposableNodeContainer(
        condition=UnlessCondition(is_container_set),
        name='image_proc_container',
        namespace='',
        package='rclcpp_components',
        executable='component_container',
    )
```

Perhaps there's a way to do it without a sentinel value. However, I think it would be much nicer not to have to do this evaluation myself in the first place, hence this feature request.

jacobperron:
Resolves #452

These are useful for conditioning launch file actions based on the values of launch configurations.

### DIFF
--- a/launch/launch/conditions/launch_configuration_equals.py
+++ b/launch/launch/conditions/launch_configuration_equals.py
@@ -1,0 +1,67 @@
+# Copyright 2020 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Module for LaunchConfigurationEquals class."""
+
+from typing import Optional
+from typing import Text
+
+from ..condition import Condition
+from ..launch_context import LaunchContext
+from ..some_substitutions_type import SomeSubstitutionsType
+from ..utilities import normalize_to_list_of_substitutions
+from ..utilities import perform_substitutions
+
+
+class LaunchConfigurationEquals(Condition):
+    """
+    Condition on the value of a launch configuration.
+
+    This condition takes an optional string expression that is compared with the value of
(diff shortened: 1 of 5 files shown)


launch/launch/conditions/launch_configuration_equals.py
```diff
@@ -0,0 +1,67 @@
...
+from ..utilities import normalize_to_list_of_substitutions
+from ..utilities import perform_substitutions
+
+
+class LaunchConfigurationEquals(Condition):
+    """
+    Condition on the value of a launch configuration.
+
+    This condition takes an optional string expression that is compared with the value of
+    a launch configuraiton.
```
wjwwood, line 32:
```suggestion
    a launch configuration.
```

jacobperron: 62886a3

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11968/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7002/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9752/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11906/)

Issue #114, opened 2018-07-12
[launch] consider refactoring ExecuteProcess into Execute and Executable


wjwwood:
I believe it might be wise to separate the description of a process to be executed and its state while running from the Action which executes it.

Currently all of this information is encapsulated in the `ExecuteProcess` action, and my proposal would be to replace that with an action called `Execute` that takes an `Executable` (not an action or launch description entity, just a new kind of class). The `Executable` would contain any information needed by `Execute` to actually execute the executable described or introspect it. There could be sub classes of both `Execute` and `Executable`. For instance, a sub class of `Executable` might be `Node` or `LifecycleNode`, and a sub class of `Execute` might be `ExecuteRemotely` (execute on a different machine).

The reasons for doing this include:

- using `Executable`s (or subclasses thereof) with other actions
- cleaner abstractions
  - i.e. the `ExecuteProcess` class is really big now and it would be nice to break into smaller parts
- more readable
  - e.g. `Execute(Executable(...))` or `Execute(Node(...))` rather than `ExecuteProcess(...)` and `Node(...)`

The first reason is the most impactful, because it allows for a few use cases that aren't currently easy to implement:

- Collect or generate `Executable` instances and later decide whether to pass to `Execute`, `ExecuteRemotely`, or `ExecuteInDocker` (just ideas)
- Pass `ComponentNode` (possibly a subclass of `Executable` which represents a node that can be loaded dynamically) to either `Execute()` or `ExecuteNodesInSharedProcess()` (name needs work)
  - where the latter would take several `ComponentNode` instances and run them all in a single process

I still have some design questions in mind, like:

- should `Node` inherit from `Executable`
- should `Executable` contain process related things like `pid` and produce events for stdout and process exit, etc...
  -  or should there be a `Process` that also inherits from `Executable`
  - because `Node` might not be associated with a process, or rather a process might not be associated with a single `Node`
- what is the contract between `Executable` and `Execute`?
  - which creates event handlers and stores information?
  - what exactly are the responsibilities of `Execute`?

ivanpauno:
I thought a little bit about this, and about how it would look in the new frontends (examples only in XML here).

I would have the following classes:
- A `Node` class, that describes a node (name, namespace, remappings, parameters).
  - A `ComposableNode` class.
  - A `LifecycleNode` class.
- An `Executable` class (or `Process` class), that describes an executable (with process information).
[...]

Issue #394, opened 2020-03-24
Passing a string to Node arguments behaves unintuitively


rotu:
In Node constructor, arguments takes an arguments of list of strings. Passing a string to arguments typechecks but provides unexpected results, treating the string as a sequence of single-character arguments.

e.g the following code:
```
    driver_node = LifecycleNode(package='ros2_ouster',
                                node_executable='ouster_driver',
                                node_name=node_name,
                                output='screen',
                                emulate_tty=True,
                                parameters=[parameter_file],
                                node_namespace='/',
                                arguments='__log_level:=debug'
                                )
```

This displays no error message but the mistake is visible when the process dies.
```
[ERROR] [ouster_driver-1]: process has died [pid 309341, exit code -15, cmd '/opt/ros/master/install/lib/ros2_ouster/ouster_driver _ _ l o g _ l e v e l : = d e b u g --ros-args -r __node:=ouster_driver --params-file /opt/ros/master/install/share/ros2_ouster/params/os1.yaml -r __ns:=/'].
```

Passing a string to `arguments` (or similar places) should either:

1. Split the string on spaces
2. Alert the user (`ArgumentError` seems apt)

ivanpauno:
`cmd` is expected to be a list of `Substitutions`, https://github.com/ros2/launch/blob/ae12701630c7acd3d78806bc764bf96b54d6ebe8/launch/launch/actions/execute_process.py#L90

where a substitution can be the following:
https://github.com/ros2/launch/blob/ae12701630c7acd3d78806bc764bf96b54d6ebe8/launch/launch/some_substitutions_type.py#L24-L28

Strings are iterable in `Python`, so this will think that each letter in the string is an argument of the command:
https://github.com/ros2/launch/blob/ae12701630c7acd3d78806bc764bf96b54d6ebe8/launch/launch/actions/execute_process.py#L203

I think we should detect if the input is an string, and raise an error.

ivanpauno:
This is related too: https://github.com/ros2/launch/issues/263.

rotu:
I think you're right that it should raise an error. It seems wrong to have `ExecuteProcess(cmd='echo', arguments='hello          world')` print 'hello world' or to have `ExecuteProcess(cmd='ls', arguments='-l -A')` crash, so an error seems sensible.

PR #454, opened 2020-08-19
Adding Executable description class

roger-strain:
This is an initial commit toward the process described in ros2/design#272. It should not be merged to master yet, and probably needs a feature branch instead.

Haven't added unit testing yet, but also not as familiar as I'd like with the testing setup, so guidance toward getting appropriate tests would be greatly appreciated.

Part of implementation of refactor for #114

Distro A; OPSEC #2893

Signed-off-by: Roger Strain <rstrain@swri.org>

### DIFF
--- a/launch/launch/descriptions/executable.py
+++ b/launch/launch/descriptions/executable.py
@@ -1,0 +1,141 @@
+"""Module for a description of an Executable."""
+
+import os
+import shlex
+import threading
+from typing import Dict
+from typing import Iterable
+from typing import List
+from typing import Optional
+from typing import Tuple
+
+from launch.some_substitutions_type import SomeSubstitutionsType
+from launch.substitution import Substitution
+from launch.launch_context import LaunchContext
+from launch.utilities import normalize_to_list_of_substitutions
+from launch.utilities import perform_substitutions
+
+_global_process_counter_lock = threading.Lock()
+_global_process_counter = 0  # in Python3, this number is unbounded (no rollover)
+
+class Executable:
+    """Describes an executable (typically a single process) which may be run by the launch system."""
+
+    def __init__(
+        self, *,
+        cmd: Iterable[SomeSubstitutionsType],
+        name: Optional[SomeSubstitutionsType] = None,
+        cwd: Optional[SomeSubstitutionsType] = None,
+        env: Optional[Dict[SomeSubstitutionsType, SomeSubstitutionsType]] = None,
+        additional_env: Optional[Dict[SomeSubstitutionsType, SomeSubstitutionsType]] = None,
+    ) -> None:
+        """
+        Initialize an Executable description.
+
+        :param cmd: A list where the first item is the executable and the rest are
+            arguments to the executable, each item may be a string or a list of strings
+            and Substitutions to be resolved at runtime
+        :param name: The label used to represent the process, as a string or a Substitution
+            to be resolved at runtime, defaults to the basename of the executable
+        :param cwd: The directory in which to run the executable
+        :param env: Dictionary of environment variables to be used, starting from a clean
+            environment. If None, the current environment is used.
+        :param additional_env: Dictionary of environment variables to be added. If env was
+            None, they are added to the current environment. If not, env is updated with
+            additional_env.
+        """
+        self.__cmd = [normalize_to_list_of_substitutions(x) for x in cmd]
+        self.__name = name if name is None else normalize_to_list_of_substitutions(name)
+        self.__cwd = cwd if cwd is None else normalize_to_list_of_substitutions(cwd)
+        self.__env = None  # type: Optional[List[Tuple[List[Substitution], List[Substitution]]]]
+        if env is not None:
+            self.__env = []
+            for key, value in env.items():
+                self.__env.append((
+                    normalize_to_list_of_substitutions(key),
+                    normalize_to_list_of_substitutions(value)))
+        self.__additional_env: Optional[List[Tuple[List[Substitution], List[Substitution]]]] = None
+        if additional_env is not None:
+            self.__additional_env = []
+            for key, value in additional_env.items():
+                self.__additional_env.append((
+                    normalize_to_list_of_substitutions(key),
+                    normalize_to_list_of_substitutions(value)))
+
+    @property
+    def name(self):
+        """Getter for name."""
+        return self.__name
+
+    @property
+    def cmd(self):
+        """Getter for cmd."""
+        return self.__cmd
+
+    @property
+    def cwd(self):
+        """Getter for cwd."""
+        return self.__cwd
+
+    @property
+    def env(self):
+        """Getter for env."""
+        return self.__env
+
+    @property
+    def additional_env(self):
+        """Getter for additional_env."""
+        return self.__additional_env
+
+    @property
+    def process_details(self):
+        """Getter for the substituted executable details, e.g. cmd, cwd, env, or None if substitutions have not been performed."""
+        return self.__process_event_args
+
+    def apply_context(self, context: LaunchContext):
+        """
+        Prepares an executable description for execution in a given environment.
+
+        This does the following:
+        - performs substitutions on various properties
+        """
+        self.__expand_substitutions(context)
+        process_event_args = self.__process_event_args
+        if process_event_args is None:
+            raise RuntimeError('process_event_args unexpectedly None')
+
+    def __expand_substitutions(self, context):
+        # expand substitutions in arguments to async_execute_process()
+        cmd = [perform_substitutions(context, x) for x in self.__cmd]
+        name = os.path.basename(cmd[0]) if self.__name is None \
+            else perform_substitutions(context, self.__name)
+        with _global_process_counter_lock:
+            global _global_process_counter
+            _global_process_counter += 1
+            self.__name = '{}-{}'.format(name, _global_process_counter)
+        cwd = None
+        if self.__cwd is not None:
+            cwd = ''.join([context.perform_substitution(x) for x in self.__cwd])
+        env = None
+        if self.__env is not None:
+            env = {}
+            for key, value in self.__env:
+                env[''.join([context.perform_substitution(x) for x in key])] = \
+                    ''.join([context.perform_substitution(x) for x in value])
+        if self.__additional_env is not None:
+            if env is None:
+                env = dict(os.environ)
+            for key, value in self.__additional_env:
+                env[''.join([context.perform_substitution(x) for x in key])] = \
+                    ''.join([context.perform_substitution(x) for x in value])
+        # store packed kwargs for all ProcessEvent based events
+        self.__process_event_args = {
+            'description': self,
+            'name': self.__name,
+            'cmd': cmd,
+            'cwd': cwd,
+            'env': env,
+            # pid is added to the dictionary in the connection_made() method of the protocol.
+        }
+
+
--- a/launch/launch/__init__.py
+++ b/launch/launch/__init__.py
@@ -16,6 +16,7 @@
 
 from . import actions
 from . import conditions
+from . import descriptions
 from . import events
 from . import frontend
 from . import logging
@@ -40,6 +41,7 @@ from .substitution import Substitution
 __all__ = [
     'actions',
     'conditions',
+    'descriptions',
     'events',
     'frontend',
     'logging',
--- a/launch/launch/descriptions/__init__.py
+++ b/launch/launch/descriptions/__init__.py
@@ -1,0 +1,9 @@
+
+"""descriptions Module."""
+
+from .executable import Executable
+
+
+__all__ = [
+    'Executable'
+]


roger-strain:
@ivanpauno @hidmic Getting started on the implementation of this. Wanted to just put a quick first cut of one class up first to make sure I get the process right. Looking forward to feedback and suggestions, especially with respect to testing.

launch/launch/descriptions/executable.py
```diff
@@ -0,0 +1,141 @@
...
+from typing import Tuple
+
+from launch.some_substitutions_type import SomeSubstitutionsType
+from launch.substitution import Substitution
+from launch.launch_context import LaunchContext
+from launch.utilities import normalize_to_list_of_substitutions
+from launch.utilities import perform_substitutions
+
+_global_process_counter_lock = threading.Lock()
+_global_process_counter = 0  # in Python3, this number is unbounded (no rollover)
```
ivanpauno, line 19: this is going to be a problem until `ExecuteProcess` doesn't internally use this new description, as there is going to be two global counters (and thus, names may collide).

This does not need to be fixed now.

launch/launch/descriptions/executable.py
```diff
@@ -0,0 +1,141 @@
...
+    def apply_context(self, context: LaunchContext):
+        """
+        Prepares an executable description for execution in a given environment.
+
+        This does the following:
+        - performs substitutions on various properties
+        """
+        self.__expand_substitutions(context)
+        process_event_args = self.__process_event_args
+        if process_event_args is None:
```
ivanpauno, line 104: nit: this can be done in a single line
```suggestion
        if self.__process_event_args is None:
```

launch/launch/descriptions/executable.py
```diff
@@ -0,0 +1,141 @@
...
+
+_global_process_counter_lock = threading.Lock()
+_global_process_counter = 0  # in Python3, this number is unbounded (no rollover)
+
+class Executable:
+    """Describes an executable (typically a single process) which may be run by the launch system."""
+
+    def __init__(
+        self, *,
+        cmd: Iterable[SomeSubstitutionsType],
```
ivanpauno, line 26: A bit unrelated: maybe we should take advantage of this new description to fix:
- https://github.com/ros2/launch/issues/263
- https://github.com/ros2/launch/issues/394
?

Using a single string instead than a list of strings plays better with substitutions.

launch/launch/descriptions/executable.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def __expand_substitutions(self, context):
+        # expand substitutions in arguments to async_execute_process()
+        cmd = [perform_substitutions(context, x) for x in self.__cmd]
+        name = os.path.basename(cmd[0]) if self.__name is None \
+            else perform_substitutions(context, self.__name)
+        with _global_process_counter_lock:
+            global _global_process_counter
+            _global_process_counter += 1
+            self.__name = '{}-{}'.format(name, _global_process_counter)
```
ivanpauno, line 115: nit: use f-strings

launch/launch/descriptions/executable.py
ivanpauno: is this check even needed?

ivanpauno:
I will create a branch `feature/refactor-execute-process` when this is ready, and we can start merging PRs there.

launch/launch/descriptions/executable.py
```diff
@@ -0,0 +1,141 @@
...
+            self.__additional_env = []
+            for key, value in additional_env.items():
+                self.__additional_env.append((
+                    normalize_to_list_of_substitutions(key),
+                    normalize_to_list_of_substitutions(value)))
+
+    @property
+    def name(self):
+        """Getter for name."""
+        return self.__name
```
hidmic, line 68: In an unrelated (or maybe not quite unrelated) note, `launch` and `launch_ros` actions are inconsistent w.r.t how action properties interact with the action lifecycle.

- `launch.actions.ExecuteProcess.name` gets you a normalized substitution, always.
- `launch_ros.actions.Node.node_name` raises if you try to get it before executing the action.

`launch_ros.parameter_description.Parameter.name`, though arguably not an action, introduces yet another variation and gets you a normalized substitution before evaluating the parameter and the actual name after evaluating it. 

@ivanpauno @wjwwood perhaps we should settle this here and stick to it.

roger-strain: @hidmic Would it be worth introducing more explicitly named parameters instead, to clarify which thing you're requesting? Maybe `name` always returns the value as it was originally specified, without applying substitutions, and `applied_name` returns the one with substitutions applied to it, raising an exception if substitutions haven't been performed yet?

ivanpauno: > @hidmic Would it be worth introducing more explicitly named parameters instead, to clarify which thing you're requesting? Maybe name always returns the value as it was originally specified, without applying substitutions, and applied_name returns the one with substitutions applied to it, raising an exception if substitutions haven't been performed yet?

I think that's ok.
bikeshed: I'm not convinced about naming, maybe `final_name` sounds better than `applied_name`. I don't mind much.

roger-strain: I've taken a stab at this in and updated. In reviewing how the objects were used, I don't think this was the right place to be creating the packed dictionary anyway, so I've removed that; it'll probably come back in `ExecuteLocal`. For all the parameters which can have substitutions applied, there's now a "`final_`" version which can be accessed. Right now, that isn't throwing an error if the substitutions haven't been applied, but I can make it do so if that's the standard approach.

launch/launch/descriptions/executable.py
roger-strain: I've changed the name to a local-specific name regardless. There's no need for those to conflict.

launch/launch/descriptions/executable.py
roger-strain: Removed the identified check. `__expand_substitutions` is still modifying the object, but it's doing is by adding newly-defined values instead of changing the existing ones. Not sure if it's really different enough for this concern, so if we want to continue tweaking it, please let me know. If we really would rather have a stateful object to pass around instead of changing this one, I think I can see the outlines of doing that, but I'm not sure if we're gaining much (other than complexity). Open to suggestions/thoughts/comments, of course.

launch/launch/descriptions/executable.py
roger-strain: I've taken a stab at addressing this. Also added a few unit test cases to evaluate how it's handling the different cases described in #263 itself. I'd very much like to add more tests to make sure we're getting the expected results, so if you have any additional suggestions, I'd love to hear them.

roger-strain:
@ivanpauno I've taken a first stab at trying to refactor the main execute_process class into the new execute_local. I spent quite a while trying to figure out how to write some unit tests to actually verify it's doing what it should, but in all honesty, this part of the system is still pretty unfamiliar to me, and I wasn't able to come up with any good samples to follow. The code itself is by and large copied straight from the original class, so I'm very happy to try to clean things up and improve them from this point, so any direction there would be appreciated, as well as suggestions on how to put some useful tests together.

ivanpauno (commented):
This looks good to me.

The only missing thing for merging is refactoring `ExecuteProcess` to reuse `ExecuteLocal` and `Executable`. If all previous tests pass, I think this would be ready to merge.

It would also be great to have some direct testing of `ExecuteLocal` (and more extended testing of `Executable`)

launch/test/test_executable.py
```diff
@@ -0,0 +1,32 @@
...
+def test_cmd_strings_in_list():
+    exe = Executable(cmd=['ls', '"my/subdir/with spaces/"'])
+    exe.apply_context(LaunchContext())
+    assert all([a == b for a, b in zip(exe.final_cmd, ['ls', 'my/subdir/with spaces/'])])
+
+
+def test_cmd_multiple_arguments_in_string():
+    exe = Executable(cmd=['ls', '-opt1', '-opt2', '-opt3'])
+    exe.apply_context(LaunchContext())
+    assert all([a == b for a, b in zip(exe.final_cmd, ['ls', '-opt1', '-opt2', '-opt3'])])
```
ivanpauno, line 32: you could also have tests for `final_name`, `final_cwd` and `final_env`.

ivanpauno:
Note: Linters aren't passing, check http://build.ros2.org/job/Rpr__launch__ubuntu_focal_amd64/82/.

roger-strain:
> The only missing thing for merging is refactoring `ExecuteProcess` to reuse `ExecuteLocal` and `Executable`. If all previous tests pass, I think this would be ready to merge.

That's the plan for the next step, but I wanted to make sure things looked sane to this point first.
 
> It would also be great to have some direct testing of `ExecuteLocal` (and more extended testing of `Executable`)

I would absolutely love to get more testing in place, but I'm kind of lost on how to best accomplish that. If there are some samples I could work from that show testing execution of things, I would really love to dig into that so I can apply it to this. Actually spinning up external processes as part of unit testing seems a little dicey, especially considering the variety of platforms the system should work on. If you've got any guidance, it'd be a huge help.

launch/launch/descriptions/executable.py
```diff
@@ -100,22 +124,22 @@ def apply_context(self, context: LaunchContext):
...
-        process_event_args = self.__process_event_args
-        if process_event_args is None:
-            raise RuntimeError('process_event_args unexpectedly None')
 
     def __expand_substitutions(self, context):
         # expand substitutions in arguments to async_execute_process()
-        cmd = [perform_substitutions(context, x) for x in self.__cmd]
+        cmd = ' '.join([perform_substitutions(context, x) for x in self.__cmd])
+        cmd = shlex.split(cmd)
+        self.__final_cmd = cmd
```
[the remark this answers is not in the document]
roger-strain: I think in the interest of keeping this PR simpler, I'll err on the side of just retaining the old style for `cmd`. I'd like to be able to help pick up the other issue as well, but it's probably better to let that be a separate task down the line.

roger-strain:
@ivanpauno Pushed part of this a couple of weeks ago, and then commented on the wrong PR. Sorry about that. I think the unit tests are happy now; I'm still not 100% confident, but I'm not sure I ever will be. When you have a chance, give it a look over and see what you think. I'm going to take this out of draft status at this point.

launch/launch/actions/execute_process.py
```diff
@@ -205,50 +144,9 @@ def __init__(
...
-        self._subprocess_transport = None
-        self.__completed_future = None  # type: Optional[asyncio.Future]
-        self.__shutdown_future = None  # type: Optional[asyncio.Future]
-        self.__sigterm_timer = None  # type: Optional[TimerAction]
-        self.__sigkill_timer = None  # type: Optional[TimerAction]
-        self.__stdout_buffer = io.StringIO()
-        self.__stderr_buffer = io.StringIO()
-
-        self.__executed = False
+        self.__executable = Executable(cmd=cmd, prefix=prefix, name=name, cwd=cwd, env=env,
```
hidmic, line 147: @roger-strain nit: it'd seem to me that `self.__executable` is redundant with `self.process_description`

launch/test/launch/test_executable.py
```diff
@@ -0,0 +1,47 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# DISTRIBUTION A. Approved for public release; distribution unlimited.
+# OPSEC #4584.
+#
+# Delivered to the U.S. Government with Unlimited Rights, as defined in DFARS
+# Part 252.227-7013 or 7014 (Feb 2014).
+#
+# This notice must appear in all copies of this file and its derivatives.
```
hidmic, line 21: @tfoote would you give this preamble a look? I'm not well versed in such licensing qualifications.

launch/test/launch/test_executable.py
```diff
@@ -0,0 +1,47 @@
...
+def test_cmd_strings_in_list():
+    exe = Executable(cmd=['ls', '"my/subdir/with spaces/"'])
+    exe.apply_context(LaunchContext())
+    assert all([a == b for a, b in zip(exe.final_cmd, ['ls', '"my/subdir/with spaces/"'])])
+
+
+def test_cmd_multiple_arguments_in_string():
+    exe = Executable(cmd=['ls', '-opt1', '-opt2', '-opt3'])
+    exe.apply_context(LaunchContext())
+    assert all([a == b for a, b in zip(exe.final_cmd, ['ls', '-opt1', '-opt2', '-opt3'])])
```
hidmic, line 47: @roger-strain tests for other arguments besides `cmd` and/or using substitutions would be most welcome.

launch/launch/descriptions/executable.py
hidmic: It's OK for now. It's not up to you to solve these discrepancies anyways.

launch/launch/actions/execute_local.py
```diff
@@ -0,0 +1,663 @@
...
+                )
+            if not self.__shutdown_future.done():
+                context.asyncio_loop.create_task(self.__execute_process(context))
+                return
+        self.__cleanup()
+
+    def prepare(self, context: LaunchContext):
+        """Prepare the action for execution."""
+        self.__process_description.apply_context(context)
+        self.__expand_substitutions(context)
```
hidmic, line 592: @roger-strain meta, safe to ignore: I'm not a huge fan of methods with side effects like this. Purity is more expensive but less surprising.

hidmic:
@roger-strain I think we'll have a better grasp of the refactor once a somewhat larger portion of it lands. Are you looking forward to continue your contributions in the near future?

roger-strain:
> @roger-strain I think we'll have a better grasp of the refactor once a somewhat larger portion of it lands. Are you looking forward to continue your contributions in the near future?

@hidmic Yes, I'm hoping to get the next section of this next week. This is the only PR that will be targeting `launch` itself; the rest are going to be in the `launch_ros` side of things. When we were discussing the timeline/order of operations for the refactor, though, it was suggested I get this side of things sorted first, and backward compatible, so that was the first focus.

hidmic:
@roger-strain friendly ping

roger-strain:
> @roger-strain friendly ping

Sorry, holidays caught me off guard and I'm now in the process of getting my laptop rebuilt. I believe I have the code finished for the next update on this and the first corresponding PR over in launch_ros, but I need to get more unit tests in place before I'm ready to officially push them up. Targeting later this week, assuming nothing else blows up on me in the process.

roger-strain:
@hidmic Finally finished a big chunk of work over in ros2/launch_ros#215, and updated this one to go alongside it. Please take a look and see how you think it's looking. Still need to go in and do the bits for Composable and Lifecycle nodes, and I'm sure there are other bits that I've missed, but hopefully this is pretty close to what you're looking for otherwise.

launch/test/launch/test_executable.py
hidmic: @tfoote friendly ping

roger-strain:
@hidmic Some more updates over here; little bit of cleanup, and also part of the refactor discussed in ros2/launch_ros#215 regarding the proper place for arguments; they now are handled at the `Executable` level.

launch/test/launch/test_executable.py
clalancette: @roger-strain Does this notice have to be on these files?  It is much easier for us if this is just not there; then it looks the same as all of the rest of the licensing in our code and isn't a special snowflake.  If it absolutely has to be there for some reason, I'm going to have to check with some others about whether this is allowed in this codebase.

hidmic (commented):
@roger-strain my since apologies for not coming back to this earlier. Too many things to do. 

Nothing specifically added by this patch calls my attention in a negative way. Its PR job doesn't seem happy though. There's plenty of failing tests. Mind to take a look?

launch/test/launch/test_executable.py
clalancette: It's really the whole thing.  The basic problem is that we are not lawyers, and so that anything that deviates from the standard license text is a problem.

Please let me know what you decide.  Once you've come to your decision, I'll pass it along to those more in the know on this side.  Thanks.

roger-strain: Just to follow up, we're still in active discussions about how to sort this out. Appreciate your patience.

roger-strain:
@hidmic @clalancette Finally hoping to get this set of PRs ready to go. I've been given other tasks at the moment, but @mlanting has been part of the discussions from early on, and knows what's going on. I believe from all the various discussions that have been going on that we have a solution on the license language issue. Please check his latest commit and see if we're ready to put this one to bed, then move on to bringing the `launch_ros` side back up to current functionality.

clalancette:
The license changes look good to me.  I'm not familiar with the content of this PR, so I'll leave that up to @hidmic.

launch/launch/actions/execute_local.py
```diff
@@ -0,0 +1,662 @@
...
+        emulate_tty: bool = False,
+        output: Text = 'log',
+        output_format: Text = '[{this.process_description.final_name}] {line}',
+        log_cmd: bool = False,
+        on_exit: Optional[Union[
+            SomeActionsType,
+            Callable[[ProcessExited, LaunchContext], Optional[SomeActionsType]]
+        ]] = None,
+        respawn: bool = False,
+        respawn_delay: Optional[float] = None,
```
hidmic, line 100: @roger-strain @mlanting meta: looking at this with fresh eyes, none of these arguments (from line 92 to to line 100) are specific to the local execution method. The line is a bit blurry though. Perhaps, these could be part of the process description or at least be pushed to an abstract base class for all execution actions. 

I'm fine with postponing that decision to a follow-up PR though.

launch/launch/actions/execute_process.py
```diff
@@ -1,885 +1,318 @@
...
+_global_process_counter_lock = threading.Lock()
+_global_process_counter = 0  # in Python3, this number is unbounded (no rollover)
+
+
+@expose_action('executable')
+class ExecuteProcess(ExecuteLocal):
+    """Action that begins executing a process and sets up event handlers for the process."""
+
+    def __init__(
+            self,
```
hidmic, line 43: @roger-strain @mlanting indentation here seems to have been doubled. I wonder why the linter did not complain.

launch/launch/descriptions/executable.py
```diff
@@ -0,0 +1,192 @@
...
+
+_executable_process_counter_lock = threading.Lock()
+_executable_process_counter = 0  # in Python3, this number is unbounded (no rollover)
+
+
+class Executable:
+    """Describes an executable (usually a single process) which may be run by the launch system."""
+
+    def __init__(
+            self, *,
```
hidmic, line 45: @roger-strain @mlanting same comment about indentation

launch/launch/event_handlers/on_process_exit.py
```diff
@@ -45,15 +45,15 @@ class OnProcessExit(BaseEventHandler):
     def __init__(
         self,
         *,
-        target_action: 'ExecuteProcess' = None,
+        target_action: 'ExecuteLocal' = None,
```
hidmic, line 48: @roger-strain @mlanting meta: yeap, this definitely calls for an abstract `Execute` action.

hidmic (approved):
Overall LGTM. I'd really like to see `ExecuteLocal` subclass an abstract `Execute` action, but I'm fine with a short-term deferral too. @ivanpauno @wjwwood for last minute feedback.

hidmic:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14999/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9738/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12686/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15161/)

launch/launch/actions/execute_process.py
mlanting: I think it's because these are function arguments rather than the function code.

launch/launch/actions/execute_local.py
mlanting: I would prefer to postpone this for now. Our plan is to add an ExecuteRemote class to support multi-machine launching once we get these changes through. That process should clarify how a base class should look and we can include it in that PR.

hidmic:
@roger-strain @mlanting friendly ping ! Linters are not happy.

hidmic:
@mlanting friendly ping

mlanting:
Sorry for the delay; I've been busy supporting a couple other project, but finally had time to take at look this week. The prefix filtering changes should now be properly incorporated into the refactor.

launch/launch/descriptions/executable.py
```diff
@@ -0,0 +1,206 @@
...
+            configuration is provided.
+        :param name: The label used to represent the process, as a string or a Substitution
+            to be resolved at runtime, defaults to the basename of the executable
+        :param cwd: The directory in which to run the executable
+        :param env: Dictionary of environment variables to be used, starting from a clean
+            environment. If None, the current environment is used.
+        :param additional_env: Dictionary of environment variables to be added. If env was
+            None, they are added to the current environment. If not, env is updated with
+            additional_env.
+        :param arguments: list of extra arguments for the executable
```
wjwwood, line 75: I know this issue is quite old, but what's the purpose of this argument? Why could they not be part of the `cmd` argument?

PR #456, opened 2020-08-20
add bsd license to launch due to files from roslaunch

### LINKED ISSUE #439: clarify inconsistent license
See https://github.com/ros2/launch/pull/209#discussion_r451097332

wjwwood:
fixes #439


dirk-thomas:
Should the `LICENSE` file in the root of the repo also reflect this?

wjwwood:
Yeah, but also I screwed up the package.xml, it should be a different package. I'll fix it up.

PR #457, opened 2020-08-26
[foxy backport] Add new conditions for checking launch configuration values (#453)

jacobperron:
Backports #453 to Foxy.

It's not a bug fix, but it makes writing launch files a lot easier so I'd like to take advantage of it in Foxy.

### DIFF
--- a/launch/launch/conditions/__init__.py
+++ b/launch/launch/conditions/__init__.py
@@ -17,11 +17,15 @@
 from .evaluate_condition_expression_impl import evaluate_condition_expression
 from .if_condition import IfCondition
 from .invalid_condition_expression_error import InvalidConditionExpressionError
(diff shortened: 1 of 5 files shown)


jacobperron:
@ivanpauno We don't need to backport this. It is primarily motivated by https://github.com/ros-perception/image_pipeline/pull/583, which can't target Foxy without this change. Otherwise, I can modify the downstream launch file changes to not use this feature (but this seemed like the cleaner and easier option).

jacobperron:
IMO, backporting this particular feature is low-risk since it is entirely new code (not interacting with existing `launch` code).

ivanpauno:
> IMO, backporting this particular feature is low-risk since it is entirely new code (not modifying with existing launch code).

If one extra reviewer agrees I'm fine with the backport :+1:

PR #461, opened 2020-09-29
Fix error with dollar symbols in launch frontend substitution grammar

ivanpauno:
Fixes small error with dollar symbols that aren't part of a substitution expression.

From [re docs](https://docs.python.org/3/library/re.html):

```
(?=...)
Matches if ... matches next, but doesn’t consume any of the string. This is called a lookahead assertion. For example, Isaac (?=Asimov) will match 'Isaac ' only if it’s followed by 'Asimov'.

(?!...)
Matches if ... doesn’t match next. This is a negative lookahead assertion. For example, Isaac (?!Asimov) will match 'Isaac ' only if it’s not followed by 'Asimov'.
```

A regression test was also added.
Reading those regexp is hard :grimacing:.


jacobperron:
With this change, does existing code that escaped `$` characters still behave the same? I.e. are the following two strings equivalent:

`$foo $bar` vs `\$foo \$bar`

ivanpauno:
> With this change, does existing code that escaped $ characters still behave the same? I.e. are the following two strings equivalent:
> 
> $foo $bar vs \$foo \$bar

Yes, both work.

The substitution parser does something a bit "unusual".
It replaces all the escaped characters (e.g. '\x') by the character being escaped ('x'),  regardless if it was an special character or not (see [here](https://github.com/ros2/launch/blob/4370b2b829946cd8c5ec5a76cbb3bb6a70c95e97/launch/launch/frontend/parse_substitution.py#L33-L35)).

jacobperron (approved):
It's hard for me to parse the regexes, but the tests look good 😄

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12540/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7509/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10252/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12461/)

hidmic (approved):
Yeap, those extra `=` are incorrect. Thanks for the fix @ivanpauno.

LGTM once CI goes green.

jacobperron:
@ivanpauno If you have time, opening up backport PRs for this change would be good.

PR #462, opened 2020-09-29
Expose Timer action in launch xml

ivanpauno:
Example usage:
```xml
<launch>
  <timer period="5">
    <node pkg="turtlesim" exec="turtlesim_node" name="sim"/>
  </timer>
</launch>
```

Other changes:
- `cancel_on_shutdown` now can be a substitution.
- `period` can still be a substitution, but not a string literal (e.g. `TextSubstitution(text='5')` is valid, but '5' isn't).
- Used new `type_utils` functions to handle "typed" substitutions.

### DIFF
--- a/launch/launch/actions/timer_action.py
+++ b/launch/launch/actions/timer_action.py
@@ -34,6 +34,9 @@ from ..action import Action
 from ..event_handler import EventHandler
 from ..events import Shutdown
 from ..events import TimerEvent
+from ..frontend import Entity
+from ..frontend import expose_action
+from ..frontend import Parser
 from ..launch_context import LaunchContext
 from ..launch_description_entity import LaunchDescriptionEntity
 from ..some_actions_type import SomeActionsType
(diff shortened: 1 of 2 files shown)


jacobperron (commented):
> period can still be a substitution, but not a string literal (e.g. TextSubstitution(text='5') is valid, but '5' isn't).

It would be nice if we could maintain backwards compatibility and continue to accept literal strings. Is this because of a limitation or because you don't think we should continue to accept literals?

ivanpauno:
> It would be nice if we could maintain backwards compatibility and continue to accept literal strings.

Done in https://github.com/ros2/launch/pull/462/commits/55a721f0da224e71b391333bf7cde0d7364ac1af.

> Is this because of a limitation or because you don't think we should continue to accept literals?

There's no technical limitation, but accepting a string literal doesn't look pythonic.

jacobperron (approved):
LGTM

ivanpauno:
CI testing launch, launch_xml, launch_yaml, launch_ros, test_launch_ros:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12563/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7531/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10274/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12485/)

Issue #338, opened 2019-09-23
[launch_frontend] Produce errors for invalid entities and attributes


jacobperron:
#### Feature description
When an invalid entity or attribute appears in a launch file, it should produce an error.
This way users are notified of mistakes in their launch files instead of them being silently ignored.
For instance, it is easy to mistype the attribute `output`.

Here's an example launch file that I would expect to fail for a couple of reasons:

```xml
<launch>
  <!-- The attribute 'outputs' should cause an error -->
  <node
    pkg="demo_nodes_cpp"
    exec="talker"
    outputs="screen"
    name="talker">
    <!-- This tag should cause an error -->
    <this_is_not_a_valid_entity />
  </node>
</launch>
```

#### Implementation considerations

Launch files already throw errors for invalid entities at the global scope. E.g. the following launch file will error:

<launch>
  <this_is_not_a_valid_entity />
</launch>

Maybe we can reuse the logic for nested scopes as well.

ivanpauno:
This issue is somewhat related to https://github.com/ros2/launch/issues/392, though not exactly the same.

ivanpauno:
I have some ideas on how to achieve this, I'm assigning self to try them.

PR #468, opened 2020-10-29
Validate unparsed attributes and subentities in launch_xml and launch_yaml

ivanpauno:
Fixes https://github.com/ros2/launch/issues/338.

Adds three methods to `launch.frontend.Entity`:
- `assert_no_children(self)`
- `assert_subentity_types(self, types: Iterable[str])`
- `assert_attribute_names(self, names: Iterable[str])`

It might make the `parse` methods a bit more verbose, but it's not hard to use.
I have modified two parsing functions to use this, and it has already caught a preexisting error!

There might be some other ways to achieve the same, so I would like to get feedback about the approach before applying this more broadly.

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -324,6 +324,12 @@ class ExecuteProcess(Action):
         :param: ignore A list of arguments that should be ignored while parsing.
             Intended for code reuse in derived classes (e.g.: launch_ros.actions.Node).
         """
+        if ignore is None:
+            entity.assert_attribute_names((
+                'cmd', 'cwd', 'name', 'launch-prefix', 'output', 'respawn', 'respawn_delay', 'shell',
+                'additional_env', 'if', 'unless'
+            ))
+            entity.assert_subentity_types(('env', ))
         _, kwargs = super().parse(entity, parser)
 
         if ignore is None:
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -324,12 +324,6 @@ class ExecuteProcess(Action):
         :param: ignore A list of arguments that should be ignored while parsing.
             Intended for code reuse in derived classes (e.g.: launch_ros.actions.Node).
         """
-        if ignore is None:
-            entity.assert_attribute_names((
-                'cmd', 'cwd', 'name', 'launch-prefix', 'output', 'respawn', 'respawn_delay', 'shell',
-                'additional_env', 'if', 'unless'
-            ))
-            entity.assert_subentity_types(('env', ))
         _, kwargs = super().parse(entity, parser)
 
         if ignore is None:
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -378,12 +378,12 @@ class ExecuteProcess(Action):
             # `unset_enviroment_variable` actions should be used.
             env = entity.get_attr('env', data_type=List[Entity], optional=True)
             if env is not None:
-                env = {
+                kwargs['additional_env'] = {
                     tuple(parser.parse_substitution(e.get_attr('name'))):
                     parser.parse_substitution(e.get_attr('value')) for e in env
                 }
-                kwargs['additional_env'] = env
-
+                for e in env:
+                    e.assert_entity_completely_parsed()
         return cls, kwargs
 
     @property
--- a/launch/launch/frontend/entity.py
+++ b/launch/launch/frontend/entity.py
@@ -83,7 +83,7 @@ class Entity:
         """
         raise NotImplementedError()
 
-    def assert_entity_complely_parsed(self):
+    def assert_entity_completely_parsed(self):
         """
         Assert that all attributes and children of the entity were parsed.
 
--- a/launch/launch/frontend/entity.py
+++ b/launch/launch/frontend/entity.py
@@ -14,6 +14,7 @@
 
 """Module for Entity class."""
 
+from typing import Iterable
 from typing import List
 from typing import Optional
 from typing import Text
@@ -80,5 +81,31 @@ class Entity:
             Only happens in frontend implementations that do type checking
         :raises `ValueError`: Attribute found but can't be coerced to one of the specified types.
             Only happens in frontend implementations that do type coercion.
+        """
+        raise NotImplementedError()
+
+    def assert_no_children(self):
+        """
+        Assert that this entity doesn't have any nested entity.
+
+        Either this method or `assert_subentity_types` should be used, but not both.
+        :raises `ValueError`: if a nested entity is found.
+        """
+        raise NotImplementedError()
+
+    def assert_subentity_types(self, types: Iterable[str]):
+        """
+        Assert that all subentities are of one of the types specified in `types`.
+
+        Either this method or `assert_subentity_types` should be used, but not both.
+        :raises `ValueError`: if a nested entity of a different type is found.
+        """
+        raise NotImplementedError()
+
+    def assert_attribute_names(self, names: Iterable[str]):
+        """
+        Assert that there are no other attributes than the provided `names`.
+
+        :raises `ValueError`: if an attribute named differently is found.
         """
         raise NotImplementedError()
--- a/launch/launch/frontend/entity.py
+++ b/launch/launch/frontend/entity.py
@@ -14,7 +14,6 @@
 
 """Module for Entity class."""
 
-from typing import Iterable
 from typing import List
 from typing import Optional
 from typing import Text
@@ -84,28 +83,11 @@ class Entity:
         """
         raise NotImplementedError()
 
-    def assert_no_children(self):
+    def assert_entity_complely_parsed(self):
         """
-        Assert that this entity doesn't have any nested entity.
+        Assert that all attributes and children of the entity were parsed.
 
-        Either this method or `assert_subentity_types` should be used, but not both.
-        :raises `ValueError`: if a nested entity is found.
-        """
-        raise NotImplementedError()
-
-    def assert_subentity_types(self, types: Iterable[str]):
-        """
-        Assert that all subentities are of one of the types specified in `types`.
-
-        Either this method or `assert_subentity_types` should be used, but not both.
-        :raises `ValueError`: if a nested entity of a different type is found.
-        """
-        raise NotImplementedError()
-
-    def assert_attribute_names(self, names: Iterable[str]):
-        """
-        Assert that there are no other attributes than the provided `names`.
-
-        :raises `ValueError`: if an attribute named differently is found.
+        This function is automatically called after the `parse(entity, parser)`
+        function completed.
         """
         raise NotImplementedError()
--- a/launch/launch/frontend/expose.py
+++ b/launch/launch/frontend/expose.py
@@ -99,7 +99,7 @@ def __expose_impl(name: Text, parse_methods_map: dict, exposed_type: Text):
                 )
             )
         if exposed_type == 'action':
-            # For actions, validate that the user didn't provide unkown attributes or children
+            # For actions, validate that the user didn't provide unknown attributes or children
             def wrapper(entity, parser):
                 ret = found_parse_method(entity, parser)
                 entity.assert_entity_complely_parsed()
--- a/launch/launch/frontend/expose.py
+++ b/launch/launch/frontend/expose.py
@@ -14,6 +14,7 @@
 
 """Module which adds methods for exposing parsing methods."""
 
+import functools
 import inspect
 from typing import Iterable
 from typing import Optional
@@ -99,6 +100,7 @@ def __expose_impl(name: Text, parse_methods_map: dict, exposed_type: Text):
             )
         if exposed_type == 'action':
             # For actions, validate that the user didn't provide unknown attributes or children
+            @functools.wraps(found_parse_method)
             def wrapper(entity, parser):
                 ret = found_parse_method(entity, parser)
                 entity.assert_entity_completely_parsed()
--- a/launch/launch/frontend/expose.py
+++ b/launch/launch/frontend/expose.py
@@ -101,7 +101,7 @@ def __expose_impl(name: Text, parse_methods_map: dict, exposed_type: Text):
             # For actions, validate that the user didn't provide unknown attributes or children
             def wrapper(entity, parser):
                 ret = found_parse_method(entity, parser)
-                entity.assert_entity_complely_parsed()
+                entity.assert_entity_completely_parsed()
                 return ret
             parse_methods_map[name] = wrapper
         else:
--- a/launch/launch/frontend/expose.py
+++ b/launch/launch/frontend/expose.py
@@ -98,7 +98,15 @@ def __expose_impl(name: Text, parse_methods_map: dict, exposed_type: Text):
                     name
                 )
             )
-        parse_methods_map[name] = found_parse_method
+        if exposed_type == 'action':
+            # For actions, validate that the user didn't provide unkown attributes or children
+            def wrapper(entity, parser):
+                ret = found_parse_method(entity, parser)
+                entity.assert_entity_complely_parsed()
+                return ret
+            parse_methods_map[name] = wrapper
+        else:
+            parse_methods_map[name] = found_parse_method
         return exposed
     return expose_impl_decorator
 
--- a/launch/launch/frontend/expose.py
+++ b/launch/launch/frontend/expose.py
@@ -70,7 +70,6 @@ def __expose_impl(name: Text, parse_methods_map: dict, exposed_type: Text):
         method in the dictionary.
     :param parse_methods_map: a dict where the parsing method will be stored.
     :param exposed_type: A string specifing the parsing function type.
-        Only used for having clearer error log messages.
     """
     # TODO(ivanpauno): Check signature of the registered method/parsing function.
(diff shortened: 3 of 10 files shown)


jacobperron (commented):
I think it would be nicer to have children and attributes explicitly declared versus calling separate assert functions. For example, by extending the `expose_action` decorator:

```python
@expose_action(
    'executable',
    allowed_attributes=(
        'cmd',
        'cwd',
        'name',
        'launch-prefix',
        'output',
        'respawn',
        'respawn_delay',
        'shell',
        'additional_env',
        'if',
        'unless'
    ),  
    allowed_children=('env', ))
class ExecuteProcess(Action):
    ...
```
If `allowed_*` is empty (or `None`), then implicitly none are allowed. This way we'll expose bugs in user launch scripts right away; the downside being that we'll break a lot of existing user code, but maybe we could tick-tock the feature, emitting a warning if they haven't explictly declared attributes/children they are accessing.

It might also be nice to have a flag to allow **any** attributes or children. I'm not sure what that would look like.

I don't know the details of the frontend implementation, so I'm not sure how feasible what I proposed is.

ivanpauno:
> I think it would be nicer to have children and attributes explicitly declared versus calling separate assert functions. For example, by extending the expose_action decorator:

I like that one.
I will still need the `assert` functions in `launch.frontend.Entity`, but those can be called directly from the description in the decorator.

I will still need to distinguish between `allow_subentities` (e.g. env in `ExecuteProcess`), and allow children (e.g. any action in `GroupAction`).

launch/launch/actions/execute_process.py
```diff
@@ -324,6 +324,12 @@ def parse(
         :param: ignore A list of arguments that should be ignored while parsing.
             Intended for code reuse in derived classes (e.g.: launch_ros.actions.Node).
         """
+        if ignore is None:
+            entity.assert_attribute_names((
+                'cmd', 'cwd', 'name', 'launch-prefix', 'output', 'respawn', 'respawn_delay', 'shell',
+                'additional_env', 'if', 'unless'
```
hidmic, line 330: @ivanpauno hmm, do we have to explicitly list base classes' attributes? What's going to happen with the assertions performed by another class that derives from this one?

launch_xml/launch_xml/entity.py
```diff
@@ -55,6 +56,28 @@ def children(self) -> List['Entity']:
...
+                f'but the following were found: {tuple(item.tag for item in self.__xml_element)}'
+            )
+
+    def assert_subentity_types(self, types: Iterable[str]):
+        for item in self.__xml_element:
+            if item.tag not in types:
+                raise ValueError(
+                    f'Found subentity of type `{item.tag}` in a `{self.__xml_element.tag}`, '
+                    f'which expects only the following subentities: {types}'
+                )
```
hidmic, line 72: @ivanpauno hmm, do we need this distinction between children and subentities? IIRC all of them are (sub)entities, whether we require them to be of an specific type or not.

hidmic (commented):
First of all, to put things in context, I think that, to solve #338, we have two problems at hand: (1.) to enforce an schema as we parse, and (2.) to correlate any errors with the given input. To complain about a bad tag _somewhere_ in, say, a three levels deep launch file hierarchy is of little use.

Now, this patch addresses the first problem. I agree with @jacobperron's suggestion, but based on my observations, @ivanpauno I (personally) think it's time to go back to that old premise about frontends being standardized 1-to-1 mappings to their Python implementation. 

I'd suggest we use introspection from within our decorators to deduce what attributes and children should be based on `__init__` arguments (along with their annotated types), for the decorated class and its base classes, recursively. As bonus, we get to enforce that 1-to-1 mapping by the same price.

launch_yaml/test/launch_yaml/test_executable.py
```diff
@@ -32,7 +32,7 @@ def test_executable():
                 name: my_ls
                 shell: true
                 output: log
-                launch_prefix: $(env LAUNCH_PREFIX)
+                'launch-prefix': $(env LAUNCH_PREFIX '')
```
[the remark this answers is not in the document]
ivanpauno: The correct name is `launch-prefix` (not `launch_prefix`), and after adding the assert functions this test wasn't passing anymore.

launch/launch/actions/execute_process.py
ivanpauno: That's why I used the trick `if ignore is None:`.
Not a beauty, but I don't know how to do better.

launch_xml/launch_xml/entity.py
ivanpauno: They aren't exactly the same in launch yaml AFAIR:

```yaml
launch:
  group:
    - executable:
      env:
        - ...
    - executable: ...
```

ivanpauno:
> I (personally) think it's time to go back to that old premise about frontends being standardized 1-to-1 mappings to their Python implementation.
>
> I'd suggest we use the full power of Python's introspection machinery from within our decorators to deduce which attributes and children are valid from __init__ arguments (along with their annotated types) for the decorated class and its base classes, recursively. As a bonus, we get to enforce that 1-to-1 mapping.

I don't disagree, but we aren't mapping them 1-to-1 right now, and start doing that would be a breaking change.
In the case of a class with base classes the mechanism isn't obvious (e.g. node accepts different arguments that executable, but some are directly forwarded to the base class).
In that last case, the decorator will still need some user input (a solution to this is to never use `kwargs`, and always forward arguments explicitly in constructors).

We could still add a more intelligent mechanism that deduce the attribute names from the constructor arguments in the future, I think this improvement wouldn't block that (specially if following @jacobperron suggestion of providing the valid attributes and sub-entities in the decorator).

ivanpauno:
It's also not obvious how the decorator would guess what's an attribute (name, executable, plugin, etc), what are children subentities (nested actions in group_action) and what are complex attributes (env, param, etc). Maybe there could be some heuristic based on type annotations, but I'm not sure if that will always work,

hidmic:
> I don't disagree, but we aren't mapping them 1-to-1 right now, and start doing that would be a breaking change.

Hmm, aren't we? IIRC that's what we strived for. I'd suggest we don't delay in correcting the deviation. User provided aliases can aid deprecation cycles.

> We could still add a more intelligent mechanism that deduce the attribute names from the constructor arguments in the future, I think this improvement wouldn't block that

I agree schema deduction is not strictly necessary. To be clear, I meant it as a proposal to solve the problem. I'm sure there are other solutions, perhaps better ones. We can discuss them. But that is not to say I agree with this patch being an improvement as long as it requires replicating information across inheritance chains like this.

> In the case of a class with base classes the mechanism isn't obvious (e.g. node accepts different arguments that executable, but some are directly forwarded to the base class).

One solution to that problem (and perhaps an alternative, explicit, solution to the problem w/o using introspection) is to progressively tag input (at least for validation purposes) as it is accessed. If by the time you've processed an entire subtree, a part of it remains unacknowledged, then you know the system didn't recognize it.

> the decorator will still need some user input

This is correct. The canonical example being arbitrary children mapped to an specific argument. I'm not against user input, but against introducing unnecessary coupling in the system.

> It's also not obvious how the decorator would guess what's an attribute (name, executable, plugin, etc), what are children subentities (nested actions in group_action) and what are complex attributes (env, param, etc). Maybe there could be some heuristic based on type annotations, but I'm not sure if that will always work,

Once it is clear which (and if an) argument maps to all nested entities (which IIRC cannot be coexist with complex attributes), you can deduce the all attributes at first depth from type annotations. Of course, it cannot deduce the structure of complex attributes -- whose shape may be the result of a parsing function's implementation e.g. nested dictionaries.

ivanpauno:
> One solution to that problem (and perhaps an alternative, explicit, solution to the problem w/o using introspection) is to progressively tag input (at least for validation purposes) as it is accessed. If by the time you've processed an entire subtree, a part of it remains unacknowledged, then you know the system didn't recognize it.

Yeah, I thought about that one.
I will try to implement something like that and see how it works.

ivanpauno:
I have updated the PR with the approach suggested above, good news is that we don't need to update any parsing function.

launch/launch/frontend/expose.py
```diff
@@ -98,7 +98,15 @@ def expose_impl_decorator(exposed):
                     name
                 )
             )
-        parse_methods_map[name] = found_parse_method
+        if exposed_type == 'action':
+            # For actions, validate that the user didn't provide unkown attributes or children
```
jacobperron, line 102:
```suggestion
            # For actions, validate that the user didn't provide unknown attributes or children
```

launch_xml/launch_xml/entity.py
```diff
@@ -53,8 +55,23 @@ def parent(self) -> Optional['Entity']:
...
         """Get the Entity's children."""
+        self.__read_children = {item.tag for item in self.__xml_element}
         return [Entity(item) for item in self.__xml_element]
 
+    def assert_entity_complely_parsed(self):
+        unparsed_nested_tags = {item.tag for item in self.__xml_element} - self.__read_children
+        if unparsed_nested_tags:
+            raise ValueError(
+                f'Found the following nested tags of `{self.__xml_element.tag}` that were not '
+                f'parsed: {unparsed_nested_tags}'
```
jacobperron, line 66: Tiny nitpick: from a user perspective, I don't know if "parsed" is the right word. I suggest rewording to something like,  `"Unexpected nested tag(s) found in '{self.__xml_element.tag}': {unparsed_nested_tags}"`

launch_xml/launch_xml/entity.py
```diff
@@ -53,8 +55,23 @@ def parent(self) -> Optional['Entity']:
...
+            )
+        unparsed_attributes = set(self.__xml_element.attrib.keys()) - self.__read_attributes
+        if unparsed_attributes:
+            raise ValueError(
+                f'Found the following attributes of `{self.__xml_element.tag}` that were not '
+                f'parsed: {unparsed_attributes}'
```
jacobperron, line 72: Same nitpick here.

launch_yaml/launch_yaml/entity.py
```diff
@@ -54,24 +56,46 @@ def parent(self) -> Optional['Entity']:
...
+        if not isinstance(self.__element, (dict, list)):
+            raise TypeError(
+                f'Expected a dict or list, got {type(self.element)}:'
+                f'\n---\n{self.__element}\n---'
+            )
         if isinstance(self.__element, dict):
             if 'children' not in self.__element:
-                raise KeyError('Missing `children` key in Entity {}'.format(self.__type_name))
+                raise ValueError(
+                    f'Expected entity {self.__type_name} to have children entities.'
```
jacobperron, line 68:
```suggestion
                    f"Expected entity '{self.__type_name}' to have children entities."
```

launch/launch/frontend/expose.py
```diff
@@ -98,7 +98,15 @@ def expose_impl_decorator(exposed):
                     name
                 )
             )
-        parse_methods_map[name] = found_parse_method
+        if exposed_type == 'action':
```
jacobperron, line 101: Do we need to have this condition on the `exposed_type`? According the docstring, `exposed_type` is only used for clearer error logging.

jacobperron (commented):
Generally looks good. I've left some minor comments

launch/launch/frontend/expose.py
ivanpauno: Fixed in https://github.com/ros2/launch/pull/468/commits/09871dd4925a29a27ac8f6b67a7add84f120d7c8.

launch_xml/launch_xml/entity.py
ivanpauno: Improved this one and similar error messages in https://github.com/ros2/launch/pull/468/commits/6ad4a3a47e0b50ed9d7ea581a159ccbffec821df.

launch/launch/frontend/expose.py
ivanpauno: There are two kinds of parsing functions, one for substitutions and one for actions.
The wrapper only apply to actions (because substitutions parsing functions are different).

ivanpauno: Removed the old comment in https://github.com/ros2/launch/pull/468/commits/716e15d729495d6b5e3211cf585600eebe8cdc28.

jacobperron (approved):
LGTM. It'd be good to have @hidmic take a look too.

ivanpauno:
CI, testing all downstream packages of `launch`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12945/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7894/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10662/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12894/)

launch/launch/frontend/entity.py
```diff
@@ -82,3 +82,12 @@ def get_attr(
             Only happens in frontend implementations that do type coercion.
         """
         raise NotImplementedError()
+
+    def assert_entity_complely_parsed(self):
```
hidmic, line 86: @ivanpauno did you mean:
```suggestion
    def assert_entity_completely_parsed(self):
```
?

launch/launch/frontend/expose.py
```diff
@@ -98,7 +97,15 @@ def expose_impl_decorator(exposed):
...
+            # For actions, validate that the user didn't provide unknown attributes or children
+            def wrapper(entity, parser):
```
hidmic, line 102: @ivanpauno consider:
```suggestion
            @functools.wraps(found_parse_method)
            def wrapper(entity, parser):
```

launch_xml/launch_xml/entity.py
```diff
@@ -53,8 +55,23 @@ def parent(self) -> Optional['Entity']:
     @property
     def children(self) -> List['Entity']:
         """Get the Entity's children."""
+        self.__read_children = {item.tag for item in self.__xml_element}
         return [Entity(item) for item in self.__xml_element]
 
+    def assert_entity_complely_parsed(self):
+        unparsed_nested_tags = {item.tag for item in self.__xml_element} - self.__read_children
```
hidmic, line 62: @ivanpauno shouldn't we assert recursively? I'm thinking of nested structures that are parsed by the same parsing function e.g. nested `launch_ros.actions.Node` parameters. Thinking out loud, perhaps we can keep some form of context in the parser to avoid full traversals at each node of the tree, but that's arguably early optimization.

ivanpauno: > shouldn't we assert recursively?

That doesn't sound like a good idea to me. I mean for things like `group` or `timer` it shouldn't be recursive, but for `executable` or `node` that could work.

>  I'm thinking of nested structures that are parsed by the same parsing function e.g. nested launch_ros.actions.Node parameters

Yeah that's true, maybe in those places we need to call `assert_entity_completely_parsed` manually (as `param`, `env`, etc don't have an explicit parsing function that can be wrapped).

launch/launch/frontend/entity.py
ivanpauno: ups, fixed in https://github.com/ros2/launch/pull/468/commits/dc8fc265f5ef4c7be2645210621d969d2dd0d7f8

launch/launch/frontend/expose.py
ivanpauno: Done in https://github.com/ros2/launch/pull/468/commits/7367cf9119dc6aa6f6edf1f556a8e1ad2d44233f

launch_xml/launch_xml/entity.py
ivanpauno: See https://github.com/ros2/launch/pull/468/commits/caf4b0cf338ee55218cc3c737d17c58d861f17af

ivanpauno: and https://github.com/ros2/launch_ros/pull/198

launch_xml/test/launch_xml/test_arg.py
```diff
@@ -34,3 +36,18 @@ def test_arg():
...
+        <launch>
+            <arg name="my_arg" whats_this="hello" default="asd" description="something"/>
+        </launch>
+        """
+    xml_file = textwrap.dedent(xml_file)
+    root_entity, parser = Parser.load(io.StringIO(xml_file))
+    with pytest.raises(ValueError) as excinfo:
+        parser.parse_description(root_entity)
+    assert '`arg`' in str(excinfo.value)
+    assert 'whats_this' in str(excinfo.value)
```
hidmic, line 53: @ivanpauno can we add a test for a more complex (i.e. nested) description?

hidmic (approved):
LGTM pending green CI!

launch_xml/test/launch_xml/test_arg.py
ivanpauno: See https://github.com/ros2/launch/pull/468/commits/9dfb1a7a7cf575b4193c32b9753fbc958cdaddb9.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12960/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7909/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10678/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12909/)

jacobperron (approved):

adityapande-1995:
@Mergifyio backport foxy

Issue #473, opened 2020-11-18
Shutdown hangs on SIGINT


almaslov:
## Bug report

**Required Info:**

- Operating System:
  - DISTRIB_ID=Ubuntu
  - DISTRIB_RELEASE=20.04
  - DISTRIB_CODENAME=focal
  - DISTRIB_DESCRIPTION="Ubuntu 20.04.1 LTS"
- Installation type:
  - binaries
- Version or commit hash:
  - ros-foxy-ros2launch                             0.10.3-1focal.20201008.001853
- DDS implementation:
  - Fast-RTPS

#### Steps to reproduce issue
Any rarely logging binary/script, e.g. `/tmp/simple_script.bash`:
```
#!/usr/bin/env bash
sleep 20
echo 'Hello'
sleep 20
```

A launch file `/tmp/simple.launch.py`:
```
from launch import LaunchDescription
import launch.actions


def generate_launch_description():
    return LaunchDescription([
        launch.actions.ExecuteProcess(cmd=['/tmp/simple_script.bash'], output='screen')
    ])
```

Run the launch file with debug option:
```
ros2 launch -d /tmp/simple.launch.py
```

In separate terminal tab send SIGINT to launcher process:
```
pkill -2 ros2
```
#### Expected behavior
In best case I expect launcher would forward SIGINT to my bash subprocess and both processes terminate gracefully. At least I expect launcher would terminate in finite time.
#### Actual behavior
Nothing happens until subprocess echoes. Looks like Shutdown event stuck in queue. In the log below you can see there's about 19 second between emitting Shutdown event and processing one:
```
1605703730.0072842 [DEBUG] [launch.launch_context]: emitting event synchronously: 'launch.events.IncludeLaunchDescription'
1605703730.0082226 [INFO] [launch]: All log files can be found below /home/user/.ros/log/2020-11-18-12-48-49-887166-27703d5967b1-7901
1605703730.0083606 [INFO] [launch]: Default logging verbosity is set to DEBUG
1605703730.0106397 [DEBUG] [launch]: processing event: '<launch.events.include_launch_description.IncludeLaunchDescription object at 0x7f462fc195b0>'
1605703730.0108380 [DEBUG] [launch]: processing event: '<launch.events.include_launch_description.IncludeLaunchDescription object at 0x7f462fc195b0>' ✓ '<launch.event_handlers.on_include_launch_description.OnIncludeLaunchDescription object at 0x7f462fc28e80>'
[...]

PR #475, opened 2020-12-16
Support non-interactive launch.LaunchService runs

hidmic:
Connected to #473. This patch allows the user to modify the underlying assumption `launch` currently makes about Ctrl+C behavior i.e. sent to all processes in the process group. In the same vein as `bash` interactive/non-interactive shells (see [here](https://tldp.org/LDP/abs/html/intandnonint.html)).

### DIFF
--- a/launch/launch/launch_context.py
+++ b/launch/launch/launch_context.py
@@ -34,7 +34,8 @@ class LaunchContext:
     """Runtime context used by various launch entities when being visited or executed."""
 
     def __init__(
-        self, *,
+        self,
+        *,
         argv: Optional[Iterable[Text]] = None,
         noninteractive: bool = False
     ) -> None:
--- a/launch/launch/launch_context.py
+++ b/launch/launch/launch_context.py
@@ -33,13 +33,20 @@ from .substitution import Substitution
 class LaunchContext:
     """Runtime context used by various launch entities when being visited or executed."""
 
-    def __init__(self, *, argv: Optional[Iterable[Text]] = None) -> None:
+    def __init__(
+        self, *,
+        argv: Optional[Iterable[Text]] = None,
+        noninteractive: bool = False
+    ) -> None:
         """
         Create a LaunchContext.
 
         :param: argv stored in the context for access by the entities, None results in []
+        :param: noninteractive if True (not default), this service will assume it has
+            no terminal associated e.g. it is being executed from a non interactive script
         """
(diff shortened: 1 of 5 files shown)


hidmic:
I have not added regression tests yet. I plan to. At the very least, check the right events get sent.

wjwwood (approved):
Makes sense.

launch/launch/launch_context.py
```diff
@@ -33,13 +33,20 @@
 class LaunchContext:
     """Runtime context used by various launch entities when being visited or executed."""
 
-    def __init__(self, *, argv: Optional[Iterable[Text]] = None) -> None:
+    def __init__(
+        self, *,
```
wjwwood, line 37: nitpick:

```suggestion
        self,
        *,
```

ivanpauno (approved):

launch/launch/launch_context.py
hidmic: See https://github.com/ros2/launch/pull/475/commits/f985529e48fc0d299426de300075302d425326d2.

hidmic:
CI up to `launch`, `launch_ros`, `test_launch_ros`, `test_communication`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13306/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8239/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11039/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13349/)

wjwwood (approved):

ivanpauno (approved):

hidmic:
Thanks for the review guys!

hidmic:
@jacobperron how do you feel about backporting this to Foxy?

jacobperron:
> @jacobperron how do you feel about backporting this to Foxy?

Sounds okay to me :+1:

felixdivo:
Who would be responsible for that? Is there more involved then cherry-picking these three commits to a new branch targeted at branch `foxy`?

jacobperron:
> Who would be responsible for that? Is there more involved then cherry-picking these three commits to a new branch targeted at branch foxy?

Anyone is welcome to propose a backport PR targeting the `foxy` branch (just as you've described; cherry-picking the relevant commits). If you do, feel free to mention me and/or hidmic.

PR #476, opened 2020-12-16
Handle signals within the asyncio loop.

hidmic:
Connected to #473. This patch uses `signal.set_wakeup_fd` to delegate signal handling to the running loop. Similar to [`asyncio._UnixSelectorEventLoop.add_signal_handler()`](https://github.com/python/cpython/blob/66d3b589c44fcbcf9afe1e442d9beac3bd8bcd34/Lib/asyncio/unix_events.py#L88) implementation, but not limited to Unix platforms.

For additional context, without this patch, in a scenario like the one in #473, events emitted within a signal handler do not seem to wake up the loop.

### DIFF
--- a/launch/launch/utilities/signal_management.py
+++ b/launch/launch/utilities/signal_management.py
@@ -34,8 +34,10 @@ def is_winsock_handle(fd):
     try:
         # On Windows, WinSock handles and regular file handles
         # have disjoint APIs. This test leverages the fact that
-        # attempting to os.dup a WinSock handle will fail.
-        os.close(os.dup(fd))
+        # attempting to get an MSVC runtime file handle from a
+        # WinSock handle will fail.
+        import msvcrt
+        msvcrt.get_osfhandle(fd)
         return False
     except OSError:
         return True
--- a/launch/launch/utilities/signal_management.py
+++ b/launch/launch/utilities/signal_management.py
@@ -14,131 +14,52 @@
 
 """Module for the signal management functionality."""
 
-import platform
+import os
+import socket
 import signal
-import threading
-
-import launch.logging
-
-__signal_handlers_installed_lock = threading.Lock()
-__signal_handlers_installed = False
-__custom_sigint_handler = None
-__custom_sigquit_handler = None
-__custom_sigterm_handler = None
 
 
-def on_sigint(handler):
-    """
-    Set the signal handler to be called on SIGINT.
+class AsyncSafeSignalManager:
 
-    Pass None for no custom handler.
+    def __init__(self, loop):
+        self.__loop = loop
+        self.__handlers = {}
+        self.__prev_wsock_fd = -1
+        self.__wsock, self.__rsock = socket.socketpair()
+        self.__wsock.setblocking(False)
+        self.__rsock.setblocking(False)
 
-    install_signal_handlers() must have been called in the main thread before.
-    It is called automatically by the constructor of `launch.LaunchService`.
-    """
-    global __custom_sigint_handler
-    if handler is not None and not callable(handler):
-        raise ValueError('handler must be callable or None')
-    __custom_sigint_handler = handler
+    def __enter__(self):
+        self.__loop.add_reader(self.__rsock.fileno(), self.__handle_signal)
+        self.__prev_wsock_fd = signal.set_wakeup_fd(self.__wsock.fileno())
+        return self
 
+    def __exit__(self, type_, value, traceback):
+        assert self.__wsock.fileno() == signal.set_wakeup_fd(self.__prev_wsock_fd)
+        self.__loop.remove_reader(self.__rsock.fileno())
 
-def on_sigquit(handler):
-    """
-    Set the signal handler to be called on SIGQUIT.
+    def __handle_signal(self):
+        while True:
+            try:
+                data = self.__rsock.recv(4096)
+                if not data:
+                    break
+                for signum in data:
+                    if signum not in self.__handlers:
+                        continue
+                    self.__handlers[signum](signum)
+                if self.__prev_wsock_fd != -1:
+                    os.write(self.__prev_wsock_fd, data)
+            except InterruptedError:
+                continue
+            except BlockingIOError:
+                break
 
-    Note Windows does not have SIGQUIT, so it can be set with this function,
-    but the handler will not be called.
-
-    Pass None for no custom handler.
-
-    install_signal_handlers() must have been called in the main thread before.
-    It is called automatically by the constructor of `launch.LaunchService`.
-    """
-    global __custom_sigquit_handler
-    if handler is not None and not callable(handler):
-        raise ValueError('handler must be callable or None')
-    __custom_sigquit_handler = handler
-
-
-def on_sigterm(handler):
-    """
-    Set the signal handler to be called on SIGTERM.
-
-    Pass None for no custom handler.
-
-    install_signal_handlers() must have been called in the main thread before.
-    It is called automatically by the constructor of `launch.LaunchService`.
-    """
-    global __custom_sigterm_handler
-    if handler is not None and not callable(handler):
-        raise ValueError('handler must be callable or None')
-    __custom_sigterm_handler = handler
-
-
-def install_signal_handlers():
-    """
-    Install custom signal handlers so that hooks can be setup from other threads.
-
-    Calling this multiple times does not fail, but the signals are only
-    installed once.
-
-    If called outside of the main-thread, a ValueError is raised, see:
-    https://docs.python.org/3.6/library/signal.html#signal.signal
-
-    Also, if you register your own signal handlers after calling this function,
-    then you should store and forward to the existing signal handlers, because
-    otherwise the signal handlers registered by on_sigint(), on_sigquit, etc.
-    will not be run.
-    And the signal handlers registered with those functions are used to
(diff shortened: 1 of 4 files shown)


hidmic:
This is lacking tests and quite a bit of documentation. But I wanted to share this POC as soon as possible for thorough scrutiny.

wjwwood (approved):
The changes make sense to me, but I haven't dug deeply into the signal handler provided by the loop and I haven't tried it out locally.

This is good for me, if we can show it works well on all the platforms.

Have you tried it when the launch service is not in the main thread?

wjwwood:
And yes, it needs docs and tests if we can manage to make them (can be hard with these kind of things).

hidmic:
> Have you tried it when the launch service is not in the main thread?

It won't work, as signal handling API is restricted to the main thread. But that's a limitation we already had. See [here](https://github.com/ros2/launch/blob/55ca523ff0053e1c7932debaa7366e33180c23a6/launch/launch/launch_service.py#L309-L310) and [here](https://github.com/ros2/launch/blob/55ca523ff0053e1c7932debaa7366e33180c23a6/launch/launch/launch_service.py#L65-L68)). Also https://github.com/ros2/launch/issues/126. 

---

FWIW I think we should eventually go away with signal handling in `launch.LaunchService`. Some async-signal-safe APIs for a user to setup their own handlers would pose less restrictions on usage. Moreover, if we could replace signals with a less rudimentary IPC to monitor and interrupt subprocesses, that'd lift all concurrency limitations and improve `launch` behavior on Windows.

hidmic:
CI up to `launch`, `test_launch_ros`, and `test_communication`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13321/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8251/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11052/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13365/)

hidmic:
If this patch gets merged, I'd like to see if it fixes https://github.com/ros2/build_farmer/issues/248.

launch/launch/utilities/signal_management.py
```diff
@@ -12,133 +12,102 @@
...
+        :param signum: number of the signal to be handled
+        :param handler: callback taking a signal number
+          as its sole argument, or None
+        :return: previous handler if any, otherwise None
+        """
+        signum = signal.Signals(signum)
+        if handler is not None and not callable(handler):
+            raise ValueError('signal handler must be a callable')
+        old_handler = self.__handlers.get(signum, None)
+        self.__handlers[signum] = handler
```
ivanpauno, line 112: IIUC `handler` can be `None` here, which can be problematic with the above:

```python3
                    if signum not in self.__handlers:
                        continue
                    self.__handlers[signum](signum)
```

maybe the correct code is
```suggestion
        if handler is not None:
          self.__handlers[signum] = handler
        else:
          self.__handlers.pop(signum, None)
```

ivanpauno (approved):

launch/launch/utilities/signal_management.py
hidmic: Oof, yes. See 3c753e60b1d4a3c16269b4aeb257287c17b632cc. I extended the test to cover this.

ivanpauno (approved):

ivanpauno (approved):
Code too handle the windows quirks seems reasonable too me!

hidmic:
> Code too handle the windows quirks seems reasonable too me!

Of course, it wasn't enough... See https://github.com/ros2/launch/pull/476/commits/71ab3f2c2851333ee19fa62f5fc676c39f84c257.

hidmic:
CI up to `launch`, `test_launch_ros`, and `test_communication`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13326/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8256/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11057/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13375/)

launch/launch/utilities/signal_management.py
```diff
@@ -12,133 +12,157 @@
...
+            self.__prev_wakeup_handle = prev_wakeup_handle
+        assert self.__wsock.fileno() == signal.set_wakeup_fd(self.__prev_wakeup_handle)
+        if self.__background_loop:
+            self.__background_loop.call_soon_threadsafe(self.__background_loop.stop)
+            self.__background_thread.join()
+            self.__background_loop.close()
+        else:
+            self.__loop.remove_reader(self.__rsock.fileno())
+
+    def __handle_signal(self):
```
ivanpauno, line 126: Now on Windows, this is being called from a different thread than the main event loop thread.
Are the callbacks we're registering thread safe? e.g.: is `LaunchService._shutdown()` thread safe? (which is the method called from most signal handlers we're registering).

launch/launch/utilities/signal_management.py
```diff
@@ -12,133 +12,157 @@
...
 
+def is_winsock_handle(fd):
+    """Check if the given file descriptor is WinSock handle."""
+    if platform.system() != 'Windows':
+        return False
+    try:
+        # On Windows, WinSock handles and regular file handles
+        # have disjoint APIs. This test leverages the fact that
+        # attempting to os.dup a WinSock handle will fail.
+        os.close(os.dup(fd))
```
ivanpauno, line 38: is there a better way to test this?

launch/launch/utilities/signal_management.py
```diff
@@ -12,133 +12,157 @@
...
+                asyncio.set_event_loop(self.__background_loop)
+                self.__background_loop.run_forever()
+
+            self.__background_thread = threading.Thread(target=run_background_loop)
+            self.__background_thread.start()
+        self.__prev_wakeup_handle = signal.set_wakeup_fd(self.__wsock.fileno())
+        if self.__prev_wakeup_handle != -1 and is_winsock_handle(self.__prev_wakeup_handle):
+            # On Windows, os.write will fail on a WinSock handle. There is no WinSock API
+            # in the standard library either. Thus we wrap it in a socket.socket instance.
+            self.__prev_wakeup_handle = socket.socket(fileno=self.__prev_wakeup_handle)
```
ivanpauno, line 109: maybe always wrapping the previous `fd` in a `socket.socket` works, without testing `is_winsock_handle`

launch/launch/utilities/signal_management.py
hidmic: I wish there was.

launch/launch/utilities/signal_management.py
hidmic: Thing is, wakeup file descriptors might be regular files too.

launch/launch/utilities/signal_management.py
hidmic: If you look at line 94 above, you'll that the background loop calls back the foreground loop. Or that was my intention.

launch/launch/utilities/signal_management.py
ivanpauno: maybe https://docs.python.org/3/library/msvcrt.html#msvcrt.get_osfhandle?
otherwise I'm ok with the `os.close(os.dup(fd))` combo

launch/launch/utilities/signal_management.py
ivanpauno: makes sense

launch/launch/utilities/signal_management.py
ivanpauno: Makes sense, I didn't see that.

launch/launch/utilities/signal_management.py
hidmic: That might actually work. In fact, `msvcrt.open_osfhandle` may help to wrap socket handles.

hidmic: > In fact, `msvcrt.open_osfhandle` may help to wrap socket handles.

Scratch that, it doesn't work :) But I did replace `os.dup()` with `msvcrt.get_osfhandle()` in https://github.com/ros2/launch/pull/476/commits/584df2fdcdf84edaedeeb24130c3e086bd3815d0. Thanks !

hidmic:
CI up to `launch`, `test_launch_ros`, `test_communication`, and `osrf_pycommon`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13409/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8329/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11134/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13468/)

hidmic:
Going in! As a side note, we should sort out `asyncio` usage in `launch` at some point. It goes great lengths to ensure loop locality, and `asyncio` API is set up for global loops.

ciandonovan:
@hidmic This merge broke SIGTERM handling in Launch by removing the signal handler and not replacing it.

SIGINT wasn't broken only because [Python](https://docs.python.org/3/library/signal.html) installs a signal handler for it by default, translating it into a `KeyboardInterrupt` exception.

Issue with further details: https://github.com/ros2/launch/issues/666
Pull request fixing it: https://github.com/ros2/launch/pull/712

PR #479, opened 2021-01-13
Workaround asyncio signal handling on Unix

hidmic:
Unix `asyncio` loops override existing signal wakeup file descriptors with no regards for existing ones -- set by user code or by another loop. For further reference, see [here](https://github.com/python/cpython/blob/187f76def8a5bd0af7ab512575cad30cfe624b05/Lib/asyncio/unix_events.py#L95). This is a problem when the child watcher gets lazy initialized (e.g. on first async subprocess execution, see [here](https://github.com/python/cpython/blob/187f76def8a5bd0af7ab512575cad30cfe624b05/Lib/asyncio/unix_events.py#L188) and [here](https://github.com/python/cpython/blob/187f76def8a5bd0af7ab512575cad30cfe624b05/Lib/asyncio/unix_events.py#L927)).

CI up to `launch`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13422/)
* CentOS [Build Status](https://ci.ros2.org/job/ci_linux-centos/99/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8337/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11143/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13478/)

### DIFF
--- a/launch/launch/utilities/signal_management.py
+++ b/launch/launch/utilities/signal_management.py
@@ -98,12 +98,31 @@ class AsyncSafeSignalManager:
         self.__rsock.setblocking(False)
 
     def __enter__(self):
+        self.__add_signal_readers()
+        try:
+            self.__install_signal_writers()
+        except Exception:
+            self.__remove_signal_readers()
+            raise
+        self.__chain()
+        return self
+
+    def __exit__(self, exc_type, exc_value, exc_traceback):
+        try:
+            try:
+                self.__uninstall_signal_writers()
+            finally:
+                self.__remove_signal_readers()
+        finally:
+            self.__unchain()
+
+    def __add_signal_readers(self):
         try:
             self.__loop.add_reader(self.__rsock.fileno(), self.__handle_signal)
         except NotImplementedError:
             # Some event loops, like the asyncio.ProactorEventLoop
             # on Windows, do not support asynchronous socket reads.
-            # So we emulate it.
+            # Emulate it.
             self.__background_loop = asyncio.SelectorEventLoop()
             self.__background_loop.add_reader(
                 self.__rsock.fileno(),
@@ -114,26 +133,49 @@ class AsyncSafeSignalManager:
                 asyncio.set_event_loop(self.__background_loop)
                 self.__background_loop.run_forever()
 
-            self.__background_thread = threading.Thread(target=run_background_loop)
+            self.__background_thread = threading.Thread(
+                target=run_background_loop, daemon=True)
             self.__background_thread.start()
-        self.__chain_wakeup_handle(self.__set_wakeup_fd(self.__wsock.fileno()))
+
+    def __remove_signal_readers(self):
+        if self.__background_loop:
+            self.__background_loop.call_soon_threadsafe(self.__background_loop.stop)
+            self.__background_thread.join()
+            self.__background_loop.close()
+            self.__background_loop = None
+        else:
+            self.__loop.remove_reader(self.__rsock.fileno())
+
+    def __install_signal_writers(self):
+        prev_wakeup_handle = self.__set_wakeup_fd(self.__wsock.fileno())
+        try:
+            self.__chain_wakeup_handle(prev_wakeup_handle)
+        except Exception:
+            own_wakeup_handle = self.__set_wakeup_fd(prev_wakeup_handle)
+            assert self.__wsock.fileno() == own_wakeup_handle
+            raise
+
+    def __uninstall_signal_writers(self):
+        prev_wakeup_handle = self.__chain_wakeup_handle(-1)
+        own_wakeup_handle = self.__set_wakeup_fd(prev_wakeup_handle)
+        assert self.__wsock.fileno() == own_wakeup_handle
+
+    def __chain(self):
         self.__parent = AsyncSafeSignalManager.__current
         AsyncSafeSignalManager.__current = self
         if self.__parent is None:
             # Do not trust signal.set_wakeup_fd calls within context.
             # Overwrite handle at the start of the managers' chain.
-            signal.set_wakeup_fd = self.__chain_wakeup_handle
-        return self
+            def modified_set_wakeup_fd(signum):
+                if threading.current_thread() is not threading.main_thread():
+                    raise ValueError(
+                        'set_wakeup_fd only works in main'
+                        ' thread of the main interpreter'
+                    )
+                return self.__chain_wakeup_handle(signum)
+            signal.set_wakeup_fd = modified_set_wakeup_fd
 
-    def __exit__(self, type_, value, traceback):
-        prev_wakeup_handle = self.__chain_wakeup_handle(-1)
-        assert self.__wsock.fileno() == self.__set_wakeup_fd(prev_wakeup_handle)
-        if self.__background_loop:
-            self.__background_loop.call_soon_threadsafe(self.__background_loop.stop)
-            self.__background_thread.join()
-            self.__background_loop.close()
-        else:
-            self.__loop.remove_reader(self.__rsock.fileno())
+    def __unchain(self):
         if self.__parent is None:
             signal.set_wakeup_fd = self.__set_wakeup_fd
         AsyncSafeSignalManager.__current = self.__parent
--- a/launch/launch/utilities/signal_management.py
+++ b/launch/launch/utilities/signal_management.py
@@ -68,6 +68,10 @@ class AsyncSafeSignalManager:
     descriptor, if any.
     """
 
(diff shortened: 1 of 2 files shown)


hidmic:
I'll open another ticket upstream. `asyncio` signal handling is awful.

launch/launch/utilities/signal_management.py
```diff
@@ -104,26 +109,38 @@ def run_background_loop():
...
-        if self.__prev_wakeup_handle != -1 and is_winsock_handle(self.__prev_wakeup_handle):
-            # On Windows, os.write will fail on a WinSock handle. There is no WinSock API
-            # in the standard library either. Thus we wrap it in a socket.socket instance.
-            self.__prev_wakeup_handle = socket.socket(fileno=self.__prev_wakeup_handle)
+        self.__chain_wakeup_handle(self.__set_wakeup_fd(self.__wsock.fileno()))
+        self.__parent, self.__current = self.__current, self
+        if self.__parent is None:
+            # Do not trust signal.set_wakeup_fd calls within context.
+            # Overwrite handle at the start of the managers' chain.
+            signal.set_wakeup_fd = self.__chain_wakeup_handle
```
ivanpauno, line 117: ?

I'm not sure if this is ok or not ...

launch/launch/utilities/signal_management.py
```diff
@@ -104,26 +109,38 @@ def run_background_loop():
...
             self.__loop.remove_reader(self.__rsock.fileno())
+        if self.__parent is None:
+            signal.set_wakeup_fd = self.__set_wakeup_fd
+        self.__current = self.__parent
+
+    def __chain_wakeup_handle(self, wakeup_handle):
+        prev_wakeup_handle = self.__prev_wakeup_handle
+        if isinstance(prev_wakeup_handle, socket.socket):
+            # Detach (Windows) socket and retrieve the raw OS handle.
+            prev_wakeup_handle, _ = prev_wakeup_handle.fileno(), prev_wakeup_handle.detach()
```
ivanpauno, line 137: isn't 

```suggestion
            prev_wakeup_handle = prev_wakeup_handle.detach()
```

the same thing?

launch/launch/utilities/signal_management.py
```diff
@@ -104,26 +109,38 @@ def run_background_loop():
...
+        assert self.__wsock.fileno() == self.__set_wakeup_fd(prev_wakeup_handle)
         if self.__background_loop:
             self.__background_loop.call_soon_threadsafe(self.__background_loop.stop)
             self.__background_thread.join()
             self.__background_loop.close()
         else:
             self.__loop.remove_reader(self.__rsock.fileno())
+        if self.__parent is None:
+            signal.set_wakeup_fd = self.__set_wakeup_fd
+        self.__current = self.__parent
```
ivanpauno, line 131: do you meant to set the global `__current`?

launch/launch/utilities/signal_management.py
```diff
@@ -104,26 +109,38 @@ def run_background_loop():
 
             self.__background_thread = threading.Thread(target=run_background_loop)
             self.__background_thread.start()
-        self.__prev_wakeup_handle = signal.set_wakeup_fd(self.__wsock.fileno())
-        if self.__prev_wakeup_handle != -1 and is_winsock_handle(self.__prev_wakeup_handle):
-            # On Windows, os.write will fail on a WinSock handle. There is no WinSock API
-            # in the standard library either. Thus we wrap it in a socket.socket instance.
-            self.__prev_wakeup_handle = socket.socket(fileno=self.__prev_wakeup_handle)
+        self.__chain_wakeup_handle(self.__set_wakeup_fd(self.__wsock.fileno()))
+        self.__parent, self.__current = self.__current, self
```
ivanpauno, line 113: global `__current`?

ivanpauno (commented):
I'm not really comfortable with the fact that this PR overwrites `signal.set_wakeup_fd`, but the code looks reasonable and I don't know how to fix this in another way.

It's worth writing a big comment explaining why this is being done in the code.

LGTM with nits addressed.

launch/launch/utilities/signal_management.py
hidmic: It is not, but until `asyncio` signal handling is fixed, it's the best workaround I could come up with.

ivanpauno: Is the a way to pre-initialize the child watcher so that happens before we install a signal handler?

launch/launch/utilities/signal_management.py
hidmic: It is! Not sure why I thought it returned `None`.

launch/launch/utilities/signal_management.py
hidmic: Oops, yes, using `self` for class variables works on `__getattribute__`, not `__setattribute__`.

launch/launch/utilities/signal_management.py
hidmic: Yeap

hidmic:
@ivanpauno PTAL @ https://github.com/ros2/launch/pull/479/commits/a15958abbca507dcef2931d78bc1da48240bf64d

ivanpauno:
> I'll open another ticket upstream. asyncio signal handling is awful.

does this PR mean that signal handlers added with [asyncio.loop.add_signal_handler](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.add_signal_handler) are ignored if you later happen to asynchronously launch a process?

Really, that's pretty broken :smile: ....

hidmic:
> does this PR mean that signal handlers added with asyncio.loop.add_signal_handler are ignored if you later happen to asynchronously launch a process?

`signal.set_wakeup_fd` is called every time you do `add_signal_handler`, so not ignored, no. Problem is that `asyncio` loops do not cooperate with the rest of the system (or with each other). They own `signal.set_wakeup_fd`. If a signal handler gets added at the wrong time, whatever was set is gone.

hidmic:
> what does happen in a multi-threaded scenario?

[`signal.set_wakeup_fd`](https://docs.python.org/3/library/signal.html#signal.set_wakeup_fd) can only be used in the main thread.

But now that I think about it, I should enforce it for the `AsyncSafeSignalManager` too.

ivanpauno:
> signal.set_wakeup_fd is called every time you do add_signal_handler, so not ignored, no. Problem is that asyncio loops do not cooperate with the rest of the system (or with each other). They own signal.set_wakeup_fd. If a signal handler gets added at the wrong time, whatever was set is gone.



I mean, does something like:

```python3
loop.add_signal_handler(...)
# first process launch in next line, child watcher initialized there
process = asyncio.create_subprocess_exec(...)
```

work, or is that also broken?

If that's not broken, if they would support `add_signal_handler` on Windows everything would be more happy ...

hidmic:
> work, or is that also broken?

If it's the same loop for both calls, it works. If it's different loops, last one wins :D

hidmic:
https://github.com/ros2/launch/pull/479/commits/f90f8bcc1bb568b9e91536bf31e8d142e3d5ca6b should take care of multi-threaded scenarios. It's (*I think*) exception-safe too.

hidmic:
CI up to `launch`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13426/)
* CentOS [Build Status](https://ci.ros2.org/job/ci_linux-centos/100/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8342/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11147/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13483/)

ivanpauno (approved):

launch/launch/utilities/signal_management.py
```diff
@@ -114,26 +133,49 @@ def run_background_loop():
                 asyncio.set_event_loop(self.__background_loop)
                 self.__background_loop.run_forever()
 
-            self.__background_thread = threading.Thread(target=run_background_loop)
+            self.__background_thread = threading.Thread(
+                target=run_background_loop, daemon=True)
```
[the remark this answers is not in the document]
hidmic: To deal with the largely improbable scenario in which we fail to stop the `asyncio` loop that's running there (e.g. if `loop.call_soon_threadsafe` raises before we get to `threading.Thread.join`).

hidmic:
Alright, all's green and reviewer's happy. Going in!

PR #494, opened 2021-03-26
Remove is_winsock_handle() and instead test if wrapping the handle in a socket.socket() works

ivanpauno:
When investigating failures in the nightly windows debug job I was getting a lot of debug assertions dialogs about invalid file descriptions.
I don't think this was causing the CI failures, but it was pretty anoying when debugging locally.

Testing `os.dup(os.close())` had a similar issue.
This seems to work fine :crossed_fingers: .

### DIFF
--- a/launch/launch/utilities/signal_management.py
+++ b/launch/launch/utilities/signal_management.py
@@ -16,7 +16,6 @@
 
 import asyncio
 import os
-import platform
 import signal
 import socket
 import threading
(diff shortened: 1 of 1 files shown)


hidmic (approved):
LGTM :crossed_fingers:

ivanpauno:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14194/)

PR #498, opened 2021-03-30
Only try to wrap the fd in a socket on Windows

ivanpauno:
After rereading https://github.com/ros2/launch/pull/494 I noticed that we shouldn't be trying to wrap the `fd` in a socket on non-Windows platforms.

I'm not sure if that could cause a bug or not, but this is more clean.

### DIFF
--- a/launch/launch/utilities/signal_management.py
+++ b/launch/launch/utilities/signal_management.py
@@ -16,6 +16,7 @@
 
 import asyncio
 import os
+import platform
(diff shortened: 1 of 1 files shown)


hidmic (approved):
Before #494, socket wrapping could only happen on Windows. Perhaps that's why it was overlooked. 

Anyhow, LGTM (pending green CI)!

PR #497, opened 2021-03-30
Close the socket pair used for signal management

ivanpauno:
I saw some log warnings related to this when testing things on Windows (using `python_d`).

PR #506, opened 2021-05-04
[Foxy] Handle signals within the asyncio loop. (#476)

mjeronimo:
Signed-off-by: Michel Hidalgo <michel@ekumenlabs.com>

Backporting https://github.com/ros2/launch/pull/476 to Foxy. 
Fixes: https://github.com/ros2/launch_ros/issues/223

### DIFF
--- a/launch/launch/utilities/__init__.py
+++ b/launch/launch/utilities/__init__.py
@@ -20,6 +20,7 @@ from .ensure_argument_type_impl import ensure_argument_type
 from .normalize_to_list_of_substitutions_impl import normalize_to_list_of_substitutions
 from .perform_substitutions_impl import perform_substitutions
 from .signal_management import AsyncSafeSignalManager
+from .signal_management import install_signal_handlers, on_sigint, on_sigquit, on_sigterm
 from .visit_all_entities_and_collect_futures_impl import visit_all_entities_and_collect_futures
 
 __all__ = [
@@ -30,6 +31,10 @@ __all__ = [
     'ensure_argument_type',
     'perform_substitutions',
     'AsyncSafeSignalManager',
+    'install_signal_handlers',
+    'on_sigint',
+    'on_sigquit',
+    'on_sigterm',
     'normalize_to_list_of_substitutions',
     'visit_all_entities_and_collect_futures',
 ]
--- a/launch/launch/utilities/__init__.py
+++ b/launch/launch/utilities/__init__.py
@@ -19,10 +19,7 @@ from .create_future_impl import create_future
 from .ensure_argument_type_impl import ensure_argument_type
 from .normalize_to_list_of_substitutions_impl import normalize_to_list_of_substitutions
 from .perform_substitutions_impl import perform_substitutions
-from .signal_management import install_signal_handlers
-from .signal_management import on_sigint
-from .signal_management import on_sigquit
-from .signal_management import on_sigterm
+from .signal_management import AsyncSafeSignalManager
 from .visit_all_entities_and_collect_futures_impl import visit_all_entities_and_collect_futures
 
 __all__ = [
@@ -32,10 +29,7 @@ __all__ = [
     'create_future',
     'ensure_argument_type',
     'perform_substitutions',
-    'install_signal_handlers',
-    'on_sigint',
-    'on_sigquit',
-    'on_sigterm',
+    'AsyncSafeSignalManager',
     'normalize_to_list_of_substitutions',
     'visit_all_entities_and_collect_futures',
 ]
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -18,6 +18,7 @@ import asyncio
 import collections.abc
 import contextlib
 import logging
+import platform
 import signal
 import threading
 import traceback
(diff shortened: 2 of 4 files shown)


hidmic:
If we backport #476, we have to backport #479, #494, #497, and #498 too. Signal handling is tricky.

mjeronimo:
@hidmic  Do you recommend proceeding with those for Foxy? Is it worth it for this issue?

hidmic:
@mjeronimo all those PRs patch #476 in one way or another. #479 deals with an issue in `asyncio` that became a problem when signal management implementation changed in #476. #494, #497, and #498 fix issues in #476 in various ways (mainly for Windows).

mjeronimo:
@jacobperron  What do you think about backporting this set of changes to Foxy? Would like to get your input before proceeding.

jacobperron:
I'm okay with backporting all of the changes pointed out, as long as @hidmic doesn't have any major reservations about them. On the surface of it, it looks like it's worth the effort to improve things in Foxy, but I don't know enough about the signal handling changes to say if we're going to break users code.

Of all of the changes, the only thing I would change for Foxy would be to keep this function (unused) to maintain API compatibility: 

https://github.com/ros2/launch/pull/494/files#diff-5dcef8ac9a56749cd4f33305fea85b4c918383b74c8c1369cce3a22286d84ee9L30

hidmic:
> as long as @hidmic doesn't have any major reservations about them

No particular reservations. It's sensitive code, that's all.

> Of all of the changes, the only thing I would change for Foxy would be to keep this function (unused) to maintain API compatibility:

That function didn't exist before #476, so it doesn't exist in Foxy. On the other hand, this patch removes quite a few functions in [`signal_management.py`](https://github.com/ros2/launch/pull/506/files#diff-5dcef8ac9a56749cd4f33305fea85b4c918383b74c8c1369cce3a22286d84ee9). Off the top of my head, I don't think we can keep them around. Maybe as a no-op?

felixdivo:
I can probably offer to help with testing these changes. It can't be too hard to set it up properly.

felixdivo:
The fix would be important for me / my team. Can I assist you in some way?

Keeping the old API and having it do nothing seems like the only sensible way (the only one with moderate coding effort). One should perhaps also print a warning pointing to this PR or otherwise explaining that the functionality has moved and the function won't do anything any more. That way, these is actually merit in keeping it.

launch/launch/utilities/__init__.py
```diff
@@ -32,10 +29,7 @@
     'create_future',
     'ensure_argument_type',
     'perform_substitutions',
-    'install_signal_handlers',
-    'on_sigint',
-    'on_sigquit',
-    'on_sigterm',
```
jacobperron, line 38: Removing these APIs will break anyone using them directly.

jacobperron (commented):
I overlooked the changes in the PR originally (and only looked at the follow-up fixes).

I'm very hesitant about backporting this to Foxy since it breaks API. I think we should investigate either:

1. Making an API-compatible fix here (or in launch_ros).
2. Keeping the old API, and keep it functional.

I don't like the idea of making the old API a noop (even if it produces a warning). This can cause a subtle break for the user versus a hard break (if the API does not exist). If we are going to produce a warning, we may as well remove the old API since it forces the user to update instead of potentially having this break subtly.

hidmic:
> Keeping the old API, and keep it functional.

Hmm, perhaps we can use the fact that `asyncio` loops are thread-local globals and re-implement this API with a global `AsyncSignalSafeManager` instance. I haven't thought through all the implications though.

felixdivo:
I just want to ping this issue and find out what the roadmap might be. Is there someone willing to tackle this? I of course unterstand, that this is a rather nasty change and personally also do not want to dig into it.

So if the decision is taken that there simply will not be a fix and foxy will be left with the bug ☹️, it would be great to spell it out in this issue. That way, any stakeholder knows what to do. I my case, I will probably have to hack something together in our product until the next ROS2 LTS is released.

hidmic:
I just put together a POC in #511. It needs testing, and Foxy ROS Boss approval.

hidmic:
Full Foxy CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14824/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9574/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12490/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14979/)

felixdivo:
As with #511, this PR should probably be added to the [Foxy Patch Release 5](https://github.com/orgs/ros2/projects/30)

hidmic:
Test failures seem unrelated. Are we good to go @jacobperron @mjeronimo ?

mjeronimo:
Looks good to me. Thanks for taking on this complicated issue.

hidmic:
@jacobperron I'd like your blessing before I merge this.

felixdivo:
Where should I look for a possible release date of the [Foxy Patch Release 5](https://github.com/orgs/ros2/projects/30) that this will be part of?

hidmic (approved):

jacobperron (approved):

hidmic:
Alright, moving forward !

PR #511, opened 2021-06-16
Complete #506 backport and deprecate signal_management APIs

hidmic:
Precisely what the title says. Connected to #506. Needs https://github.com/osrf/osrf_pycommon/pull/75.

### DIFF
--- a/launch/launch/utilities/signal_management.py
+++ b/launch/launch/utilities/signal_management.py
@@ -16,7 +16,6 @@
 
 import asyncio
 import os
-import platform
 import signal
 import socket
 import threading
@@ -26,21 +25,10 @@ from typing import Optional
 from typing import Tuple  # noqa: F401
 from typing import Union
 
-
-def is_winsock_handle(fd):
-    """Check if the given file descriptor is WinSock handle."""
-    if platform.system() != 'Windows':
-        return False
-    try:
-        # On Windows, WinSock handles and regular file handles
-        # have disjoint APIs. This test leverages the fact that
-        # attempting to get an MSVC runtime file handle from a
-        # WinSock handle will fail.
-        import msvcrt
-        msvcrt.get_osfhandle(fd)
-        return False
-    except OSError:
-        return True
+try:
+    _WindowsError = WindowsError
+except NameError:
+    _WindowsError = None
 
 
 class AsyncSafeSignalManager:
@@ -185,10 +173,14 @@ class AsyncSafeSignalManager:
         if isinstance(prev_wakeup_handle, socket.socket):
             # Detach (Windows) socket and retrieve the raw OS handle.
             prev_wakeup_handle = prev_wakeup_handle.detach()
-        if wakeup_handle != -1 and is_winsock_handle(wakeup_handle):
+        if wakeup_handle != -1:
             # On Windows, os.write will fail on a WinSock handle. There is no WinSock API
             # in the standard library either. Thus we wrap it in a socket.socket instance.
-            wakeup_handle = socket.socket(fileno=wakeup_handle)
+            try:
+                wakeup_handle = socket.socket(fileno=wakeup_handle)
+            except _WindowsError as e:
+                if e.winerror != 10038:  # WSAENOTSOCK
+                    raise
         self.__prev_wakeup_handle = wakeup_handle
         return prev_wakeup_handle
 
--- a/launch/launch/utilities/signal_management.py
+++ b/launch/launch/utilities/signal_management.py
@@ -17,6 +17,7 @@
 import asyncio
 from contextlib import ExitStack
 import os
+import platform
 import signal
 import socket
 import threading
@@ -25,11 +26,6 @@ from typing import Callable
 from typing import Optional
 from typing import Tuple  # noqa: F401
 from typing import Union
-
-try:
-    _WindowsError = WindowsError
-except NameError:
-    _WindowsError = None
 
 
 class AsyncSafeSignalManager:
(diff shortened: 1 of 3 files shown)


hidmic:
@jacobperron this is what the change to Foxy would look like. Unit tests pass locally, though I have not run full Foxy CI just yet.

felixdivo:
@hidmic I could not find any docs on how to install this locally while overriding the stuff installed via apt. Do I simply clone and `pip install .` it?

hidmic:
> @hidmic I could not find any docs on how to install this locally while overriding the stuff installed via apt. Do I simply clone and pip install . it globally (for the time being)?

The simplest approach is probably to build an overlay workspace with `osrf_pycommon` and `launch` in it, checked out at the right branches.

felixdivo:
Okay, I managed to install it in a more hacky way but it worked. The PR seems the fix the issue, everything seems to work fine when I issue a SIGINT via `process.send_signal(SIGINT)`:

```
[WARNING] [launch]: user interrupted with ctrl-c (SIGINT)
[DEBUG] [launch.launch_context]: emitting event synchronously: 'launch.events.Shutdown'
[DEBUG] [launch]: processing event: '<launch.events.shutdown.Shutdown object at 0x7fe5c507a910>'
[DEBUG] [launch]: processing event: '<launch.events.shutdown.Shutdown object at 0x7fe5c507a910>' ✓ '<launch.event_handlers.on_shutdown.OnShutdown object at 0x7fe5c507a220>'
[DEBUG] [launch.launch_context]: emitting event synchronously: 'launch.events.process.SignalProcess'
[DEBUG] [launch]: processing event: '<launch.events.shutdown.Shutdown object at 0x7fe5c507a910>' ✓ '<launch.event_handlers.on_shutdown.OnShutdown object at 0x7fe5c519e190>'
[DEBUG] [launch]: processing event: '<launch.events.process.signal_process.SignalProcess object at 0x7fe5c507ae80>'
[DEBUG] [launch]: processing event: '<launch.events.process.signal_process.SignalProcess object at 0x7fe5c507ae80>' ✓ '<launch.event_handler.EventHandler object at 0x7fe5c4684fa0>'
[INFO] [mynode-1]: sending signal 'SIGINT' to process[mynode-1]
[INFO] [mynode-1]: process has finished cleanly [pid 16868]
```

hidmic:
Full Foxy CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14788/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9537/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12454/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14935/)

hidmic:
Full Foxy CI after https://github.com/ros2/launch/pull/511/commits/4b7e29b5616ad990943340ddda88546e64fb3a0d:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14794/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9543/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12460/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14941/)

felixdivo:
Since this is soon coming to an end thanks to @hidmic, shouldn't this be added to [Foxy Patch Release 5
](https://github.com/orgs/ros2/projects/30)?

launch/launch/utilities/signal_management.py
```diff
@@ -240,3 +241,107 @@ def handle(
...
+
+    Pass None for no custom handler.
+
+    install_signal_handlers() must have been called in the main thread before.
+
+    .. deprecated:: Foxy
+
+      Use AsyncSafeSignalManager instead
+    """
+    warnings.warn(
```
jacobperron, line 265: I'd rather not release deprecation warnings into a long-standing ROS distro to avoid disruption. Besides, users who have already switched to Galactic would have already run into the API change without a tick-tock cycle. There may be other opinions.

jacobperron (commented):
Looks okay to me; one concern about the deprecation warnings.

launch/launch/utilities/signal_management.py
felixdivo: I can only give one user's opinion (mine/my team's): I'd like to get warned of breakages as soon a possible. We'd have to change it eventually in any way since we move over to newer versions of ROS2 eventually (LTS -> LTS). And if not, a `warnings.simplefilter()` ist easy to implement if one does not want to see the message for the time being.

But you know best what's common practice in ROS. This isn't Rolling Ridley, after all.

clalancette: I have to agree with @jacobperron ; I don't think we should introduce new deprecation warnings into a stable distribution.  Having new warnings show up for Foxy users as a result of `sudo apt-get update` is really user-unfriendly.

When people port to Galactic or Rolling, they'll get the warnings.

hidmic: Alright, see https://github.com/ros2/launch/pull/511/commits/d8e320a42597348ad5bbc135c58954932733a8d6. 

However I will say that, while I agree that introducing warnings into a patch release of a stable distribution is not ideal, this patch is not transparent. Signals now go through wakeup file descriptors and handling is deferred to the main-thread local `asyncio` loop (assuming it is run). Depending on how the old signal management API was used (the only case in which you'd get a warning) this can potentially break those use cases -- silently as of https://github.com/ros2/launch/pull/511/commits/d8e320a42597348ad5bbc135c58954932733a8d6.

Which is why I didn't even think of backporting anything, and I probably wouldn't if these bits were not kinda broken in Foxy.

> When people port to Galactic or Rolling, they'll get the warnings.

The old API was completely removed in Galactic, so no warnings. This patch is a retrofit of the original patch.

jacobperron: I didn't fully comprehend the nature of the change. If there's a risk that we're breaking somebody, then I think issuing a warning is a good thing. I don't like the idea of releasing breaking changes in Foxy, but if you think the fix is worth the potential churn then I'd rather be very loud about it.

hidmic: So, without this fix, `launch` hangs every now and then. It's the same we've experienced with `ros2cli` tests before. And it hit @felixdivo, so it may hit others too.

I agree about being loud. I thought a warning would be OK, but perhaps it's too loud. @clalancette ?

clalancette: It seems like we are stuck between a rock and a hard place.  While I don't like introducing new warnings into a stable distribution, silently breaking users isn't good either.

In this case, I'll defer to your best judgement on what to do here.  You obviously have a much better handle on the issue and launch in general than I do.

jacobperron: Presumably, the these signal APIs are not very popular so I'd hope the number of affected users would be small.

hidmic: Alright, I put deprecation warnings back in. I'll merge this PR into #506 and run CI there.

Issue #507, opened 2021-05-10
Ability to clear current launch configurations


djchopp:
#### Feature description
Somewhat related to #313 #501

Consider the following launch file setup:

1. A base launch file has argument `arg1` with default of `base_arg`
2. Another launch file also has `arg1` (same name, different purpose) with default of `ext_arg` and is included in the base launch file

Upon running the base launch file, `arg1` within the included launch file will not have its default value of `ext_arg`, but rather it will be `base_arg`.

By default there is no scoping and the base value of `arg1` will be passed on to the included launch file (overwriting the default). Surrounding the included launch file with `PushLaunchConfigurations()` and `PopLaunchConfigurations()` will not change this as these do not clear configurations, but only scope modifications to the existing context. Same goes for putting the incude in a `GroupAction`. One could push and  `UnsetLaunchConfiguration('arg_name')` for each configuration they do not want, but this is tedious for included launch files with large numbers of configurations.

I would suggest a `ClearContextConfiguration` action that can be called after a `PushLaunchConfigurations()`, or as the first action in a `GroupAction`.  Extremely simple implementation below, but could be expanded to have whitelist and blacklist functionality. 

Is this feature wanted/inline with recommended design patterns? Is there another way to accomplish this without a new addition?

Its simple enough I could probably do the PR for it.

```python
class ClearContextConfiguration(Action):
    """
    Action that clears the launch configurations from the context.
    """

    def __init__(self, **kwargs) -> None:
        """Create a ClearContextConfiguration action."""
        super().__init__(**kwargs)

    def execute(self, context: LaunchContext):
        """Execute the action."""
        context.launch_configurations.clear()
```

hidmic:
> Upon running the base launch file, arg1 within the included launch file will not have its default value of ext_arg, but rather it will be base_arg.

I agree this is a problem, and an awfully subtle one. 

> Is this feature wanted/inline with recommended design patterns? Is there another way to accomplish this without a new addition?

Your proposal is straightforward. A `ClearLaunchConfigurations` would be in line with `Push*` and `Pop*`. If you could also update include and group actions to support opt-out scoping and explicit scope forwarding (i.e. to push/clear/pop by default) I think that'd make for a less surprising UX. CC @ivanpauno @wjwwood 

Would you be willing to take a stab at it @djchopp?

djchopp:
@hidmic Sure, I'll see what I can do.

hidmic:
Friendly ping @djchopp.

[...]

PR #515, opened 2021-06-21
Feature clear launch configs

djchopp:
Adds the feature/functionality described in issue #507 such that launch configurations can be cleared within a current scope. The `ClearLaunchConfigurations` action compliments the `Push*` and `Pop*` actions with the following use pattern:

```
PushLaunchConfigurations()
ClearLaunchConfigurations()
# Scoped and isolated actions here
PopLaunchConfigurations()
```

Additionally the `GroupAction` has been updated to also support clearing of the launch configurations utilizing the pattern above.


launch/launch/actions/clear_launch_configurations.py
```diff
@@ -0,0 +1,53 @@
...
+
+from ..utilities import perform_substitutions
+
+
+class ClearLaunchConfigurations(Action):
+    """Action that clears the launch configurations from the context."""
+
+    def __init__(
+        self,
+        forwarded_configurations: Optional[List[SomeSubstitutionsType]] = None,
```
hidmic, line 32: @djchopp nit: why talk about `forwarded_configurations` here?
```suggestion
        launch_configurations: Optional[List[SomeSubstitutionsType]] = None,
```

launch/launch/actions/clear_launch_configurations.py
```diff
@@ -0,0 +1,53 @@
...
+        """Create a ClearLaunchConfigurations action."""
+        super().__init__(**kwargs)
+        self.__forwarded_configurations = forwarded_configurations
+
+    def execute(self, context: LaunchContext):
+        """Execute the action."""
+        if self.__forwarded_configurations is not None:
+            forwarded_keys = []
+            for cfg in self.__forwarded_configurations:
+                key = perform_substitutions(context, cfg.variable_name)
```
hidmic, line 44: @djchopp since this assumes `cfg` is an instance of `LaunchConfiguration`, consider updating its type annotation accordingly.

launch/launch/actions/group_action.py
```diff
@@ -48,12 +49,14 @@ def __init__(
         *,
         scoped: bool = True,
         launch_configurations: Optional[Dict[SomeSubstitutionsType, SomeSubstitutionsType]] = None,
+        forwarded_configurations: Optional[List[SomeSubstitutionsType]] = None,
```
hidmic, line 52: @djchopp hmm, I perceive some redundancy between `launch_configurations` and `forwarded_configurations`. IIUC, the only use case not covered by the current `GroupAction` is a non-forwarding, scoped group.

Can we achieve the same using a `forwarding` boolean, set to `True` by default?

launch/launch/actions/clear_launch_configurations.py
wjwwood: Maybe more like `launch_configurations_to_be_cleared`. A docstring enty explaining the purpose of this argument would be nice too. I had to read the code to figure out what it was doing.

wjwwood: Or, now that I'm looking more closely, I'm not even sure at the moment what the type of these "forwarded_configurations" should be... so definitely need a docstring and maybe even an example in the docstring of how it is meant to be used.

wjwwood: Looks more like `launch_configurations_to_not_be_cleared`.

wjwwood (commented):
Concept seems solid to me, as well as implementation. If we can clear up some documentation/interface concerns and add a few tests, I think it will be good to go.

launch/launch/actions/group_action.py
djchopp: My understanding of the current `GroupAction` is that it does not support the "scoped and non-forwarding" or "scoped and partial-forwarding" cases. I think it is possible to merge `forwarded_configurations` into `launch_configurations` and add add the boolean for `forwarding` while supporting "scoped and non-forwarding", "scoped and partial-forwarding", "scoped and full-forwarded", and "unscoped" using the following table:

| Value of `scoped` | Value of `forwarded` | Key/Value pair in `launch_configurations` | Result |
| --- | --- | --- | --- |
| `True` | `True` or `False` | `Arg1: 'NewValue'` | `Arg1` will have value `NewValue` in the scoped context regardless if it existed or not |
| `True` | `True` | `Arg1: None` | `Arg1` will be passed through to the scoped context with its existing value |
| `True` | `False` | `Arg1: None` | No action is taken, but the end result is `Arg1` still being passed through to the scoped context |
| `True` | `True` | Key for an existing launch configuration does not exist in the `launch_configurations` dict | The key/value will be cleared from the scoped context |
| `True` | `False` | Key for an existing launch configuration does not exist in the `launch_configurations` dict | The key/value will still pass through to the scoped context unmodified |
| `False` | Does not matter |  | The group actions are not scoped so no context launch configurations are modified |

Using `Arg1: None` to signify forwarding may be a little confusing, but I am not sure how else to support partial-forwarding without it. Open to idea's. In any case I can provide a docstring to help describe the behavior

launch/launch/actions/clear_launch_configurations.py
djchopp: @hidmic Good point `launch_configurations` probably makes more sense although @wjwwood suggestion of  `launch_configurations_to_not_be_cleared` is the most descriptive of what it does. 

I modeled its use after `UnsetLaunchConfiguration` where you pass it a list of `LaunchConfiguration` objects as seen in this example:

      arg1 = LaunchConfiguration('arg1')

      return LaunchDescription([
        DeclareLaunchArgument(
          'arg1',
          default_value='default_1',
          description='Argument'
        ),
        # Some actions where arg1 is set to 'default_1'
        ClearLaunchConfigurations(launch_configurations_to_not_be_cleared=[arg1])
        # Some more actions where arg1 is not set
      ])

Should I be using a list of strings (keys) instead?

djchopp: Disregard my last comment on using `LaunchConfiguration` objects. I misunderstood how `UnsetLaunchConfiguration` worked. I will change the implementation to match the style of using the key name as a string.

launch/launch/actions/clear_launch_configurations.py
djchopp: Updating to support any substitution that results in a key name (string). This is more in line with how `UnsetLaunchConfiguration` works. Existing data annotation of `Optional[List[SomeSubstitutionsType]]` will then be accurate. I will add usage to the docstring.

launch/launch/actions/clear_launch_configurations.py
```diff
@@ -0,0 +1,62 @@
...
+
+    If a member of launch_configurations_to_not_be_cleared matches a key
+    in the existing context, it will be preserved in the current context.
+    All other launch configurations will be cleared from the current.
+    context.
+    """
+
+    def __init__(
+        self,
+        launch_configurations_to_not_be_cleared: Optional[List[SomeSubstitutionsType]] = [],
```
wjwwood, line 43: I would avoid defaulting to `[]`, and instead use `None`, due to a common python gotcha referred to as the "mutable default argument anti-pattern", see: https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

wjwwood (changes requested):
Documentation looks good to me, and implementation makes sense.

I did request one change related to the default argument, but I think it can trivially be fixed by using `None`.

I would still like to see some tests for this new action and extensions to the group action tests.

launch/launch/actions/clear_launch_configurations.py
djchopp: Thanks for the link, that makes sense. I will make the change and add in tests.

launch/launch/actions/group_action.py
```diff
@@ -34,39 +35,72 @@
...
+    When scoped is set to True, changes to launch configurations are
+    limited to the scope of actions in the group action.
+
+    When scoped is set to True and forwarded is set to True, all existing
+    launch configurations are available in the scoped context.
+
+    When scoped is set to True and forwarded is set to False, all existing
+    launch configurations are removed from the scoped context unless it is
+    explicitly forwarded by adding its key to the launch_configurations
+    dictionary with a None value.
```
wjwwood, line 55: This feels weird to me, maybe this is what you were talking about when @hidmic suggested deduplicating the two lists. I'll wait for feedback from others before saying what I think we should do (not sure yet).

wjwwood (commented):
One point for discussion, but looking good otherwise, maybe time to bring it out of draft?

launch/launch/actions/group_action.py
hidmic: @djchopp I'd suggest a simpler approach. Groups (and includes) can be non-scoped, scoped and non-forwarding, and scoped and forwarding. "Partial-forwarding" (`'Arg1': None` here) and explicit argument passing (`'Arg1': LaunchConfiguration('Arg1')`) are functionally equivalent and the latter reads better, which is why I think we do not need the former nor two lists.

launch/launch/actions/group_action.py
```diff
@@ -34,39 +35,72 @@
...
+        )
+        ...
     """
 
     def __init__(
         self,
         actions: Iterable[Action],
         *,
         scoped: bool = True,
+        forwarded: bool = True,
```
hidmic, line 75: @djchopp nit: I'd suggest `forwarding` -- it's a forwarding group, not a forwarded group.

launch/launch/actions/clear_launch_configurations.py
hidmic: Oh, I did not really grok `ClearLaunchConfigurations`. As is stands, `ClearLaunchConfigurations(['arg1'])` means exactly the opposite. I'd avoid that. Following https://github.com/ros2/launch/pull/515#discussion_r658055646, we may not even need this clear-all-but-a-few action.

launch/launch/actions/group_action.py
djchopp: I agree the later is more intuitive, but unfortunately when forwarding is `False` the context would already be pushed and cleared by the time the resulting `SetLaunchConfiguration` from `launch_configurations={'Arg1': LaunchConfiguration('Arg1')}` is evaluated. The resulting list of actions would be:

    # scoped=True pushes the configurations at the start of the group action
    PushLaunchConfigurations(),

    # forwarding=False clears 'Arg1' from the launch configurations
    ClearLaunchConfigurations(), 

    # launch_configurations={'Arg1': LaunchConfiguration('Arg1')} will
    # add this SetLaunchConfiguration action to the list of actions,
    # but LaunchConfiguration('Arg1') cannot be resolved as it no longer exists
    SetLaunchConfiguration('Arg1', LaunchConfiguration('Arg1')), 

    # scoped=True pops the configurations at the end of the group action
    PopLaunchConfigurations()

Therefore ending in an error: `launch.substitutions.substitution_failure.SubstitutionFailure: launch configuration 'Arg1' does not exist`. 

One possibility around this is to evaluate the substitutions before clearing the context. Essentially evaluate the substitutions into strings that way they are not tied to a `LaunchConfiguration` object by the time the `SetLaunchConfiguration` is executed. Do you foresee any issue with this approach?

The issue I see is that I don't think this pattern would be possible outside of the group action since it requires evaluation of substitutions in a non-standard order compared to how the base actions (`Push`, `Clear`, `Set`, `Pop`) operate.

launch/launch/actions/group_action.py
djchopp: Good with me, I'll update it.

launch/launch/actions/clear_launch_configurations.py
djchopp: Indeed the more descriptive name for this would be `ClearLaunchConfigurationsExcept`.

See my concern about removing pass through functionality in my [other comment](https://github.com/ros2/launch/pull/515#discussion_r658226793). Specifically not being able to replicate the pass through functionality when using a clear-all instead of a clear-all-but-a-few outside of the group action with special substitution evaluation ordering.

My original use case for this feature was a complex launch system with several layers of depth (includes). Each launch description has many launch configurations. By the time you get to the deepest include, the context launch configurations can be incredibly cluttered. In that deepest include, I want to use mostly default values for the launch configurations. Currently I would have to add an `UnsetLaunchDescription` for each one. This is ok for a couple, but inconvenient past that. The clear-all action works to make sure all defaults are used, but still does not cover wanting to pass just a couple launch configurations through.

If the pass through of just a couple can be accomplished without a clear-all-but-a-few action, then it can definitely be dropped in favor of the clear-all action. I just have not figured out how to do that yet.

launch/launch/actions/group_action.py
hidmic: >  but unfortunately when forwarding is False the context would already be pushed and cleared by the time the resulting SetLaunchConfiguration from launch_configurations={'Arg1': LaunchConfiguration('Arg1')} is evaluated.

You're absolutely right. There's dragons in handling scope as a global dictionary.

> The issue I see is that I don't think this pattern would be possible outside of the group action since it requires evaluation of substitutions in a non-standard order compared to how the base actions (Push, Clear, Set, Pop) operate.

Hmm, that's also true. What about collapsing the `ClearLaunchConfigurations` and `SetLaunchConfiguration` actions into a single `ResetLaunchConfigurations` action that can force a given context? We could bridge "outer" and "inner" contexts in there, and `ClearLaunchConfigurations` would become `ResetLaunchConfigurations` with no arguments. Your draft of a partial-forwarding scoped group implementation would then be:

```
# scoped=True pushes the configurations at the start of the group action
PushLaunchConfigurations(),

# forwarding=False resets the context to match the given 
# launch_configurations={'Arg1': LaunchConfiguration('Arg1')} will.
# If no launch_configurations are given, the clear-all effect is achieved
ResetLaunchConfigurations(launch_configurations),

# scoped=True pops the configurations at the end of the group action
PopLaunchConfigurations()
```

I'm not amazed by the idea but the gap seems small enough. WDYT @djchopp?

djchopp: @hidmic I like it better than my other idea which was to allow access to launch configurations up the stack with a new `OutOfScopeLaunchConfiguration` substitution type. For the record it did work and would look like  `launch_configurations={'Arg1': OutOfScopeLaunchConfiguration('Arg1', depth=1)}`, but it just feels pretty far in the weeds / playing with fire. I will work on a `ResetLaunchConfigurations` action and see if I find any issues.

launch/launch/actions/group_action.py
djchopp: Weirdness hopefully resolved with the new `ResetLaunchConfigurations` action which does not use special flags (ie None) to indicate a pass through.

launch/launch/actions/group_action.py
```diff
@@ -65,24 +85,40 @@ def parse(cls, entity: Entity, parser: Parser):
...
+                        *self.__actions_to_return,
+                        PopLaunchConfigurations()
+                    ]
+            else:
                 self.__actions_to_return = [
-                    PushLaunchConfigurations(),
-                    *self.__actions_to_return,
-                    PopLaunchConfigurations()
+                    *configuration_sets,
+                    *self.__actions_to_return
```
wjwwood, line 121: nitpick: this logic might be reduced by keeping the:

```python
if self.__scoped:
    self.__actions_to_return = [
         PushLaunchConfigurations(),
         *self.__actions_to_return,
         PopLaunchConfigurations(),
    ]
```

And doing that unconditionally, rather than doing it in both the cases where forwarding is true or false. Dunno if that made sense... something like this:

```python
if self.__forwarding:
    self.__actions_to_return = [
        *configuration_sets,
        *self.__actions_to_return,
    ]
else:
    self.__actions_to_return = [
        ResetLaunchConfigurations(self.__launch_configurations),
        *self.__actions_to_return,
    ]

if self.__scoped:
    self.__actions_to_return = [
        PushLaunchConfigurations(),
        *self.__actions_to_return,
        PopLaunchConfigurations(),
    ]
```

launch/launch/actions/reset_launch_configurations.py
```diff
@@ -0,0 +1,67 @@
...
+class ResetLaunchConfigurations(Action):
+    """
+    Action that resets launch configurations in the current context.
+
+    This action can be used to clear the launch configurations from the
+    context it was called in. It optionally can be given a dictionary
+    with launch configurations to be set after clearing. Launch
+    configurations given in the dictionary are evaluated before the
+    context launch configurations are cleared. This allows launch
+    configurations to be passed through the clearing of the context.
```
wjwwood, line 36: nitpick: please use one sentence per line

> - [any] Each sentence must start on a new line.
>   - Rationale: For longer paragraphs a single change in the beginning makes the diff unreadable since it carries forward through the whole paragraph.

https://docs.ros.org/en/rolling/Contributing/Code-Style-Language-Versions.html#markdown-restructured-text-docblocks

Same as above.

launch/launch/actions/reset_launch_configurations.py
```diff
@@ -0,0 +1,67 @@
...
+
+    If launch_configurations=None or {} then all launch configurations are cleared.
+
+    If launch_configurations has entries (ie {'foo': 'FOO'}) then they will be
```
wjwwood, line 40:
```suggestion
    If launch_configurations has entries (i.e. {'foo': 'FOO'}) then they will be
```

launch/launch/actions/reset_launch_configurations.py
```diff
@@ -0,0 +1,67 @@
...
+        """Create an ResetLaunchConfigurations action."""
+        super().__init__(**kwargs)
+        self.__launch_configurations = launch_configurations
+
+    def execute(self, context: LaunchContext):
+        """Execute the action."""
+        if self.__launch_configurations is None:
+            context.launch_configurations.clear()
+        else:
+            evaled_configurations = {}
```
wjwwood, line 58: nitpick: I tend to avoid abbreviated variable names, partially because they're harder for people who don't speak English, e.g. `evaluated_configurations` translates better than `evaled_configurations`.

wjwwood (commented):
lgtm, several nitpicks but nothing serious

launch/launch/actions/group_action.py
djchopp: @wjwwood I am ok with that and like the idea of cleaning up the logic. It does expose the case of `forwarding=False` and `scoped=False` which was not possible with the previous implementation. The end result would be clearing launch configurations in the current scope which will clear launch configurations for anything that comes after the group action. Is that a use case that should be exposed to users? If so your suggestion works well and I can add the documentation for it. 

Thoughts?

launch/launch/actions/reset_launch_configurations.py
djchopp: @wjwwood Abbreviations expanded in [9d05cc6](https://github.com/ros2/launch/pull/515/commits/9d05cc6f75fb642cb6dd2629e1813e6bfa37c4cd)

launch/launch/actions/reset_launch_configurations.py
djchopp: Fixed in [68b9ffe](https://github.com/ros2/launch/pull/515/commits/68b9ffeb6dff6ce532b2b297da781e44396201a7)

launch/launch/actions/group_action.py
wjwwood: I think that scoped needs to be true for anything in forwarding to be done, so maybe the logic needs to be nested as it is, or the else case for forwarding needs to be `elif scoped`.

wjwwood: That is to say, I don't think anything should happen when both scoped and forwarding are false.

Let me know what you will do here, either change it or leave it as-is, so I know when it's ready for another review.

launch/launch/actions/reset_launch_configurations.py
```diff
@@ -0,0 +1,69 @@
...
+    This action can be used to clear the launch configurations from the
+    context it was called in.
+    It optionally can be given a dictionary with launch configurations
+    to be set after clearing.
+    Launch configurations given in the dictionary are evaluated before
+    the context launch configurations are cleared.
+    This allows launch configurations to be passed through the clearing
+    of the context.
+
+    If launch_configurations=None or {} then all launch configurations are cleared.
```
hidmic, line 40: @djchopp nit^2:
```suggestion
    If launch_configurations is None or an empty dict then all launch configurations will be cleared.
```

launch/launch/actions/reset_launch_configurations.py
```diff
@@ -0,0 +1,69 @@
...
+
+    If launch_configurations has entries (i.e. {'foo': 'FOO'}) then they will be
```
hidmic, line 42: @djchopp nit^2:
```suggestion
    If launch_configurations has entries (i.e. {'foo': 'FOO'}) then these will be
```

launch/launch/actions/reset_launch_configurations.py
```diff
@@ -0,0 +1,69 @@
...
+            evaluated_configurations = {}
+            for k, v in self.__launch_configurations.items():
+                evaluated_k = perform_substitutions(context, normalize_to_list_of_substitutions(k))
+                evaluated_v = perform_substitutions(context, normalize_to_list_of_substitutions(v))
+                evaluated_configurations[evaluated_k] = evaluated_v
+
+            context.launch_configurations.clear()
+
+            for k, v in evaluated_configurations.items():
+                context.launch_configurations[k] = v
```
hidmic, line 69: @djchopp nit: 
```suggestion
            context.launch_configurations.update(evaluated_configurations)
```

launch/test/launch/actions/test_group_action.py
```diff
@@ -101,3 +127,30 @@ def test_group_action_execute():
...
+    assert isinstance(result[0], PushLaunchConfigurations)
+    assert isinstance(result[1], ResetLaunchConfigurations)
+    assert isinstance(result[2], Action)
+    assert isinstance(result[3], PopLaunchConfigurations)
+    result[0].visit(lc1)  # Push
+    assert 'foo' in lc1.launch_configurations.keys()  # Copied to new scope, before Reset
+    assert 'bar' in lc1.launch_configurations.keys()  # Copied to new scope, before Reset
+    result[1].visit(lc1)  # Reset
+    assert 'foo' not in lc1.launch_configurations.keys()  # Cleared from scope in Reset
+    assert 'bar' in lc1.launch_configurations.keys()  # Evaluated and forwarded in Reset
```
hidmic, line 148: @djchopp nit: shall we check the launch configuration value as well?

launch/test/launch/actions/test_reset_launch_configurations.py
```diff
@@ -0,0 +1,78 @@
...
+
+    # Pass through an existing launch configuration
+    lc4 = LaunchContext()
+    assert len(lc4.launch_configurations) == 0
+    lc4.launch_configurations['foo'] = 'FOO'
+    lc4.launch_configurations['bar'] = 'BAR'
+    assert len(lc4.launch_configurations) == 2
+    ResetLaunchConfigurations(launch_configurations={'foo': LaunchConfiguration('foo')}).visit(lc4)
+    assert len(lc4.launch_configurations) == 1
+    assert 'bar' not in lc4.launch_configurations.keys()
```
hidmic, line 69: @djchopp would be good to check that `foo` is there and still equals `'FOO'`.

hidmic (commented):
Overall I like this patch. Left a few minor comments here and there, but otherwise this is pretty much ready to fly.

launch/launch/actions/group_action.py
djchopp: @wjwwood I think I am going to leave it as is. I have tried a couple different versions and they are not much cleaner than the existing. Besides documentation, it could be mitigated with either printing a warning or raising an error if scoped and forwarding are both False. Printing a warning creates a dependency on `launch.logging` but allows execution to continue, raising an error would stop execution. Thoughts?

    def execute(self, context: LaunchContext) -> Optional[List[LaunchDescriptionEntity]]:
        """Execute the action."""
        if self.__scoped is False and self.__forwarding is False:
            launch.logging.get_logger('GroupAction').warn(
                'forwarding = False will be ignored when scoped = False.')
            
            # or

            raise RuntimeError('GroupAction cannot have forwarding = False while scoped = False')

        return self.get_sub_entities()

launch/test/launch/actions/test_reset_launch_configurations.py
djchopp: Added in [c7b55cf](https://github.com/ros2/launch/pull/515/commits/c7b55cfc5ea664d9d353d049beabfb7c950d0c4a)

launch/test/launch/actions/test_group_action.py
djchopp: Added in [c7b55cf](https://github.com/ros2/launch/pull/515/commits/c7b55cfc5ea664d9d353d049beabfb7c950d0c4a)

launch/launch/actions/reset_launch_configurations.py
djchopp: Fixed in [1501e9d](https://github.com/ros2/launch/pull/515/commits/1501e9de65781db3afdaeff636833bdc62f85694)

hidmic (approved):
LGTM ! CI up to `launch`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15085/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9817/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12770/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15252/)

djchopp:
@hidmic Can you rerun the CI? Latest commit should fix the flake errors from the previous CI builds.

hidmic:
Sure thing. Take two:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15094/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9826/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12781/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15261/)

hidmic:
Alright, all green. We should probably run a more comprehensive CI now. @ivanpauno @wjwwood mind to take a look before I do that?

launch/launch/actions/reset_launch_configurations.py
```diff
@@ -0,0 +1,68 @@
...
+from typing import Optional
+
+from ..action import Action
+from ..launch_context import LaunchContext
+from ..some_substitutions_type import SomeSubstitutionsType
+from ..utilities import normalize_to_list_of_substitutions
+from ..utilities import perform_substitutions
+
+
+class ResetLaunchConfigurations(Action):
```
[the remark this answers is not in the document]
ivanpauno: > Hmm, one possible, not terribly surprising nor great approach would be:

That sounds okay to me, I don't have a better idea.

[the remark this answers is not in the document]
ivanpauno: Though, I'm not sure if there's any benefit in using `keep` instead of `arg`.
If there's not I would rather just us `arg`.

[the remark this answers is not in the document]
ivanpauno: @djchopp I think you can use the [children()](https://github.com/ros2/launch/blob/dc050b2c425c894773bf39a704a4386314608d3d/launch/launch/frontend/entity.py#L40) method and filter the ones named `arg`/`keep` to be used as launch configurations.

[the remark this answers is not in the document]
djchopp: Sounds good, I will give that a try. `include_launch_description` uses the `arg` tag to pass launch configurations so I will stick with that for now.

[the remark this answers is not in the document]
djchopp: I have committed a first attempt at this. Both `reset_launch_configurations` and `group_action` use the same chunk of code for parsing:

```python
args = entity.get_attr('arg', data_type=List[Entity], optional=True)
if args is not None:
    kwargs['launch_configurations'] = dict()
    for e in args:
        arg_name = tuple(parser.parse_substitution(e.get_attr('name')))
        arg_value = parser.parse_substitution(e.get_attr('value'))
        kwargs['launch_configurations'][arg_name] = arg_value
```

which allows using `arg` tags to specify launch configurations. 

examples:
```xml
<reset>
    <arg name="foo" value="FOO2"/>
    <arg name="bar" value="BAR2"/>
</reset>
```

```xml
<group scoped="true" forwarding="false">
    <arg name="$(var foo)$(var bar)" value="foo"/>
    <arg name="bar" value="$(var bar)"/>
    <arg name="baz" value="BAZ"/>
    <log message="In Group (Scoped True, Forwarding False)"/>
</group>
```

Some notes:
- The `arg` tag matches the usage in `include_launch_description`, however this does not match how the `arg` tag is used everywhere else. When using as a launch argument there should not be a `value` field.
- arg_name is converted from the list produced by `parser.parse_substitution` to a tuple so that it is a hashable type (required for dict key). This allows using a combination of configuration values to build the names of other configurations for example `<arg name="$(var foo)$(var bar)" value="FOOBAR"/>`. I have no clue what the use case for that may be, but the option is available.

Thoughts?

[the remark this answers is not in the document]
hidmic: It looks reasonable to me. Adding tests would be ideal.

---

On a related note, 

> The arg tag matches the usage in include_launch_description, however this does not match how the arg tag is used everywhere else. When using as a launch argument there should not be a value field.

Yeah, the use of `arg` in this context is a bit unusual and its effect is not immediately obvious when reading the XML. `keep` is not great but it is less confusing than `arg` IMHO.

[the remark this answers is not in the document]
djchopp: @hidmic Is `launch_xml/test/launch_xml` the right place for these tests? Should yaml tests be created too?

> Yeah, the use of arg in this context is a bit unusual and its effect is not immediately obvious when reading the XML. keep is not great but it is less confusing than arg IMHO.

Agreed. I will change to `keep`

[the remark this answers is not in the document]
hidmic: >  Is `launch_xml/test/launch_xml` the right place for these tests? Should yaml tests be created too?

Yes and yes.

launch_yaml/test/launch_yaml/test_reset.py
```diff
@@ -0,0 +1,71 @@
...
+def test_reset():
+    yaml_file = \
+        """\
+        launch:
+            - let:
+                name: 'foo'
+                value: 'FOO'
+            - let:
+                name: 'bar'
+                value: &bar 'BAR'
```
hidmic, line 34: @djchopp although PyYAML does support node anchors, I'd suggest we use launch substitution syntax i.e. `$(var ...)` to actually check things are working as expected. Same elsewhere.

launch/launch/actions/group_action.py
```diff
@@ -88,19 +88,20 @@ def parse(cls, entity: Entity, parser: Parser):
...
             kwargs['forwarding'] = forwarding
-        if args is not None:
-            kwargs['launch_configurations'] = dict()
-            for e in args:
-                arg_name = tuple(parser.parse_substitution(e.get_attr('name')))
-                arg_value = parser.parse_substitution(e.get_attr('value'))
-                kwargs['launch_configurations'][arg_name] = arg_value
+        if keeps is not None:
+            kwargs['launch_configurations'] = {
+                    tuple(parser.parse_substitution(e.get_attr('name'))):
```
hidmic, line 98: @djchopp nit^(N!): I'd use `as` or instead of `name`. @wjwwood @ivanpauno sanity check this.

hidmic (commented):
Overall this looks great to me @djchopp.

Thanks for the work and the patience.

launch_yaml/test/launch_yaml/test_reset.py
djchopp: @hidimic Good call, fixed in [c31a37c](https://github.com/djchopp/launch/commit/c31a37cb95a453a0fb4f96967d4f67d5dbd76e90)

hidmic (approved):
LGTM pending green CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15325/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10036/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13005/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15518/)

Thank you for patience @djchopp!

launch/launch/actions/group_action.py
djchopp: @hidmic Should I put this in as well? I don't have strong feelings to which one should be used. I was waiting for input from @wjwwood and @ivanpauno but have not seen anything.

hidmic: No need to. If nobody else feels strongly about it, perhaps it's not worth the change.

hidmic:
@djchopp would you mind rebasing on top of `master`? That seems to be the cause behind CI failures.

djchopp:
@hidmic This is nearing the edge of my git knowledge so hopefully I did that correctly. Let me know if I did something stupid.

hidmic:
@djchopp did you merge `master` into your branch along the way at some point?  I see commits that do not belong here.

djchopp:
@hidimic Very possibly. Steps I did:

- Fetch upstream in github
- Pull origin master branch
- Checkout local feat branch
- rebase on master
- Pull (merge) origin feat branch
- Push local feat branch
- Forget to signoff, rebase again with signoff
- Work through merge conflicts
- Push local feat branch, force with lease

If you know how to fix, feel free to do so

hidmic:
@djchopp 

> Pull (merge) origin feat branch
> Push local feat branch
> Forget to signoff, rebase again with signoff
> Work through merge conflicts

those extra steps seem like the culprit. If you kept your original branch, just rebase and push -f. If you nuked it, I'd *think* you might be able to re-establish it by dropping all commits before ff4c4f92e63f15f27a3d96cbc967e05c1fa4a928. 

As someone who has had trouble with this in the past, I'd strongly advice you to always branch to backup before rewriting history.

djchopp:
@hidmic Thanks for the lesson! I had a copy saved off so was able to recover. Let me know if that looks better now.

hidmic:
Thanks @djchopp !

Firing up CI again:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15338/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10049/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13018/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15532/)

launch/launch/actions/group_action.py
wjwwood: I don't mind either.

wjwwood (approved):

wjwwood:
Does this close https://github.com/ros2/launch/issues/507?

hidmic:
> Does this close #507?

Indeed :)

PR #516, opened 2021-06-25
Make each parser extension provide a set of file extensions

christophebedard:
This is an implementation of an idea from this comment: https://github.com/ros2/launch_ros/issues/149#issuecomment-713670587. I think asking the parser implementations for extensions instead of basing it on the name of the extension definitely makes sense. For example, we could now easily add `launch.yml` (vs `launch.yaml`) as a recognized extension.

It can go alonside https://github.com/ros2/launch_ros/pull/251.

### DIFF
--- a/launch/launch/frontend/parser.py
+++ b/launch/launch/frontend/parser.py
@@ -18,12 +18,14 @@
 import itertools
 import os.path
 from typing import List
+from typing import Optional
 from typing import Set
 from typing import Text
 from typing import TextIO
 from typing import Type
 from typing import TYPE_CHECKING
 from typing import Union
+import warnings
 
 try:
     import importlib.metadata as importlib_metadata
@@ -119,6 +121,32 @@ class Parser:
         """Return the registered extensions."""
         cls.load_parser_implementations()
         return cls.frontend_parsers.keys()
+
+    @classmethod
+    def is_extension_valid(
+        cls,
+        extension: Text,
+    ) -> bool:
+        """Return an entity loaded with a markup file."""
+        warnings.warn(
+            'Parser.is_extension_valid is deprecated, use Parser.is_filename_valid instead')
+        cls.load_parser_implementations()
+        return extension in cls.frontend_parsers
+
+    @classmethod
+    def get_parser_from_extension(
+        cls,
+        extension: Text,
+    ) -> Optional[Type['Parser']]:
+        """Return an entity loaded with a markup file."""
+        warnings.warn(
+            'Parser.get_parser_from_extension is deprecated, '
+            'use Parser.get_parsers_from_filename instead')
+        cls.load_parser_implementations()
+        try:
+            return cls.frontend_parsers[extension]
+        except KeyError:
+            raise RuntimeError('Not recognized frontend implementation')
 
     @classmethod
     def may_parse(
--- a/launch/launch/frontend/parser.py
+++ b/launch/launch/frontend/parser.py
@@ -198,5 +198,5 @@ class Parser:
 
     @classmethod
     def get_file_extensions(cls) -> Set[Text]:
-        """Return a set of file extensions for this parser."""
+        """Return the set of file extensions known to this parser."""
         return {}
--- a/launch/launch/frontend/parser.py
+++ b/launch/launch/frontend/parser.py
@@ -81,11 +81,12 @@ class Parser:
     def load_parser_implementations(cls):
         """Load all the available frontend entities."""
         if cls.frontend_parsers is None:
-            cls.frontend_parsers = {
+            parsers = {
                 entry_point.name: entry_point.load()
                 for entry_point in importlib_metadata.entry_points().get(
                         'launch.frontend.parser', [])
             }
+            cls.frontend_parsers = dict(sorted(parsers.items()))
 
     def parse_action(self, entity: Entity) -> Action:
         """Parse an action, using its registered parsing method."""
@@ -176,7 +177,7 @@ class Parser:
         """Return a list of parsers which entity loaded with a markup file."""
         cls.load_parser_implementations()
         return [
-            parser for _, parser in sorted(cls.frontend_parsers.items())
+            parser for _, parser in cls.frontend_parsers.items()
             if parser.may_parse(filename)
         ]
 
@@ -216,7 +217,7 @@ class Parser:
             filename = getattr(fileobj, 'name', '')
             implementations = cls.get_parsers_from_filename(filename)
             implementations += [
-                parser for _, parser in sorted(cls.frontend_parsers.items())
+                parser for _, parser in cls.frontend_parsers.items()
                 if parser not in implementations
             ]
 
--- a/launch/launch/frontend/parser.py
+++ b/launch/launch/frontend/parser.py
@@ -15,9 +15,11 @@
 
 """Module for Parser class and parsing methods."""
 
+import itertools
 import os.path
 from typing import List
 from typing import Optional
+from typing import Set
 from typing import Text
 from typing import TextIO
 from typing import Type
@@ -56,8 +58,9 @@ class Parser:
     Abstract class for parsing launch actions, substitutions and descriptions.
 
     Implementations of the parser class, should override the load method.
-    They could also override the parse_substitution method, or not.
-    load_launch_extensions, parse_action and parse_description are not suposed to be overrided.
+    They could also override the parse_substitution and/or get_file_extensions methods, or not.
+    load_launch_extensions, parse_action, parse_description and get_file_extensions_from_parsers
+    are not suposed to be overrided.
     """
 
     extensions_loaded = False
@@ -139,6 +142,15 @@ class Parser:
             raise RuntimeError('Not recognized frontend implementation')
 
     @classmethod
+    def get_file_extensions_from_parsers(cls) -> Set[Type['Parser']]:
+        """Return a set of file extensions gathered from the parser implementations."""
+        cls.load_parser_implementations()
+        return set(itertools.chain.from_iterable([
+            parser_extension.get_file_extensions()
+            for parser_extension in cls.frontend_parsers.values()
+        ]))
+
+    @classmethod
     def load(
         cls,
         file: Union[FilePath, TextIO],
@@ -183,3 +195,8 @@ class Parser:
         finally:
             if didopen:
                 fileobj.close()
+
+    @classmethod
+    def get_file_extensions(cls) -> Set[Text]:
+        """Return a set of file extensions for this parser."""
+        return {}
--- a/launch/launch/frontend/parser.py
+++ b/launch/launch/frontend/parser.py
@@ -18,7 +18,6 @@
 import itertools
 import os.path
 from typing import List
-from typing import Optional
 from typing import Set
 from typing import Text
 from typing import TextIO
@@ -121,29 +120,42 @@ class Parser:
         return cls.frontend_parsers.keys()
 
     @classmethod
-    def is_extension_valid(
+    def is_filename_valid_for_parser(
         cls,
-        extension: Text,
+        filename: Text,
+        parser: Type['Parser'],
     ) -> bool:
-        """Return an entity loaded with a markup file."""
-        cls.load_parser_implementations()
-        return extension in cls.frontend_parsers
+        """Return `true` if the filename is valid for the given parser."""
+        return any(
+            filename.endswith('.' + parser_ext) for parser_ext in parser.get_file_extensions())
 
     @classmethod
-    def get_parser_from_extension(
+    def is_filename_valid(
         cls,
-        extension: Text,
-    ) -> Optional[Type['Parser']]:
-        """Return an entity loaded with a markup file."""
+        filename: Text,
+    ) -> bool:
+        """Return `true` if the filename is valid for any parser."""
         cls.load_parser_implementations()
-        try:
-            return cls.frontend_parsers[extension]
-        except KeyError:
-            raise RuntimeError('Not recognized frontend implementation')
+        return any(
+            cls.is_filename_valid_for_parser(filename, parser)
+            for parser in cls.frontend_parsers.values()
+        )
+
+    @classmethod
+    def get_parsers_from_filename(
+        cls,
+        filename: Text,
+    ) -> List[Type['Parser']]:
+        """Return a list of parsers which entity loaded with a markup file."""
+        cls.load_parser_implementations()
+        return [
+            parser for _, parser in sorted(cls.frontend_parsers.items())
+            if cls.is_filename_valid_for_parser(filename, parser)
+        ]
 
     @classmethod
     def get_file_extensions_from_parsers(cls) -> Set[Type['Parser']]:
-        """Return a set of file extensions gathered from the parser implementations."""
+        """Return a set of file extensions known to the parser implementations."""
         cls.load_parser_implementations()
         return set(itertools.chain.from_iterable(
             parser_extension.get_file_extensions()
@@ -175,12 +187,9 @@ class Parser:
 
         try:
             filename = getattr(fileobj, 'name', '')
-            # file extension without leading '.'
-            extension = os.path.splitext(filename)[1][1:]
-
-            sorted_parsers = sorted(cls.frontend_parsers.items())
-            implementations = [v for k, v in sorted_parsers if k == extension] + [
-                v for k, v in sorted_parsers if k != extension
+            implementations = cls.get_parsers_from_filename(filename)
+            implementations += [
+                parser for parser in cls.frontend_parsers.values() if parser not in implementations
             ]
 
             exceptions = []
@@ -190,7 +199,9 @@ class Parser:
                 except Exception as ex:
                     exceptions.append(ex)
                     fileobj.seek(0)
-            extension = '' if not cls.is_extension_valid(extension) else extension
+            # file extension without leading '.'
+            extension = os.path.splitext(filename)[1][1:]
+            extension = '' if not cls.is_filename_valid(filename) else extension
             raise InvalidFrontendLaunchFileError(extension, likely_errors=exceptions)
         finally:
             if didopen:
--- a/launch/launch/frontend/parser.py
+++ b/launch/launch/frontend/parser.py
@@ -58,8 +58,9 @@ class Parser:
 
     Implementations of the parser class, should override the load method.
     They could also override the parse_substitution and/or get_file_extensions methods, or not.
-    load_launch_extensions, parse_action, parse_description and get_file_extensions_from_parsers
-    are not suposed to be overrided.
+    load_launch_extensions, parse_action, parse_description, get_available_extensions, may_parse,
+    is_filename_valid, get_parsers_from_filename and get_file_extensions_from_parsers are not
+    suposed to be overrided.
     """
 
     extensions_loaded = False
@@ -120,14 +121,12 @@ class Parser:
         return cls.frontend_parsers.keys()
 
     @classmethod
-    def is_filename_valid_for_parser(
+    def may_parse(
         cls,
         filename: Text,
-        parser: Type['Parser'],
     ) -> bool:
-        """Return `true` if the filename is valid for the given parser."""
-        return any(
-            filename.endswith('.' + parser_ext) for parser_ext in parser.get_file_extensions())
+        """Return `true` if the filename is valid for this parser."""
+        return any(filename.endswith('.' + ext) for ext in cls.get_file_extensions())
 
     @classmethod
     def is_filename_valid(
@@ -137,7 +136,7 @@ class Parser:
         """Return `true` if the filename is valid for any parser."""
         cls.load_parser_implementations()
         return any(
-            cls.is_filename_valid_for_parser(filename, parser)
+            parser.may_parse(filename)
             for parser in cls.frontend_parsers.values()
         )
 
@@ -150,7 +149,7 @@ class Parser:
         cls.load_parser_implementations()
         return [
             parser for _, parser in sorted(cls.frontend_parsers.items())
-            if cls.is_filename_valid_for_parser(filename, parser)
+            if parser.may_parse(filename)
         ]
 
     @classmethod
--- a/launch/launch/frontend/parser.py
+++ b/launch/launch/frontend/parser.py
@@ -145,10 +145,10 @@ class Parser:
     def get_file_extensions_from_parsers(cls) -> Set[Type['Parser']]:
         """Return a set of file extensions gathered from the parser implementations."""
         cls.load_parser_implementations()
-        return set(itertools.chain.from_iterable([
+        return set(itertools.chain.from_iterable(
             parser_extension.get_file_extensions()
             for parser_extension in cls.frontend_parsers.values()
-        ]))
+        ))
 
     @classmethod
     def load(
--- a/launch/launch/frontend/parser.py
+++ b/launch/launch/frontend/parser.py
@@ -189,7 +189,8 @@ class Parser:
             filename = getattr(fileobj, 'name', '')
             implementations = cls.get_parsers_from_filename(filename)
             implementations += [
-                parser for parser in cls.frontend_parsers.values() if parser not in implementations
+                parser for _, parser in sorted(cls.frontend_parsers.items())
+                if parser not in implementations
             ]
 
             exceptions = []
--- a/launch/launch/frontend/parser.py
+++ b/launch/launch/frontend/parser.py
@@ -62,7 +62,7 @@ class Parser:
     They could also override the parse_substitution and/or get_file_extensions methods, or not.
     load_launch_extensions, parse_action, parse_description, get_available_extensions, may_parse,
     is_filename_valid, get_parsers_from_filename and get_file_extensions_from_parsers are not
-    suposed to be overrided.
+    supposed to be overriden.
     """
 
     extensions_loaded = False
@@ -154,7 +154,7 @@ class Parser:
         cls,
         filename: Text,
     ) -> bool:
-        """Return `true` if the filename is valid for this parser."""
+        """Return `True` if the filename is valid for this parser."""
         return any(filename.endswith('.' + ext) for ext in cls.get_file_extensions())
 
     @classmethod
@@ -162,7 +162,7 @@ class Parser:
         cls,
         filename: Text,
     ) -> bool:
-        """Return `true` if the filename is valid for any parser."""
+        """Return `True` if the filename is valid for any parser."""
         cls.load_parser_implementations()
         return any(
             parser.may_parse(filename)
@@ -177,7 +177,7 @@ class Parser:
         """Return a list of parsers which entity loaded with a markup file."""
         cls.load_parser_implementations()
         return [
-            parser for _, parser in cls.frontend_parsers.items()
+            parser for parser in cls.frontend_parsers.values()
             if parser.may_parse(filename)
         ]
 
@@ -217,7 +217,7 @@ class Parser:
             filename = getattr(fileobj, 'name', '')
             implementations = cls.get_parsers_from_filename(filename)
             implementations += [
-                parser for _, parser in cls.frontend_parsers.items()
+                parser for parser in cls.frontend_parsers.values()
                 if parser not in implementations
             ]
 
--- a/launch/launch/launch_description_sources/any_launch_file_utilities.py
+++ b/launch/launch/launch_description_sources/any_launch_file_utilities.py
@@ -39,14 +39,15 @@ def get_launch_description_from_any_launch_file(
     :raise `ValueError`: Invalid file. The file may not be a text file.
     """
     loaders = [get_launch_description_from_frontend_launch_file]
-    extension = os.path.splitext(launch_file_path)[1]
+    launch_file_name = os.path.basename(launch_file_path)
+    extension = os.path.splitext(launch_file_name)[1]
     if extension:
         extension = extension[1:]
     if extension == 'py':
         loaders.insert(0, get_launch_description_from_python_launch_file)
     else:
         loaders.append(get_launch_description_from_python_launch_file)
-        extension = '' if not Parser.is_extension_valid(extension) else extension
+        extension = '' if not Parser.is_filename_valid(launch_file_name) else extension
     exceptions = []
     for loader in loaders:
         try:
--- a/launch_xml/launch_xml/parser.py
+++ b/launch_xml/launch_xml/parser.py
@@ -15,6 +15,8 @@
 """Module for XML Parser class."""
 
 import io
+from typing import Set
+from typing import Text
 from typing import Union
 import xml.etree.ElementTree as ET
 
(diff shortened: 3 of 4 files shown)


launch/launch/frontend/parser.py
```diff
@@ -138,6 +141,15 @@ def get_parser_from_extension(
...
             raise RuntimeError('Not recognized frontend implementation')
 
+    @classmethod
+    def get_file_extensions_from_parsers(cls) -> Set[Type['Parser']]:
+        """Return a set of file extensions gathered from the parser implementations."""
+        cls.load_parser_implementations()
+        return set(itertools.chain.from_iterable([
+            parser_extension.get_file_extensions()
+            for parser_extension in cls.frontend_parsers.values()
+        ]))
```
hidmic, line 151: @christophebedard nit: no need for the temporary list
```suggestion
        return set(itertools.chain.from_iterable(
            parser_extension.get_file_extensions()
            for parser_extension in cls.frontend_parsers.values()
        ))
```

launch/launch/frontend/parser.py
```diff
@@ -183,3 +195,8 @@ def load(
         finally:
             if didopen:
                 fileobj.close()
+
+    @classmethod
+    def get_file_extensions(cls) -> Set[Text]:
+        """Return a set of file extensions for this parser."""
```
hidmic, line 201: @christophebedard nit: 
```suggestion
        """Return the set of file extensions known to this parser."""
```

launch_yaml/launch_yaml/parser.py
```diff
@@ -51,3 +53,8 @@ def load(
         finally:
             if didopen:
                 fileobj.close()
+
+    @classmethod
+    def get_file_extensions(cls) -> Set[Text]:
+        """Return a set of file extensions for this parser."""
```
hidmic, line 59: @christophebedard nit: 
```suggestion
        """Return the set of file extensions known to this parser."""
```

hidmic (commented):
Overall LGTM pending green CI

launch_yaml/launch_yaml/parser.py
```diff
@@ -51,3 +53,8 @@ def load(
         finally:
             if didopen:
                 fileobj.close()
+
+    @classmethod
+    def get_file_extensions(cls) -> Set[Text]:
```
hidmic, line 58: @christophebedard we should use these extensions on `Parser.load()` as well.

hidmic:
Also, `Parser.get_available_extensions()` and `Parser.get_extensions_from_parser()` need a change.

launch/launch/frontend/parser.py
christophebedard: fixed in 7fae1d4e3ac33ab32b9626ddb53c85cb1c0f5191

launch/launch/frontend/parser.py
christophebedard: changed in bb5538ffc4ed9cc5bfbf9d78a999d3a29c79a688

launch_yaml/launch_yaml/parser.py
christophebedard: changed in bb5538ffc4ed9cc5bfbf9d78a999d3a29c79a688 (I also applied this change for the launch_xml parser)

christophebedard:
> Also, `Parser.get_available_extensions()` and `Parser.get_extensions_from_parser()` need a change.

@hidmic I'm not sure I understand what needs to be changed there

christophebedard:
> @hidmic I'm not sure I understand what needs to be changed there

Ah I think I get it. "Extension" is used to mean both "extension point" and "file extension" in a lot of places here.

You want me to change `Parser.load_parser_implementations()` so that its `frontend_parsers` dict is `{file_extension: parser type}`, is that correct?

In the case of `get_available_extensions()`, it refers to the name of the extension points, and it becomes a bit useless with this change. It was only used when building the list of valid file extensions, so it won't be needed anymore; it's kind of replaced by `get_file_extensions_from_parsers()`. Assuming it's an internal/non-public API, should it be removed?

`Parser.is_extension_valid()` (or, rather, its callers) would need to be modified as well, since we would be comparing "semantic" file extensions (i.e. extensions that may contain a period, like `launch.xml` instead of `xml`). That means we would probably need the whole file name and check that it endswith one of the file extensions from the parsers.

christophebedard:
This is turning into a bigger change than I originally thought :sweat_smile: but here's my proposal to make the whole `Parser` more uniform/consistent again: e526f9cbdf04a7ce20dea2e37e4234d68b5e0253

I switched to comparing whole file names instead of just extensions because "extension" meant literally "the text after the last period," which doesn't really work with `launch.xml`, `launch.yaml`, etc.

@hidmic let me know what you think :smiley:

launch_yaml/launch_yaml/parser.py
christophebedard: done as part of e526f9cbdf04a7ce20dea2e37e4234d68b5e0253

hidmic (commented):
> You want me to change Parser.load_parser_implementations() so that its frontend_parsers dict is {file_extension: parser type}, is that correct?

Indeed.

> In the case of get_available_extensions(), it refers to the name of the extension points, and it becomes a bit useless with this change. It was only used when building the list of valid file extensions, so it won't be needed anymore; it's kind of replaced by get_file_extensions_from_parsers(). Assuming it's an internal/non-public API, should it be removed?

As I said below, it'd be best to deprecate instead of removing if at all possible.

> Parser.is_extension_valid() (or, rather, its callers) would need to be modified as well, since we would be comparing "semantic" file extensions (i.e. extensions that may contain a period, like launch.xml instead of xml). That means we would probably need the whole file name and check that it endswith one of the file extensions from the parsers.

Checking the entire filename makes sense to me. It opens the door for other, more thorough validity checks in the future.

> Also, if we want to support more generic extensions (like .launch), we need to allow multiple extensions to declare the exact same file extension as being known to them.

I'm not entirely convinced. Checking for filename extensions is just a heuristic to speed up the parser lookup process. If all parsers specify `.launch` as an acceptable file extension, it is as good as if none did, and I'd just spare the duplication.

launch/launch/frontend/parser.py
```diff
@@ -118,25 +120,47 @@ def get_available_extensions(cls) -> List[Text]:
         return cls.frontend_parsers.keys()
 
     @classmethod
-    def is_extension_valid(
```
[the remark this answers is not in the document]
christophebedard: I ended up assuming that this wasn't part of the public API. I will try to add it back with a warning message.

christophebedard:
> I'm not entirely convinced. Checking for filename extensions is just a heuristic to speed up the parser lookup process. If all parsers specify `.launch` as an acceptable file extension, it is as good as if none did, and I'd just spare the duplication.

That's true, but I'm also thinking about having to put `.launch` *somewhere* for the autocompletion feature (https://github.com/ros2/launch_ros/issues/149). I thought that putting it here would make it more intentional compared to adding it to the set after the fact (in `launch_ros`, e.g. https://github.com/ros2/launch_ros/pull/251). We do it for `launch.py`, but it's not related to these frontend parsers.

Also, I don't want to do any kind of premature abstraction/optimization, but perhaps in the future we could have N-1 out of N frontend parsers that support the same `.ext` extension? In that case, we wouldn't really have trivial duplication.

Anyway, I can remove this & simplify the logic a bit, but, like I said, I'm also thinking about the autocompletion part.

launch/launch/frontend/parser.py
```diff
@@ -163,12 +187,9 @@ def load(
...
             filename = getattr(fileobj, 'name', '')
-            # file extension without leading '.'
-            extension = os.path.splitext(filename)[1][1:]
-
-            sorted_parsers = sorted(cls.frontend_parsers.items())
-            implementations = [v for k, v in sorted_parsers if k == extension] + [
-                v for k, v in sorted_parsers if k != extension
+            implementations = cls.get_parsers_from_filename(filename)
+            implementations += [
+                parser for parser in cls.frontend_parsers.values() if parser not in implementations
```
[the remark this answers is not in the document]
christophebedard: fixed in 3b05e1ed26e22e989508fdb227036105d517aa70

launch/launch/frontend/parser.py
```diff
@@ -118,25 +120,47 @@ def get_available_extensions(cls) -> List[Text]:
...
+    def is_filename_valid_for_parser(
         cls,
-        extension: Text,
+        filename: Text,
+        parser: Type['Parser'],
     ) -> bool:
-        """Return an entity loaded with a markup file."""
+        """Return `true` if the filename is valid for the given parser."""
+        return any(
+            filename.endswith('.' + parser_ext) for parser_ext in parser.get_file_extensions())
```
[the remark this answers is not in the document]
christophebedard: done in 3922b7781e91b2d417a1c8118af3ff3e5ca003a0. It has to be a `@classmethod` though.

launch/launch/frontend/parser.py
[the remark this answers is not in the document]
christophebedard: done in 4d55b414943246447837b376c2f06cfdf84d3034

launch/launch/frontend/parser.py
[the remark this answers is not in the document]
christophebedard: I changed the sorting so that it's done up front/only once: ca09eb54fc249527a7b98f132f16404f9872439a

launch/launch/frontend/parser.py
```diff
@@ -132,12 +140,56 @@ def get_parser_from_extension(
...
             return cls.frontend_parsers[extension]
         except KeyError:
             raise RuntimeError('Not recognized frontend implementation')
 
+    @classmethod
+    def may_parse(
+        cls,
+        filename: Text,
+    ) -> bool:
+        """Return `true` if the filename is valid for this parser."""
```
hidmic, line 157: @christophebedard nit:
```suggestion
        """Return `True` if the filename is valid for this parser."""
```

launch/launch/frontend/parser.py
```diff
@@ -132,12 +140,56 @@ def get_parser_from_extension(
...
+
+    @classmethod
+    def get_parsers_from_filename(
+        cls,
+        filename: Text,
+    ) -> List[Type['Parser']]:
+        """Return a list of parsers which entity loaded with a markup file."""
+        cls.load_parser_implementations()
+        return [
+            parser for _, parser in cls.frontend_parsers.items()
```
hidmic, line 180: @christophebedard nit: 
```suggestion
            parser for parser in cls.frontend_parsers.values()
```

launch/launch/frontend/parser.py
```diff
@@ -163,12 +215,10 @@ def load(
...
             filename = getattr(fileobj, 'name', '')
-            # file extension without leading '.'
-            extension = os.path.splitext(filename)[1][1:]
-
-            sorted_parsers = sorted(cls.frontend_parsers.items())
-            implementations = [v for k, v in sorted_parsers if k == extension] + [
-                v for k, v in sorted_parsers if k != extension
+            implementations = cls.get_parsers_from_filename(filename)
+            implementations += [
+                parser for _, parser in cls.frontend_parsers.items()
```
hidmic, line 220: @christophebedard nit: 
```suggestion
                parser for parser in cls.frontend_parsers.values()
```

hidmic (approved):
LGTM pending green CI. Also, it'd would be good if @ivanpauno or @wjwwood took a look as well.

launch/launch/frontend/parser.py
christophebedard: fixed in db9f1f2e9fce7b2268daa441bf13b0f9346206ac

launch/launch/frontend/parser.py
christophebedard: fixed in db9f1f2e9fce7b2268daa441bf13b0f9346206ac

launch/launch/frontend/parser.py
christophebedard: fixed in db9f1f2e9fce7b2268daa441bf13b0f9346206ac

christophebedard:
CI `--packages-above launch`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14839/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9589/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12506/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14996/)

launch/launch/frontend/parser.py
```diff
@@ -163,12 +215,10 @@ def load(
...
-            # file extension without leading '.'
-            extension = os.path.splitext(filename)[1][1:]
-
-            sorted_parsers = sorted(cls.frontend_parsers.items())
-            implementations = [v for k, v in sorted_parsers if k == extension] + [
-                v for k, v in sorted_parsers if k != extension
+            implementations = cls.get_parsers_from_filename(filename)
+            implementations += [
+                parser for parser in cls.frontend_parsers.values()
+                if parser not in implementations
```
ivanpauno, line 221: So `implementations` is a list with the same items as `cls.frontend_parsers.values()`, but reordered so the first ones are the ones that support that extension, right?

I think that before we were doing the same (trying with all the parsers), so it's fine, but maybe if the extension is exactly known (e.g. yaml), we could only try that one and fail.

launch/launch/launch_description_sources/any_launch_file_utilities.py
```diff
@@ -39,14 +39,15 @@ def get_launch_description_from_any_launch_file(
     :raise `ValueError`: Invalid file. The file may not be a text file.
     """
     loaders = [get_launch_description_from_frontend_launch_file]
-    extension = os.path.splitext(launch_file_path)[1]
+    launch_file_name = os.path.basename(launch_file_path)
```
ivanpauno, line 42: is this line needed?

launch_xml/launch_xml/parser.py
```diff
@@ -33,3 +35,8 @@ def load(
     ) -> (Entity, 'Parser'):
         """Return entity loaded from XML file."""
         return (Entity(ET.parse(file).getroot()), cls())
+
+    @classmethod
+    def get_file_extensions(cls) -> Set[Text]:
+        """Return the set of file extensions known to this parser."""
+        return {'launch.xml'}
```
ivanpauno, line 42: maybe we should add also `xml` here.
I know there are packages that are naming launch files like `_launch.xml`.

christophebedard: we could also test for `'_' + ext`, but I'll just add `xml`: 74934f64093f6e7d9d75cb38c31e5c6e61ea6c5c

launch/launch/frontend/parser.py
christophebedard: > So `implementations` is a list with the same items as `cls.frontend_parsers.values()`, but reordered so the first ones are the ones that support that extension, right?
> 
> I think that before we were doing the same (trying with all the parsers), so it's fine,

yeah

> but maybe if the extension is exactly known (e.g. yaml), we could only try that one and fail.

I can create a follow-up issue if you want.

launch/launch/launch_description_sources/any_launch_file_utilities.py
christophebedard: the filename is used below for the call to `Parser.is_filename_valid()`, even if in practice (because we're only checking the end of the filename) using the file path wouldn't change anything

launch/launch/frontend/parser.py
hidmic: > maybe if the extension is exactly known (e.g. yaml), we could only try that one and fail.

I lightly disagree. Extensions are non-binding IMHO.

launch_xml/launch_xml/parser.py
hidmic: We should probably do the same for `yaml`, no?

launch/launch/frontend/parser.py
ivanpauno: > I lightly disagree. Extensions are non-binding IMHO.

I don't understand what you mean.
There could be multiple extensions supporting a `.yaml` extension, and in that case we should try loading the file with them all.
What I mean is that it doesn't make sense to try to load a `.yaml` file with the XML frontend, I don't think anybody is going to name a xml file `my_launch_file.yaml`.


If we go ahead with this more strict approch, there are some extensions all plugins should support: no extension, .launch.
I agree this is unrelated to the PR though

launch_xml/launch_xml/parser.py
ivanpauno: > we could also test for '_' + ext, but I'll just add xml: 74934f6

:+1:
I think we should do the same for yaml.

christophebedard: > :+1:

Is this a :+1: for checking for `filename.endswith('_' + ext)` as well as `filename.endswith('.' + ext)`? :sweat_smile:

> I think we should do the same for yaml.

Should I also add `yml`?

ivanpauno: > Is this a +1 for checking for filename.endswith('_' + ext) as well as filename.endswith('.' + ext)? sweat_smile

Sorry, I meant the current commit was ok (i.e. adding xml and not only launch.xml to the accepted extensions).

> I'll add it. Should I also add yml?

yes, for all extensions we should support .launch.ext and .ext IMO.

christophebedard: done in aa3b33144dc42763b178876ed845f387b163f4e9

christophebedard: > yes, for all extensions we should support .launch.ext and .ext IMO.

Makes sense. Strictly implementation-wise, it is kind of redundant, but explicitly writing the extensions out like that makes sense.

ivanpauno (approved):

launch_xml/launch_xml/parser.py
christophebedard: > I know there are packages that are naming launch files like `_launch.xml`.

I even recently saw one that uses `-launch.py` (https://github.com/KumarRobotics/imu_vn_100/blob/dashing/launch/imu_vn_100-launch.py), but I think that's from before there was any kind of discussion about picking a standard extension.

christophebedard:
Another CI run with the new commits:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14843/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9593/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12510/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15001/)

launch/launch/frontend/parser.py
hidmic: > What I mean is that it doesn't make sense to try to load a `.yaml` file with the XML frontend, I don't think anybody is going to name a `xml` file `my_launch_file.yaml`.

I think that's a reasonable enough convention to warrant the heuristic we have, but I wouldn't put restrictions we don't need. If e.g. a downstream user needs to annotate their launch XMLs in any way and use a different extension for whatever reason, I don't see why the XML frontend should be blocked from parsing it. And while it does make sense to be strict for some very ubiquitous extensions like `.yaml` or `.xml` (so common that using them for anything else is misleading), I don't love special rules.

hidmic:
The test that failed is known flake in CI. Merging! Thanks for the patience @christophebedard.

christophebedard:
thanks for the reviews @hidmic and @ivanpauno!

wjwwood:
It seems that since this was merged I have a test that is failing for a pr I'm working on (https://github.com/osrf/ros2launch_security/pull/4), see:

```
13:54:21 test 1
13:54:21     Start 1: test_fake_imu_launch
13:54:21 
13:54:21 1: Test command: /Users/osrf/jenkins-agent/workspace/ci_osx/venv/bin/python3 "-u" "/Users/osrf/jenkins-agent/workspace/ci_osx/ws/install/ament_cmake_test/share/ament_cmake_test/cmake/run_test.py" "/Users/osrf/jenkins-agent/workspace/ci_osx/ws/build/ros2launch_security_examples/test_results/ros2launch_security_examples/test_fake_imu_launch.xunit.xml" "--package-name" "ros2launch_security_examples" "--output-file" "/Users/osrf/jenkins-agent/workspace/ci_osx/ws/build/ros2launch_security_examples/launch_test/test_fake_imu_launch.txt" "--env" "RCL_ASSERT_RMW_ID_MATCHES=rmw_cyclonedds_cpp" "RMW_IMPLEMENTATION=rmw_cyclonedds_cpp" "--command" "/Users/osrf/jenkins-agent/workspace/ci_osx/venv/bin/python3" "-m" "launch_testing.launch_test" "/Users/osrf/jenkins-agent/workspace/ci_osx/ws/src/osrf/ros2launch_security/ros2launch_security_examples/test/test_fake_imu_launch.py" "--junit-xml=/Users/osrf/jenkins-agent/workspace/ci_osx/ws/build/ros2launch_security_examples/test_results/ros2launch_security_examples/test_fake_imu_launch.xunit.xml" "--package-name=ros2launch_security_examples"
13:54:21 1: Test timeout computed to be: 60
13:54:22 1: -- run_test.py: extra environment variables:
13:54:22 1:  - RCL_ASSERT_RMW_ID_MATCHES=rmw_cyclonedds_cpp
13:54:22 1:  - RMW_IMPLEMENTATION=rmw_cyclonedds_cpp
13:54:22 1: -- run_test.py: invoking following command in '/Users/osrf/jenkins-agent/workspace/ci_osx/ws/build/ros2launch_security_examples':
13:54:22 1:  - /Users/osrf/jenkins-agent/workspace/ci_osx/venv/bin/python3 -m launch_testing.launch_test /Users/osrf/jenkins-agent/workspace/ci_osx/ws/src/osrf/ros2launch_security/ros2launch_security_examples/test/test_fake_imu_launch.py --junit-xml=/Users/osrf/jenkins-agent/workspace/ci_osx/ws/build/ros2launch_security_examples/test_results/ros2launch_security_examples/test_fake_imu_launch.xunit.xml --package-name=ros2launch_security_examples
13:54:22 1: [INFO] [launch]: All log files can be found below /Users/jenkins/.ros/log/2021-07-12-13-54-22-395317-lore-37109
13:54:22 1: [INFO] [launch]: Default logging verbosity is set to INFO
13:54:22 1: ERROR:asyncio:Task exception was never retrieved
13:54:22 1: future: <Task finished name='Task-2' coro=<LaunchService._process_one_event() done, defined at /Users/osrf/jenkins-agent/workspace/ci_osx/ws/install/launch/lib/python3.8/site-packages/launch/launch_service.py:226> exception=InvalidLaunchFileError('')>
13:54:22 1: Traceback (most recent call last):
13:54:22 1:   File "/Users/osrf/jenkins-agent/workspace/ci_osx/ws/install/launch/lib/python3.8/site-packages/launch/launch_service.py", line 228, in _process_one_event
13:54:22 1:     await self.__process_event(next_event)
13:54:22 1:   File "/Users/osrf/jenkins-agent/workspace/ci_osx/ws/install/launch/lib/python3.8/site-packages/launch/launch_service.py", line 248, in __process_event
13:54:22 1:     visit_all_entities_and_collect_futures(entity, self.__context))
13:54:22 1:   File "/Users/osrf/jenkins-agent/workspace/ci_osx/ws/install/launch/lib/python3.8/site-packages/launch/utilities/visit_all_entities_and_collect_futures_impl.py", line 45, in visit_all_entities_and_collect_futures
13:54:22 1:     futures_to_return += visit_all_entities_and_collect_futures(sub_entity, context)
13:54:22 1:   File "/Users/osrf/jenkins-agent/workspace/ci_osx/ws/install/launch/lib/python3.8/site-packages/launch/utilities/visit_all_entities_and_collect_futures_impl.py", line 45, in visit_all_entities_and_collect_futures
13:54:22 1:     futures_to_return += visit_all_entities_and_collect_futures(sub_entity, context)
13:54:22 1:   File "/Users/osrf/jenkins-agent/workspace/ci_osx/ws/install/launch/lib/python3.8/site-packages/launch/utilities/visit_all_entities_and_collect_futures_impl.py", line 45, in visit_all_entities_and_collect_futures
13:54:22 1:     futures_to_return += visit_all_entities_and_collect_futures(sub_entity, context)
13:54:22 1:   [Previous line repeated 1 more time]
13:54:22 1:   File "/Users/osrf/jenkins-agent/workspace/ci_osx/ws/install/launch/lib/python3.8/site-packages/launch/utilities/visit_all_entities_and_collect_futures_impl.py", line 38, in visit_all_entities_and_collect_futures
13:54:22 1:     sub_entities = entity.visit(context)
13:54:22 1:   File "/Users/osrf/jenkins-agent/workspace/ci_osx/ws/install/launch/lib/python3.8/site-packages/launch/action.py", line 108, in visit
13:54:22 1:     return self.execute(context)
13:54:22 1:   File "/Users/osrf/jenkins-agent/workspace/ci_osx/ws/install/launch/lib/python3.8/site-packages/launch/actions/include_launch_description.py", line 127, in execute
13:54:22 1:     launch_description = self.__launch_description_source.get_launch_description(context)
13:54:22 1:   File "/Users/osrf/jenkins-agent/workspace/ci_osx/ws/install/launch/lib/python3.8/site-packages/launch/launch_description_source.py", line 84, in get_launch_description
13:54:22 1:     self._get_launch_description(self.__expanded_location)
13:54:22 1:   File "/Users/osrf/jenkins-agent/workspace/ci_osx/ws/install/launch/lib/python3.8/site-packages/launch/launch_description_sources/any_launch_description_source.py", line 53, in _get_launch_description
13:54:22 1:     return get_launch_description_from_any_launch_file(location)
13:54:22 1:   File "/Users/osrf/jenkins-agent/workspace/ci_osx/ws/install/launch/lib/python3.8/site-packages/launch/launch_description_sources/any_launch_file_utilities.py", line 57, in get_launch_description_from_any_launch_file
13:54:22 1:     raise InvalidLaunchFileError(extension, likely_errors=exceptions)
13:54:22 1: launch.invalid_launch_file_error.InvalidLaunchFileError: The launch file may have a syntax error, or its format is unknown
13:54:22 1: Processes under test stopped before tests completed
13:54:22 1: -- run_test.py: return code 1
```

-- from https://ci.ros2.org/job/ci_osx/12541/consoleFull

One test includes a launch file as part of a `launch_testing` test, and one invokes `ros2 launch` directly as part of the test, but either way the launch file is called `fake_imu.launch.xml`:

- https://github.com/osrf/ros2launch_security/pull/4/files#diff-f4389ac3ca2a7ca0532c9a082695f17f1614abec42416bbd90c7aaf304b5356d

The test was passing last week and locally for me, which lead me to this change.

I boiled it down to the filename is considered invalid in this line:

https://github.com/ros2/launch/blob/2a97e806ead3b0b6428bd7c1f648c3a01bb9e231/launch/launch/launch_description_sources/any_launch_file_utilities.py#L50

I think this is a bug, but I'm not sure. Does my launch file name no longer work? Do we have any tests to cover this case (i.e. `foo.launch.xml` works)?

christophebedard:
> I think this is a bug, but I'm not sure. Does my launch file name no longer work?

Yeah that should work. I'll look into it.

> Do we have any tests to cover this case (i.e. foo.launch.xml works)?

No unit tests as far as I know (I didn't have to modify any tests for this PR), but I assume there are integration/e2e tests somewhere.

christophebedard:
Huh for some reason `launch_xml` (or `launch_yaml`) aren't being built with `--packages-up-to ros2launch_security_examples`, so the parser extensions aren't found.

christophebedard:
If I make sure that `launch_xml` is built (`--packages-up-to ros2launch_security_examples launch_xml`), then the `xml` parser is found and your test runs fine.

At this point we should open a new issue, but shouldn't something ensure that the parsers are always built/available if we're using `launch`? I don't see how this PR could have affected dependencies.

wjwwood:
Interesting, I would have expected that xml and yaml support were always available when using launch...

christophebedard:
> Interesting, I would have expected that xml and yaml support were always available when using launch...

Yeah same here. Maybe this could be improved, somehow. A quick fix would be to have `launch_ros` and/or `launch_testing_ros` depend on `launch_xml`/`launch_yaml`, but maybe there's a reason why they don't?

hidmic:
> but shouldn't something ensure that the parsers are always built/available if we're using `launch`?

`launch_xml` and `launch_yaml` are themselves dependent on `launch`, we can't bring them in as `launch` dependencies.

> A quick fix would be to have `launch_ros` and/or `launch_testing_ros` depend on `launch_xml`/`launch_yaml`

We could but it seems a bit inadequate. `launch` frontend extensions are unrelated to `launch` ROS support.

> Interesting, I would have expected that xml and yaml support were always available when using launch...

If we want to force `launch_xml` and `launch_yaml` to always be available, I think a couple `<exec_depend>` tags in the `ros2launch` package should do. Alternatively, we could implicitly target all available frontend packages using group dependencies. Those don't play nice with `bloom` though. Downstream users will still have to install binaries for frontend extensions separately. Thoughts?

christophebedard:
> If we want to force `launch_xml` and `launch_yaml` to always be available, I think a couple `<exec_depend>` tags in the `ros2launch` package should do. Alternatively, we could implicitly target all available frontend packages using group dependencies. Those don't play nice with `bloom` though. Downstream users will still have to install binaries for frontend extensions separately. Thoughts?

Ah I see. Then I don't think there's a need for anything complex, dependency-wise, as you said.

Adding `<exec_depend>`s in `ros2launch` is a simple solution for CLI stuff. However, if you're using launch through `launch_testing[_ros]` (what @wjwwood is doing), you're not using `ros2launch` and you'll still be facing this problem (unless you also happen to depend on `ros2launch`). It's still a good improvement though.

hidmic:
For @wjwwood specific case, we'd be covered -- `ros2launch` is a dependency of `ros2launch_security`.

For the general case, we could do the same for `launch_testing` I guess. It's a sibling package for testing only, should be fine.

christophebedard:
> For @wjwwood specific case, we'd be covered -- `ros2launch` is a dependency of `ros2launch_security`.

Yeah but it's more of a coincidence in this case! Someone might be doing launch tests on a non-launch related package.

> For the general case, we could do the same for `launch_testing` I guess. It's a sibling package for testing only, should be fine.

:+1:

wjwwood:
> Alternatively, we could implicitly target all available frontend packages using group dependencies. Those don't play nice with bloom though. Downstream users will still have to install binaries for frontend extensions separately. Thoughts?

This is precisely the kind of scenario group dependencies were designed to solve, imo. You can _also_ have hard dependencies as well in something like `ros2launch` or `launch_testing`. That doesn't resolve any circular dependencies, but it makes a convenient way to include frontends with something like `ros2launch`.

I don't know if we have it documented anywhere, but we should recommend depending on `ros2launch` always, it seems. Otherwise important parts will not be included.

rviz does this a bit different where `rviz2` depends on "plugins" (group or directly) and they depend on `rviz_common`. We don't have that kind of layout for launch, but we can still get some millage out of the group dependencies.

christophebedard:
> This is precisely the kind of scenario group dependencies were designed to solve, imo. You can _also_ have hard dependencies as well in something like `ros2launch` or `launch_testing`. That doesn't resolve any circular dependencies, but it makes a convenient way to include frontends with something like `ros2launch`.

Oh yeah that would be the best of both worlds! I opened https://github.com/ros2/launch_ros/issues/255 for this and I'll mention adding a group dependency as well.

> I don't know if we have it documented anywhere, but we should recommend depending on `ros2launch` always, it seems. Otherwise important parts will not be included.

I agree. I often have to re-build because I forgot to build `ros2launch`. I'll look into mentioning this somewhere in the docs.

PR #520, opened 2021-07-15
Declare frontend group dependency & use explicit dependencies in launch_testing

christophebedard:
Part of https://github.com/ros2/launch_ros/issues/255

As suggested/discussed under https://github.com/ros2/launch/pull/516#issuecomment-880342455, this declares a group dependency for the launch frontends and declares explicit dependencies on `launch_xml` and `launch_yaml` for `launch_testing`.

Using both a group dependency and hard/explicit dependencies allows this to work fine with binaries (with bloom) and also allows us to include any future launch frontend (or any user's custom frontend) in source builds. This is similar to how `rcl` depends on both the `rcl_logging_packages` group and the default logging implementation explicitly.

### DIFF
--- a/launch_testing/package.xml
+++ b/launch_testing/package.xml
@@ -15,6 +15,9 @@
 
   <exec_depend>ament_index_python</exec_depend>
   <exec_depend>launch</exec_depend>
+  <!-- explicitly depend on the main launch frontends -->
+  <exec_depend>launch_xml</exec_depend>
+  <exec_depend>launch_yaml</exec_depend>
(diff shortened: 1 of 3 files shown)


wjwwood (approved):
lgtm

christophebedard:
@wjwwood I think this can be merged. This PR was tested alongside https://github.com/ros2/launch_ros/pull/256, see https://github.com/ros2/launch_ros/pull/256#issuecomment-966667695

jacobperron (approved):

PR #521, opened 2021-07-15
Added support for handling launch-prefix injection from CLI argument

camm73:
Added support for passing the `--launch-prefix` CLI argument that will be added via [launch_ros PR 254](https://github.com/ros2/launch_ros/pull/254) to `LaunchService` and `LaunchContext`. This `launch-prefix`, when provided, is then directly injected to all `ExecuteProcess` commands.

Testing was also added to `LaunchContext` for the new constructor argument.


camm73:
Just to note, this PR should not be considered until implementation design decisions are confirmed in ros2/launch_ros#254.

camm73:
Closing this PR since it is no longer required for ros2/launch_ros#254.

PR #522, opened 2021-08-18
Add filtering mechanism for executable prefix application

camm73:
In order to support the `--launch-prefix-filter` argument proposed in https://github.com/ros2/launch_ros/issues/257, the regex filter supplied via the `launch-prefix-filter` LaunchConfiguration must be matched against the executable name. In the case of a match, the prefix passed via the `launch-prefix` LaunchConfiguration is prepended to the executable command. If a match is not found, the prefix is not applied.

Additionally, if the user provides a specific prefix for an executable in the launch file, the filter will be overridden and the prefix will be applied as expected.

Corresponding PR: https://github.com/ros2/launch_ros/pull/261 will implement the `--launch-prefix-filter` argument into `launch_ros`.

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -231,10 +231,9 @@ class ExecuteProcess(Action):
         self.__prefix = normalize_to_list_of_substitutions(
             LaunchConfiguration('launch-prefix', default='') if prefix is None else prefix
         )
-        self.__override_filter = False if prefix is None else True
         self.__prefix_filter = normalize_to_list_of_substitutions(
             LaunchConfiguration('launch-prefix-filter', default='')
-        )
+        ) if prefix is None else None
         self.__output = os.environ.get('OVERRIDE_LAUNCH_PROCESS_OUTPUT', output)
         self.__output_format = output_format
 
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -189,8 +189,9 @@ class ExecuteProcess(Action):
         :param: prefix a set of commands/arguments to preceed the cmd, used for
             things like gdb/valgrind and defaults to the LaunchConfiguration
             called 'launch-prefix'. Note that a non-default prefix provided in
-            a launch file will override the prefix provided via the `--launch-prefix` CLI
-            argument regardless of whether `--launch-prefix-filter` is provided.
+            a launch file will override the prefix provided via the `launch-prefix`
+            launch configuration regardless of whether the `launch-prefix-filter` launch
+            configuration is provided.
         :param: output configuration for process output logging. Defaults to 'log'
             i.e. log both stdout and stderr to launch main log file and stderr to
             the screen.
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -230,6 +230,9 @@ class ExecuteProcess(Action):
         self.__prefix = normalize_to_list_of_substitutions(
             LaunchConfiguration('launch-prefix', default='') if prefix is None else prefix
         )
+        self.__prefix_filter = normalize_to_list_of_substitutions(
+            LaunchConfiguration('launch-prefix-filter', default='')
+        )
         self.__output = os.environ.get('OVERRIDE_LAUNCH_PROCESS_OUTPUT', output)
         self.__output_format = output_format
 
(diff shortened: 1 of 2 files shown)


launch/launch/actions/execute_process.py
```diff
@@ -681,7 +686,19 @@ def __expand_substitutions(self, context):
...
+        filter_str = perform_substitutions(context, self.__prefix_filter)
+        regex_filter = None
+        if len(filter_str):
+            regex_filter = re.compile(filter_str)
+        if regex_filter is not None and regex_filter.match(os.path.basename(cmd[0])):
+            cmd = shlex.split(perform_substitutions(context, self.__prefix)) + cmd
+
+        # Apply to all if no filter is provided or filter is overridden
+        if regex_filter is None or self.__override_filter:
+            cmd = shlex.split(perform_substitutions(context, self.__prefix)) + cmd
```
hidmic, line 700: @camm73 meta: hmm, if I read this correctly the prefix filter is irrelevant when a prefix is explicitly given. Nevermind the potential code simplification, that's actually a bit counter-intuitive when attempting to target an specific executable.

camm73: My thinking behind this was based on the fact that whenever a user sets a specific prefix in their launch file it currently overrides any `launch-prefix` provided in the launch configuration and applies the specific prefix: https://github.com/ros2/launch/blob/f891aed9f904df6397ef554f7e0b36bb37b30529/launch/launch/actions/execute_process.py#L230-L232

Since the specific prefix already overrides a prefix passed via `--launch-prefix`, I figured it should also do the same when `--launch-prefix` and `--launch-prefix-filter` are used in conjunction, for the sake of consistency.

If you think it's more intuitive to remove the `override_filter` feature though, I can definitely do that.

hidmic: That's fair. Should be fine so long as we document it.

launch/launch/actions/execute_process.py
```diff
@@ -230,6 +231,10 @@ def __init__(
         self.__prefix = normalize_to_list_of_substitutions(
             LaunchConfiguration('launch-prefix', default='') if prefix is None else prefix
         )
+        self.__override_filter = False if prefix is None else True
+        self.__prefix_filter = normalize_to_list_of_substitutions(
+            LaunchConfiguration('launch-prefix-filter', default='')
+        )
```
hidmic, line 237: @camm73 nit: since `self.__prefix_filter` is irrelevant when `prefix is not None`, then:
```suggestion
        self.__prefix_filter = normalize_to_list_of_substitutions(
            LaunchConfiguration('launch-prefix-filter', default='')
        ) if prefix is None else None
```
conveys the same amount of information.

launch/launch/actions/execute_process.py
```diff
@@ -681,7 +686,19 @@ def __expand_substitutions(self, context):
         cmd = [perform_substitutions(context, x) for x in self.__cmd]
         name = os.path.basename(cmd[0]) if self.__name is None \
             else perform_substitutions(context, self.__name)
-        cmd = shlex.split(perform_substitutions(context, self.__prefix)) + cmd
+
+        # Peform filtering for prefix application
```
hidmic, line 690: @camm73 typo: 
```suggestion
        # Perform filtering for prefix application
```

launch/launch/actions/execute_process.py
hidmic, line 700: @camm73 following above's comment, this can be simplified a bit: 
```suggestion
        should_apply_prefix = True  # by default
        if self.__prefix_filter is not None:  # no prefix given on construction
            prefix_filter = perform_substitutions(context, self.__prefix_filter)
            # Apply if filter regex matches (empty regex matches all strings)
            should_apply_prefix = re.match(prefix_filter, os.path.basename(cmd[0]))
        if should_apply_prefix:
            cmd = shlex.split(perform_substitutions(context, self.__prefix)) + cmd
```

launch/test/launch/test_execute_process.py
```diff
@@ -168,3 +171,40 @@ def generate_launch_description():
...
+        'launch-prefix-filter', 'no-match').visit(lc)
+    assert len(lc.launch_configurations) == 2
+
+    test_process = ExecuteProcess(
+        cmd=[sys.executable, '-c', "print('action')"],
+        output='screen'
+    )
+
+    test_process.execute(lc)
+    assert 'time' not in test_process.process_details['cmd']
```
hidmic, line 210: @camm73 it'd be good to have a test where a prefix is specified for `ExecuteProcess` (and make sure it behaves as expected).

launch/launch/actions/execute_process.py
hidmic: Specifically, let's document this behavior in `ExecuteProcess.__init__()` docstring.

launch/launch/actions/execute_process.py
```diff
@@ -187,7 +188,9 @@ def __init__(
             'emulate_tty' configuration does not represent a boolean.
         :param: prefix a set of commands/arguments to preceed the cmd, used for
             things like gdb/valgrind and defaults to the LaunchConfiguration
-            called 'launch-prefix'
+            called 'launch-prefix'. Note that a non-default prefix provided in
+            a launch file will override the prefix provided via the `--launch-prefix` CLI
+            argument regardless of whether `--launch-prefix-filter` is provided.
```
hidmic, line 193: @camm73 nit: I think it may be slightly more accurate to talk about launch configurations rather than CLI arguments. `ros2launch` is not the only way one can execute launch files.

hidmic (approved):
Overall LGTM

CI up to `launch`, `launch_ros`, and `test_launch_ros`: 
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15099/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9831/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12786/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15267/)

launch/launch/actions/execute_process.py
camm73: Good point. I'll adjust the wording of this

hidmic (approved):

hidmic:
Alright, going in!

PR #525, opened 2021-08-27
Add examples to ExecuteProcess docs

ivanpauno:


launch/launch/actions/execute_process.py
```diff
@@ -84,7 +84,98 @@
...
+    """
+    Action that begins executing a process and sets up event handlers for the process.
+
+    Simple example:
+
+        .. doctest::
+
+            >>> ld = LaunchDescription([
+            ...     ExecuteProcess(
+            ...         cmd=['ls', '-las'],
```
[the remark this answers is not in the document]
ivanpauno: mmm, I didn't think about that.
It really doesn't matter IMO.

The idea is to provide a minimal example, and the doctest is just checking that the constructor doesn't fail, it isn't actually trying to run a launch service.

launch/launch/actions/execute_process.py
```diff
@@ -84,7 +84,98 @@
...
+
+            <launch>
+                <arg name="open_gui" description="when truthy, the gui will be opened"/>
+                <executable cmd="my_cmd --open-gui" if="$(var open_gui)"/>
+                <executable cmd="my_cmd" unless="$(var open_gui)"/>
+                <!--If you have a launch configuration named 'open_gui_arg' that is equal
+                    to "--open-gui" when the "open_gui" argument is truthy and "" when it's
+                    falsy, you'd think that avoiding conditions is possible and you can only
+                    use one executable action with `cmd="my_cmd $(var open_gui_arg)"`.
+                    Unfortunately, that doesn't work, use conditions as in this example.-->
```
hidmic, line 157: Hmm, why not? This is unusual.

hidmic (approved):

launch/launch/actions/execute_process.py
ivanpauno: That ends up being an empty argument passed to the process, depending on the argument parser that might be an issue.
e.g. I think that `rcl` has problems with that (if it appears after `--ros-args`).

hidmic: I see. It should be straightforward to filter them before process invocation though (not to do that now, only pointing out we could easily fix that).

ivanpauno: okay, I will expand the comment explaining the difference between the two options.

ivanpauno: I deleted the comment, as I think it was confusing.
PTAL

ivanpauno:
I'm not sure why I never merged this, CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15320/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10031/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13000/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15513/)

Issue #524, opened 2021-08-26
[launch_xml] Fix how choices is parsed in <arg> tag


ivanpauno:
Current implementation looks wrong https://github.com/ros2/launch/pull/523#discussion_r696130989.
There're currently no tests and no launch file using it.

PR #529, opened 2021-09-09
Fix `DeclareLaunchArgument` xml parsing and constructor

ivanpauno:
Fixes https://github.com/ros2/launch/issues/524 and another bug I found while writing the added test case.

### DIFF
--- a/launch_xml/test/launch_xml/test_arg.py
+++ b/launch_xml/test/launch_xml/test_arg.py
@@ -54,7 +54,7 @@ def test_arg_with_choices():
     arg = ld.entities[0]
     assert 'my_arg' == arg.name
     assert 'asd' == ''.join([x.perform(None) for x in arg.default_value])
-    assert "something Valid choices are: ['asd', 'bsd']" == arg.description
+    assert "something. Valid choices are: ['asd', 'bsd']" == arg.description
     assert ['asd', 'bsd'] == arg.choices
 
 
--- a/launch_xml/test/launch_xml/test_arg.py
+++ b/launch_xml/test/launch_xml/test_arg.py
@@ -38,7 +38,6 @@ def test_arg():
     assert 'something' == arg.description
 
 
-
 def test_arg_with_choices():
     xml_file = \
         """\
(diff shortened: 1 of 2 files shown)


launch_xml/test/launch_xml/test_arg.py
```diff
@@ -38,6 +38,27 @@ def test_arg():
...
+            </arg>
+        </launch>
+        """
+    xml_file = textwrap.dedent(xml_file)
+    root_entity, parser = Parser.load(io.StringIO(xml_file))
+    ld = parser.parse_description(root_entity)
+    arg = ld.entities[0]
+    assert 'my_arg' == arg.name
+    assert 'asd' == ''.join([x.perform(None) for x in arg.default_value])
+    assert "something Valid choices are: ['asd', 'bsd']" == arg.description
```
hidmic, line 58: @ivanpauno nit^N: I know this is pre-existing code, but could we ensure there's a period between the description and the list of choices:
```suggestion
    assert "something. Valid choices are: ['asd', 'bsd']" == arg.description
```
?

hidmic (approved):
LGTM pending green CI

launch_xml/test/launch_xml/test_arg.py
ivanpauno: the test doesn't path with that change, I would need to change the source code for that

hidmic: Right, that's what I meant by ensuring there's a period :sweat_smile: 

Sorry for the noisy comms.

ivanpauno: > Right, that's what I meant by ensuring there's a period sweat_smile

Sorry, I didn't read the first part of your original comment, I only saw the suggestion.
Fixed in https://github.com/ros2/launch/pull/529/commits/5d860a8693aed53ee80b0ec63cb45a8ffbd005ec

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15192/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9912/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12872/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15373/)

hidmic (approved):

PR #530, opened 2021-09-10
[foxy] Backport parser fixes and type_utils refactor

jacobperron:
Backport #414 (prerequisite for #438) (6fd0e83)
Backport #438 (175c466)
Add back functions removed in #438 for backwards compatibility (7ce628f)

For context, this is part of an effort to fix https://github.com/ros2/launch_ros/issues/214 for Foxy.
This PR will enable us to easily backport https://github.com/ros2/launch_ros/pull/137

I don't think there are any significant changes to API, but it's a large change so I may be mistaken.

### DIFF
--- a/launch/launch/frontend/entity.py
+++ b/launch/launch/frontend/entity.py
@@ -14,11 +14,13 @@
 
 """Module for Entity class."""
 
-from typing import Any
 from typing import List
 from typing import Optional
 from typing import Text
 from typing import Union
+
+from launch.utilities.type_utils import AllowedTypesType
+from launch.utilities.type_utils import AllowedValueType
 
 
 class Entity:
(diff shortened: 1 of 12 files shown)


hidmic (approved):
Only skimmed through the PR. API compatibility holds. 

LGTM pending green Foxy CI

ivanpauno:
> Only skimmed through the PR. API compatibility holds.

It does change behavior though, there are values that are parsed different after this PR.
It doesn't affect the most common launch files, but it might break some uncommon ones.

wjwwood:
> It doesn't affect the most common launch files, but it might break some uncommon ones.

Question is, do we consider the old behavior a bug or valid but different? I was leaning towards this is a bugfix (correction to intended behavior), but I don't know.

ivanpauno:
https://github.com/ros2/launch/pull/443 also needs to be backport (https://github.com/ros2/launch/pull/438 was causing issues in CI).

> Question is, do we consider the old behavior a bug or valid but different? I was leaning towards this is a bugfix (correction to intended behavior), but I don't know.

I don't remember exactly, I think that some cases were valid but different (some cases were maybe a bug).
e.g.

```xml
<param name="asd" value="''3''"/>
```

was a parameter with a value `'3'`, using double single quotes was valid and the outer ones were eliminated.
Now that string isn't valid, you have to instead use:

```xml
<param name="asd" value="3" type="str"/>
```

hidmic:
> #443 also needs to be backport (#438 was causing issues in CI).

Good catch.

> Now that string isn't valid

Good point, and this is indeed breaking behavior @jacobperron. But it's also true that quoting was a bit of an obscure corner of `launch` frontends. I mean, we knew that:

> using double single quotes was valid and the outer ones were eliminated.

but I don't think we ever documented that behavior. Or at least I don't remember seeing it written anywhere.

ivanpauno:
> Having said that, it's also true that quoting was a bit of an obscure corner of launch frontends.

True. I'm ok with backporting this, but it's a potentially breaking change.

jacobperron:
This is primarily an attempt to fix https://github.com/ros2/launch_ros/issues/214 (which is a bug in the parser), but it seemed not trivial to fix. I hadn't considered the behavior change. It would be good to know what corner cases this breaks exactly before proceeded. If we decide to merge this, then hopefully we can enumerate the cases this breaks (and the preferred syntax).

ivanpauno:
> This is primarily an attempt to fix [ros2/launch_ros#214](https://github.com/ros2/launch_ros/issues/214) (which is a bug in the parser), but it seemed not trivial to fix. I hadn't considered the behavior change. It would be good to know what corner cases this breaks exactly before proceeding. If we decide to merge this, then hopefully we can enumerate the cases this breaks (and show the preferred syntax).

The only one I remember is the one commented before https://github.com/ros2/launch/pull/530#issuecomment-919154634, I'm not aware of other corner cases but there might be.

jacobperron:
Added backport of https://github.com/ros2/launch/pull/443 (85f01b9).

> The only one I remember is the one commented before #530 (comment), I'm not aware of other corner cases but there might be.

```
<param name="asd" value="''3''"/>
```
IMO, this was a workaround because of a bug. I think it's worth fixing this with a release note that folks no longer need to do this extra quoting. If everyone agrees with this, I can add a note to the Foxy release page (for the upcoming patch release).

ivanpauno (approved):

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15340/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10051/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13020/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15535/)

Blast545:
:man_farmer: I think this backport caused a test regression in all the foxy jobs.
Reference: https://build.ros2.org/view/Fci/job/Fci__nightly-release_ubuntu_focal_amd64/335/

This is the test is failing: https://github.com/ros2/launch_ros/blob/c9b6b7057d19b3cba67072818d499370cb14f163/test_launch_ros/test/test_launch_ros/frontend/test_node_frontend.py#L167
Should the test be updated to match the PR or is this an unexpected side effect?

PTAL @jacobperron

jacobperron:
@Blast545 My mistake, I forgot to merge the connected PR https://github.com/ros2/launch_ros/pull/265, which I think should resolve the issue.

Blast545:
No prob, thanks for taking a look!

PR #537, opened 2021-10-22
Add pathlib.Path to SomeSubstitutionsType

DLu:
I'm making this change relatively blindly as a way to start discussion. 

In https://github.com/ament/ament_index/pull/73 I added the ability to load the path to a package as a `pathlib.Path`. I'd like to be able to directly use that in my launch files without manually converting to `str`. Consider the following example:

```
from ament_index_python.packages import get_package_share_path

from launch.actions import IncludeLaunchDescription

    pkg_path = get_package_share_path('super_launch_package')
    launch_py_1 = PythonLaunchDescriptionSource(str(pkg_path / 'launch/launch_1.launch.py'))
    launch_py_2 = PythonLaunchDescriptionSource(str(pkg_path / 'launch/launch_2.launch.py'))
    ld.add_action(IncludeLaunchDescription(launch_py_1))
    ld.add_action(IncludeLaunchDescription(launch_py_2))
    return ld
```

Questions I have:
 * Where are there repercussions of this that I'm missing? 
 * Where should tests be added?

### DIFF
--- a/launch/launch/some_substitutions_type.py
+++ b/launch/launch/some_substitutions_type.py
@@ -15,6 +15,7 @@
 """Module for SomeSubstitutionsType type."""
 
 import collections.abc
+import pathlib
 from typing import Iterable
 from typing import Text
 from typing import Union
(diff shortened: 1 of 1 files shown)


ivanpauno:
> Hmm, good point. @wjwwood @ivanpauno WDYT?

`SomeSubstitutionsType` is also used for things that are not a path, so I don't think we should modify that.

We could have add a `SomePathType` definition, similar to what's proposed here.
We would also need a method to resolve that `SomePathType` "substitution" when a context is provided.

DLu:
I'm closing this for now, as calling `str()` is not too onerous.

PR #552, opened 2021-11-03
* [launch_pytest] Fix issue when colcon --retest-until-fail flag is used

ivanpauno:
Fixes https://github.com/ros2/launch/pull/528#issuecomment-959769891.

test_module_scope.py / test_function_scope.py don't work when `--count=N` is passed, because now the test order is different.
I moved them to the example folder and I'm now running then with the pytester plugin to avoid the issue.

I'm also running the other examples we have in that folder, ignoring one that would introduce a dependency on launch_ros (we have to move that one somewhere else later).
I've also fixed some issues in the existing examples.


launch_pytest/test/launch_pytest/examples/pytest_hello_world.py
```diff
@@ -50,7 +50,7 @@ def validate_output(output):
         # this function can use assertions to validate the output or return a boolean.
         # pytest generates easier to understand failures when assertions are used.
         assert output == 'hello_world\n', 'process never printed hello_world'
-    assert process_tools.wait_for_output_sync(
```
Blast545, line 53: Curious question, why was this assert removed?

Blast545 (approved):
LGTM after green CI

launch_pytest/test/launch_pytest/examples/pytest_hello_world.py
ivanpauno: The function always returns `None` in this case, though I think that's confusing.

ivanpauno: See https://github.com/ros2/launch/pull/552/commits/a9cc81cbca37d730773a28cae9671550c2ee41b0

Blast545 (approved):

ivanpauno:
CI using `--retest-until-fail 2`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15569/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10253/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13237/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15784/)

ivanpauno:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15789/)

launch_pytest/test/launch_pytest/examples/pytest_hello_world.py
hidmic: @ivanpauno meta: we clearly need a less terse idiom to write predicates. If we were to assume the predicate passed if it returns `None`, that'd be practical yet surprising. If we were to simply return the result of evaluating the expression, we'd lose detailed assertion message (most of which are provided `pytest` rewriting logic). So how about splitting use cases?

```py
# Wait for incoming process output with a timeout
output = wait_for_output_sync(process, timeout=10)
output = await wait_for_output(process, timeout=10)

# Wait for incoming process output matching a predicate with a timeout
def predicate(output):
    return 'foo' in output
output = wait_for_output_sync(process, predicate, timeout=10)
output = await wait_for_output(process, predicate, timeout=10)

# Assert a statement on process output with a timeout 
def statement(output):
    assert output == 'foo'
assert_output_sync(process, statement, timeout=10)
await assert_output(process, statement, timeout=10)
```

In this case, it would read:

```py
def hello_world_found():
     assert output.splitlines() == ['hello_world'], 'process never printed hello_world'
assert_output_sync(launch_context, hello_world_proc, hello_world_found, timeout=5)
```

hidmic (approved):
LGTM but for one meta comment

launch_pytest/test/launch_pytest/examples/pytest_hello_world.py
ivanpauno: That sounds like the right approach to me.
I will address it in a follow-up PR so we solve the CI issue.

hidmic: Agreed, merge!

Issue #313, opened 2019-08-16
Propagate launch arguments in included launch descriptions


ivanpauno:
#### Feature description
There have been some previous discussion in https://github.com/ros2/launch/issues/252 https://github.com/ros2/launch/issues/249 too.
I've been discussing this a little with @hidmic, here are some of the items we discussed:

- We're currently doing a [best effort checking](https://github.com/ros2/launch/blob/89de94182947ed8fa61fbe7554866e871d8a5121/launch/launch/actions/include_launch_description.py#L139-L152) of arguments when visiting `IncludeLaunchDescription` action. [Launch arguments](https://github.com/ros2/launch/blob/89de94182947ed8fa61fbe7554866e871d8a5121/launch/launch/launch_description.py#L101-L128) of nested launch descriptions are ignored in the process. We're also only showing those arguments when executing `ros2 launch -s <PACKAGE> <LAUNCH_FILE>`. But [we're not checking in ros2 launch](https://github.com/ros2/launch_ros/blob/7e4f43156c729839cc8e6dd986f32bcac169a25b/ros2launch/ros2launch/api/api.py#L126-L160) if the passed arguments were declared by the launch description or not. So currently, the launch arguments are "automatically propagated" to nested launch files, though they aren't showed to the user.
- Checking if all the launch arguments are satisfied when visiting a `IncludeLaunchDescription` action, including arguments in nested launch descriptions, is hard, because some information may not be available at that moment:
  - [launch file name](https://github.com/ros2/launch/blob/89de94182947ed8fa61fbe7554866e871d8a5121/launch/launch/actions/include_launch_description.py#L121-L124) is maybe not known.
  - [name of passed launch arguments](https://github.com/ros2/launch/blob/89de94182947ed8fa61fbe7554866e871d8a5121/launch/launch/actions/include_launch_description.py#L68-L70) to nested `IncludeLaunchDescription` may not be known.
  - A `SetLaunchConfiguration` action may satisfy the requirement before the nested `IncludeLaunchDescritpion` is visited. If we do the verification anticipated, we can't be aware of that.
- If `IncludeLaunchDescription` aren't scoped (as is currently happening), we can run in the following problems:
  ```python3
  def get_launch_description():
    return LaunchDescription([
      IncludeLaunchDescription(
        <LAUNCH_DESCRIPTION_SOURCE_1>,
        launch_arguments={'<ARG1_NAME>': '<ARG1_VALUE>'}.items()
      ),
      IncludeLaunchDescription(
[...]

PR #556, opened 2021-11-10
Improve launch arguments introspection

ivanpauno:
Currently you can set an argument of a nested launch description from the top level, without the need of redeclaring the argument.
This updates the `get_launch_arguments()` method to return a more complete result (currently it doesn't return arguments from included launch descriptions).
It also updates the `include_launch_description` launch arguments validation by tracking all the include launch description actions in the middle (this basically avoids reintroducing https://github.com/ros2/launch/issues/248).

Related discussion https://github.com/ros2/launch/issues/313.

### DIFF
--- a/launch/launch/actions/include_launch_description.py
+++ b/launch/launch/actions/include_launch_description.py
@@ -137,9 +137,9 @@ class IncludeLaunchDescription(Action):
             ]
         except Exception as exc:
             self.__logger.debug(
-                'Failed to get launch arguments names for launch description: '
-                f'{self.__launch_description_source.location}'
-                f'exception: {str(exc)}'
+                'Failed to get launch arguments names for launch description '
+                f"'{self.__launch_description_source.location}', "
+                f'with exception: {str(exc)}'
             )
         return None
 
--- a/launch/launch/actions/include_launch_description.py
+++ b/launch/launch/actions/include_launch_description.py
@@ -135,10 +135,11 @@ class IncludeLaunchDescription(Action):
                 perform_substitutions(context, normalize_to_list_of_substitutions(arg_name))
                 for arg_name, arg_value in self.__launch_arguments
             ]
-        except Exception:
+        except Exception as exc:
             self.__logger.debug(
                 'Failed to get launch arguments names for launch description: '
                 f'{self.__launch_description_source.location}'
+                f'exception: {str(exc)}'
             )
         return None
 
--- a/launch/launch/actions/include_launch_description.py
+++ b/launch/launch/actions/include_launch_description.py
@@ -21,6 +21,8 @@ from typing import Optional
 from typing import Tuple
 from typing import Union
 
+import launch.logging
+
 from .set_launch_configuration import SetLaunchConfiguration
 from ..action import Action
 from ..frontend import Entity
@@ -77,6 +79,7 @@ class IncludeLaunchDescription(Action):
             launch_description_source = AnyLaunchDescriptionSource(launch_description_source)
         self.__launch_description_source = launch_description_source
         self.__launch_arguments = () if launch_arguments is None else tuple(launch_arguments)
+        self.__logger = launch.logging.get_logger(__name__)
 
     @classmethod
     def parse(cls, entity: Entity, parser: Parser):
@@ -125,6 +128,20 @@ class IncludeLaunchDescription(Action):
         ret = self.__launch_description_source.try_get_launch_description_without_context()
         return [ret] if ret is not None else []
 
+    def _try_get_arguments_names_without_context(self):
+        try:
+            context = LaunchContext()
+            return [
+                perform_substitutions(context, normalize_to_list_of_substitutions(arg_name))
+                for arg_name, arg_value in self.__launch_arguments
+            ]
+        except Exception:
+            self.__logger.debug(
+                'Failed to get launch arguments names for launch description: '
+                f'{self.__launch_description_source.location}'
+            )
+        return None
+
     def execute(self, context: LaunchContext) -> List[LaunchDescriptionEntity]:
         """Execute the action."""
         launch_description = self.__launch_description_source.get_launch_description(context)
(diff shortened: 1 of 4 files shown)


ivanpauno:
@jacobperron I'm still thinking if this is the right approach, but feedback would be appreciated.

launch/launch/actions/include_launch_description.py
```diff
@@ -125,6 +128,20 @@ def get_sub_entities(self):
...
+        try:
+            context = LaunchContext()
+            return [
+                perform_substitutions(context, normalize_to_list_of_substitutions(arg_name))
+                for arg_name, arg_value in self.__launch_arguments
+            ]
+        except Exception:
+            self.__logger.debug(
+                'Failed to get launch arguments names for launch description: '
+                f'{self.__launch_description_source.location}'
```
wjwwood, line 141: Is it possible to give a bit more context here? maybe just `str()` the exception at the end?

launch/launch/launch_description.py
```diff
@@ -76,6 +76,25 @@ def get_launch_arguments(self, conditional_inclusion=False) -> List[DeclareLaunc
...
+        return [
+            item[0] for item in
+            self.get_launch_arguments_with_include_launch_descriptions(conditional_inclusion)
+        ]
+
+    def get_launch_arguments_with_include_launch_descriptions(
+        self, conditional_inclusion=False
+    ) -> List[DeclareLaunchArgument]:
+        """
+        Return a list of (:py:class:`launch.actions.DeclareLaunchArgument`, Optional[List[:py:class:`launch.actions.IncludeLaunchDescription`]]) pairs.
```
wjwwood, line 91: nitpick: should probably wrap this line somehow.

I see you're suppressing that warning, but I don't know why this cannot be wrapped.

launch/launch/launch_description.py
```diff
@@ -76,6 +76,25 @@ def get_launch_arguments(self, conditional_inclusion=False) -> List[DeclareLaunc
...
 
+        See :py:method:`get_launch_arguments_with_launch_descriptions()`
+        for more details.
+        """
+        return [
+            item[0] for item in
+            self.get_launch_arguments_with_include_launch_descriptions(conditional_inclusion)
+        ]
+
+    def get_launch_arguments_with_include_launch_descriptions(
```
wjwwood, line 87: nitpick: would `get_launch_arguments_with_include_launch_description_actions()` be a better name?

launch/launch/launch_description.py
```diff
@@ -97,31 +116,40 @@ def get_launch_arguments(self, conditional_inclusion=False) -> List[DeclareLaunc
...
-                    if entity.name in (e.name for e in declared_launch_arguments):
+                    if entity.name in (e.name for e, _ in declared_launch_arguments):
                         continue
                     # Stuff this contextual information into the class for
                     # potential use in command-line descriptions or errors.
                     entity._conditionally_included = _conditional_inclusion
                     entity._conditionally_included |= entity.condition is not None
-                    declared_launch_arguments.append(entity)
+                    declared_launch_arguments.append((entity, nested_ild_actions))
+                if isinstance(entity, ResetLaunchConfigurations):
```
wjwwood, line 134: This is an interesting side-effect of this action, good catch.

wjwwood (approved):
Generally looks good to me. It's best effort only, but the best we can do under the circumstances.

The only other thing I would mention is that when displaying them on the CLI it would be nice to differentiate between arguments that are conditional and ones that are certain.

ivanpauno:
> The only other thing I would mention is that when displaying them on the CLI it would be nice to differentiate between arguments that are conditional and ones that are certain.

That's already possible, see:

https://github.com/ros2/launch/blob/8bb79dd1fa64e84b734475f27c16714d788206d7/launch/launch/launch_description.py#L131-L132

and:

https://github.com/ros2/launch_ros/blob/b84b72748e22d8fd648151d96478d19305c96c82/ros2launch/ros2launch/api/api.py#L104-L106

Stuffing the attribute in the original action is a bit hacky though, we could maybe be returning a:

```
List[Tuple[DeclareLaunchArgument, List[IncludeLaunchDescription], bool]]
```

where the last boolean indicates if the `DeclareLaunchArgument` action is being conditionally included or not.

launch/launch/launch_description.py
ivanpauno: If I wrap the line I get:

>      82 ./launch/launch_description.py:95:1: D205 1 blank line required between summary line and description
>      86 ./launch/launch_description.py:95:1: D400 First line should end with a period

I solved it by shortening the summary line, see https://github.com/ros2/launch/pull/556/commits/2c49f810cd6bc7b851ab785bb9ade9146aabcde3.

launch/launch/actions/include_launch_description.py
ivanpauno: Sounds good, fixed in https://github.com/ros2/launch/pull/556/commits/2aa41db6636a23c34257acf4af15f1f7e3b89b86.

launch/launch/launch_description.py
ivanpauno: fixed in https://github.com/ros2/launch/pull/556/commits/32113b20649a4a64e12ba66526f6c5bc61b7f486

launch/launch/actions/include_launch_description.py
```diff
@@ -125,6 +128,21 @@ def get_sub_entities(self):
...
+            context = LaunchContext()
+            return [
+                perform_substitutions(context, normalize_to_list_of_substitutions(arg_name))
+                for arg_name, arg_value in self.__launch_arguments
+            ]
+        except Exception as exc:
+            self.__logger.debug(
+                'Failed to get launch arguments names for launch description: '
+                f'{self.__launch_description_source.location}'
+                f'exception: {str(exc)}'
```
wjwwood, line 142: I think this will result in the word 'exception' and the path getting concatenated. What about this (two colons was kind of weird):

```suggestion
                'Failed to get launch arguments names for launch description '
                f"'{self.__launch_description_source.location}', "
                f'with exception: {str(exc)}'
```

Or at least something like this:

```suggestion
                'Failed to get launch arguments names for launch description: '
                f'{self.__launch_description_source.location}, '
                f'exception: {str(exc)}'
```

wjwwood (approved):
> where the last boolean indicates if the DeclareLaunchArgument action is being conditionally included or not.

I don't mind how it's done really (what you proposed sounds fine), but I just care that the user can see it on the CLI (as in this information is actually being used).

launch/launch/actions/include_launch_description.py
ivanpauno: Right :man_facepalming:.
I applied the first suggestion.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15625/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10306/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13293/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15842/)
Triggering a new build of ci_linux

wjwwood (approved):

PR #558, opened 2021-11-18
Document TimerAction params

christophebedard:
Mainly to explicitly mention that the period value is in seconds.


jacobperron (approved):
Thanks for the improvment!

PR #566, opened 2021-12-02
Propagate exceptions of completed actions to launch service main loop

tumtom:
Hi,

exceptions raised by launch actions (e.g. RuntimeError raised by action.DeclareLaunchArgument) are not handled by the "except Exception" block in https://github.com/tumtom/launch/blob/98bcc874fe6f459ade8571697be32c8c74f6b045/launch/launch/launch_service.py#L337.

This is, because the exception is not picked up from the completed Future (https://docs.python.org/3/library/asyncio-future.html#asyncio.Future.exception)

Instead, only the exception handler function in https://github.com/tumtom/launch/blob/98bcc874fe6f459ade8571697be32c8c74f6b045/launch/launch/launch_service.py#L280 is called then, which leads to the following message being printed

```
Task exception was never retrieved
future: <Task finished name='Task-2' coro=<LaunchService._process_one_event() done, defined at ..../opt/ros/foxy/lib/python3.8/site-packages/launch/launch_service.py:226> exception=RuntimeError("Included launch description missing required argument 'argument (description: 'A description), given: []")>
```

This commit changes the completed action's exception being handled by the "except Exception" block and being printed by the logger as:

`[ERROR] [launch]: Caught exception in launch (see debug for traceback): Included launch description missing required argument 'argument (description: 'A description), given: []`

Signed-off-by: Thomas Rehner <thomas.rehner@aed-engineering.com>

### DIFF
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -319,10 +319,14 @@ class LaunchService:
                     entity_futures.append(process_one_event_task)
 
                     # Wait on events and futures
-                    await asyncio.wait(
+                    completed_tasks, _ = await asyncio.wait(
                         entity_futures,
                         return_when=asyncio.FIRST_COMPLETED
                     )
+                    # Propagate exception from completed task
+                    for completed_task in completed_tasks:
+                        if completed_task.exception():
+                            raise completed_task.exception()
 
                 except KeyboardInterrupt:
                     continue
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -327,7 +327,7 @@ class LaunchService:
                     completed_tasks_exceptions = [task.exception() for task in completed_tasks]
                     completed_tasks_exceptions = list(filter(None, completed_tasks_exceptions))
                     if completed_tasks_exceptions:
-                        self.__logger.debug("An exception was raised in an async action/event")
+                        self.__logger.debug('An exception was raised in an async action/event')
                         # in case there is more than one completed_task, log other exceptions
                         for completed_tasks_exception in completed_tasks_exceptions[1:]:
                             self.__logger.error(completed_tasks_exception)
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -323,10 +323,15 @@ class LaunchService:
                         entity_futures,
                         return_when=asyncio.FIRST_COMPLETED
                     )
-                    # Propagate exception from completed task
-                    for completed_task in completed_tasks:
-                        if completed_task.exception():
-                            raise completed_task.exception()
+                    # Propagate exception from completed tasks
+                    completed_tasks_exceptions = [task.exception() for task in completed_tasks]
+                    completed_tasks_exceptions = list(filter(None, completed_tasks_exceptions))
+                    if completed_tasks_exceptions:
+                        self.__logger.debug("An exception was raised in an async action/event")
+                        # in case there is more than one completed_task, log other exceptions
+                        for completed_tasks_exception in completed_tasks_exceptions[1:]:
+                            self.__logger.error(completed_tasks_exception)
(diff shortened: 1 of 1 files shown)


launch/launch/launch_service.py
```diff
@@ -319,10 +319,14 @@ def _on_exception(loop, context):
...
                     # Wait on events and futures
-                    await asyncio.wait(
+                    completed_tasks, _ = await asyncio.wait(
                         entity_futures,
                         return_when=asyncio.FIRST_COMPLETED
                     )
+                    # Propagate exception from completed task
+                    for completed_task in completed_tasks:
+                        if completed_task.exception():
+                            raise completed_task.exception()
```
hidmic, line 329: @tumtom if more than one exception is thrown, won't we get the same "exception not retrieved" errors?

hidmic (commented):
Thanks for your contribution @tumtom ! I only have one question. Then I'll run CI.

launch/launch/launch_service.py
tumtom: Hi @hidmic
good question. You're probably right. Do you think this could happen and should be handled?
If so, I can think of different options:
- generate a new exception, in which we sum up the exceptions (modifies exception type!)
- log all exceptions that were retrieved, then raise the first (I think this would be better, right?)

The second approach could be implemented like this:
````python
completed_tasks_exceptions = [task.exception() for task in completed_tasks]
completed_tasks_exceptions  = list(filter(None, completed_tasks_exceptions))
if completed_tasks_exceptions:
    self.__logger.error(completed_tasks_exceptions)
    raise completed_tasks_exceptions[0]
````

hidmic: @tumtom I'm subjectively inclined towards the second approach. Perhaps we can do better when it comes to logging them, but I'm onboard with the idea.

tumtom: Ok, let me know what you think!
I'm not sure this code will ever execute due to `return_when=asyncio.FIRST_COMPLETED`, but if more than one action would complete and raise, the exceptions won't be lost now

hidmic (approved):
LGTM pending green CI

CI up to and above `launch`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15761/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10442/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13427/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16013/)

hidmic:
Let's give it another shot:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15779/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10457/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13441/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16033/)

hidmic:
@tumtom `flake8` is still not happy.

tumtom:
Hi @hidmic sorry, I had some troubles executing the linters locally before pushing. Should be fine now, I hope! Do you want me to squash the commits to keep the history clean?

hidmic:
@tumtom third's the charm!

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15839/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10516/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13504/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16131/)

> Do you want me to squash the commits to keep the history clean?

No need to. We'll squash before merging.

hidmic:
Alright, CI's happy. Going in!

Thank you for your contribution (and patience) @tumtom.

Issue #366, opened 2020-01-03
Add a Command substitution


ivanpauno:
#### Feature description

In ROS1, it was possible to load a parameter value with the output of a command:

```xml
<param name="robot_description" command="$(find xacro)/xacro --inorder
                       '$(find my_package)/path/to/xacro/file'"/>
```

In order to support similar functionality in ROS2 launch, a `Command` substitution can be added.
Adding a substitution instead of adding an specific way of loading a parameter value with a command output will allow other use cases.

Example usage:

```python
Node(
  node_executable = 'my_node',
  package='my_package',
  parameters=[{'robot_description': Command(
    [
      PathJoinSubstitution([FindPackagePrefix('xacro'), 'xacro']),
      ' --inorder ',
      PathJoinSubstitution([FindPackagePrefix('my_package'), 'path', 'to', 'xacro', 'file']),
    ]
  }]
)
```

#### Implementation considerations

Implementation skeleton:

```python3

class CommandSubstitution(Substitution):

  def __init__(self, command: SomeSubstitutionsType):
    ...

  def perform(self, context: LaunchContext) -> Text:
    ...
```

ivanpauno:
@wjwwood @hidmic can you review the proposal?

hidmic:
LGTM, though we should do something about ergonomics when this gets mapped to XML/YAML - - it's crazy verbose :sweat_smile:.

wjwwood:
lgtm too.

ivanpauno:
> LGTM, though we should do something about ergonomics when this gets mapped to XML/YAML - - it's crazy verbose .

About this: I think that most of the verbosity comes from the need of the `PathJoinSubstitution` stuff.
Forgetting about that, it would look in xml like:

```xml
<param name="robot_description" value="$(command '$(find_pkg_prefix xacro)/xacro --inorder
                       $(find my_package)/path/to/xacro/file'"/>
```

(note that the command is between quotes, because I think it's the most natural way to match what we do in Python)

That's not super verbose.
We could also have directly an `output` attribute for params like in ROS 1, to make it less verbose.
Another options is having a specific substitution for `xacro` (which is pretty common).

Paths are a problem here (if we want the launch file to work in all the platforms).
It's hard to accept posix paths here without adding a substitution.
The first part of the command (`$(find_pkg_prefix xacro)/xacro`) is known to be a path, so we can avoid a substitution there, and accept a posix path.
The problem is if there are other paths in the arguments of the command, a substitution is required there.

[...]

PR #570, opened 2021-12-10
Add XacroFileContents substitution class

bergercookie:
Add a substitution class for loading a xacro file during evaluation.

P.S. Midway through writing this I discovered that I could use the `Command` substitution to call to the `xacro` executable (https://github.com/ros2/launch/issues/366). However I still believe there's value in having a dedicated class for in that
* it should be faster to call a pythn function than to invoke subprocess and
* it should be more straightforward for the developer at hand to discover and also use this substitution type for manipulating xacro files instead of concatenating the individual command parts

### DIFF
--- a/launch/launch/substitutions/xacro_file_contents.py
+++ b/launch/launch/substitutions/xacro_file_contents.py
@@ -1,0 +1,46 @@
+"""Module for the XacroFile substitution."""
+
+from pathlib import Path
+from typing import Text
+
+import xacro
+from launch.launch_context import LaunchContext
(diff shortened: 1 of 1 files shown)


clalancette (commented):
Note that in order to do this, you would have to make `xacro` a dependency of this package, and that would mean pulling it into the ROS 2 core.  Overall, I don't think this is a good move; while I know xacro is popular, it isn't the only way to do things.  So I don't think it should be "special" and I think use of the Command substitution could just be better documented.

bergercookie:
@clalancette, ok, thanks for the feedback, closing this.

PR #580, opened 2022-01-12
Mention that ready_fn in generate_test_description is deprecated

Bi0T1N:
Foxy already shows:
*Warning: Passing ready_fn as an argument to generate_test_description will be removed in a future release.  Include a launch_testing.actions.ReadyToTest action in the LaunchDescription instead.*

see #346


hidmic:
@Bi0T1N thanks for the patch! However, `ready_fn` is already gone in Humble and Rolling. If anything, this applies to Galactic. Consider re-targeting your PR if interested.

PR #584, opened 2022-01-14
[launch_testing_ament_cmake] Add test label

### LINKED ISSUE #579: [launch_testing_ament_cmake] Feature request: add test label
#### Feature description

In the current implementation, no test label is set for `add_launch_test()`. This causes the results of the `colcon test-result` command to show the results of launch_testing without labels.
Also, in my use case, I would like to run each test with a test label, but I cannot specify a label for `launch_testing` because it is not set.
When I use `ament_add_gtest()`, the LABEL is set to `gtest` ([See here](https://github.com/ament/ament_cmake/blob/master/ament_cmake_gtest/cmake/ament_add_gtest_test.cmake#L103)), but when I use `add_launch_test()`, the LABEL is not set.(https://github.com/ros2/launch/blob/master/launch_testing_ament_cmake/cmake/add_launch_test.cmake#L125)

Can you please specify LABEL for `add_launch_test()` as well, to make the behavior consistent?

I think this function can be achieved by writing the following code after `amend_add_test()` in the existing `add_launch_test.cmake`.

```
  set_tests_properties(
    "${_launch_test_TARGET}"
    PROPERTIES
    LABELS "launch_test"
  )
```

Thanks in advance.

KeisukeShima:
Closes #579 

Signed-off-by: Keisuke Shima <19993104+KeisukeShima@users.noreply.github.com>

### DIFF
--- a/launch_testing_ament_cmake/cmake/add_launch_test.cmake
+++ b/launch_testing_ament_cmake/cmake/add_launch_test.cmake
@@ -52,7 +52,7 @@ macro(parse_launch_test_arguments namespace filename)
   cmake_parse_arguments(${namespace}
     ""
     "TARGET;TIMEOUT;PYTHON_EXECUTABLE"
-    "ARGS"
+    "ARGS;LABELS"
     ${ARGN})
 
   if(NOT ${namespace}_TIMEOUT)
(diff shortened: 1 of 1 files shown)


jacobperron (approved):
LGTM, thanks for the addition!

Testing everything above `launch_testing_ament_cmake`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16020/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10687/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16351/)

PR #589, opened 2022-02-01
[launch_testing] remove deprecated `ready_fn` feature

wjwwood:
This feature was deprecated in https://github.com/ros2/launch/pull/346 (foxy) and should be removed by now, so we'll do it in Humble.

We realized this due to this pr (thanks for that btw): https://github.com/ros2/launch/pull/580

~~I'm opening in draft because I quickly did this in the browser and I'm not sure if the related `ReadyAggregator` class needs to be removed. So I'll let CI run first.~~

CI is good, but I'm still not sure about the `ReadyAggregator` class, looking for input on that.

### DIFF
--- a/launch_testing/README.md
+++ b/launch_testing/README.md
@@ -44,21 +44,6 @@ The launch description needs to include a `ReadyToTest` action to signal to the 
 
 In the above example, there is no need to delay the start of the tests so the `ReadyToTest` action is a peer to the process under test and will signal to the framework that it's safe to start around the same time the `ExecuteProcess` action is run.
 
-In older style tests, a function called `ready_fn` is declared as an argument to `generate_test_description` and must be plumbed into the launch description with an `OpaqueFunction`.
-
-```python
-def generate_test_description(ready_fn):
-
-    return launch.LaunchDescription([
-        launch.actions.ExecuteProcess(
-            cmd=[path_to_process],
-        ),
-
-        # Start tests right away - no need to wait for anything in this example
-        launch.actions.OpaqueFunction(function=lambda context: ready_fn()),
-    ])
-```
-
 #### Active Tests
 
 Any classes that inherit from `unittest.TestCase` and not decorated with the `post_shutdown_test` descriptor will be run concurrently with the proccess under test.
--- a/launch_testing/launch_testing/loader.py
+++ b/launch_testing/launch_testing/loader.py
@@ -42,49 +42,34 @@ def _normalize_ld(launch_description_fn):
             return result, {}
 
     def wrapper(**kwargs):
+        # This is a new-style launch_description which should contain a ReadyToTest action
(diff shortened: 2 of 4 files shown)


wjwwood:
Preliminary CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16045/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10700/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16369/)

adityapande-1995:
Should this line be removed as well ? https://github.com/ros2/launch/blob/73fa05efa142da20a704aea4003e61d08e99e728/launch_testing/launch_testing/test_runner.py#L74

wjwwood:
The short answer is: "I don't know"

I looked at that line, but I couldn't figure out if it is part of the test runner machinery that is unfortunately named the same as the old feature users would use, or not. That's why I ran CI, I figured that if it was the old feature I'm removing some tests should fail, but maybe that's not the case.

I also sort of traced the logic and I believe that it is still used, but the difference is that the `ready_fn` is passed to the wrapper and not the `launch_description_fn`, as it used to be.

Maybe someone more familiar with `launch_testing` like @ivanpauno or @hidmic could comment on it.

hidmic:
> Should this line be removed as well ?

@adityapande-1995 @wjwwood It shouldn't. That's the actual callable that the `ReadyToTest` action eventually invokes. 

> I looked at that line, but I couldn't figure out if it is part of the test runner machinery that is unfortunately named the same as the old feature users would use, or not.

TBH this whole thing is messy.

hidmic (approved):
@wjwwood LGTM but 

https://github.com/ros2/launch/blob/b59bb6b88b40eaa82870542d93bdfb1a033d05c3/launch_testing/launch_testing/test_runner.py#L295-L296

should be removed too.

adityapande-1995 (approved):

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16163/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10791/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16501/)

hidmic:
@wjwwood linters were still failing in that last CI.

wjwwood:
New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16539/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11125/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16913/)

ivanpauno (approved):
LGTM!

audrow (approved):
Looks good to me!

PR #591, opened 2022-02-09
Add in pytest 7 compatibility.

### LINKED ISSUE #590: :man_farmer: Windows test regressions, deprecated: `fspath argument for Node constructors replaced with pathlib.Path`

clalancette:
In pytest 7, a bunch of APIs around hooks were deprecated.
Since we use those APIs, we are getting warnings every time
we invoke the APIs in question.  This commit adds in a compatibility
layer so that we can support pytest from 4.6.9 (which is in
Ubuntu 20.04), up to pytest 7.0.0 (which is what is currently
in pip).

Closes #590

You may notice that I have done some serious hackery in here.  In particular, I now have a global `pytest_major` where I parse out the pytest major version from the `pytest.__version__` string, and then use that to figure out which signature of `pytest_pycollect_makemodule` to overload.  That works, and is ugly, but I didn't know a better way to do this without resorting to `ast` hackery or similar.  I'm open to other ways to do that if someone gives me a pointer.

Also, I've currently only tested this on the `launch_testing` module; I need to do a more full-fledged test before I consider it ready for review.  Still, I wanted to open it now to get some feedback.

### DIFF
--- a/launch_testing/launch_testing/pytest/hooks.py
+++ b/launch_testing/launch_testing/pytest/hooks.py
@@ -23,6 +23,9 @@ from ..loader import LoadTestsFromPythonModule
 from ..test_runner import LaunchTestRunner
 
 
+pytest_major = int(pytest.__version__.split('.')[0])
+
+
 class LaunchTestFailure(Exception):
 
     def __init__(self, message, results):
@@ -122,10 +125,22 @@ class LaunchTestItem(pytest.Item):
         return self.fspath, 0, 'launch tests: {}'.format(self.name)
 
 
-class LaunchTestModule(pytest.File):
+class LaunchTestModulePytest7(pytest.File):
 
-    def __init__(self, parent, *, fspath):
-        super().__init__(parent=parent, fspath=fspath)
+    def makeitem(self, *args, **kwargs):
+        return LaunchTestItem.from_parent(*args, **kwargs)
+
+    def collect(self):
+        module = self.fspath.pyimport()
+        yield self.makeitem(
+            name=module.__name__, parent=self,
+            test_runs=LoadTestsFromPythonModule(
+                module, name=module.__name__
+            )
+        )
+
+
+class LaunchTestModule(pytest.File):
 
     @classmethod
     def from_parent(cls, parent, *, fspath):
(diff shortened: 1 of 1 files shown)


ivanpauno (approved):
LGTM with green CI!

sloretz:
Gah, you just beat me #592. I'll try this one locally

launch_testing/launch_testing/pytest/hooks.py
```diff
@@ -122,10 +125,22 @@ def reportinfo(self):
         return self.fspath, 0, 'launch tests: {}'.format(self.name)
 
 
-class LaunchTestModule(pytest.File):
+class LaunchTestModulePytest7(pytest.File):
 
-    def __init__(self, parent, *, fspath):
-        super().__init__(parent=parent, fspath=fspath)
```
sloretz, line 128: Seeing a warning in `LaunchROSTestModule`

```
./../../../../../../../usr/lib/python3/dist-packages/pydocstyle/config.py:6
  Warning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working

../../../../../../.local/lib/python3.8/site-packages/_pytest/nodes.py:140
../../../../../../.local/lib/python3.8/site-packages/_pytest/nodes.py:140
../../../../../../.local/lib/python3.8/site-packages/_pytest/nodes.py:140
  Warning: The (fspath: py.path.local) argument to LaunchROSTestModule is deprecated. Please use the (path: pathlib.Path) argument instead.
  See https://docs.pytest.org/en/latest/deprecations.html#fspath-argument-for-node-constructors-replaced-with-pathlib-path
```

using

```
pip3 install pytest==7.0 && colcon test --event-handlers console_direct+ --packages-select ros2param
```

I think filtering the `fspath` argument if it's passed to the constructor may fix it as done here

https://github.com/ros2/launch/pull/592/files#diff-ae6811e267c65bf66fd7dfb5d9142aa58f70d41c354ba3c652078ee77ea2c271R140-R144

```python
       if _pytest_version_ge(7):
            if 'fspath' in kwargs:
                if kwargs['fspath'] is not None:
                    kwargs['path'] = pathlib.Path(kwargs['fspath'])
                del kwargs['fspath']
```

clalancette:
Yep, you are right.  I also like your solution better in #592.  Closing this.

Issue #469, opened 2020-10-31
eval substitution errors


Thodoris1999:
## Bug report

**Required Info:**

- Operating System:
  Ubuntu 20.04
- Installation type:
 Binaries
- Version or commit hash:
  foxy latest release
- DDS implementation:
  Fast-RTPS
- Client library (if applicable):
  N/A

#### Steps to reproduce issue
Create and launch a xml launch file like this
```
<launch>
  <let name="variable" value="val1"/>
  <node pkg="demo_nodes_cpp" exec="talker if="$(eval variable == 'val1')" output="screen" />
</launch>
```

#### Expected behavior
talker starts running

#### Actual behavior
I get this error: `TypeError: eval substitution expects 1 argument`

#### Additional information
It seems that spaces confuse the `eval` substitution. Even if I remove the spaces, launch fails with
```
launch.invalid_launch_file_error.InvalidLaunchFileError: Caught exception when trying to load file of format [xml]: No terminal defined for ''' at line 1 col 20

$(eval variable=='val1')
                 ^

Expecting: {'UNQUOTED_RSTRING', 'SPACE', 'RPAR', 'DOLLAR'}
```
which I guess means that it doesn't like the single quote character

hidmic:
> I get this error: TypeError: eval substitution expects 1 argument

That's because the Python expression substitution [expects a single argument](https://github.com/ros2/launch/blob/d648e19590d425e5d38411fe83fdb3b59f56d586/launch/launch/substitutions/python_expression.py#L52-L56). Making the argument a single string and escaping single quotes accordingly should work:

```xml
<launch>
  <let name="variable" value="val1"/>
  <node pkg="demo_nodes_cpp" exec="talker if="$(eval 'variable ==\ 'val1\'')" output="screen" />
</launch>
```

You may also use double quotes via XML escaping, but it's a bit clunky (i.e. `&quot;`).

> if I remove the spaces, launch fails with

Hmm, that looks like a bug to me.

Thodoris1999:
> > I get this error: TypeError: eval substitution expects 1 argument
> 
> That's because the Python expression substitution [expects a single argument](https://github.com/ros2/launch/blob/d648e19590d425e5d38411fe83fdb3b59f56d586/launch/launch/substitutions/python_expression.py#L52-L56). Making the argument a single string and escaping single quotes accordingly should work:
[...]

PR #600, opened 2022-03-02
Extend PythonExpression substitution to simplify the notation

kenji-miyake:
Related to #469, I supported the notation of `$(eval $(var value) == 'abc')`.
I'm not sure whether I should support `$(eval value == 'abc')` as well, but I guess it's a bit difficult because the current system seems to automatically drop `''`.

I'll send this Draft PR in order to get reviews from maintainers.


kenji-miyake:
@ivanpauno Hello, would you kindly review this, please? :pray:

launch/launch/substitutions/python_expression.py
```diff
@@ -68,5 +78,18 @@ def describe(self) -> Text:
 
     def perform(self, context: LaunchContext) -> Text:
         """Perform the substitution by evaluating the expression."""
-        from ..utilities import perform_substitutions
-        return str(eval(perform_substitutions(context, self.expression), {}, math.__dict__))
+        expressions = [perform_substitutions(context, [exp]) for exp in self.expression]
+
+        if len(expressions) == 3 and (expressions[1] in ['==', '!=']):
```
hidmic, line 83: @kenji-miyake I understand it's the easy path forward, but we can't merge a barely-implemented feature. IMHO, if we want `$(eval)` to be able to use launch configurations as variables, we need to inject launch configurations in `eval()`'s  local scope.

CC @ivanpauno @wjwwood

hidmic (changes requested):

launch/launch/substitutions/python_expression.py
ivanpauno: > we need to inject launch configurations in eval()'s local scop

Yeah that sounds better, the current proposal is too ad-hoc.

> I'm not sure whether I should support $(eval value == 'abc') as well, but I guess it's a bit difficult because the current system seems to automatically drop ''.

Maybe this is a problem (?).
We probably need to modify the launch substitutions grammar to make it more flexible, which will require a bit of thinking.

kenji-miyake: Thank you for your reviews!

> we need to inject launch configurations in eval()'s local scop

Is there any sample code that does this? Unfortunately, I'm not so familiar with the design concept of `launch`. :cry: 

Anyway, I'll drop expanding LaunchConfiguration from this PR for now.

kenji-miyake: Reverted by 4cc38be12994f5933c3f54f07e58f0468cb9e6f2.
@hidmic @ivanpauno Would you please teach me what else should I change? :pray:

kenji-miyake: @hidmic @ivanpauno Friendly ping. Could you review this again? :bow:

launch/launch/substitutions/python_expression.py
```diff
@@ -47,15 +50,15 @@ def __init__(self, expression: SomeSubstitutionsType) -> None:
...
-        from ..utilities import normalize_to_list_of_substitutions
         self.__expression = normalize_to_list_of_substitutions(expression)
 
     @classmethod
     def parse(cls, data: Iterable[SomeSubstitutionsType]):
         """Parse `PythonExpression` substitution."""
-        if len(data) != 1:
-            raise TypeError('eval substitution expects 1 argument')
-        return cls, {'expression': data[0]}
+        if len(data) == 1:
```
hidmic, line 58: @kenji-miyake we should still catch `len(data) == 0` and raise an exception.

launch/launch/substitutions/python_expression.py
hidmic: @kenji-miyake my apologies for the delay. I still think this is special casing to workaround the fact that, as @ivanpauno  pointed out, the substitution grammar wasn't designed to deal with arbitrary Python expressions. What if an expression is supposed to expand to, the name of some constant like `pi`? That implicit quoting you're adding in line 79 would block that. 

IMO to have this feature we need to revisit that grammar. And that's separate from simplifying `$(var name)` to `name` when used within `$(eval ...)`.

kenji-miyake: @hidmic Thank you for your comment!

> What if an expression is supposed to expand to, the name of some constant like pi?

Hmm, I didn't care about the case. But I guess just adding an eval with try-cache can support it? :thinking: 

> IMO to have this feature we need to revisit that grammar.

Yes, I agree with that. (I've changed this to a draft for now.)
How can I build a consensus on the syntax and move things forward?
Are there any design concept documents or plans for the eval syntax?

kenji-miyake: Or, how about adding `eq` and `ne` substitutions like #598?

```
$(eq $(var foo) 'bar')
$(ne $(var foo) 'bar')
```

launch/launch/substitutions/python_expression.py
kenji-miyake: Fixed in 205e7ccc04577d75098538ec916aae419ca58926.
And added a test.

launch/launch/substitutions/python_expression.py
hidmic: @kenji-miyake first of all, my apologies for the insane delay. I've been uber busy.

Now, circling back to this.

> How can I build a consensus on the syntax and move things forward?

A concrete proposal is the way to go. You want to simplify notation by making `$(var name)` and `name` equivalent in `$(eval ...)` context. We agree with the intent, but we need a fully functional implementation i.e. not just for a few operators. One possible approach is to inject launch configurations in `eval` scope. 

That doesn't solve the quoting issue though. In `launch` substitutions' grammar, arguments are delimited by spaces and so you need quotes to pass a single argument that includes a space e.g. `'/my/very/unhelpful path'`. Quotes must be dropped in these cases. When an `$(eval)` expression is not quoted, the parser will try to split `expression` into space delimited arguments, potentially dropping quotes from string literals, and the rest follows.

The only easy way I see we could avoid this is by instructing the substitution parser to skip the entire `$(eval)` expression. That means, no nested substitutions, tracking left and right parenthesis, and coming up with a special escaping mechanism for parenthesis so that something like `$(anon $(eval data() + ')'))` doesn't fall apart. 

> Are there any design concept documents or plans for the eval syntax?

IIRC there are no design documents about the syntax itself. Only the [grammar](https://github.com/ros2/launch/blob/master/launch/share/launch/frontend/grammar.lark).

> how about adding `eq` and `ne` substitutions 

That's definitely a simpler (better?) path forward.

kenji-miyake: @hidmic Hi, welcome back! No problem, I know ROS maintainers are quite busy. :cry: 
And thank you for your detailed explanation. But it's super difficult for me... I have to learn the grammar a lot.

> That's definitely a simpler (better?) path forward.

Oh, it's good to hear that. Can I send a new PR to add `eq/ne`?

kenji-miyake: `eq/ne` seem to be implemented in https://github.com/ros2/launch/pull/649.

kenji-miyake:
I believe I can close this thanks to https://github.com/ros2/launch/pull/649.

Issue #501, opened 2021-04-03
Unexpected value of launch argument in 'on_exit' clause


siteks:
## Bug report

- Ubuntu 20.04 ROS2 Foxy
- Installed from package binaries:

#### Steps to reproduce issue

```
# test1.launch.py
from launch import LaunchDescription
from launch.actions import IncludeLaunchDescription
from launch.launch_description_sources import PythonLaunchDescriptionSource

def generate_launch_description():
    ld = LaunchDescription()
    for r in ['r1', 'r2']:
        ld.add_action(IncludeLaunchDescription(
            PythonLaunchDescriptionSource('test2.launch.py'),
                launch_arguments = {'r' : r}.items()
        ))
    return ld

# test2.launch.py
from launch import LaunchDescription
from launch.actions import ExecuteProcess
from launch.actions import DeclareLaunchArgument
from launch.substitutions import LaunchConfiguration

def generate_launch_description():
    ld = LaunchDescription()
    ld.add_action(DeclareLaunchArgument('r'))
    ld.add_action(
        ExecuteProcess(
            cmd     = ['echo', 'first', LaunchConfiguration('r')],
            output  = 'screen',
            on_exit = ExecuteProcess(
                cmd     = ['echo', 'second', LaunchConfiguration('r')],
                output  = 'screen'
            )
        )
    )
    return ld


ros2 launch test1.launch.py
```

#### Expected behavior
echo commands print:
```
first r1
first r2
second r1
second r2
```
#### Actual behavior
```
[INFO] [echo-1]: process started with pid [1686]
[INFO] [echo-1]: process has finished cleanly [pid 1686]
[INFO] [echo-2]: process started with pid [1688]
[INFO] [echo-2]: process has finished cleanly [pid 1688]
[echo-1] first r1
[echo-2] first r2
[INFO] [echo-3]: process started with pid [1690]
[INFO] [echo-3]: process has finished cleanly [pid 1690]
[echo-3] second r2
[INFO] [echo-4]: process started with pid [1692]
[INFO] [echo-4]: process has finished cleanly [pid 1692]
[echo-4] second r2
```
#### Additional information
[...]

PR #601, opened 2022-03-03
Support scoping environment variables

### LINKED ISSUE #597: Support scoping environment variables in GroupAction
#### Feature description

Currently, the `SetEnvironmentVariable` action applies globally. Consider the following launch file:

```python
import launch
from launch.actions import GroupAction
from launch.actions import ExecuteProcess
from launch.actions import SetEnvironmentVariable


        SetEnvironmentVariable(name='my_env_var', value='1'),
        ExecuteProcess(cmd=['echo', '$my_env_var'], output='screen', shell=True),
        GroupAction([
            SetEnvironmentVariable(name='my_env_var', value='2'),
        ]),
    ])
```

Though I would expect it to output `1`, it actually outputs `2`.

My expectation is similar to what is described in the [XML launch file design doc](https://design.ros2.org/articles/roslaunch_xml.html#set_env-tag). Note, front-end launch files inherit the same behavior has the above Python example.

FWIW, I noticed that using the `EnvironmentVariable` substitution works as expected. For example,

```python
import launch
from launch.actions import GroupAction
from launch.actions import ExecuteProcess
from launch.actions import SetEnvironmentVariable
from launch.substitutions import EnvironmentVariable


        SetEnvironmentVariable(name='my_env_var', value='1'),
        ExecuteProcess(cmd=['echo', EnvironmentVariable('my_env_var')], output='screen')
        GroupAction([
            SetEnvironmentVariable(name='my_env_var', value='2'),
        ]),
    ])
```

outputs `1` as I would expect. So, it seems the issue is specific to processes.

jacobperron:
Resolves #597

Wherever actions were previously using os.environ, now get environment variables from the launch
context. There are a few exceptions where os.environ is still used since a launch context is not
available (e.g. [logging directory](https://github.com/ros2/launch/blob/7d0d4b41208b5f3d3d6d45f3297c37e7382f83dc/launch/launch/logging/__init__.py#L59-L61) and [overriding launch process output](https://github.com/ros2/launch/blob/7d0d4b41208b5f3d3d6d45f3297c37e7382f83dc/launch/launch/actions/execute_local.py#L194)).

Maintain a stack of environment variables in the launch context, similar to launch configurations.
Add new actions PushEnvironment and PopEnvironment to interact with the stack.
Push and pop the environment inside the GroupAction to support scoping.

Add a new ResetEnvironment action that sets the contexts environment
back to it's initial state (os.environ).
Return ResetEnvironment as part of GroupAction when forwarding=False.

### DIFF
--- a/launch/launch/launch_context.py
+++ b/launch/launch/launch_context.py
@@ -63,7 +63,7 @@ class LaunchContext:
         self.__launch_configurations = {}  # type: Dict[Text, Text]
 
         self.__environment_stack = []  # type: List[Dict[Text, Text]]
-        self.__environment = dict(os.environ)  # type: Dict[Text, Text]
+        self._reset_environment()
 
         self.__is_shutdown = False
         self.__asyncio_loop = None  # type: Optional[asyncio.AbstractEventLoop]
@@ -168,6 +168,9 @@ class LaunchContext:
         if not self.__environment_stack:
             raise RuntimeError('environment stack unexpectedly empty')
         self.__environment = self.__environment_stack.pop()
+
+    def _reset_environment(self):
+        self.__environment = dict(os.environ)  # type: Dict[Text, Text]
 
     def _push_launch_configurations(self):
         self.__launch_configurations_stack.append(self.__launch_configurations.copy())
--- a/launch/launch/launch_context.py
+++ b/launch/launch/launch_context.py
@@ -21,6 +21,7 @@ from typing import Any
 from typing import Dict
 from typing import Iterable
 from typing import List  # noqa: F401
+from typing import Mapping
 from typing import Optional
 from typing import Text
 
@@ -62,8 +63,9 @@ class LaunchContext:
         self.__launch_configurations_stack = []  # type: List[Dict[Text, Text]]
         self.__launch_configurations = {}  # type: Dict[Text, Text]
 
-        self.__environment_stack = []  # type: List[Dict[Text, Text]]
-        self._reset_environment()
+        self.__environment_stack = []  # type: List[Mapping[Text, Text]]
+        # We will reset to this copy when "reset_environment" is called
+        self.__environment_reset = os.environ.copy()  # type: Mapping[Text, Text]
 
         self.__is_shutdown = False
         self.__asyncio_loop = None  # type: Optional[asyncio.AbstractEventLoop]
@@ -162,15 +164,15 @@ class LaunchContext:
         return AttributeDict(self._get_combined_locals())
 
     def _push_environment(self):
-        self.__environment_stack.append(self.__environment.copy())
+        self.__environment_stack.append(os.environ.copy())
 
     def _pop_environment(self):
         if not self.__environment_stack:
             raise RuntimeError('environment stack unexpectedly empty')
-        self.__environment = self.__environment_stack.pop()
+        os.environ = self.__environment_stack.pop()
 
     def _reset_environment(self):
-        self.__environment = dict(os.environ)  # type: Dict[Text, Text]
+        os.environ = self.__environment_reset.copy()
 
     def _push_launch_configurations(self):
         self.__launch_configurations_stack.append(self.__launch_configurations.copy())
@@ -186,9 +188,9 @@ class LaunchContext:
         return self.__launch_configurations
 
     @property
-    def environment(self) -> Dict[Text, Text]:
+    def environment(self) -> Mapping[Text, Text]:
         """Getter for environment variables dictionary."""
-        return self.__environment
+        return os.environ
 
     def would_handle_event(self, event: Event) -> bool:
         """Check whether an event would be handled or not."""
--- a/launch/launch/launch_context.py
+++ b/launch/launch/launch_context.py
@@ -16,6 +16,7 @@
 
 import asyncio
 import collections
+import os
 from typing import Any
 from typing import Dict
 from typing import Iterable
@@ -60,6 +61,9 @@ class LaunchContext:
 
         self.__launch_configurations_stack = []  # type: List[Dict[Text, Text]]
         self.__launch_configurations = {}  # type: Dict[Text, Text]
+
+        self.__environment_stack = []  # type: List[Dict[Text, Text]]
+        self.__environment = dict(os.environ)  # type: Dict[Text, Text]
 
         self.__is_shutdown = False
         self.__asyncio_loop = None  # type: Optional[asyncio.AbstractEventLoop]
@@ -157,8 +161,16 @@ class LaunchContext:
 
         return AttributeDict(self._get_combined_locals())
 
+    def _push_environment(self):
+        self.__environment_stack.append(self.__environment.copy())
+
+    def _pop_environment(self):
+        if not self.__environment_stack:
+            raise RuntimeError('environment stack unexpectedly empty')
+        self.__environment = self.__environment_stack.pop()
+
     def _push_launch_configurations(self):
-        self.__launch_configurations_stack.append(dict(self.__launch_configurations))
+        self.__launch_configurations_stack.append(self.__launch_configurations.copy())
 
     def _pop_launch_configurations(self):
         if not self.__launch_configurations_stack:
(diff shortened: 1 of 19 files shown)


jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16285/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10905/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16652/)

jacobperron:
I'm also interested in any objections to backporting this change (which was my plan). I think it should be fine, but I imagine there may be consequences I can't think of since we're changing how launch interacts with `os.environ`.

jacobperron:
This change breaks a test in `tracetools_launch` that was assuming launch would set `os.environ`, https://ci.ros2.org/job/ci_linux-aarch64/10905/testReport/tracetools_launch.tracetools_launch.test.tracetools_launch.test_ld_preload_action/TestLdPreloadAction/test_action/

While it's easy to fix the test, it's an example of how this change may break assumptions if we backport it.

launch/launch/launch_context.py
```diff
@@ -170,6 +185,11 @@ def launch_configurations(self) -> Dict[Text, Text]:
         """Getter for launch_configurations dictionary."""
         return self.__launch_configurations
 
+    @property
+    def environment(self) -> Dict[Text, Text]:
+        """Getter for environment variables dictionary."""
+        return self.__environment
```
jacobperron, line 191: I suppose we could push/pop `os.environ` directly.. then it should be backwards compatible :thinking:

jacobperron:
Another test that breaks because of this change, see https://github.com/ros2/rcutils/pull/354

I'll see if I can modify this PR so that these tests don't break.

jacobperron:
Okay, with [3d2a695](https://github.com/ros2/launch/pull/601/commits/3d2a6950cb0ae0424e273838968e2ff495ba68a3), we're now operating directly on `os.environ` so we don't need #354 or any changes to ros2_tracing.

In retrospect, I think this is much simpler and technically we don't need the `context.environment` abstraction (though I kind of like it). Let me know what you think.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16286/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10906/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16653/)

launch/launch/launch_context.py
jacobperron: see [3d2a695](https://github.com/ros2/launch/pull/601/commits/3d2a6950cb0ae0424e273838968e2ff495ba68a3)

hidmic (commented):
Code itself looks good, have next to no objections about it.

Now, as to whether I think we should do this or not. It seems to me that we're turning environment variables into a less feature-rich equivalent of launch configurations (ie. string values only). Implementation is nearly a duplicate. There's overlap in functionality between `let`/`var` and  `set_env`/`env`. 

I'm inclined to think that, if we want a means to force environment variables to a group of processes, we can use launch configurations like we do for node parameters and remapping rules. Special casing for environment variables will make it harder to properly solve scoping later on (which goes out the window when even handlers enter the room, see https://github.com/ros2/launch/issues/501).

wjwwood:
Can we discuss this at the next ROS 2 meeting and/or a dedicated meeting?

jacobperron:
@hidmic @wjwwood To try and summarize our offline discussion:

- There is a bug, likely due to a race, related to environment variables and launch subprocesses that is separate from the feature of scoping. I've captured it here: https://github.com/ros2/launch/issues/602
- We could decide that scoping environment variables is not supported in launch and be loud about it in the documentation, however users will likely expect it and it seems reasonable to support it.
- Reducing duplication with LaunchConfigurations would be nice, since maintaining a separate stack for environment variables will complicate implementation when we fix scoping in other situations (I think scoping only works for group actions at the moment). However, given that we are currently operating on the global `os.environ`, it can result in arguably confusing behavior.
    - Consider that existing code that calls `context.launch_configurations.clear()` would remove **all** environment variables. I would say we should recommend against doing this or override that method to ensure that the value for environment variables is reset to the state when launch started.
    - Maybe this is doable, I'm willing to spend a little time investigating reusing launch configurations.
- In the future, we would like to consider maintaining a separate set of environment variables instead of modifying `os.environ`.
    - The idea would be to only support new launch actions for modifying the environment passed to subprocesses.
    - Consider deprecating the existing actions, `SetEnvironmentVariable`, `UnsetEnvironmentVariable`, and `AppendEnvironmentVariable`.

Please provide any corrections to my summary or add important points I may have missed.

hidmic (approved):
@jacobperron LGTM (as per offline discussion) 

Mind to re-run CI on Windows? Too many test failures, not sure if unrelated.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16398/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11006/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16770/)

Blast545:
:man_farmer: I think this PR introduced some test regressions in the repeated jobs buildfarm:

[launch.test.launch.actions.test_append_environment_variable.test_append_environment_variable_execute[2-3]](https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/2602/testReport/junit/launch.test.launch.actions/test_append_environment_variable/test_append_environment_variable_execute_2_3_/)
[launch.test.launch.actions.test_append_environment_variable.test_append_environment_variable_execute[3-3]](https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/2602/testReport/junit/launch.test.launch.actions/test_append_environment_variable/test_append_environment_variable_execute_3_3_/)
[launch.test.launch.actions.test_group_action.test_group_action_execute[2-3]](https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/2602/testReport/junit/launch.test.launch.actions/test_group_action/test_group_action_execute_2_3_/)
[launch.test.launch.actions.test_group_action.test_group_action_execute[3-3]](https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/2602/testReport/junit/launch.test.launch.actions/test_group_action/test_group_action_execute_3_3_/)
[launch.test.launch.actions.test_set_and_unset_environment_variable.test_set_and_unset_environment_variable_execute[2-3]](https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/2602/testReport/junit/launch.test.launch.actions/test_set_and_unset_environment_variable/test_set_and_unset_environment_variable_execute_2_3_/)
[launch.test.launch.actions.test_set_and_unset_environment_variable.test_set_and_unset_environment_variable_execute[3-3]](https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/2602/testReport/junit/launch.test.launch.actions/test_set_and_unset_environment_variable/test_set_and_unset_environment_variable_execute_3_3_/)

First time appearing: https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/2602/
Failing constantly since then.
AFAIK @sloretz is looking into this.

jacobperron:
Backport PRs are open:
- Foxy: https://github.com/ros2/launch/pull/630
- Galactic: https://github.com/ros2/launch/pull/631

launch/test/launch/actions/test_push_and_pop_environment.py
```diff
@@ -0,0 +1,81 @@
...
+    PopEnvironment()
+    PushEnvironment()
+
+
+def test_push_and_pop_environment_execute():
+    """Test the execute() of the PopEnvironment and PushEnvironment classes."""
+    context = LaunchContext()
+
+    # does not change empty state
+    context.environment.clear()
```
jacobperron, line 33: This modifies `os.environ`, which can negatively effect other tests in the same pytest process relying on it.
This bug was discovered and fixed in https://github.com/ros2/launch/pull/652

PR #603, opened 2022-03-15
Fix awaiting shutdown in launch context

jacobperron:
The method '_shutdown' is not a coroutine, and optionally returns a coroutine.
Guard against the case where it returns None, otherwise we can get the following error from launch:

    TypeError: object NoneType can't be used in 'await' expression

I noticed this error when a launch action raises an error during shutdown.

### DIFF
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -399,11 +399,7 @@ class LaunchService:
                 self.__context.emit_event_sync(shutdown_event)
             elif self.__loop_from_run_thread == asyncio_event_loop:
                 # If in the thread of the loop.
-                async def emit_event_async():
-                    nonlocal shutdown_event
-                    return await self.__context.emit_event(shutdown_event)
-
-                return emit_event_async
+                retval = self.__context.emit_event(shutdown_event)
             else:
                 # Otherwise in a different thread, so use the thread-safe method.
                 self.emit_event(shutdown_event)
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -290,7 +290,10 @@ class LaunchService:
                     # the queue
                     is_idle = self._is_idle()  # self._entity_future_pairs is pruned here
                     if not self.__shutting_down and shutdown_when_idle and is_idle:
-                        ret = await self._shutdown(reason='idle', due_to_sigint=False)
+                        coroutine = self._shutdown(reason='idle', due_to_sigint=False)
+                        ret = None
+                        if coroutine is not None:
+                            ret = await coroutine()
                         assert ret is None, ret
                         continue
 
@@ -343,7 +346,10 @@ class LaunchService:
                     msg = 'Caught exception in launch (see debug for traceback): {}'.format(exc)
                     self.__logger.debug(traceback.format_exc())
                     self.__logger.error(msg)
-                    ret = await self._shutdown(reason=msg, due_to_sigint=False)
+                    coroutine = self._shutdown(reason=msg, due_to_sigint=False)
+                    ret = None
+                    if coroutine is not None:
+                        ret = await coroutine()
                     assert ret is None, ret
                     return_code = 1
                     # keep running to let things shutdown properly
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -399,7 +399,11 @@ class LaunchService:
                 self.__context.emit_event_sync(shutdown_event)
             elif self.__loop_from_run_thread == asyncio_event_loop:
                 # If in the thread of the loop.
-                retval = self.__context.emit_event(shutdown_event)
+                async def emit_event_async():
+                    nonlocal shutdown_event
+                    return await self.__context.emit_event(shutdown_event)
+
+                return emit_event_async
             else:
                 # Otherwise in a different thread, so use the thread-safe method.
                 self.emit_event(shutdown_event)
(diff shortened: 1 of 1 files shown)


jacobperron:
CI for packages including and above launch:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16325/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10945/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16696/)

wjwwood (approved):
lgtm

It's obvious from the implementation of shutdown that sometimes it returns none, so it makes sense that this code needs to handle that case:

https://github.com/ros2/launch/blob/72968c789730e72b34f56fbd70becebb0cfbf4d5/launch/launch/launch_service.py#L386-L408

I don't know why this is happening only now. I suppose it happens in the exception handling case, like when something within launch or a launch action throws unexpectedly, so you don't see it in normal operation.

I suppose you could also mark the `self._shutdown` method as `async`, but I'm not sure what the consequence of that would be without doing more digging.

jacobperron:
I don't think we can mark `_shutdown` as async without also changing these call sites to use `await`:

https://github.com/ros2/launch/blob/8adf7deccd47eb81d773acf7d2f3c285c1c04bd2/launch/launch/launch_service.py#L196-L198


https://github.com/ros2/launch/blob/8adf7deccd47eb81d773acf7d2f3c285c1c04bd2/launch/launch/launch_service.py#L418-L421

jacobperron:
I don't think this PR actually works.. it sometimes results in RuntimeWarnings like,
```
launch_service.py:349: RuntimeWarning: coroutine 'LaunchContext.emit_event' was never awaited
```

jacobperron:
`_shutdown` was calling a coroutine (not awaiting it). Fixed in [7a7d556](https://github.com/ros2/launch/pull/603/commits/7a7d5569b578cd6c7277c17c2e59ce09cbb80b2b).
I'm not sure how this was working before. It seems like it was implicitly being awaiting because we were calling `await self._shutdown()` :confused:

launch/launch/launch_service.py
```diff
@@ -290,7 +290,10 @@ def _on_exception(loop, context):
                     # the queue
                     is_idle = self._is_idle()  # self._entity_future_pairs is pruned here
                     if not self.__shutting_down and shutdown_when_idle and is_idle:
-                        ret = await self._shutdown(reason='idle', due_to_sigint=False)
+                        coroutine = self._shutdown(reason='idle', due_to_sigint=False)
+                        ret = None
+                        if coroutine is not None:
+                            ret = await coroutine()
```
ivanpauno, line 296: shouldn't it be
```suggestion
                            ret = await coroutine
```
?

ivanpauno (commented):
@jacobperron do you have an example where the issue can be reproduced?

jacobperron:
@ivanpauno this launch file reproduces the original issue:

```
import time
import launch

class MyActionRaisesOnShutdown(launch.action.Action):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)

    def execute(self, context):
        context.add_completion_future(
            context.asyncio_loop.run_in_executor(
                None, self._foo, context
            )
        )

    def _foo(self, context):
        while True:
            if context.is_shutdown:
                raise RuntimeError('Error from MyActionRaisesOnShutdown')
            time.sleep(1.0)

        MyActionRaisesOnShutdown(),
    ])
```

launch/launch/launch_service.py
jacobperron: From my search online, I think you await a _coroutine_ with parentheses (it may take arguments) and you await a _task_ without the parentheses.

Running the following code seems to confirm part of that:

```
import asyncio

async def bar():
    print('bar')

async def foo():
    await bar

asyncio.run(foo())
```

Results in an error:

```
TypeError: object function can't be used in 'await' expression
```

jacobperron:
This is breaking downstream tests using `launch_service.shutdown`; looking into it.

launch/launch/launch_service.py
ivanpauno: `bar` and `foo` are a coroutine function, that return a coroutine object.
You need to await coroutines, and call coroutine functions before awaiting on the coroutine object.

From python [docs](https://docs.python.org/3/library/asyncio-task.html#awaitables):

> Important In this documentation the term “coroutine” can be used for two closely related concepts:
> a coroutine function: an [async def](https://docs.python.org/3/reference/compound_stmts.html#async-def) function;
> 
> a coroutine object: an object returned by calling a coroutine function.

---

Before the last [commit](https://github.com/ros2/launch/pull/603/commits/7a7d5569b578cd6c7277c17c2e59ce09cbb80b2b), `_shutdown()` was returning a coroutine object, so I think `ret = await coroutine` was the correct thing (probably, calling the coroutine object was generated the warning).
After that commit we're returning a coroutine function, so `ret = await coroutine()` is now fine

jacobperron: Ah, thanks for the clarification! That explains the odd behavior I was seeing.  Sorry, I'm not that familiar with asyncio.

I've reverted [7a7d556](https://github.com/ros2/launch/pull/603/commits/7a7d5569b578cd6c7277c17c2e59ce09cbb80b2b) and avoided calling the coroutine ([8a47880](https://github.com/ros2/launch/pull/603/commits/8a47880eb0ce51ff1192655d676b0b9acbdc5184)),

PR #605, opened 2022-04-03
Fix sphinx directive to cross-ref Launch method

aprotyas:
This PR is required for a clean rosdoc2 build of launch_yaml and
launch_xml.

`:ref:meth:` is not a valid Sphinx directive to cross-reference python
methods. This has been replaced with `:py:meth:`, as suggested in the
Sphinx documentation. Reference:
https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#role-py-meth

Signed-off-by: Abrar Rahman Protyasha <aprotyas@u.rochester.edu>

### DIFF
--- a/launch_xml/launch_xml/entity.py
+++ b/launch_xml/launch_xml/entity.py
@@ -86,7 +86,7 @@ class Entity(BaseEntity):
         """
         Access an attribute of the entity.
 
-        See :ref:meth:`launch.frontend.Entity.get_attr`.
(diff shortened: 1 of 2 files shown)


clalancette (approved):

clalancette:
Since this is just a change to documentation, I'm going to go ahead and merge this with just the Rpr job and without full CI.  Thanks for the contribution!

PR #607, opened 2022-04-06
expose sigterm_timeout and sigkill_timeout to frontend

wjwwood:
Previously, these options were not accessible to the frontends like XML.

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -346,7 +346,7 @@ class ExecuteProcess(ExecuteLocal):
             if sigterm_timeout is not None:
                 if sigterm_timeout < 0.0:
                     raise ValueError(
-                        'Attribute sigterm_timeout of Entity node expected to be '
+                        'Attribute sigterm_timeout for a process to be executed expected to be '
                         'a non-negative value but got `{}`'.format(sigterm_timeout)
                     )
                 kwargs['sigterm_timeout'] = sigterm_timeout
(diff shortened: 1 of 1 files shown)


ivanpauno (approved):
LGTM!
Though I guess we want to merge this after we branch off Humble.

It would also be nice to add some test cases.

hidmic:
> It would also be nice to add some test cases.

+1 to adding some tests

wjwwood:
How do you guys recommend I test it in `launch`? I was going to use some XML snippet to test it, but I don't have `launch_xml` or others available at that point. I don't see any other tests in `launch` that do something like this yet. Do you guys know of some?

hidmic:
@wjwwood yeah, frontend tests for entities are in frontend packages right now (e.g. [here](https://github.com/ros2/launch/blob/master/launch_xml/test/launch_xml/test_executable.py)). 

Arguably, that's not very nice. It's not even necessary. I guess it was easy at the time (easier than building markup-agnostic testing infrastructure).

wjwwood:
I can just add some tests there. But (for time travelers from the future) it would make sense to move the testing of these "exposures" into `launch` and being XML/YAML agnostic some how.

hidmic:
@wjwwood do you still want to add some tests and merge this?

wjwwood:
I would like to, but it's just been sitting in my open tabs waiting for me to have some time to do it.

adityapande-1995:
Closing this in favour of https://github.com/ros2/launch/pull/667

Issue #582, opened 2022-01-13
Make hardcoded 15 second launch_testing timeout configurable


Aposhian:
#### Feature description
Currently you only get [15 seconds](https://github.com/ros2/launch/blob/b59bb6b88b40eaa82870542d93bdfb1a033d05c3/launch_testing/launch_testing/test_runner.py#L184) to signal `ReadyToTest`. This is not configurable, which means that you can't use processes that take a long time to startup.

#### Implementation considerations
Ideally this could be configured at the launch file level. Maybe the tests themselves could say how long they are willing to wait for `ReadyToTest`?

clalancette:
Yeah, agreed that this would be nice to make configurable.  Please consider making a pull request to do so, we're happy to review.

jacobperron:
I also hit this hard-coded limit recently. It's easy to reproduce with the following launch test file:

```py
import launch
from launch.actions import TimerAction
import launch_testing.markers
from launch_testing.actions import ReadyToTest


def generate_test_description():
    return launch.LaunchDescription([
        TimerAction(period=16.0, actions=[ReadyToTest()]),
    ])
```

As a potential workaround, the alternative [`launch_pytest`](https://github.com/ros2/launch/tree/master/launch_pytest) package does not have such a limit. Here is the equivalent test file written with `launch_pytest`:

```py
import pytest

import launch
from launch.actions import TimerAction
import launch_pytest
from launch_pytest.actions import ReadyToTest


@launch_pytest.fixture
def launch_description():
    return launch.LaunchDescription([
        TimerAction(period=16.0, actions=[ReadyToTest()]),
    ])


@pytest.mark.launch(fixture=launch_description)
def test_case_1():
    pass
```

deepanshubansal01:
Closing the issue, since the feature has been implemented https://github.com/ros2/launch/pull/625

PR #622, opened 2022-06-21
Add timeout for ReadyToTest action

deepanshubansal01:
Related to #582 

Signed-off-by: deepanshu <deepanshubansal01@gmail.com>

### DIFF
--- a/launch_testing/launch_testing/launch_test.py
+++ b/launch_testing/launch_testing/launch_test.py
@@ -56,7 +56,10 @@ def add_arguments(parser):
         '--junit-xml', action='store', dest='xmlpath', default=None,
         help='Do write xUnit reports to specified path.'
     )
-
+    parser.add_argument(
+        '--timeout', type=float, default=15.0,
+        help='timeout for ReadyToTest action'
+    )
 
 def parse_arguments():
     parser = argparse.ArgumentParser(
@@ -92,7 +95,8 @@ def run(parser, args, test_runner_cls=LaunchTestRunner):
     runner = test_runner_cls(
         test_runs=test_runs,
         launch_file_arguments=args.launch_arguments,
-        debug=args.verbose
+        debug=args.verbose,
+        timeout=args.timeout
     )
 
     _logger_.debug('Validating test configuration')
--- a/launch_testing/launch_testing/launch_test.py
+++ b/launch_testing/launch_testing/launch_test.py
@@ -58,8 +58,11 @@ def add_arguments(parser):
     )
     parser.add_argument(
         '--timeout', type=float, default=15.0,
-        help='timeout for ReadyToTest action'
+        help='Maximum duration for which the ReadyToTest action waits for the processes'
+             'to start up, if processes take more time than the one specified by timeout'
+             'an error is thrown. By default timeout is 15 seconds.'
     )
+
 
 def parse_arguments():
     parser = argparse.ArgumentParser(
--- a/launch_testing/test/launch_testing/test_launch_test_runner_validation.py
+++ b/launch_testing/test/launch_testing/test_launch_test_runner_validation.py
@@ -167,3 +167,24 @@ class TestNewStyleTestDescriptions(unittest.TestCase):
 
         dut.validate()  # Make sure this passes initial validation (probably redundant with above)
         runs[0].normalized_test_description(ready_fn=lambda: None)
+
+    def test_launch_description_with_ready_action_changed_timeout(self):
+
+        def generate_test_description():
+            return launch.LaunchDescription([
+                launch.actions.TimerAction(
+                    period=17.0,  # takes 17 sec for this action to startup
+                    actions=[ReadyToTest()]
+                )
+            ])
+
+        runs = make_test_run_for_dut(generate_test_description)
+        dut = LaunchTestRunner(
+            test_runs=runs,
+            launch_file_arguments=[],
+            debug=False,
+            timeout=20.0  # increase the timeout giving the time for process to startup
+        )
+
+        dut.validate()  # Make sure this passes initial validation (probably redundant with above)
+        runs[0].normalized_test_description(ready_fn=lambda: None)
--- a/launch_testing/test/launch_testing/test_launch_test_runner_validation.py
+++ b/launch_testing/test/launch_testing/test_launch_test_runner_validation.py
@@ -33,7 +33,6 @@ def make_test_run_for_dut(generate_test_description_function):
 class TestLaunchTestRunnerValidation(unittest.TestCase):
 
     def test_catches_bad_signature(self):
-
         # If there's no ReadyToTest action, we won't catch that until later because dut.validate()
         # doesn't actually invoke the function.
         # We will still expect to reject functions with wrong name arguments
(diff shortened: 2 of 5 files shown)


adityapande-1995 (commented):
Can we have a test that uses multiple runners or Launch descriptions ? Probably a parameterized test here ? https://github.com/ros2/launch/blob/69964b72cf586be99b0b1d9fcfe8b68545123fff/launch_testing/test/launch_testing/test_launch_test_runner_validation.py#L157

launch_testing/test/launch_testing/test_launch_test_runner_validation.py
```diff
@@ -167,3 +167,24 @@ def generate_test_description(my_param):
...
+        runs = make_test_run_for_dut(generate_test_description)
+        dut = LaunchTestRunner(
+            test_runs=runs,
+            launch_file_arguments=[],
+            debug=False,
+            timeout=20.0  # increase the timeout giving the time for process to startup
+        )
+
+        dut.validate()  # Make sure this passes initial validation (probably redundant with above)
+        runs[0].normalized_test_description(ready_fn=lambda: None)
```
adityapande-1995, line 190: Flake8 needs the newline character here it seems.

deepanshubansal01:
> Reference

I can try that but is this test for checking if we can we have multiple runners/launch descriptions or for testing the timeout thing? 
Also for the params I am not sure if they are the same as cli optional arguments parsed using argsparse lib?

launch_testing/test/launch_testing/test_launch_test_runner_validation.py
```diff
@@ -153,7 +146,6 @@ def generate_test_description():
         runs[0].normalized_test_description(ready_fn=lambda: None)
 
     def test_parametrized_launch_description(self):
-
```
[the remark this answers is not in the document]
deepanshubansal01: Done

launch_testing/launch_testing/launch_test.py
```diff
@@ -56,7 +56,10 @@ def add_arguments(parser):
         '--junit-xml', action='store', dest='xmlpath', default=None,
         help='Do write xUnit reports to specified path.'
     )
-
+    parser.add_argument(
+        '--timeout', type=float, default=15.0,
+        help='timeout for ReadyToTest action'
```
jacobperron, line 61: I think we should clarify a few things in the help message:
- this is the duration between the start of the launch process and the ReadyToTest action being executed.
- what happens if the timeout occurs
- the units are seconds

launch_testing/test/launch_testing/test_launch_test_runner_validation.py
```diff
@@ -33,7 +33,7 @@ def make_test_run_for_dut(generate_test_description_function):
 class TestLaunchTestRunnerValidation(unittest.TestCase):
 
     def test_catches_bad_signature(self):
-
+
```
jacobperron, line 36: I guess linters won't like the extra whitespace introduced here.

launch_testing/test/launch_testing/test_launch_test_runner_validation.py
```diff
@@ -167,3 +167,23 @@ def generate_test_description(my_param):
 
         dut.validate()  # Make sure this passes initial validation (probably redundant with above)
         runs[0].normalized_test_description(ready_fn=lambda: None)
+
+    def test_launch_description_with_ready_action_changed_timeout(self):
```
jacobperron, line 171: I don't think this test is actually testing the timeout. It only performs validation of the launch description, but does not actually execute anything as far as I can tell.

I see a few other tests that exercise some of the other command line options, maybe we can do something similar for the `--timeout` option:

- https://github.com/ros2/launch/blob/master/launch_testing/test/launch_testing/test_print_arguments.py
- https://github.com/ros2/launch/blob/master/launch_testing/test/launch_testing/test_xml_output.py

jacobperron (commented):
The new option looks good to me :+1:

launch_testing/test/launch_testing/test_launch_test_runner_validation.py
deepanshubansal01: I see.

ivanpauno (changes requested):
IMO, it would be nice if the timeout is somehow specified in the launch file, instead of being an option for the as a command line argument.

Is there any alternative to make that happen?

deepanshubansal01:
> IMO, it would be nice if the timeout is somehow specified in the launch file, instead of being an option for the as a command line argument.
> 
> Is there any alternative to make that happen?

I tried making timeout configurable as part of launch file itself and having timeout as an argument to the `ReadyToTest` action. However, this is not possible with the current code base. This is because the `_run_test` thread which checks for `ReadyToTest` Action timeout is started in the `_RunnerWorker` class constructor and the `ReadyToTest` action itself is instantiated after the `_run_test` thread has been started which makes it futile to access the timeout attribute from `ReadyToTest` action class since the `_run_test` thread has already been started. 

The `_run_test` thread is started here: https://github.com/ros2/launch/blob/074cd2903ddccd61bce8f40a0f58da0b7c200481/launch_testing/launch_testing/test_runner.py#L60-L64
And the `ReadyToTest` object is instantiated here: https://github.com/ros2/launch/blob/074cd2903ddccd61bce8f40a0f58da0b7c200481/launch_testing/launch_testing/test_runner.py#L73-L75

deepanshubansal01:
> > IMO, it would be nice if the timeout is somehow specified in the launch file, instead of being an option for the as a command line argument.
> 
> How would the timeout parameter being added here being specified in practice? i.e. where would the users specify it? (users are not usually running launch_testing manually)

I am not sure about this right now the timeout can be passed through cli and manually to the `LaunchTestRunner` class. If the use case is not through cli then we might have to do something else? Passing it in launch file itself makes much more sense I guess as part of `ReadyToTest` action but I am not sure how can we do it with the current design.

ivanpauno:
Ideally, I think that a decorator would be a nice way to specify it:

```python3
@launch_testing.runner_processes_wait_timeout(5.3)
def generate_test_description(arg_param):
```

To have something like that, the decorator would have to add a hidden attribute to the decorated function (e.g. `generate_test_description.__runner_processes_wait_timeout`).
The loader could read that attribute, similar to how [parametrized](https://github.com/ros2/launch/blob/dcf180a0da309a1c24c210f56d381fc59ea1546a/launch_testing/launch_testing/loader.py#L164) launch descriptions are handled.
Each [TestRun](https://github.com/ros2/launch/blob/dcf180a0da309a1c24c210f56d381fc59ea1546a/launch_testing/launch_testing/loader.py#L77) could have the timeout as a property.
And then the runner would use `self._test_run.runner_processes_wait_timeout` instead of `15` [here](https://github.com/ros2/launch/blob/dcf180a0da309a1c24c210f56d381fc59ea1546a/launch_testing/launch_testing/test_runner.py#L184).

@deepanshubansal01 @jacobperron does that make sense to you?

ivanpauno:
@hidmic maybe you can provide more feedback of how to implement this

deepanshubansal01:
> @hidmic maybe you can provide more feedback of how to implement this

@ivanpauno should I proceed with the implementation that you suggested or wait for @hidmic inputs. Thanks

ivanpauno:
> @ivanpauno should I proceed with the implementation that you suggested or wait for @hidmic inputs.

I think it's fine to proceed with the implementation.

deepanshubansal01:
Closing the PR now, the feature will be implemented in this PR now https://github.com/ros2/launch/pull/625

Issue #545, opened 2021-10-27
launch not killing processes started with shell=True


v-lopez:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - binaries
- Version or commit hash:
  - Foxy 0.10.6-1focal.20210901.033555
- DDS implementation:
  - Fast
  - 
#### Steps to reproduce issue

```
import launch_testing
import unittest
import os

from launch import LaunchDescription
from launch.actions import DeclareLaunchArgument, ExecuteProcess
from launch.substitutions import PathJoinSubstitution, LaunchConfiguration
from launch_ros.actions import Node
from launch.actions import IncludeLaunchDescription
from launch.substitutions import PathJoinSubstitution
from launch_ros.substitutions import FindPackageShare
from launch.launch_description_sources import PythonLaunchDescriptionSource



def generate_test_description():
    # This is necessary to get unbuffered output from the process under test
    proc_env = os.environ.copy()
    proc_env['PYTHONUNBUFFERED'] = '1'

    dut_process = ExecuteProcess(cmd=['/usr/bin/sleep', '5']) 
    pkg_dir = FindPackageShare('gazebo_ros')
    full_path = PathJoinSubstitution([pkg_dir] + ['launch', 'gazebo.launch.py'])

    gazebo_launch = IncludeLaunchDescription(
        PythonLaunchDescriptionSource(
            full_path))

    return LaunchDescription([
        gazebo_launch,
        dut_process,

        launch_testing.actions.ReadyToTest(),
    ]), {'dut_process': dut_process}


class TestWait(unittest.TestCase):

    def test_wait_for_end(self, proc_output, proc_info, dut_process):
        # Wait until process ends
        proc_info.assertWaitForShutdown(process=dut_process, timeout=10)


@ launch_testing.post_shutdown_test()
class TestSuccessfulExit(unittest.TestCase):

    def test_exit_code(self, proc_info, dut_process):
        # Check that dut_process finishes with code 0
        launch_testing.asserts.assertExitCodes(proc_info, process=dut_process)
```

Execute it with `launch_test` or as part of `add_launch_test` in CMake.

#### Expected behavior
When it ends, gazebo should be killed.

#### Actual behavior
Gazebo is not killed.

#### Additional information
Happens with other nodes besides gazebo. But gazebo is particularly troublesome because I cannot have 2 tests running sequentially using gazebo, because the first test didn't kill the server, and only one server may be running in a machine.

[...]

PR #632, opened 2022-07-25
Kill dangling subprocesses

ivanpauno:
I used psutil to figure out children of a process recursively.
It's an easy way to handle this issue platform independently.

For posix OSs, we could send a signal to the process group, but for that we should create a new process group when launching a process, which I'm not sure if it's the best ideal.

### DIFF
--- a/launch/launch/actions/execute_local.py
+++ b/launch/launch/actions/execute_local.py
@@ -31,6 +31,8 @@ from typing import Text
 from typing import Tuple  # noqa: F401
 from typing import Union
 
+import psutil
+
 import launch.logging
 
 from osrf_pycommon.process_utils import async_execute_process
@@ -87,6 +89,8 @@ class ExecuteLocal(Action):
             'sigterm_timeout', default=5),
         sigkill_timeout: SomeSubstitutionsType = LaunchConfiguration(
             'sigkill_timeout', default=5),
+        sigkill_subprocesses_timeout: SomeSubstitutionsType = LaunchConfiguration(
+            'sigkill_subprocesses_timeout', default=5),
         emulate_tty: bool = False,
         output: SomeSubstitutionsType = 'log',
         output_format: Text = '[{this.process_description.final_name}] {line}',
@@ -158,6 +162,11 @@ class ExecuteLocal(Action):
             as a string or a list of strings and Substitutions to be resolved
             at runtime, defaults to the LaunchConfiguration called
             'sigkill_timeout'
+        :param: sigkill_subprocesses_timeout time until sending SIGKILL directly to dangling
+            subprocesses after sending SIGKILL to the process,
+            as a string or a list of strings and Substitutions to be resolved
+            at runtime, defaults to the LaunchConfiguration called
+            'sigkill_subprocesses_timeout'
         :param: emulate_tty emulate a tty (terminal), defaults to False, but can
             be overridden with the LaunchConfiguration called 'emulate_tty',
             the value of which is evaluated as true or false according to
@@ -188,6 +197,8 @@ class ExecuteLocal(Action):
         self.__shell = shell
         self.__sigterm_timeout = normalize_to_list_of_substitutions(sigterm_timeout)
         self.__sigkill_timeout = normalize_to_list_of_substitutions(sigkill_timeout)
+        self.__sigkill_subprocesses_timeout = normalize_to_list_of_substitutions(
+            sigkill_subprocesses_timeout)
         self.__emulate_tty = emulate_tty
         self.__output = os.environ.get('OVERRIDE_LAUNCH_PROCESS_OUTPUT', output)
         self.__output = normalize_to_list_of_substitutions(self.__output)
@@ -434,6 +445,11 @@ class ExecuteLocal(Action):
         sigkill_timeout = [PythonExpression(
             ('float(', *self.__sigterm_timeout, ') + float(', *self.__sigkill_timeout, ')')
         )]
+        sigkill_subprocesses_timeout = [PythonExpression(
+            (
+                'float(', *self.__sigterm_timeout, ') + float(', *self.__sigkill_timeout,
+                ') + float(', *self.__sigkill_subprocesses_timeout, ')')
+        )]
         # Setup a timer to send us a SIGTERM if we don't shutdown quickly.
         self.__sigterm_timer = TimerAction(
             period=sigterm_timeout,
@@ -464,9 +480,35 @@ class ExecuteLocal(Action):
             ],
             cancel_on_shutdown=False,
         )
+        def kill_subprocesses(
+            context,
+            timeout_substitutions,
+            children=psutil.Process(
+                self._subprocess_transport.get_pid()).children(recursive=True)
+        ):
+            process_name = context.locals.process_name
+            for p in children:
+                try:
+                    p.send_signal(signal.SIGKILL)
+                except psutil.NoSuchProcess:
+                    continue
+                self.__logger.warn(
+                    f'subprocess[pid={p.pid}] of process[{process_name}] was not terminated after '
+                    f"'{perform_substitutions(context, timeout_substitutions)}' seconds of parent "
+                    f"being killed. "
+                    'Sending SIGKILL to subprocess directly.'
+                )
+        self.__sigkill_subprocesses_timer = TimerAction(
+            period=sigkill_subprocesses_timeout,
+            actions=[OpaqueFunction(
+                function=kill_subprocesses,
+                args=(sigkill_subprocesses_timeout, ))],
+            cancel_on_shutdown=False,
+        )
         return [
             cast(Action, self.__sigterm_timer),
             cast(Action, self.__sigkill_timer),
+            cast(Action, self.__sigkill_subprocesses_timer),
         ]
 
     def __get_sigint_event(self):
--- a/launch/package.xml
+++ b/launch/package.xml
@@ -18,6 +18,7 @@
   <exec_depend>ament_index_python</exec_depend>
   <exec_depend>python3-importlib-metadata</exec_depend>
   <exec_depend>python3-lark-parser</exec_depend>
+  <exec_depend>python3-psutil</exec_depend>
   <exec_depend>python3-yaml</exec_depend>
 
   <test_depend>ament_copyright</test_depend>


launch/launch/actions/execute_local.py
```diff
@@ -434,6 +445,11 @@ def printer(context, msg, timeout_substitutions):
         sigkill_timeout = [PythonExpression(
             ('float(', *self.__sigterm_timeout, ') + float(', *self.__sigkill_timeout, ')')
         )]
+        sigkill_subprocesses_timeout = [PythonExpression(
+            (
+                'float(', *self.__sigterm_timeout, ') + float(', *self.__sigkill_timeout,
+                ') + float(', *self.__sigkill_subprocesses_timeout, ')')
+        )]
```
ivanpauno, line 452: I'm doing the same as the other timers are doing, but I think that this substitution should be resolved when the `ExecuteLocal` action is executed, and not when the timer is.

launch/launch/actions/execute_local.py
```diff
@@ -464,9 +480,35 @@ def printer(context, msg, timeout_substitutions):
...
+            children=psutil.Process(
+                self._subprocess_transport.get_pid()).children(recursive=True)
+        ):
+            process_name = context.locals.process_name
+            for p in children:
+                try:
+                    p.send_signal(signal.SIGKILL)
+                except psutil.NoSuchProcess:
+                    continue
+                self.__logger.warn(
```
ivanpauno, line 495: For the other cases (scalating SIGINT->SIGTERM and SIGTERM->SIGKILL), we're logging an error.
But I think a warning may be more appropriate, or even an INFO message.

hidmic:
@ivanpauno before commenting anything, why explicitly list all processes instead of using process groups (and whatever Windows has as an equivalent)?

ivanpauno:
> @ivanpauno before commenting anything, why explicitly list all processes instead of using process groups (and whatever Windows has as an equivalent)?

IIUC there's no equivalent to process groups in windows.
We could use them in posix-like OSs though

ivanpauno:
> @ivanpauno before commenting anything, why explicitly list all processes instead of using process groups (and whatever Windows has as an equivalent)?

I have a few reasons:

- There's no multiplatform way of doing that.
  - For linux/posix-like we could use https://docs.python.org/3/library/os.html#os.setpgrp to create a new group for the subprocess.
  - For windows, https://docs.python.org/3/library/subprocess.html#subprocess.CREATE_NEW_PROCESS_GROUP.
    - Ctrl-Break event only works for "console processes", see [here](https://docs.python.org/3/library/os.html#os.kill).
      I'm not sure if we can guarantee that launch will be run in a "console".
- A child process may run another subprocess with a different group id. If we want to kill all dangling subprocesses, I think this works better.

Maybe we shouldn't have this feature at all, and only log a warning if a "dangling" subprocess is detected.

launch/launch/actions/execute_local.py
jacobperron: +1 for consistency. If you want, you can propose a refactor in a follow-up PR.

launch/launch/actions/execute_local.py
jacobperron: A warning sounds okay to me.

jacobperron (approved):
Seems like a reasonable feature to me.

> Maybe we shouldn't have this feature at all, and only log a warning if a "dangling" subprocess is detected.

Is there a scenario you had in mind where we wouldn't want a subprocess to end?

launch/launch/actions/execute_local.py
```diff
@@ -464,9 +480,35 @@ def printer(context, msg, timeout_substitutions):
             ],
             cancel_on_shutdown=False,
         )
+        def kill_subprocesses(
+            context,
+            timeout_substitutions,
+            children=psutil.Process(
+                self._subprocess_transport.get_pid()).children(recursive=True)
```
wjwwood, line 487: Is it possible for `self._subprocess_transport.get_pid()` to return None or something like that? Maybe this should be broken into steps with proper error handling in place?

launch/launch/actions/execute_local.py
wjwwood: +1 for a warning.

launch/launch/actions/execute_local.py
```diff
@@ -464,9 +480,35 @@ def printer(context, msg, timeout_substitutions):
...
+            for p in children:
+                try:
+                    p.send_signal(signal.SIGKILL)
+                except psutil.NoSuchProcess:
+                    continue
+                self.__logger.warn(
+                    f'subprocess[pid={p.pid}] of process[{process_name}] was not terminated after '
+                    f"'{perform_substitutions(context, timeout_substitutions)}' seconds of parent "
+                    f"being killed. "
+                    'Sending SIGKILL to subprocess directly.'
```
wjwwood, line 499: Suggested rewording:

```suggestion
                    f'subprocess[pid={p.pid}] of process[{process_name}] did not terminate after '
                    f"'{perform_substitutions(context, timeout_substitutions)}' seconds since the parent "
                    f"was killed. "
                    'Sending SIGKILL to subprocess directly.'
```

wjwwood (commented):
Do we not need to wait for those children to exit?

Some like this example:

https://psutil.readthedocs.io/en/latest/index.html?highlight=children#kill-process-tree

Where it uses `psutil.wait_procs`?

It would be nice to notify the user in the launch output if the child process still did not exit after some period of time.

Furthermore, does this timer always wait until it expires and then check the subprocesses, or does it only do this if the parent process has to be sent SIGTERM/SIGKILL?

Also, what happens when all the subprocess exit cleanly and quickly? Does this timer still wait and then check them?

It feels like we're missing something here where we await these processes exiting and only send them signals if they don't exit.

ivanpauno:
> Where it uses psutil.wait_procs?
> 
> It would be nice to notify the user in the launch output if the child process still did not exit after some period of time.

I don't think we can block when executing an action, so I would need another timer to check if the processes were killed.
Anyway, I'm sending SIGKILL, which cannot be ignored.

> Furthermore, does this timer always wait until it expires and then check the subprocesses, or does it only do this if the parent process has to be sent SIGTERM/SIGKILL?

It always sends the signals to all subprocesses, event if SIGTERM/SIGKILL was not needed.

> Also, what happens when all the subprocess exit cleanly and quickly? Does this timer still wait and then check them?

Yes, and "process does not exist" errors are ignored when trying to kill the subprocesses that were previously detected in the tree for this reason.
I'm counting with the OS not reassinging the same pid shortly after.

> It feels like we're missing something here where we await these processes exiting and only send them signals if they don't exit.

Do you mean we should kill subprocesses going "level by level"?
i.e. first kill the process launched with `ExecuteLocal`.
  After waiting that process was killed, kills its subprocesses that are still running.
  Is that your idea?
  Should this be repeated recursively or only for one level?

wjwwood:
> I don't think we can block when executing an action, so I would need another timer to check if the processes were killed.

We can await the processes to exit, either with a thread that sets a future, or if possible using existing asyncio subprocess stuff.

> Anyway, I'm sending SIGKILL, which cannot be ignored.

It cannot be _caught_, but it doesn't mean that it will definitely exit, or how long that will take. I've definitely had some processes fail to exit on `kill -9` before, but usually the system is borked then. But the point would be to notify the user (so they don't have to go crawling to ps to see what stayed around), not to do anything else, since there's nothing else we could do after SIGKILL.

> It always sends the signals to all subprocesses, event if SIGTERM/SIGKILL was not needed.

> Yes, and "process does not exist" errors are ignored when trying to kill the subprocesses that were previously detected in the tree for this reason.

My point was more about the waiting. If the default timeout of `sigkill_subprocesses_timeout` is like 5 seconds, but all the processes exit immediately? Do we wait 5 seconds, send a bunch of signals we don't need to send, then finally exit?

>  I'm counting with the OS not reassinging the same pid shortly after.

Is that a safe assumption?

> Do you mean we should kill subprocesses going "level by level"?
i.e. first kill the process launched with ExecuteLocal.
After waiting that process was killed, kills its subprocesses that are still running.
Is that your idea?

No, that's not what I'm suggesting.

Instead I was thinking something like this:

- send SIGINT to the "main process" launch started
- collect all children recursively, and start monitoring their status (checking for termination)
- wait for the main process to exit, or for SIGTERM timeout
- if SIGTERM timeout, escalate to SIGKILL for main process and wait
- when the main process exits or it fails to even after SIGTERM and some time, check all child processes
- SIGTERM each child process
- await the processes, if they don't exit SIGKILL the ones alive
- when all have exited, or some time has passed after SIGKILL, finish
- log an error in launch for each process that never exited, even after SIGKILL

This process is good in my opinion because it:

- exits as soon as all processes have finished
- ensures all processes finish, or else logs that they don't
- avoids jumping to SIGKILL untill absolutely needed
- avoids trying to terminate the child processes until the main process has exited or until it failed to exit after SIGKILL
- starts tracking the child process statuses before trying to terminate them, so (maybe) you don't have to worry about the OS reassigning them, because you'll be notified (via an await or a thread waiting on them with `psutil` or similar) when they exit

Obviously the escalation process for the main process and the child processes are the same, so maybe we can generalize that, so it can be used in `ExecuteLocal` as well as with any individual process we have a pid for.

What I outlined is a bit more work to implement, but I also think it's a lot more thorough and is likely to save us (and our users) some time in the future debugging these kind of issues.

wjwwood:
> Do you mean we should kill subprocesses going "level by level"?

Another comment about this.

I wasn't thinking that, I was still thinking doing it in a batch operation, but I also don't think doing it layer by layer is a bad idea, though it might be really annoying if it takes a long time.

The reason I like it as an idea is that I want to give the called processes every chance to behave, and doing a shutdown escalation level by level (top to bottom) is the best way to do that.

I could see this as something of a configuration. If we decide to implement what I outlined above, then making it possible to do it step by step instead of in batch should be easy-ish to do. We just need to keep that in mind when working on it.

ivanpauno:
> Is that a safe assumption?

I'm not sure. If that's not a safe assumption, then we cannot do anything.
The best we can do is to notify the user "it looks like a launched process left some subprocesses alive after being killed: {pids}".

> starts tracking the child process statuses before trying to terminate them, so (maybe) you don't have to worry about the OS reassigning them

But you cannot really do this:

* asyncio only allows to interact with subprocesses that you directly launched.
* OSs don't provide a way to do this AFAIK (yes for a process you launched directly, but it's not possible for recursive subprocesses).

ivanpauno:
The alternative is to create a new group id when launching a process, and then send a signal to the created group.
The problem is that we have to handle the posix and windows case separately, but maybe we can find something it works for both.

wjwwood:
> OSs don't provide a way to do this AFAIK. It's possible for a process you launched directly, but it's not possible for recursive subprocesses (I think waitpid is limited to direct children in linux).

Then how does https://psutil.readthedocs.io/en/latest/index.html?highlight=children#psutil.wait_procs work? In their example they enumerate the children and then wait for them to exit.

ivanpauno:
> Then how does https://psutil.readthedocs.io/en/latest/index.html?highlight=children#psutil.wait_procs work? In their example they enumerate the children and then wait for them to exit.

polling
https://github.com/giampaolo/psutil/blob/379598312f60bf414afa8bf549f7f26af9e578ea/psutil/__init__.py#L1532-L1550
https://github.com/giampaolo/psutil/blob/57ed46de8a988e7ab26279c2a967fb15b05397a3/psutil/_psposix.py#L120-L125

wjwwood:
Ok, I see, so no actual blocking is realistic, and it does it one at a time, so you may "miss" the exit of one process while waiting on another.

So maybe that doesn't help with the race between process exit and pid reuse, but I still think waiting to see if they exit is a decent idea. There's nothing worse than having to go back to ps or something to figure out what was left behind, especially if launch could just tell us.

wjwwood:
> The alternative is to create a new group id when launching a process, and then send a signal to the created group.

What benefit does that give us? (curious)

ivanpauno:
> Ok, I see, so no actual blocking is realistic, and it does it one at a time, so you may "miss" the exit of one process while waiting on another.

Just in case, polling is not only used to check the status of more than one process, it's also used to "wait for a process to exit" if the process is not a child.
Waiting for a process that's not a child is just a loop with a "sleep" between tries, checking if a process with `pid==x` exists (see [here](https://github.com/giampaolo/psutil/blob/57ed46de8a988e7ab26279c2a967fb15b05397a3/psutil/_psposix.py#L126-L127)).

> but I still think waiting to see if they exit is a decent idea

Sounds good.
But given the API we have available, I think it's a good idea to use a launch timer for that.
Is that okay?
If another way is preferred, please specify which.

> And my proposed series of steps have some other advantages, even if this one doesn't pan out.

Do you mean scalating SIGINT -> SIGTERM -> SIGKILL -> "Log if subprocess(es) are still alive" for subprocesses as well?
That sounds fine to me, I can add that.

> What benefit does that give us? (curious)

You send only one signal to the group, though the part to monitor if the processes exited or not doesn't change.
So it doesn't seem quite benefitial honestly.

ivanpauno:
> > And my proposed series of steps have some other advantages, even if this one doesn't pan out.

> Do you mean scalating SIGINT -> SIGTERM -> SIGKILL -> "Log if subprocess(es) are still alive" for subprocesses as well?
That sounds fine to me, I can add that.

@wjwwood could you confirm this?

ivanpauno:
@wjwwood friendly ping

wjwwood:
> But given the API we have available, I think it's a good idea to use a launch timer for that.
Is that okay?
If another way is preferred, please specify which.

So, I was actually thinking we do something like what psutils does (perhaps using it), but in a thread or maybe as an async coroutine/task. Basically create a list of the pid you're watching, start timers to escalate their signals, then a thread/coroutine-task that iterates over all the pid that we're watching, and for each that have exited cancel the timers and remove them from the list, and then once you iterate, sleep for a fixed short period and then poll again until the list is empty. If, after doing SIGKILL and some period has passed, log it and then remove it from the list of pid to watch.

> Do you mean scalating SIGINT -> SIGTERM -> SIGKILL -> "Log if subprocess(es) are still alive" for subprocesses as well?

Correct.

ivanpauno:
> So, I was actually thinking we do something like what psutils does (perhaps using it), but in a thread or maybe as an async coroutine/task. Basically create a list of the pid you're watching, start timers to escalate their signals, then a thread/coroutine-task that iterates over all the pid that we're watching, and for each that have exited cancel the timers and remove them from the list, and then once you iterate, sleep for a fixed short period and then poll again until the list is empty. If, after doing SIGKILL and some period has passed, log it and then remove it from the list of pid to watch.

Please, take a look to https://github.com/ros2/launch/pull/632/commits/e326c0dfee10ad1beaa7c12db6d97eb15da57b74

launch/launch/actions/execute_local.py
```diff
@@ -420,27 +438,21 @@ def __on_shutdown(self, event: Event, context: LaunchContext) -> Optional[SomeAc
...
-        )]
         # Setup a timer to send us a SIGTERM if we don't shutdown quickly.
+        sigterm_timeout = self.__sigterm_timeout_value
         self.__sigterm_timer = TimerAction(
             period=sigterm_timeout,
             actions=[
                 OpaqueFunction(
                     function=printer,
-                    args=(base_msg.format('{}', '{}', 'SIGINT', 'SIGTERM'), sigterm_timeout)
+                    args=(base_msg.format('{}', sigterm_timeout, 'SIGINT', 'SIGTERM'), )
```
wjwwood, line 455: These changes are concerning to me because if `__get_shutdown_timer_actions` is ever called when not actually executing the actions then it seems like it might fail. Like can you ensure that `self.__sigterm_timeout` has been assigned a value at this point?

wjwwood: You might be right that these didn't need to be deferred as substitutions, but I'm not 100% certain of that. Can you expand on it?

launch/launch/actions/execute_local.py
```diff
@@ -580,8 +640,11 @@ async def __execute_process(self, context: LaunchContext) -> None:
                     timeout=self.__respawn_delay
                 )
             if not self.__shutdown_future.done():
-                context.asyncio_loop.create_task(self.__execute_process(context))
+                context.asyncio_loop.create_task(
+                    self.__execute_process(context))
```
wjwwood, line 644: Is this change needed?

launch/launch/actions/execute_local.py
```diff
@@ -580,8 +640,11 @@ async def __execute_process(self, context: LaunchContext) -> None:
...
                 return
+        if self.__signal_lingering_subprocesses_value:
+            await self._signal_subprocesses(context)
         self.__cleanup()
```
wjwwood, line 648: Does this mean subprocess cleanup only happens if it isn't restarted? (I'm assuming the `create_task()` call is about recursively calling this coroutine to restart the application?

It seems to me that we should wait for all subprocesses to finish before relaunching the executable no?

ivanpauno: > Does this mean subprocess cleanup only happens if it isn't restarted? (I'm assuming the create_task() call is about recursively calling this coroutine to restart the application?
>
> It seems to me that we should wait for all subprocesses to finish before relaunching the executable no?

Yes that sounds fine, I didn't think much what should happen when respawning, I will take another look.

launch/launch/actions/execute_local.py
ivanpauno: `__get_shutdown_timer_actions()` is getting `context` as an argument, so in any case the substitutions could be resolved here.
This PR is instead resolving the substitutions in `execute()`:

https://github.com/ros2/launch/blob/e326c0dfee10ad1beaa7c12db6d97eb15da57b74/launch/launch/actions/execute_local.py#L728-L729

I think that's fine, because `__get_shutdown_timer_actions()` is only called from [_shutdown_process()](https://github.com/ros2/launch/blob/e326c0dfee10ad1beaa7c12db6d97eb15da57b74/launch/launch/actions/execute_local.py#L253), which is called from two places [__on_shutdown_process_event()](https://github.com/ros2/launch/blob/e326c0dfee10ad1beaa7c12db6d97eb15da57b74/launch/launch/actions/execute_local.py#L290) and [__on_shutdown](https://github.com/ros2/launch/blob/e326c0dfee10ad1beaa7c12db6d97eb15da57b74/launch/launch/actions/execute_local.py#L434).
Those are used [here](https://github.com/ros2/launch/blob/e326c0dfee10ad1beaa7c12db6d97eb15da57b74/launch/launch/actions/execute_local.py#L700) and [here](https://github.com/ros2/launch/blob/e326c0dfee10ad1beaa7c12db6d97eb15da57b74/launch/launch/actions/execute_local.py#L715), which are event handlers defined in `execute()`.
So I think that `__sigterm_timeout_value` and `__sigkill_timeout_value` are guaranteed to have a value here.

I also think that `execute()` is the place were these substitutions should be resolved.
If they were resolved in an event handler afterwards, the value of the launch configuration could change, and that could lead to really confusing results.

launch/launch/actions/execute_local.py
```diff
@@ -340,7 +358,7 @@ def __on_process_output(
         if buffer.closed:
             # buffer was probably closed by __flush_buffers on shutdown.  Output without
             # buffering.
-            buffer.info(
+            logger.info(
```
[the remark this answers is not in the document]
ivanpauno: It's an old bug.
Buffer doesn't have an `info` method, as it's a `io.TextIOBase` instance.
That was meant to be `logger`.

I saw an error message related to that when testing and I modified it, though I can open a new PR for that fix.

launch/launch/actions/execute_local.py
```diff
@@ -158,6 +162,13 @@ def __init__(
             as a string or a list of strings and Substitutions to be resolved
             at runtime, defaults to the LaunchConfiguration called
             'sigkill_timeout'
+        :param: signal_subprocesses_timeout time until subprocesses start to be signaled directly,
```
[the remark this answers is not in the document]
ivanpauno: Thanks, will fix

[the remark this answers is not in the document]
ivanpauno: Done in https://github.com/ros2/launch/pull/632/commits/89ea1c2b5e29108676c05295cfe9b8fb47c9cdca

launch/launch/actions/execute_local.py
```diff
@@ -464,6 +477,13 @@ def printer(context, msg, timeout_substitutions):
             ],
             cancel_on_shutdown=False,
         )
+        self.__children = psutil.Process(
+            self._subprocess_transport.get_pid()).children(recursive=True)
+        # process_name = context.locals.process_name
+        # process_pid = context.locals.process_pid
+        # log_prefix_format = f'subprocess[pid={{}}] of process[{process_name}, pid={process_pid}]: '
+
+        # context.asyncio_loop.create_task(_signal_subprocesses())
```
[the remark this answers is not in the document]
ivanpauno: Removed in https://github.com/ros2/launch/pull/632/commits/12fb9701ecada5b96d4ed592325ffde170161ce6

ciandonovan:
Has this not been fixed by https://github.com/ros2/launch/pull/475? When running a ROS2 Launch in "non-interactive" mode, SIGINTs are successfully propagated to child nodes, cleanly terminating the process tree without needing Ctrl-C in a controlling terminal. SIGTERMs still aren't handled at all as I mentioned here though https://github.com/ros2/launch/issues/666

ivanpauno:
> Has this not been fixed by https://github.com/ros2/launch/pull/475? When running a ROS2 Launch in "non-interactive" mode, SIGINTs are successfully propagated to child nodes

This is not about killing subprocesses of launch, but about killing subprocesses of processes that launch spawns.
Ideally, this would be a responsability of the launch subprocess in question, and not of launch.
But there're easy ways to run into the problem (e.g. shell=True), that's why this PR.
See discussion in https://github.com/ros2/launch/issues/545.

I'm not sure if what we're doing here should be enabled by default or not, but having it as an option is really nice to have.

launch/launch/actions/execute_local.py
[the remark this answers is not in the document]
wjwwood: No that’s fine, this is fine to leave in.

launch/launch/actions/execute_local.py
wjwwood: Any comment on this one?

ivanpauno: sorry, I didn't see that before.
I can check if it can return `None` or something else if the process died.

ivanpauno: See code after https://github.com/ros2/launch/pull/632/commits/85ea3a90682751c3c817c197afb0d45d1173e807

launch/launch/actions/execute_local.py
wjwwood: Since the `__get_shutdown_timer_actions()` function now takes a `context` object, I think this change is safe because callers will need to source the context object which is only ever available from the `execute` function, but I don't think it was necessary for this work.

The point about the event handler and the right time to evaluate the substitutions is fine, but it could have been done without resolving the substitutions in this function, keeping it declarative (not requiring a context in its scope).

wjwwood (commented):
I had a few suggested changes, but the general approach seems to lgtm, though it's hard to say if this will be without bugs (it's a complicated task).

I don't know if it is feasible or worth the effort to create tests for this behavior, but it would be good I think to show it is working and guard against regressions.

launch/launch/actions/execute_local.py
```diff
@@ -513,6 +529,46 @@ def on_stdout_received(self, data: bytes) -> None:
...
+        start_time = context.asyncio_loop.time()
+        sigterm_timeout = self.__sigterm_timeout_value
+        sigkill_timeout = self.__sigterm_timeout_value + self.__sigkill_timeout_value
+        process_pid = self.process_details['pid']
+        process_name = self.process_details['name']
+        log_prefix_format = (
+            'subprocess[pid={}] of process['
+            f'{process_name}, pid={process_pid}]: ')
+        # signal_subprocesses_timeout_v = perform_substitutions(
+        #     context, signal_subprocesses_timeout)
```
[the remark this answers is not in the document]
ivanpauno: fixed in https://github.com/ros2/launch/pull/632/commits/ee89dc14c8dbe40f256b0d96d97e5535affe09a4

launch/launch/actions/execute_local.py
```diff
@@ -513,6 +529,46 @@ def on_stdout_received(self, data: bytes) -> None:
...
+            try:
+                sig, timeout = next(next_signals)
+            except StopIteration:
+                return
+            while context.asyncio_loop.time() < start_time + timeout:
+                await asyncio.sleep(0.5)
+                for p in list(signaled):
+                    log_prefix = log_prefix_format.format(p.pid)
+                    if not p.is_running():
+                        self.__logger.info(f'{log_prefix} exited')
```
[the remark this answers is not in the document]
ivanpauno: Yes, the first is correct.
Done in https://github.com/ros2/launch/pull/632/commits/ee89dc14c8dbe40f256b0d96d97e5535affe09a4.

launch/launch/actions/execute_local.py
```diff
@@ -513,6 +529,46 @@ def on_stdout_received(self, data: bytes) -> None:
...
+                )
+                signaled.append(p)
+            try:
+                sig, timeout = next(next_signals)
+            except StopIteration:
+                return
+            while context.asyncio_loop.time() < start_time + timeout:
+                await asyncio.sleep(0.5)
+                for p in list(signaled):
+                    log_prefix = log_prefix_format.format(p.pid)
```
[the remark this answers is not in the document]
ivanpauno: Done in https://github.com/ros2/launch/pull/632/commits/ee89dc14c8dbe40f256b0d96d97e5535affe09a4

launch/launch/actions/execute_local.py
```diff
@@ -513,6 +529,46 @@ def on_stdout_received(self, data: bytes) -> None:
...
+                self.__logger.info(
+                    f'{log_prefix}sending {sig.name} to subprocess directly.'
+                )
+                signaled.append(p)
+            try:
+                sig, timeout = next(next_signals)
+            except StopIteration:
+                return
+            while context.asyncio_loop.time() < start_time + timeout:
+                await asyncio.sleep(0.5)
```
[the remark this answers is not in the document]
ivanpauno: Fixed in https://github.com/ros2/launch/pull/632/commits/ee89dc14c8dbe40f256b0d96d97e5535affe09a4

ivanpauno:
@wjwwood test case was added in https://github.com/ros2/launch/pull/632/commits/5b43cdcc50108a60fbbb4b56d3eb9d88b91c52f7, I also needed to rebase to solve conclicts with `rolling`

methylDragon:
@ivanpauno just some flake8 linting issues

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17842/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12410/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18510/)

ivanpauno:
mmm, I have to double check those failures, they seem related

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17843/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12411/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18511/)

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17844/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12412/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18512/)

ivanpauno:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18522/)

ivanpauno:
This is making windows CI hang for some reason ....
I will try to set up a windows vm and see what's going on, but I'm not sure if I will have the time to complete that

launch/test/launch/test_execute_local.py
```diff
@@ -138,3 +143,39 @@ def test_execute_process_with_output_dictionary():
...
+while 1:
+    time.sleep(0.5)
+"""
+
+
+def test_kill_subprocesses():
+    """Test launching a process with an environment variable."""
+    executable = ExecuteLocal(
+        process_description=Executable(
+            cmd=['python3', '-c', f'"{PYTHON_SCRIPT}"'],
```
wjwwood, line 160: Is this sufficient to test the feature? Is it the `shell=True` part that makes this test useful?

ivanpauno: `shell=True` will create a shell process, and that shell will create a subproceses.
So the launch process will have to kill the shell subprocess, because the shell will not trap the signals and resend them to the child.

This shows that the feature works.
It's not super complete though, if you have more test case ideas I can add them.

russkel:
My use case is a process spawns some children and then exits, and I do not want launch to exit until the child processes have exited.

I have implemented this in https://github.com/Greenroom-Robotics/launch_ext/commit/850a2f4b59e7bb49857f8c5fec06112aa8e239cc#diff-7baf6e854cc3c937eaed0b127161c5f82cf86d8a8eaef0038171216a817a0c62R194-R222

My technique was to use the stdin/stdout/stderr inodes and if any shared the same inodes with the parent process they would be considered children to be waited on. I am not sure if this way is ideal but it does seem to work.

mwcondino:
Howdy @ivanpauno - any chance this could still be merged? I've encountered this issue recently, and have a workaround in place, but it would be ideal to handle the issue directly in `ExecuteLocal` with this fix :smile:

kbrameld:
Just a heads up - there is interest in getting this functionality merged in. With this change, we're able to kill gazebo simulations in our CI tests and not have lingering gazebo processes.

mjcarroll:
> Just a heads up - there is interest in getting this functionality merged in. With this change, we're able to kill gazebo simulations in our CI tests and not have lingering gazebo processes.

I'm guessing that @ivanpauno doesn't have time to move it forward these days. Would you be interested in helping get it across the line?

ahcorde:
Pulls: ros2/launch#632
Gist: https://gist.githubusercontent.com/ahcorde/ba1172c97744f3cff7ee808616bfdcb6/raw/f56a2144b4248e33928ca0d743254bdb3bfc2ade/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17641
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/26344/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/19992/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6043/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26084/)

anton-matosov:
I would like to get this functionality merged in. Are there any pending items left for this except rebase to latest and fix type issues reported by build?

anton-matosov:
Here is the fork of this PR that includes all the needed type and null check fixes https://github.com/ros2/launch/pull/936

PR #649, opened 2022-09-08
Implement Any, All, Equals, and NotEquals substitutions

methylDragon:
# Description
This PR implements new `Any` and `All` boolean substitutions. As well as new `Equals` and `NotEquals` substitutions.

The equality substitutions deprecate the `LaunchConfigurationEquals` and `LaunchConfigurationNotEquals` conditions, because they're more universal and can take in `LaunchConfiguration` substitutions.

It should be self-explanatory from the included tests (:

### DIFF
--- a/launch/launch/conditions/launch_configuration_equals.py
+++ b/launch/launch/conditions/launch_configuration_equals.py
@@ -16,6 +16,7 @@
 
 from typing import Optional
 from typing import Text
+import warnings
 
 from ..condition import Condition
 from ..launch_context import LaunchContext
@@ -43,6 +44,16 @@ class LaunchConfigurationEquals(Condition):
         launch_configuration_name: Text,
         expected_value: Optional[SomeSubstitutionsType]
     ) -> None:
+        warnings.warn(
+            "The 'LaunchConfigurationEquals' and 'LaunchConfigurationNotEquals' Conditions are "
+            " deprecated. Use the 'EqualsSubstitution' and 'NotEqualsSubstitution' substitutions "
+            "instead! E.g.:\n"
+            "  IfCondition(\n  "
+            "\tEqualsSubstitution(LaunchConfiguration('some_launch_arg'), \"some_equality_check\")"
+            "\n  )",
+            UserWarning
+        )
+
         self.__launch_configuration_name = launch_configuration_name
         if expected_value is not None:
             self.__expected_value = normalize_to_list_of_substitutions(expected_value)
--- a/launch/launch/conditions/launch_configuration_equals.py
+++ b/launch/launch/conditions/launch_configuration_equals.py
@@ -51,10 +51,10 @@ class LaunchConfigurationEquals(Condition):
         warnings.warn(
             "The 'LaunchConfigurationEquals' and 'LaunchConfigurationNotEquals' Conditions are "
             " deprecated. Use the 'EqualsSubstitution' and 'NotEqualsSubstitution' substitutions "
-            "instead! E.g.:\n"
-            "  IfCondition(\n  "
+            'instead! E.g.:\n'
+            '  IfCondition(\n  '
             "\tEqualsSubstitution(LaunchConfiguration('some_launch_arg'), \"some_equality_check\")"
-            "\n  )",
+            '\n  )',
             UserWarning
         )
 
--- a/launch/launch/conditions/launch_configuration_equals.py
+++ b/launch/launch/conditions/launch_configuration_equals.py
@@ -37,6 +37,10 @@ class LaunchConfigurationEquals(Condition):
 
     If ``None`` is provided instead of a string expression, then the condition
     evaluates to ``True`` if the launch configuration is not set.
+
+    .. deprecated:: 1.1.0
+       Replaced by the more universally usable substitutions:
+       'EqualsSubstitution' and 'NotEqualsSubstitution'
     """
 
     def __init__(
--- a/launch/launch/some_substitutions_type.py
+++ b/launch/launch/some_substitutions_type.py
@@ -34,21 +34,3 @@ SomeSubstitutionsType_types_tuple = (
     Substitution,
     collections.abc.Iterable,
 )
-
-
-def is_some_substitutions_type(input: Optional[Union[Any, Iterable[Any]]]) -> bool:
-    """
-    Check if input is instance of members of SomeSubstitutionsType.
-
-    It is not possible to do a direct comparison, and typing.get_args() doesn't seem to work, so
-    we're forced to use this very inelegant solution...
-    """
-    if isinstance(input, Iterable):
-        return all(
-            (isinstance(i, Substitution) or isinstance(i, Text)) for i in input
-        )
-    else:
-        return any([
-            isinstance(input, Substitution),
-            isinstance(input, Text)
-        ])
--- a/launch/launch/some_substitutions_type.py
+++ b/launch/launch/some_substitutions_type.py
@@ -15,7 +15,9 @@
 """Module for SomeSubstitutionsType type."""
 
 import collections.abc
+from typing import Any
 from typing import Iterable
+from typing import Optional
 from typing import Text
 from typing import Union
 
@@ -32,3 +34,21 @@ SomeSubstitutionsType_types_tuple = (
     Substitution,
     collections.abc.Iterable,
 )
+
+
+def is_some_substitutions_type(input: Optional[Union[Any, Iterable[Any]]]) -> bool:
+    """
+    Check if input is instance of members of SomeSubstitutionsType.
+
+    It is not possible to do a direct comparison, and typing.get_args() doesn't seem to work, so
+    we're forced to use this very inelegant solution...
+    """
+    if isinstance(input, Iterable):
+        return all(
+            (isinstance(i, Substitution) or isinstance(i, Text)) for i in input
+        )
+    else:
+        return any([
+            isinstance(input, Substitution),
+            isinstance(input, Text)
+        ])
--- a/launch/launch/some_substitutions_type.py
+++ b/launch/launch/some_substitutions_type.py
@@ -15,9 +15,7 @@
 """Module for SomeSubstitutionsType type."""
 
 import collections.abc
-from typing import Any
 from typing import Iterable
-from typing import Optional
 from typing import Text
 from typing import Union
 
--- a/launch/launch/substitutions/boolean_substitution.py
+++ b/launch/launch/substitutions/boolean_substitution.py
@@ -184,7 +184,7 @@ class AnySubstitution(Substitution):
 
     def describe(self) -> Text:
         """Return a description of this substitution as a string."""
-        return f'AndSubstitution({" ".join(self.args)})'
+        return f'AnySubstitution({" ".join(self.args)})'
 
     def perform(self, context: LaunchContext) -> Text:
         """Perform the substitution."""
--- a/launch/launch/substitutions/boolean_substitution.py
+++ b/launch/launch/substitutions/boolean_substitution.py
@@ -67,7 +67,7 @@ class AndSubstitution(Substitution):
     """Substitution that returns 'and' of the input boolean values."""
 
     def __init__(self, left: SomeSubstitutionsType, right: SomeSubstitutionsType) -> None:
-        """Create a AndSubstitution substitution."""
+        """Create an AndSubstitution substitution."""
         super().__init__()
 
         self.__left = normalize_to_list_of_substitutions(left)
@@ -113,7 +113,7 @@ class OrSubstitution(Substitution):
     """Substitution that returns 'or' of the input boolean values."""
 
     def __init__(self, left: SomeSubstitutionsType, right: SomeSubstitutionsType) -> None:
-        """Create a AndSubstitution substitution."""
+        """Create an OrSubstitution substitution."""
         super().__init__()
 
         self.__left = normalize_to_list_of_substitutions(left)
@@ -121,7 +121,7 @@ class OrSubstitution(Substitution):
 
     @classmethod
     def parse(cls, data: Iterable[SomeSubstitutionsType]):
-        """Parse `AndSubstitution` substitution."""
+        """Parse `OrSubstitution` substitution."""
         if len(data) != 2:
             raise TypeError('and substitution expects 2 arguments')
         return cls, {'left': data[0], 'right': data[1]}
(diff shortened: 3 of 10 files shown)


launch/launch/substitutions/not_equals_substitution.py
```diff
@@ -0,0 +1,51 @@
...
+        return f'NotEqualsSubstitution({self.left} {self.right})'
+
+    def perform(self, context: LaunchContext) -> Text:
+        """Perform the substitution."""
+        result = super().perform(context)
+
+        if result == 'true':
+            return 'false'
+        else:
+            return 'true'
```
wjwwood, line 51: nitpick: It's odd that you broke this one up, but used a shorter version of this logic in the other substitutions. Not wrong, just inconsistent. Is there a reason?

launch/launch/substitutions/equals_substitution.py
```diff
@@ -0,0 +1,87 @@
...
+            left = perform_substitutions(context, self.left)
+        else:
+            left = self.left
+
+        if is_some_substitutions_type(self.right):
+            right = perform_substitutions(context, self.right)
+        else:
+            right = self.right
+
+        return str(left == right).lower()
```
wjwwood, line 87: This leaves me wondering about special cases, like float comparison. We're obviously comparing strings here, and that works for most situations, but I think we should expand the documentation of these to clarify what we're doing and what we're returning, i.e. none of the doc blocks mention we're returning the string `true` or `false` as a result of these substitutions.

launch/launch/substitutions/equals_substitution.py
```diff
@@ -0,0 +1,87 @@
...
+from ..utilities.type_utils import perform_substitutions
+
+
+@expose_substitution('equals')
+class EqualsSubstitution(Substitution):
+    """Substitution that checks if two inputs are equal."""
+
+    def __init__(
+        self,
+        left: Optional[Union[Any, Iterable[Any]]],
```
wjwwood, line 38: I'm curios why this isn't just:

```suggestion
        left: SomeSubstitutionsType,
```

Same for the below.

launch/launch/substitutions/equals_substitution.py
```diff
@@ -0,0 +1,87 @@
...
+    def describe(self) -> Text:
+        """Return a description of this substitution as a string."""
+        return f'EqualsSubstitution({self.left} {self.right})'
+
+    def perform(self, context: LaunchContext) -> Text:
+        """Perform the substitution."""
+        if is_some_substitutions_type(self.left):
+            left = perform_substitutions(context, self.left)
+        else:
+            left = self.left
```
wjwwood, line 80: Do we need to support non-substitution inputs?

launch/launch/substitutions/not_equals_substitution.py
methylDragon: Ah, late night brainfart, I'll fix it 😬

launch/launch/substitutions/equals_substitution.py
methylDragon: I think the right arg definitely should support non-substitution inputs (e.g. the primitive Python types), since it's reasonable to see them used in the launch Python files.

Given that, I think it's nice to make it commutative and also support the left side :o

launch/launch/conditions/launch_configuration_equals.py
```diff
@@ -43,6 +44,16 @@ def __init__(
         launch_configuration_name: Text,
         expected_value: Optional[SomeSubstitutionsType]
     ) -> None:
+        warnings.warn(
```
jacobperron, line 47: I would also add a notice to the docstring above.

launch/test/launch/substitutions/test_equals_substitution.py
```diff
@@ -0,0 +1,40 @@
...
+    lc = LaunchContext()
+    assert EqualsSubstitution(None, None).perform(lc) == 'true'
+    assert EqualsSubstitution(None, "something").perform(lc) == 'false'
+    assert EqualsSubstitution(True, True).perform(lc) == 'true'
+    assert EqualsSubstitution(False, False).perform(lc) == 'true'
+    assert EqualsSubstitution(True, False).perform(lc) == 'false'
+    assert EqualsSubstitution(1, True).perform(lc) == 'true'
+    assert EqualsSubstitution(1, 1).perform(lc) == 'true'
+    assert EqualsSubstitution(1, 0).perform(lc) == 'false'
+    assert EqualsSubstitution(1, "1").perform(lc) == 'false'
```
jacobperron, line 35: I think this could be a source of confusion for users. IIUC, the following substitution will always return `'false'` since substitutions always result in a string:

```py
EqualsSubstitution(LaunchConfiguration('foo'), 1)
```
I think either we should cast non-substitution arguments to strings before doing the comparison _or_ only support substitution types as arguments.

methylDragon: Alrighty, I'll need to write some extra logic to ensure `1` and `True` (and `0` and `False`) remain equatable, since I can see that as a case that occurs frequently..

Should a non-zero integer and `True` equate also..? :thinking:
For string comparison reasons I'd like to say no.

launch/launch/substitutions/equals_substitution.py
methylDragon: I am now supporting float comparisons. I will stop short of complex numbers though :grimacing:

launch/launch/conditions/launch_configuration_equals.py
methylDragon: I added a Sphinx deprecation directive

launch/test/launch/substitutions/test_equals_substitution.py
methylDragon: [See tests](https://github.com/ros2/launch/pull/649/commits/43a9034ee0528219b36dc43ec560b7d85a7f7b85#diff-34ab44869ae2b18792077675afa5660d2bf35fa2531a62efb1a48d5698531a71)

launch/launch/some_substitutions_type.py
```diff
@@ -32,3 +34,21 @@
     Substitution,
     collections.abc.Iterable,
 )
+
+
+def is_some_substitutions_type(input: Optional[Union[Any, Iterable[Any]]]) -> bool:
```
jacobperron, line 39: This looks very similar to [`is_substitution`](https://github.com/ros2/launch/blob/76ade26d3f7bcb6a5a1b72a512a8c098e6d53f04/launch/launch/utilities/type_utils.py#L346). Can we just use `is_substitution` instead?

If not, then I think it would be better to move this function into `type_utils.py`.

launch/launch/substitutions/equals_substitution.py
```diff
@@ -0,0 +1,123 @@
...
+        left: Optional[Union[Any, Iterable[Any]]],
+        right: Optional[Union[Any, Iterable[Any]]]
+    ) -> None:
+        """Create an EqualsSubstitution substitution."""
+        super().__init__()
+
+        if is_some_substitutions_type(left):
+            self.__left = normalize_to_list_of_substitutions(left)
+        elif left is None:
+            self.__left = ""
```
jacobperron, line 70:
```suggestion
            self.__left = ''
```

launch/launch/substitutions/equals_substitution.py
```diff
@@ -0,0 +1,123 @@
...
+        elif isinstance(left, bool):
+            self.__left = str(left).lower()
+        else:
+            self.__left = str(left)
```
jacobperron, line 74: I think we can simplify the implementation if we convert the arguments here to [`TextSubstitution`s](https://github.com/ros2/launch/blob/76ade26d3f7bcb6a5a1b72a512a8c098e6d53f04/launch/launch/substitutions/text_substitution.py#L23) and then call `normalize_to_list_of_substitutions`. Then, we don't need to check if the arguments are substitutions inside of `perform`.

launch/launch/substitutions/equals_substitution.py
```diff
@@ -0,0 +1,123 @@
...
+
+        if is_some_substitutions_type(self.right) and not isinstance(self.right, Text):
+            right = perform_substitutions(context, self.right)
+        else:
+            right = self.right
+
+        if _str_is_bool(left) and _str_is_bool(right):
+            return str((left.lower() in ('true', '1')) == (right.lower() in ('true', '1'))).lower()
+        elif _str_is_float(left) and _str_is_float(right):
+            return str(math.isclose(float(left), float(right))).lower()
```
jacobperron, line 121: I'm not convinced we should add these special cases. It seems like it would be less confusing if we just clearly document that this substitution is just doing a string comparison (and that any provided arguments are cast to strings). I'm on the fence about it, but whatever we decide I think it needs to be clearly documented (ie. what does it mean for two arguments to be equal?).

launch/test/launch/substitutions/test_equals_substitution.py
```diff
@@ -0,0 +1,107 @@
...
+    def _permute_assertion(left, right, context, output):
+        assert EqualsSubstitution(left, right).perform(context) == output
+        assert EqualsSubstitution(right, left).perform(context) == output
+
+    lc = LaunchContext()
+
+    # NoneType
+    _permute_assertion(None, None, lc, 'true')
+    _permute_assertion(None, "", lc, 'true')
+    _permute_assertion(None, "something", lc, 'false')
```
jacobperron, line 35: style nit: avoid double-quotes if possible and use single quotes everywhere. I'm surprised our linter is happy with this :thinking:

launch/test/launch/substitutions/test_equals_substitution.py
jacobperron: > Should a non-zero integer and True equate also..? 

I'd say no as well, since `42 != True` in Python.

launch/test/launch/substitutions/test_boolean_substitution.py
```diff
@@ -54,3 +56,39 @@ def test_or_substitution():
...
+    assert AnySubstitution().perform(lc) == 'false'
+    assert AnySubstitution('true').perform(lc) == 'true'
+    assert AnySubstitution('false').perform(lc) == 'false'
+    assert AnySubstitution('true', 'true').perform(lc) == 'true'
+    assert AnySubstitution('true', 'false').perform(lc) == 'true'
+    assert AnySubstitution('false', 'true').perform(lc) == 'true'
+    assert AnySubstitution('false', 'false').perform(lc) == 'false'
+    assert AnySubstitution('true', 'true', 'true').perform(lc) == 'true'
+    assert AnySubstitution('true', 'true', 'false').perform(lc) == 'true'
+    assert AnySubstitution('false', 'false', 'false').perform(lc) == 'false'
```
jacobperron, line 72: If we support `'1'` as a truthy value in other substitutions, then I would expect the following to be true as well:

```
AnySubstitution('false', 'false', '1').perform(lc) == 'true'
```

Again, I'm not sure if we should support this, but we should at least be consistent.

launch/launch/substitutions/equals_substitution.py
methylDragon: I'm thinking of the cases (well, at least in ROS 1...) where I've seen people put in `'1'` instead of `'true'` that motivated the addition of those special cases.

Likewise for the Python launch API, the naming of the substitutions remind me of the Python constructs, which make me want to have the substitutions mimic that behavior.

I'm not sure what the decision should be on this though. I guess it depends on whether we think more people will use xml vs Python, and what people will find intuitive...

Till then, I think I'll add a note for these cases in the docstring and continue supporting them? :thinking:

launch/test/launch/substitutions/test_equals_substitution.py
methylDragon: Ah, gotcha. I got confused because `bool(42) == True`, but I guess that's one conversion too many :grimacing:

launch/launch/substitutions/equals_substitution.py
jacobperron: Looking around the existing code, I see we define valid "true" and "false" values here:

https://github.com/ros2/launch/blob/76ade26d3f7bcb6a5a1b72a512a8c098e6d53f04/launch/launch/conditions/evaluate_condition_expression_impl.py#L25-L26

So, it's okay to support the string `'1'` and, _maybe_, the number `1`.

For comparison, `IfCondition` and `UnlessCondition` (and I presume other conditions) expect a string. For example, the following launch file results in an exception:

```yaml
launch:
- group:
    if: 1
    children:
        - log:
            message: Hello
```

I would clearly document the expectations of the current implementation and I think that would be enough unless there's a dissenting opinion.

launch/test/launch/substitutions/test_boolean_substitution.py
jacobperron: Considering that we suppor the strings `'1'` and `'0'` for conditions, I think we should add support for them to the any and all substitutions.

methylDragon: Will do!

launch/launch/substitutions/equals_substitution.py
methylDragon: [This PR](https://github.com/ros2/launch/pull/651) should adopt '1' and '0' truthiness for the And, Or, and Not conditions.

Once that's merged in I will make use of the type coercion for the Any and All conditions

methylDragon: Implemented and merged in https://github.com/ros2/launch/pull/651

launch/launch/substitutions/equals_substitution.py
methylDragon: https://github.com/ros2/launch/pull/649/commits/bec67457dbb28d6a81e5ce3964337c082505a81e !

launch/launch/substitutions/equals_substitution.py
methylDragon: Correction, I couldn't use the implementation in #651 here because the Equals substitution cannot hard-enforce bool string inputs.

So the string checking I'm doing here persists. (Checking for 'True', 'true' and '1', and the equivalent for false only when both inputs are like that.)

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17342/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11877/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17832/)

methylDragon:
Flake8 errors. Let's try this again:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17343/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11878/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17836/)

methylDragon:
Flake8 errors. Let's try this again:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17344/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11879/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17837/)

launch/test/launch/substitutions/test_boolean_substitution.py
methylDragon: They are now implicitly supported following the merge of #651. Tests have been implemented to show that this is the case
https://github.com/ros2/launch/pull/649/commits/4de282662e85d5ded9886d87ff36d5bd6225b6a4

methylDragon:
Windows failures seem unrelated, but I'm gonna try again regardless

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17863/)

launch/launch/substitutions/boolean_substitution.py
```diff
@@ -152,3 +152,77 @@ def perform(self, context: LaunchContext) -> Text:
...
+        return cls, {'args': data}
+
+    @property
+    def args(self) -> Substitution:
+        """Getter for args."""
+        return self.__args
+
+    def describe(self) -> Text:
+        """Return a description of this substitution as a string."""
+        return f'AndSubstitution({" ".join(self.args)})'
```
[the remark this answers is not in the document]
jacobperron: This comment still stands.

jacobperron (approved):
I have one minor unresolved comment above, otherwise LGTM!

launch/launch/substitutions/boolean_substitution.py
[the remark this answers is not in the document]
methylDragon: Oops

PR #651, opened 2022-09-16
Add special cases to coerce "1" and "0" to bool when using bool coercion only

methylDragon:
# Description
Exactly as the title says.

Given that "1" and "0" are valid true and false expressions according [to the conditions](https://github.com/ros2/launch/blob/76ade26d3f7bcb6a5a1b72a512a8c098e6d53f04/launch/launch/conditions/evaluate_condition_expression_impl.py#L25-L26), I think it makes sense to allow coercion of the "1" and "0" strings to `True` and `False`.

This ONLY affects coercion when `bool` is explicitly passed to it. It's as small a change as possible.

This supports the work in https://github.com/ros2/launch/pull/649 and will be used to support the boolean substitution implementations.

# Tests
Special note: This means that [one test for the type coercion](https://github.com/ros2/launch/pull/651/files#diff-56f27934664cd12e7a3b96577955eb19f008622849f77cfd2d7c8a15cc756859L286) will no longer apply.

Beyond that, relevant tests have been implemented.
- https://github.com/ros2/launch/pull/651/files#diff-f1043644a82be146c361f43bb127588df2726dacda3264794c19441bc428fbec
- https://github.com/ros2/launch/pull/651/files#diff-56f27934664cd12e7a3b96577955eb19f008622849f77cfd2d7c8a15cc756859

### DIFF
--- a/launch/test/launch/utilities/test_type_utils.py
+++ b/launch/test/launch/utilities/test_type_utils.py
@@ -234,6 +234,8 @@ def test_coercions_given_specific_type(coerce_to_type_impl):
     assert coerce_to_type_impl('on', data_type=bool) is True
     assert coerce_to_type_impl('off', data_type=bool) is False
     assert coerce_to_type_impl('True', data_type=bool) is True
+    assert coerce_to_type_impl('1', data_type=bool) is True
+    assert coerce_to_type_impl('0', data_type=bool) is False
 
     assert coerce_to_type_impl('[.2, .1, .1]', data_type=List[float]) == [.2, .1, .1]
     assert coerce_to_type_impl('[asd, bsd, csd]', data_type=List[str]) == ['asd', 'bsd', 'csd']
--- a/launch/test/launch/utilities/test_type_utils.py
+++ b/launch/test/launch/utilities/test_type_utils.py
@@ -286,10 +286,6 @@ def test_coercion_raises_value_error(coerce_to_type_impl):
     with pytest.raises(ValueError):
         coerce_to_type_impl('Bsd', data_type=bool)
 
-    # '1' and '0' are special cases for bool coercion
-    # with pytest.raises(ValueError):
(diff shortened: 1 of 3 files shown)


methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17312/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11850/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17798/)

wjwwood (commented):
Makes sense to me, but I think @ivanpauno or @hidmic need to have a look, since they worked on this originally (I think). The coercion stuff is kind of tricky, so I could be missing something.

launch/test/launch/utilities/test_type_utils.py
```diff
@@ -283,8 +285,10 @@ def test_coercion_raises_value_error(coerce_to_type_impl):
         coerce_to_type_impl('', data_type=bool)
     with pytest.raises(ValueError):
         coerce_to_type_impl('Bsd', data_type=bool)
-    with pytest.raises(ValueError):
-        coerce_to_type_impl('1', data_type=bool)
+
+    # '1' and '0' are special cases for bool coercion
+    # with pytest.raises(ValueError):
+    #     coerce_to_type_impl('1', data_type=bool)
```
jacobperron, line 291: I think it's okay to remove these commented lines.

jacobperron (commented):
+1 from me

ivanpauno (approved):
lgtm

launch/test/launch/utilities/test_type_utils.py
methylDragon: Okie https://github.com/ros2/launch/pull/651/commits/89847240d2072199d8ae5dafcb82123cb2418323

PR #652, opened 2022-09-17
add LaunchLogDir substitution, replacing log_dir frontend only substitution

wjwwood:
This is needed in support of a new `launch_ros` action that lets you have all the node log files collect inside the launch log directory.

This changes some of the stuff done in https://github.com/ros2/launch/pull/490.

Connected to:

- https://github.com/ros2/launch_ros/pull/325

### DIFF
--- a/launch/launch/logging/__init__.py
+++ b/launch/launch/logging/__init__.py
@@ -24,14 +24,9 @@ import os
 import socket
 import sys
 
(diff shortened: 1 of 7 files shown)


wjwwood:
There are some issues with this, and I need to finish the related pr, so hold off on review for now.

PR #655, opened 2022-09-26
Pass modules to PythonExpression

RobertBlakeAnderson:
Added ability to pass Python modules to the PythonExpression substitution.
Allows eval of more expressions.

For example, my motivating use case was:
`robot_desc = PythonExpression(["process_file('", filepath, "').toprettyxml(indent='  ')"], xacro.__dict__)`

### DIFF
--- a/launch/launch/substitutions/python_expression.py
+++ b/launch/launch/substitutions/python_expression.py
@@ -38,7 +38,7 @@ class PythonExpression(Substitution):
     """
 
     def __init__(self, expression: SomeSubstitutionsType,
-                 modules: SomeSubstitutionsType = ['math']) -> None:
+                 python_modules: SomeSubstitutionsType = ['math']) -> None:
         """Create a PythonExpression substitution."""
         super().__init__()
 
@@ -49,14 +49,14 @@ class PythonExpression(Substitution):
             'PythonExpression')
 
         ensure_argument_type(
-            modules,
+            python_modules,
             (str, Substitution, collections.abc.Iterable),
-            'modules',
+            'python_modules',
             'PythonExpression')
 
         from ..utilities import normalize_to_list_of_substitutions
         self.__expression = normalize_to_list_of_substitutions(expression)
-        self.__modules = normalize_to_list_of_substitutions(modules)
+        self.__python_modules = normalize_to_list_of_substitutions(python_modules)
 
     @classmethod
     def parse(cls, data: Iterable[SomeSubstitutionsType]):
@@ -68,11 +68,11 @@ class PythonExpression(Substitution):
         if len(data) == 2:
             # We get a text subsitution from XML,
             # whose contents are comma-separated module names
-            kwargs['modules'] = []
+            kwargs['python_modules'] = []
             # Check if we got empty list from XML
             if len(data[1]) > 0:
                 modules_str = data[1][0].perform(None)
-                kwargs['modules'] = modules_str.split(', ')
+                kwargs['python_modules'] = modules_str.split(', ')
         return cls, kwargs
 
     @property
@@ -81,20 +81,20 @@ class PythonExpression(Substitution):
         return self.__expression
 
     @property
-    def modules(self) -> List[Substitution]:
+    def python_modules(self) -> List[Substitution]:
         """Getter for expression."""
-        return self.__modules
+        return self.__python_modules
 
     def describe(self) -> Text:
         """Return a description of this substitution as a string."""
         return 'PythonExpr({}, [{}])'.format(
             ' + '.join([sub.describe() for sub in self.expression]),
-            ', '.join([sub.describe() for sub in self.modules]))
+            ', '.join([sub.describe() for sub in self.python_modules]))
 
     def perform(self, context: LaunchContext) -> Text:
         """Perform the substitution by evaluating the expression."""
         from ..utilities import perform_substitutions
-        module_names = [context.perform_substitution(sub) for sub in self.modules]
+        module_names = [context.perform_substitution(sub) for sub in self.python_modules]
         module_objects = [importlib.import_module(name) for name in module_names]
         locals = {}
         for module in module_objects:
--- a/launch/launch/substitutions/python_expression.py
+++ b/launch/launch/substitutions/python_expression.py
@@ -37,7 +37,7 @@ class PythonExpression(Substitution):
     It also may contain math symbols and functions.
     """
 
-    def __init__(self, expression: SomeSubstitutionsType, locals: dict=math.__dict__) -> None:
+    def __init__(self, expression: SomeSubstitutionsType, locals: dict = math.__dict__) -> None:
         """Create a PythonExpression substitution."""
         super().__init__()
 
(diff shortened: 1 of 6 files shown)


methylDragon:
Could you add tests?
@RobertBlakeAnderson

RobertBlakeAnderson:
Sure thing! I'll get to it shortly.

RobertBlakeAnderson:
Done. I also fixed a couple tiny issues that I found while creating and running the tests.

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17827/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12394/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18496/)

launch/launch/substitutions/python_expression.py
```diff
@@ -62,11 +66,16 @@ def expression(self) -> List[Substitution]:
...
         return self.__expression
 
+    @property
+    def locals(self) -> dict:
+        """Getter for locals."""
+        return self.__locals
+
     def describe(self) -> Text:
         """Return a description of this substitution as a string."""
         return 'PythonExpr({})'.format(' + '.join([sub.describe() for sub in self.expression]))
```
methylDragon, line 76: I think the description should be updated to reflect the modules

methylDragon: Likewise, the parse up there, otherwise the XML frontend won't work

See: https://github.com/ros2/launch/blob/rolling/launch/launch/substitutions/launch_configuration.py#L68
In this case the length of data might be 1 or 2. If it's 1, be sure to set 'modules' to the default value in the parse!

methylDragon (changes requested):

launch/launch/substitutions/python_expression.py
RobertBlakeAnderson: I could store the module names in order to print them as part of the Describe output, as printing `__locals` would be way too long for many modules. But then that got me thinking that we might also want to do something like this at Line 53.

> self.__locals = {}
        for x in modules:
            self.__locals[x.__name__] = vars(x)

This way, the expression locals would remain keyed for their respective modules. Then an expression would read `sys.getrefcount(str("hello world!"))` instead of `getrefcount(str("hello world!"))`, for example. This seems good in terms of clarity and avoiding name clashes, especially if someone wants to pass a custom module.

However, doing that would make this PR a breaking change, since PythonExpression currently accepts `math` definitions without the module name. That seems like a reasonable break to make as part of a new ROS2 release, but I wanted to run this idea by you all. I think it would be a good change in the longer run.

RobertBlakeAnderson: Thinking about it even more, the new PythonExpression constructor probably ought to take the module names as a SomeSubstitutionsType, to be consistent with how substitutions generally work.

methylDragon: If we do SomeSubstitutionsType we'll most likely have to take in strings as opposed to module objects directly. I am fine with this since in that case the xml frontend can be unified. I'm not too fussed about this point though.

What will have to change is the parse:
We'll need importing logic, since there isn't a way to pass modules directly in xml.

Pardon the style/naming. I'm just trying to workshop the idea
```
import importlib

    @classmethod
        """Parse `PythonExpression` substitution."""
        if not len(data) >= 1 and not len(data) <= 2:
            raise TypeError('eval substitution expects 1 to 2 arguments')
        modules = []
        for i in data[1].split("SOME_DELIMITER_PROBABLY_COMMA"):
            module = importlib.import_module(i)
            modules.append(module)
        # the name 'modules' could change
        return cls, {'expression': data[0], 'modules': module}
```

methylDragon: As for breaking, I think we can have our cake and eat it too. We can just print a warning instead and support both cases.

We can store modules as well as any local names.

More precisely, we can:
```
# Modules is probably not a good name at this point
for x in modules:
    if type(x) is module:
        if any(i for i in vars(x) in in self__locals):
            # PRINT A WARNING
        self.__locals.update(vars(x))
    elif:
        if x.__name__ in self__locals:
            # PRINT A WARNING
        self.__locals[x.__name__] = x
```

methylDragon: Furthermore, if we really really wanted to, we could subdivide self.__locals:

```
self.__locals = {'modules': {}, 'anon_names': {}}
```

And then merge the two dictionaries (and do a warning printout in the case of name clashes) when performing the eval.
(I think this is more efficient than doing the check for each module that is added.

And then when describing the locals dict, you can print the names of any modules, and then all of the anonymous non-module names, to shorten the printout

RobertBlakeAnderson:
Got it working with Substitutions in the constructor. Updated tests and added new ones for the XML. I'll now look into the matter of keying the definitions to their modules, with backwards compatibility. As for naming, maybe `python_modules` or `python_deps` would be less ambiguous.

RobertBlakeAnderson:
Done. The expressions will now require the module names before the definitions you use. For backwards compatibility, math definitions will work with or without the module name, and the tests include cases for both. If you wish, the implicit references could be considered deprecated for removal at a future date.

I think this PR is ready to go now.

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17848/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12416/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18516/)

:fire:

launch/launch/substitutions/python_expression.py
```diff
@@ -47,26 +48,61 @@ def __init__(self, expression: SomeSubstitutionsType) -> None:
...
+        kwargs = {}
+        kwargs['expression'] = data[0]
+        if len(data) == 2:
+            # We get a text subsitution from XML,
+            # whose contents are comma-separated module names
+            kwargs['python_modules'] = []
+            # Check if we got empty list from XML
+            if len(data[1]) > 0:
+                modules_str = data[1][0].perform(None)
+                kwargs['python_modules'] = modules_str.split(', ')
```
methylDragon, line 75: Might want to split by "," and then `strip()` each element (so it supports both "a, b, c" and "a,b,c")

methylDragon (changes requested):

launch/launch/substitutions/python_expression.py
methylDragon, line 75:
```suggestion
                kwargs['python_modules'] = [module.strip() for module in modules_str.split(',')]
```

RobertBlakeAnderson: Good idea. Added that along with a test case.

methylDragon (approved):

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17850/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12418/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18518/)

methylDragon:
Flake 8 issues :o

https://ci.ros2.org/job/ci_linux-aarch64/12418/testReport/launch.test/test_flake8/test_flake8/

I think just change the name `locals`

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17854/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12422/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18523/)

PR #656, opened 2022-10-03
Fix the restoring of os.environ to maintain type.

clalancette:
During the LaunchContext constructor or _push_environment(), we take a copy of the environment via os.environ.copy().  That function returns a dict, but it turns out that os.environ itself is an os._Environ object, not a dict.  Thus, when restoring the environment in either _pop_environment() or _reset_environment(), we cannot just set os.environ to that value.  Instead, we clear out the environment and then set the key/value pairs one-by-one into the environment.  That maintains os.environ as a os._Environ object while also performing the desired pop or reset of the environment variables.

### DIFF
--- a/launch/launch/launch_context.py
+++ b/launch/launch/launch_context.py
@@ -169,10 +169,27 @@ class LaunchContext:
     def _pop_environment(self):
         if not self.__environment_stack:
             raise RuntimeError('environment stack unexpectedly empty')
-        os.environ = self.__environment_stack.pop()
+
+        # Note that we cannot just assign os.environ to the copy of the
+        # environment that we saved during _push_environment() above.
+        # That's because os.environ.copy() returns a dict, while os.environ
+        # is a os._Environ object.  They act the same in many, but not all
+        # circumstances.  Instead, when we go to pop the environment, we
+        # clear out the current environment and push the values from the
+        # copy one-by-one into os.environ.
+
+        old_env = self.__environment_stack.pop()
+        os.environ.clear()
+        for k, v in old_env.items():
+            os.environ[k] = v
 
     def _reset_environment(self):
-        os.environ = self.__environment_reset.copy()
+        # See the comment in _pop_environment for why we do this dance.
+
+        old_env = self.__environment_reset.copy()
+        os.environ.clear()
+        for k, v in old_env.items():
+            os.environ[k] = v
 
     def _push_launch_configurations(self):
         self.__launch_configurations_stack.append(self.__launch_configurations.copy())
--- a/launch/launch/launch_context.py
+++ b/launch/launch/launch_context.py
@@ -180,15 +180,12 @@ class LaunchContext:
 
         old_env = self.__environment_stack.pop()
         os.environ.clear()
-        for k, v in old_env.items():
-            os.environ[k] = v
+        os.environ.update(old_env)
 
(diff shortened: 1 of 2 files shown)


clalancette:
FYI, this should be backported to all distributions.

clalancette:
Full CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17441/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11966/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17953/)

launch/launch/launch_context.py
```diff
@@ -169,10 +169,27 @@ def _push_environment(self):
...
+        for k, v in old_env.items():
+            os.environ[k] = v
 
     def _reset_environment(self):
-        os.environ = self.__environment_reset.copy()
+        # See the comment in _pop_environment for why we do this dance.
+
+        old_env = self.__environment_reset.copy()
+        os.environ.clear()
+        for k, v in old_env.items():
```
jacobperron, line 191: I think a copy is not necessary here:
```suggestion
        os.environ.clear()
        for k, v in self.__environment_reset.items():
```

jacobperron (approved):
LGTM pending CI

launch/launch/launch_context.py
clalancette: Oh, good point.  Will fix that.

clalancette: Fixed in 8a78e41

clalancette:
New full CI after fixes:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17442/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11967/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17954/)

launch/launch/launch_context.py
```diff
@@ -169,10 +169,26 @@ def _push_environment(self):
...
+        # That's because os.environ.copy() returns a dict, while os.environ
+        # is a os._Environ object.  They act the same in many, but not all
+        # circumstances.  Instead, when we go to pop the environment, we
+        # clear out the current environment and push the values from the
+        # copy one-by-one into os.environ.
+
+        old_env = self.__environment_stack.pop()
+        os.environ.clear()
+        for k, v in old_env.items():
+            os.environ[k] = v
```
sloretz, line 184: nit, I think after `clear()` this for loop could be replaced with `os.environ.update(old_env)`.

sloretz (commented):
nitpicks, feel free to ignore.

launch/launch/launch_context.py
jacobperron: [1d690bd](https://github.com/ros2/launch/pull/656/commits/1d690bd7848baabad3f9f823f3847e7f3c687b31)

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17444/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11970/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17956/)

clalancette:
@Mergifyio backport humble galactic foxy

PR #667, opened 2022-10-26
Expose sigterm_timeout and sigkill_timeout to xml frontend

adityapande-1995:
This PR exposes ``sigkill_timeout`` and ``sigterm_timeout`` arguments to xml frontends. Added a simple test for parsing using xml.

Signed-off-by: Aditya <aditya050995@gmail.com>

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -352,6 +352,26 @@ class ExecuteProcess(ExecuteLocal):
                     )
                 kwargs['respawn_delay'] = respawn_delay
 
+        if 'sigkill_timeout' not in ignore:
+            sigkill_timeout = entity.get_attr('sigkill_timeout', data_type=float, optional=True)
+            if sigkill_timeout is not None:
+                if sigkill_timeout < 0.0:
+                    raise ValueError(
+                        'Attribute sigkill_timeout of Entity node expected to be '
+                        'a non-negative value but got `{}`'.format(sigkill_timeout)
+                    )
+                kwargs['sigkill_timeout'] = sigkill_timeout
+
+        if 'sigterm_timeout' not in ignore:
+            sigterm_timeout = entity.get_attr('sigterm_timeout', data_type=float, optional=True)
+            if sigterm_timeout is not None:
+                if sigterm_timeout < 0.0:
+                    raise ValueError(
+                        'Attribute sigterm_timeout of Entity node expected to be '
+                        'a non-negative value but got `{}`'.format(sigterm_timeout)
+                    )
+                kwargs['sigterm_timeout'] = sigterm_timeout
+
         if 'shell' not in ignore:
             shell = entity.get_attr('shell', data_type=bool, optional=True)
             if shell is not None:
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -360,7 +360,7 @@ class ExecuteProcess(ExecuteLocal):
                         'Attribute sigkill_timeout of Entity node expected to be '
                         'a non-negative value but got `{}`'.format(sigkill_timeout)
                     )
-                kwargs['sigkill_timeout'] = parser.parse_substitution(str(sigkill_timeout))
+                kwargs['sigkill_timeout'] = str(sigkill_timeout)
 
         if 'sigterm_timeout' not in ignore:
             sigterm_timeout = entity.get_attr('sigterm_timeout', data_type=float, optional=True)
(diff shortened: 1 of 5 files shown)


clalancette (changes requested):
Besides the test failures, can we also add a test for this to `launch_yaml`?

adityapande-1995:
> Besides the test failures, can we also add a test for this to `launch_yaml`?

Done !

wjwwood:
Redundant? https://github.com/ros2/launch/pull/607

launch/launch/actions/execute_process.py
```diff
@@ -352,6 +352,26 @@ def parse(
...
 
+        if 'sigkill_timeout' not in ignore:
+            sigkill_timeout = entity.get_attr('sigkill_timeout', data_type=float, optional=True)
+            if sigkill_timeout is not None:
+                if sigkill_timeout < 0.0:
+                    raise ValueError(
+                        'Attribute sigkill_timeout of Entity node expected to be '
+                        'a non-negative value but got `{}`'.format(sigkill_timeout)
+                    )
+                kwargs['sigkill_timeout'] = parser.parse_substitution(str(sigkill_timeout))
```
wjwwood, line 363: Why does this need `parse_substitution`?

wjwwood:
Since this has tests, maybe we should close https://github.com/ros2/launch/pull/607?

launch/launch/actions/execute_process.py
adityapande-1995: Hmm if I just use ``kwargs['signkill_timeout'] = sigkill_timeout`` here, the tests fail with : 

```

    def normalize_to_list_of_substitutions(subs: SomeSubstitutionsType) -> List[Substitution]:
        """Return a list of Substitutions given a variety of starting inputs."""
        # Avoid recursive import
        from ..substitutions import TextSubstitution
    
        def normalize(x):
            if isinstance(x, Substitution):
                return x
            if isinstance(x, str):
                return TextSubstitution(text=x)
            raise TypeError(
                "Failed to normalize given item of type '{}', when only "
    
        if isinstance(subs, str):
            return [TextSubstitution(text=subs)]
        if is_a_subclass(subs, Substitution):
            return [cast(Substitution, subs)]
>       return [normalize(y) for y in cast(Iterable, subs)]
E       TypeError: 'float' object is not iterable

```

adityapande-1995: Probably because the ``ExecuteLocal`` class is expecting a ``SubstitutionsType`` for ``sigkill_timeout`` : https://github.com/ros2/launch/blob/edf68913560ebb127441aca513519b4f3a2559bd/launch/launch/actions/execute_local.py#L86-L89

ivanpauno: Yeah, but you don't need to use `parse_substitution()` though, only you need the `str()` conversion.

```suggestion
                kwargs['sigkill_timeout'] = str(sigkill_timeout)
```

launch/launch/actions/execute_process.py
```diff
@@ -352,6 +352,26 @@ def parse(
...
+
+        if 'sigterm_timeout' not in ignore:
+            sigterm_timeout = entity.get_attr('sigterm_timeout', data_type=float, optional=True)
+            if sigterm_timeout is not None:
+                if sigterm_timeout < 0.0:
+                    raise ValueError(
+                        'Attribute sigterm_timeout of Entity node expected to be '
+                        'a non-negative value but got `{}`'.format(sigterm_timeout)
+                    )
+                kwargs['sigterm_timeout'] = parser.parse_substitution(str(sigterm_timeout))
```
ivanpauno, line 373: same

```suggestion
                kwargs['sigterm_timeout'] = str(sigterm_timeout)
```

adityapande-1995: Removed : https://github.com/ros2/launch/pull/667/commits/d45ea31bd89dbf4b12660384eaedbf062a95e9f2

launch/launch/actions/execute_process.py
adityapande-1995: Fixed : https://github.com/ros2/launch/pull/667/commits/d45ea31bd89dbf4b12660384eaedbf062a95e9f2

ivanpauno (approved):
LGTM with green CI

adityapande-1995:
CI :
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17609/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12151/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18179/)

PR #678, opened 2022-11-27
Add AndCondition, OrCondition to conditions

LukeSchmitt96:
This pull request adds `AndCondition` and `OrCondition` to the conditions module. These classes can be used to evaluate a list or tuple of strings or `Conditions`. Implementing this change would tidy up somewhat messy conditional statements like

```python
condition=(IfCondition(condition1) and IfCondition(condition2))
```

simplifying them to something like

```python
condition=AndCondition([condition1, condition2])
```

This change addresses the third "combine multiple conditions" option in #105.

### DIFF
--- a/launch/launch/conditions/__init__.py
+++ b/launch/launch/conditions/__init__.py
@@ -14,18 +14,22 @@
 
 """conditions Module."""
 
+from .and_condition import AndCondition
 from .evaluate_condition_expression_impl import evaluate_condition_expression
 from .if_condition import IfCondition
 from .invalid_condition_expression_error import InvalidConditionExpressionError
 from .launch_configuration_equals import LaunchConfigurationEquals
 from .launch_configuration_not_equals import LaunchConfigurationNotEquals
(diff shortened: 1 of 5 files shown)


methylDragon:
Use the And and Or substitutions instead

https://github.com/ros2/launch/blob/rolling/launch/test/launch/substitutions/test_boolean_substitution.py

LukeSchmitt96:
Didn't catch those when reviewing — thanks for the heads up.

eliasdc:
> Use the And and Or substitutions instead
> 
> https://github.com/ros2/launch/blob/rolling/launch/test/launch/substitutions/test_boolean_substitution.py

Not sure if these are equivalent, you can't link existing Conditions with And and  Or substitutions.

Example:
```
AndCondition([
    LaunchConfigurationNotEquals("param_1", "value_1"),
    LaunchConfigurationNotEquals("param_2", "value_2")
]),
```
or how would this be solved?

Issue #676, opened 2022-11-17
Type checking with mypy


haudren-woven:
## Bug report

I am sorry to open an issue in such short order, but as part of my work at Woven Planet, I'm trying to use the ROS2 launch system. Unfortunately, I have found that it is currently unusable from an external package, thus I'd like to improve it.

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - Source
- Version or commit hash:
  - 6783ca8
- DDS implementation:
  - CycloneDDS
- Client library (if applicable):
  - N/A

#### Steps to reproduce issue
1. Create a snippet with this content:
```python
import launch

context = launch.LaunchContext()
reveal_type(context)
```
2. Install mypy
3. Run the following:
```
MYPYPATH=$PATH_TO_WS/install/launch/lib/python3.8/site-packages/ mypy -m launch
```
(mypy does not use PYTHONPATH by default)

#### Expected behavior
```
test.py:4: note: Revealed type is 'launch.launch_context.LaunchContext'
```

#### Actual behavior
```
test.py:1: error: Cannot find implementation or library stub for module named 'launch'
test.py:1: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
test.py:4: note: Revealed type is 'Any'
```

#### Feature description
I'd like to make the type checking in launch (and potentially other ROS2 core libraries) a little bit stricter. This would allow to enforce a quality standard, and root out bugs at the base. It would also make the code easier to use from the outside, since type checking would prevent some common bugs in launch files, such as using a substitution where a condition is expected. Since type-checking is strictly opt-in, there would be no impact on the broader community.

#### Implementation considerations
- I have already implemented adding the `py.typed` file to mark the `launch` package as typed to the broader system, and can submit a PR promptly
- However, there are many type-checking issues in the current codebase, including bugs. For example, mypy reported that [this function](https://github.com/ros2/launch/blob/6783ca8988b3ee1910e3eb6de7388407c8c0ce0d/launch/launch/event_handler.py#L92) does not have a return statement, contradicting its signature. Similarly, [this statement](https://github.com/ros2/launch/blob/6783ca8988b3ee1910e3eb6de7388407c8c0ce0d/launch/launch/frontend/parser.py#L213) has invalid type syntax. Probably someone meant `Tuple`? In total mypy reports 224 errors in 47 files.
[...]

PR #679, opened 2022-12-01
Improve type checking

haudren-woven:
This is still a work in progress, but following @clalancette 's comments on #676 , I made this PR to:
- Introduce the mypy type checker, and check how long it takes to run (On my machine a couple of seconds)
- Show some examples of fixes that are necessary for some files to pass

Note that we will probably need to tweak the mypy configuration a little bit, since currently non-typed imports are treated as errors.

### DIFF
--- a/launch/doc/source/conf.py
+++ b/launch/doc/source/conf.py
@@ -32,6 +32,7 @@
 
 
 # -- Project information -----------------------------------------------------
+# type: ignore
 
 project = 'launch'
 copyright = '2018, Open Source Robotics Foundation, Inc.'  # noqa
--- a/launch/examples/disable_emulate_tty_counters.py
+++ b/launch/examples/disable_emulate_tty_counters.py
@@ -37,10 +37,9 @@ def generate_launch_description():
     ld.add_action(launch.actions.SetLaunchConfiguration('emulate_tty', 'false'))
 
     # Wire up stdout from processes
-    def on_output(event: launch.Event) -> None:
+    def on_output(event: launch.events.process.ProcessIO) -> None:
         for line in event.text.decode().splitlines():
-            print('[{}] {}'.format(
-                cast(launch.events.process.ProcessIO, event).process_name, line))
+            print('[{}] {}'.format(event.process_name, line))
 
     ld.add_action(launch.actions.RegisterEventHandler(launch.event_handlers.OnProcessIO(
         on_stdout=on_output,
--- a/launch/examples/disable_emulate_tty_counters.py
+++ b/launch/examples/disable_emulate_tty_counters.py
@@ -24,7 +24,6 @@ capabilities disabled."
 
 import os
 import sys
-from typing import cast
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..'))  # noqa
 
 import launch  # noqa: E402
--- a/launch/examples/launch_counters.py
+++ b/launch/examples/launch_counters.py
@@ -24,7 +24,6 @@ For an example of expected output, see the file next to this one called
 import os
 import platform
 import sys
-from typing import cast
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..'))  # noqa
 
 import launch  # noqa: E402
--- a/launch/examples/launch_counters.py
+++ b/launch/examples/launch_counters.py
@@ -55,10 +55,9 @@ def main(argv=sys.argv[1:]):
 
     # Setup a custom event handler for all stdout/stderr from processes.
     # Later, this will be a configurable, but always present, extension to the LaunchService.
-    def on_output(event: launch.Event) -> None:
+    def on_output(event: launch.events.process.ProcessIO) -> None:
         for line in event.text.decode().splitlines():
-            print('[{}] {}'.format(
-                cast(launch.events.process.ProcessIO, event).process_name, line))
+            print('[{}] {}'.format(event.process_name, line))
 
     ld.add_action(launch.actions.RegisterEventHandler(launch.event_handlers.OnProcessIO(
         # this is the action     ^              and this, the event handler ^
--- a/launch/launch/__init__.py
+++ b/launch/launch/__init__.py
@@ -32,8 +32,8 @@ from .launch_description_entity import LaunchDescriptionEntity
 from .launch_description_source import LaunchDescriptionSource
 from .launch_introspector import LaunchIntrospector
 from .launch_service import LaunchService
-from .some_actions_type import SomeActionsType
-from .some_actions_type import SomeActionsType_types_tuple
+from .some_entities_type import SomeEntitiesType
+from .some_entities_type import SomeEntitiesType_types_tuple
 from .some_substitutions_type import SomeSubstitutionsType
 from .some_substitutions_type import SomeSubstitutionsType_types_tuple
 from .substitution import Substitution
@@ -57,8 +57,8 @@ __all__ = [
     'LaunchDescriptionSource',
     'LaunchIntrospector',
     'LaunchService',
-    'SomeActionsType',
-    'SomeActionsType_types_tuple',
+    'SomeEntitiesType',
+    'SomeEntitiesType_types_tuple',
     'SomeSubstitutionsType',
     'SomeSubstitutionsType_types_tuple',
     'Substitution',
--- a/launch/launch/action.py
+++ b/launch/launch/action.py
@@ -14,12 +14,12 @@
 
 """Module for Action class."""
 
+from typing import Dict
 from typing import Iterable
 from typing import List
 from typing import Optional
 from typing import Text
 from typing import Tuple
-from typing import Dict
 
 from .condition import Condition
 from .launch_context import LaunchContext
--- a/launch/launch/action.py
+++ b/launch/launch/action.py
@@ -19,6 +19,8 @@ from typing import List
 from typing import Optional
 from typing import Text
 from typing import Tuple
+from typing import Dict
+from typing import cast
 
 from .condition import Condition
 from .launch_context import LaunchContext
@@ -60,9 +62,9 @@ class Action(LaunchDescriptionEntity):
         # Import here for avoiding cyclic imports.
         from .conditions import IfCondition
         from .conditions import UnlessCondition
-        if_cond = entity.get_attr('if', optional=True)
-        unless_cond = entity.get_attr('unless', optional=True)
-        kwargs = {}
+        if_cond = cast(str, entity.get_attr('if', optional=True))
+        unless_cond = cast(str, entity.get_attr('unless', optional=True))
+        kwargs: Dict[str, Condition] = {}
         if if_cond is not None and unless_cond is not None:
             raise RuntimeError("if and unless conditions can't be used simultaneously")
         if if_cond is not None:
--- a/launch/launch/action.py
+++ b/launch/launch/action.py
@@ -20,7 +20,6 @@ from typing import Optional
 from typing import Text
 from typing import Tuple
 from typing import Dict
-from typing import cast
 
 from .condition import Condition
 from .launch_context import LaunchContext
@@ -62,8 +61,8 @@ class Action(LaunchDescriptionEntity):
         # Import here for avoiding cyclic imports.
         from .conditions import IfCondition
         from .conditions import UnlessCondition
-        if_cond = cast(str, entity.get_attr('if', optional=True))
-        unless_cond = cast(str, entity.get_attr('unless', optional=True))
+        if_cond = entity.get_attr('if', optional=True)
+        unless_cond = entity.get_attr('unless', optional=True)
         kwargs: Dict[str, Condition] = {}
         if if_cond is not None and unless_cond is not None:
             raise RuntimeError("if and unless conditions can't be used simultaneously")
--- a/launch/launch/actions/append_environment_variable.py
+++ b/launch/launch/actions/append_environment_variable.py
@@ -17,7 +17,6 @@
 import os
 from typing import List
 from typing import Union
-from typing import cast
 
 from ..action import Action
 from ..frontend import Entity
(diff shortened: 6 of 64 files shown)


haudren-woven:
The test indeed failed, and the run time is indeed around 1 or 2 seconds:
```
10:30:23 test/test_flake8.py .                                                    [  1%]
10:30:31 test/test_mypy.py F                                                      [  2%]
10:30:32 test/test_pep257.py .                                                    [  2%]
```

methylDragon:
If the test is failing, could you set this PR to draft until you have a working version?

haudren-woven:
@methylDragon : Sorry about that, I should have made it draft. I just wanted to address @wjwwood 's concerns quickly. I will convert to draft, and try to fix the type issues, please hold on :pray:

haudren-woven:
@methylDragon : First of all sorry for the long delay. I got really busy in work stuff, then went on holidays. I'm resuming work on this, and one sticky method has been the `Entity.get_attr`. Its behavior (with regards to typing) is a little hard to capture:
- `data_type` in the original signature can be only one of `AllowedTypesType`, which is pretty much either a scalar or a list of scalars. However, in the code it's very frequently used with a `List[Entity]`. Could you maybe shed some light on what is the real type bound on the data types of `get_attr`?
- `can_be_str` is not documented in the abstract class. I understand well that in the case of a scalar, it means that the return type would be a `Union[Scalar, str]`, but what about a list? Can only the individual elements of the list be left as string (e.g. `List[Union[Scalar, str]]`) or is it the whole list (e.g. `Union[List[Scalar], str]`).
- In terms of interaction between `can_be_str` and `optional`, I'd like to make sure that my understanding is correct: when both are set to True, is it fair to say that we will return None if the attribute is absent, a string if the attribute is present but cannot be coerced to the correct type, the type otherwise?

Thank you!

haudren-woven:
Another few questions:

1. Surrounding `OnProcessExit`: in that file, a number of `cast` are performed to basically perform an upcast from functions taking the narrow type `ExecuteLocal` into more general functions taking the abstract `Action` type. I don't quite understand how this can work since `OnActionEventBase` seems to call the action matcher callable with the action itself, without checking if the action that generated the event is indeed a subclass of `target_action_cls`.
2. For the `OpaqueCoroutine`, there seems to be a great deal of confusion between a coroutine and a coroutine function. As far as I can tell, the documentation (and the type annotations) seem to think this `foo` object is a coroutine:

    ```python
        async def foo() -> None:
           pass
    ```
    However, in this case `foo` is a coroutine function, e.g. a `Callable` which returns a `Coroutine`. I can fixup the type annotations fairly easily, but that issue is present in a few places, so I'd like to hear everyone's opinion on the best course here.
3. I was quite surprised to see this code in `some_actions_type`:
    ```python
    SomeActionsType = Union[
        LaunchDescriptionEntity,
        Iterable[LaunchDescriptionEntity],
    ]
    ```
    This causes issues in some places where the return type is still `Iterable[Action]`, which is incompatible with `Iterable[LaunchDescriptionEntity]`. I guess this is the result of some refactoring, but how should we address the situation: keep the name `SomeActionsType` or rename it to `SomeEntitiesType`?

haudren-woven:
@adityapande-1995 @methylDragon @wjwwood @clalancette : any comments / remarks on the above points?

haudren-woven:
@adityapande-1995 @methylDragon @wjwwood @clalancette : Sorry to bother everyone, but could someone take a quick look at the above remarks? Then I can resume work on this PR

methylDragon:
Hey @haudren-woven , sorry for dropping this!

Unfortunately, I don't think I have the necessary context/experience with the launch frontend to give definitive answers for this, but I can give it a shot... (though I think @wjwwood would have more of the necessary context)

> * `data_type` in the original signature can be only one of `AllowedTypesType`, which is pretty much either a scalar or a list of scalars. However, in the code it's very frequently used with a `List[Entity]`. Could you maybe shed some light on what is the real type bound on the data types of `get_attr`?

I think in this case, let's expand the list of allowable types to what's actually being used in source.

Consider the usage of data_type for get_attrs for the `keep` attribute in this xml example (since `keep` attribute access is one of the cases [where the entity returns a list of entities](https://github.com/ros2/launch/blob/6a2bf20fb952da15002b2bca61f56feeb6c0f46f/launch/launch/actions/reset_launch_configurations.py#L65)):
```
        <launch>
            <let name="foo" value="FOO"/>
            <let name="bar" value="BAR"/>
            <reset>
                <keep name="bar" value="$(var bar)"/>
                <keep name="baz" value="BAZ"/>
            </reset>
        </launch>
```

If either `bar` or `baz` were not strings, then to get all the `keep` children of the `reset` entity, the get_attrs method would have to be able to return a **heterogeneous** list of children.

So in this case I think it would be reasonable to allow for `List[Entity]` to **also** be an AllowedTypesType.
Let's then set `data_type` as `Union[AllowedTypesType, List[Entity]]`, I think.

> * `can_be_str` is not documented in the abstract class. I understand well that in the case of a scalar, it means that the return type would be a `Union[Scalar, str]`, but what about a list? Can only the individual elements of the list be left as string (e.g. `List[Union[Scalar, str]]`) or is it the whole list (e.g. `Union[List[Scalar], str]`).

Consider the other available docstrings for `can_be_str` in the other methods:
> In the case a value can be either an instance of a type or a substitution, the `can_be_str` argument of `get_attr` must be used
> :param can_be_str: when True, non-uniform lists mixed with strings are allowed.

I think in no cases do we want to convert a single list entity into a string in its entirety. So I think we want `List[Union[Scalar, str]]`.

The reason for this is to think about how the Entities are expected to be used (parsed entities to be manipulated programatically). If we convert a list entity into a string, then it'd require a re-parse, which I don't think is what the intended use is. I might be wrong about this though.

Also: Ideally we'd end up documenting the `can_be_str` attribute for the abstract class, but I'm not sure if my understanding is correct. I'd defer to @wjwwood for this, but in case this needs to be merged now, I think it'd be wise to add a note/todo, while adding the current understanding of the usage. (Where the Entity can be either a string or a scalar, in the case of a list of entities, a heterogeneous list of scalars/entities AND/OR strings.)

> * In terms of interaction between `can_be_str` and `optional`, I'd like to make sure that my understanding is correct: when both are set to True, is it fair to say that we will return None if the attribute is absent, a string if the attribute is present but cannot be coerced to the correct type, the type otherwise?

I think this is correct.

haudren-woven:
Thank you very much for the thoughtful answers! I will proceed with these, but it might take a little while... Please hold on :pray:

methylDragon:
> Thank you very much for the thoughtful answers! I will proceed with these, but it might take a little while... Please hold on 🙏

Heyyy no worries about this! Take your time!
Feel free to ping me again if I don't do timely replies, and I'll help best I can. Though I'm not holding all of the context

haudren-woven:
@methylDragon : The tests passed :face_exhaling: That was a bit harder that I expected but at least now a lot of things are checked :partying_face: I think the annotations can be improved over time, as there are a number of places that are probably too permissive, but somehow mypy is happy (and found some bugs I inadvertently introduce). I think however that it's a good place to start, and we can progressively improve the type checking in the future.

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18105/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12637/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18751/)

haudren-woven:
@methylDragon : The tests are failing in dependent packages due to being unable to import `SomeActionsType` which I have renamed to `SomeEntitiesType`. One probable culprit is in ros2/launch_ros : https://github.com/ros2/launch_ros/blob/6daacbce4bade7ed40f86f16a30a08b4d7ee9272/launch_ros/launch_ros/event_handlers/on_state_transition.py#L23 .

How would you like to handle the change? I was thinking of reintroducing `some_actions_type` but with a deprecation warning?

methylDragon:
> @methylDragon : The tests are failing in dependent packages due to being unable to import `SomeActionsType` which I have renamed to `SomeEntitiesType`. One probable culprit is in ros2/launch_ros : https://github.com/ros2/launch_ros/blob/6daacbce4bade7ed40f86f16a30a08b4d7ee9272/launch_ros/launch_ros/event_handlers/on_state_transition.py#L23 .
> 
> How would you like to handle the change? I was thinking of reintroducing `some_actions_type` but with a deprecation warning?

I think doing a tick-tock with a deprecation warning makes sense, yep!

haudren-woven:
@methylDragon : I added back the `SomeActionsType` which will hopefully fix the issues in the underlying packages :crossed_fingers:

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18112/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12643/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18760/)

methylDragon:
A bunch of linting fixes to do :>, but it's good news that the tests now pass!

methylDragon:
A bunch of linting fixes to do :>, but it's good news that the tests now pass!

haudren-woven:
Oh no I had to disable my default formatter (black) to work on this codebase, so I forgot to run the other checks... BRB

haudren-woven:
I'm looking into how to run the tests, but locally I can't reproduce the CI errors... Despite running it inside a clean `ros:humble` docker container :thinking: Do you have any tips? `colcon test` reports success:
```
test/test_flake8.py . 
```

But I obviously haven't fixed all the errors :thinking:

haudren-woven:
Ok got it! I need to install `python3-flake8-import-order`, which is somehow not a rosdep? Anyhow, that's great :)

haudren-woven:
@methylDragon 32fa4cc should have fixed the CI errors :)

methylDragon:
> Ok got it! I need to install `python3-flake8-import-order`, which is somehow not a rosdep? Anyhow, that's great :)

:O

haudren-woven:
I meant it doesn't seem to be a key inside the `package.xml` so `rosdep install --from-paths src` doesn't seem to pick it up :thinking: Not sure how it's installed in CI :thinking:

methylDragon:
> I meant it doesn't seem to be a key inside the `package.xml` so `rosdep install --from-paths src` doesn't seem to pick it up 🤔 Not sure how it's installed in CI 🤔

OH! I think you're meant to call `ament_flake8` instead of running flake8 manually :>

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18118/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12649/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18769/)

methylDragon:
Just three more lint errors!

haudren-woven:
Oh no... I'm sorry I really need to learn how to run the linters locally...

haudren-woven:
Oh no... I'm sorry I really need to learn how to run the linters locally...

haudren-woven:
Should be better now :)

methylDragon:
:crossed\_fingers: 

\* Linux \[!\[Build Status\](http://ci.ros2.org/buildStatus/icon?job=ci_linux&build=18123)](http://ci.ros2.org/job/ci_linux/18123/)  
\* Linux-aarch64 \[!\[Build Status\](http://ci.ros2.org/buildStatus/icon?job=ci_linux-aarch64&build=12654)](http://ci.ros2.org/job/ci_linux-aarch64/12654/)  
\* Windows \[!\[Build Status\](http://ci.ros2.org/buildStatus/icon?job=ci_windows&build=18774)](http://ci.ros2.org/job/ci_windows/18774/)

haudren-woven:
Seems to have failed due to random connection dropping...

> 09:55:12 Caused: hudson.remoting.ChannelClosedException: Channel "hudson.remoting.Channel@50a4fcb8:JNLP4-connect connection from ip-10-0-1-61.ec2.internal/10.0.1.61:46536": Remote call on JNLP4-connect connection from ip-10-0-1-61.ec2.internal/10.0.1.61:46536 failed. The channel is closing down or has closed down

methylDragon:
[Build Status](https://ci.ros2.org/job/ci_linux-aarch64/12667/)

haudren-woven:
I see what's going on:
```
12:16:04 test\launch\utilities\test_signal_management.py:55:14: error: Module has no attribute "SIGUSR1"
12:16:04 test\launch\utilities\test_signal_management.py:56:22: error: Module has no attribute "SIGUSR2"
```

This means that these statements are protected by a guard that mypy does not recognize...

haudren-woven:
@methylDragon : Seems to have passed on my machine using `mypy --platform win32`. Now time to :pray: 

I also noticed that `ament_mypy` doesn't allow to pass arbitrary mypy command-line flags, I'm wondering if it's something we should add...

methylDragon:
Windows: \[!\[Build Status\](https://ci.ros2.org/job/ci_windows/18816/badge/icon)](https://ci.ros2.org/job/ci_windows/18816/)

haudren-woven:
Hum two test failures on Windows, but they don't seem super related :thinking: I'll have a closer look but are any flaky tests known?

methylDragon:
No harm in trying again

[Build Status](https://ci.ros2.org/job/ci_windows/18821/)

methylDragon:
Okay, I'm just gonna say that they're unrelated flaky tests that are failing, since its a different set of test failures now :grimacing: 

Merging!! Thanks for the incredible work! :rocket:

methylDragon:
Okay, I'm just gonna say that they're unrelated flaky tests that are failing, since its a different set of test failures now :grimacing: 

Merging!! Thanks for the incredible work! :rocket:

emersonknapp:
@mergifyio backport humble

PR #694, opened 2023-03-06
Remove the import of Literal from entity.py.

clalancette:
Literal is not supported in Python 3.6, which is what is in RHEL-8.  Just remove this bit for now, since I don't actually believe it is needed anyway.

This particular issue was introduced in #679 .  FYI @haudren-woven @methylDragon

### DIFF
--- a/launch/launch/frontend/entity.py
+++ b/launch/launch/frontend/entity.py
@@ -15,7 +15,6 @@
 """Module for Entity class."""
 
 from typing import List
-from typing import Literal
 from typing import Optional
 from typing import overload
 from typing import Text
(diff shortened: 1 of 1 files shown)


mjcarroll (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18168/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12706/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18830/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/337/)

clalancette:
The unstable Windows build has been fixed elsewhere (https://github.com/ros2/rclcpp/pull/2117).  The unstable RHEL tests are known issues (but likely to go away when we switch to RHEL-9).  So I'm going to go ahead and merge this in.

haudren-woven:
So in the context of launch, this ended up being fine but I needed the cast. The problem with removing literal and associated overloads is that you’re relaxing the type contract: when `optional` is False, this function has to return the type. In a few places of the code this shows up, but nothing that a couple of `is not None` checks can’t solve.

Let’s just hope we can stop supporting Python 3.6 soonish…

PR #696, opened 2023-03-07
Add maximum times for a process to respawn

Santti4go:
Hi, on this PR I'm adding the following feature:

- Ability to set a maximum number of times for a process to respawn. `respawn_max_retries`

**Why?**
Sometimes I want a process (a Node) to respawn a few times in case there were any trouble within the system. But if the Node keeps failing over and over again then I want it to stop spawning or it would overflow my logs. To accomplish that I added an optional parameter to set up the max number of retries. By default, i.e. without setting this parameter, the behavior remains the same as it was, infinite retries.

**What changed?**
Added the parameter `respawn_max_retries` to both `ExecuteProcess` and `ExecuteLocal`.
Type: `int` [OPTIONAL]. Default = None, which ends up on infinite respawns.

**How I tested**

I modified the `talker_listener.launch.py` [launch file](https://github.com/ros2/demos/blob/rolling/demo_nodes_cpp/launch/topics/talker_listener.launch.py) inside `demo_nodes_cpp/launch/topics` folder with the following:
```
    return LaunchDescription([
        launch_ros.actions.Node(
            package='demo_nodes_cpp', executable='talker', output='screen'),
        launch_ros.actions.Node(
            package='demo_nodes_cpp', executable='listener', output='screen', respawn=True, respawn_max_retries=1),
    ])
```

1. Run `ros2 launch demo_nodes_cpp talker_listener.launch.py`
Two process spawned, `talker` and `listener` (to verify `ps -a`).

2. Now I killed the listener with `kill -SEGV $(pidof listener)`
On the publisher/listener terminal you should see the listener process died.

3. Verify that the listener process is respawned. Note that I settled `respawn_max_retries`=1.
4. Kill the process again  `kill -SEGV $(pidof listener)`
Now since the process was already respawned 1 time, it won't respawn anymore.

EDIT:
To run the tests added you can 
`colcon test --packages-select launch --return-code-on-test-failure --event-handlers=console_direct+`
Or if you just want to run the tests I added, called `test_execute_process_with_respawn_max_retries` you can run
`colcon test --packages-select launch --return-code-on-test-failure --event-handlers=console_direct+ --pytest-args -k max_retries` and this would inly run the functions I've created.

Signed-off-by: Santiago Aupi

### DIFF
--- a/launch/launch/actions/execute_local.py
+++ b/launch/launch/actions/execute_local.py
@@ -618,10 +618,10 @@ class ExecuteLocal(Action):
                 and not self.__shutdown_future.done()\
                 and self.__respawn and abs(self.__respawn_max_retries) > 0:
             # if self.__respawn_max_retries == None, we respawn the process infinite times
-            # by using abs(self.__respawn_max_retries) we can track the 
+            # by using abs(self.__respawn_max_retries) we can track the
             # total amount of respawns using negative values.
             # discount 1 to self.__respawn_max_retries
-            self.__respawn_max_retries -= 1 
+            self.__respawn_max_retries -= 1
             if self.__respawn_delay is not None and self.__respawn_delay > 0.0:
                 # wait for a timeout(`self.__respawn_delay`) to respawn the process
                 # and handle shutdown event with future(`self.__shutdown_future`)
--- a/launch/launch/actions/execute_local.py
+++ b/launch/launch/actions/execute_local.py
@@ -217,9 +217,11 @@ class ExecuteLocal(Action):
                 self.__respawn_max_retries = respawn_max_retries
             else:
                 self.__logger = launch.logging.get_logger(
-                                self.__process_description.final_name)
-                self.__logger.warning("respawn_max_retries should be a positive integer.\
-                    Setting respawn_max_retries to infinite as default")
+                                    self.__process_description.final_name)
+                self.__logger.warning(
+                    "respawn_max_retries should be a positive integer.\
+                    Setting respawn_max_retries to infinite as default."
+                    )
                 self.__respawn_max_retries = -1
 
         self.__process_event_args = None  # type: Optional[Dict[Text, Any]]
--- a/launch/launch/actions/execute_local.py
+++ b/launch/launch/actions/execute_local.py
@@ -183,7 +183,8 @@ class ExecuteLocal(Action):
         :param: respawn if 'True', relaunch the process that abnormally died.
             Either a boolean or a Substitution to be resolved at runtime. Defaults to 'False'.
         :param: respawn_delay a delay time to relaunch the died process if respawn is 'True'.
-        :param: respawn_max_retries number of times to respawn the process. Defaults to infinite times.
+        :param: respawn_max_retries number of times to respawn the process.
+            Defaults to infinite times.
         """
         super().__init__(**kwargs)
         self.__process_description = process_description
--- a/launch/launch/actions/execute_local.py
+++ b/launch/launch/actions/execute_local.py
@@ -98,7 +98,7 @@ class ExecuteLocal(Action):
         ]] = None,
         respawn: Union[bool, SomeSubstitutionsType] = False,
         respawn_delay: Optional[float] = None,
-        respawn_amount: Optional[int] = None,
+        respawn_max_retries: Optional[int] = None,
         **kwargs
     ) -> None:
         """
@@ -183,7 +183,7 @@ class ExecuteLocal(Action):
         :param: respawn if 'True', relaunch the process that abnormally died.
             Either a boolean or a Substitution to be resolved at runtime. Defaults to 'False'.
         :param: respawn_delay a delay time to relaunch the died process if respawn is 'True'.
-        :param: respawn_amount number of times to respawn the process. Defaults to infinite times.
+        :param: respawn_max_retries number of times to respawn the process. Defaults to infinite times.
         """
         super().__init__(**kwargs)
         self.__process_description = process_description
@@ -209,16 +209,18 @@ class ExecuteLocal(Action):
         self.__respawn = normalize_typed_substitution(respawn, bool)
         self.__respawn_delay = respawn_delay
 
-        if not respawn_amount:
-            # a negative value for self.__respawn_amount means infinite respawns
-            self.__respawn_amount = -1
+        if not respawn_max_retries:
+            # a negative value for self.__respawn_max_retries means infinite respawns
+            self.__respawn_max_retries = -1
         else:
-            if respawn_amount >= 0:
-                self.__respawn_amount = respawn_amount
+            if respawn_max_retries >= 0:
+                self.__respawn_max_retries = respawn_max_retries
             else:
-                self.__logger.warning("respawn_amount should be a positive integer.\
-                    Setting respawn_amount to infinite as default")
-                self.__respawn_amount = -1
+                self.__logger = launch.logging.get_logger(
+                                self.__process_description.final_name)
+                self.__logger.warning("respawn_max_retries should be a positive integer.\
+                    Setting respawn_max_retries to infinite as default")
+                self.__respawn_max_retries = -1
 
         self.__process_event_args = None  # type: Optional[Dict[Text, Any]]
         self._subprocess_protocol = None  # type: Optional[Any]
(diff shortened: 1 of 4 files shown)


launch/launch/actions/execute_local.py
```diff
@@ -207,6 +209,21 @@ def __init__(
...
         self.__respawn_delay = respawn_delay
 
+        if not respawn_max_retries:
+            # a negative value for self.__respawn_max_retries means infinite respawns
+            self.__respawn_max_retries = -1
+        else:
+            if respawn_max_retries >= 0:
+                self.__respawn_max_retries = respawn_max_retries
+            else:
+                self.__logger = launch.logging.get_logger(
```
Santti4go, line 219: What would be the proper way to log this?
I used this because it was used on the same file, but if you think there is a better way please let me know.

ivolis:
[👍](https://emojipedia.org/es/pulgar-hacia-arriba/)

Santti4go:
Friendly ping :) @wjwwood @methylDragon @adityapande-1995

Santti4go:
Hey @wjwwood @methylDragon @adityapande-1995 is there anything I can do from my end to push this PR? 
Please let me know If I should assignee it to someone else.

Santti4go:
> @adityapande-1995 Can you please take a look?

Hey @adityapande-1995 feel free to let me know if there is anything I can do from my end to push this PR!

Santti4go:
Hi @clalancette @adityapande-1995 wondering if you would be interesting on having this feature in current release. Do you need me to do a rebase?

launch/launch/actions/execute_local.py
```diff
@@ -98,6 +98,7 @@ def __init__(
         ]] = None,
         respawn: Union[bool, SomeSubstitutionsType] = False,
         respawn_delay: Optional[float] = None,
+        respawn_max_retries: Optional[int] = None,
```
clalancette, line 101: Since we are using a sentinel value for max_retries, this doesn't have to be optional and can always be an int.  This will also simplify some of the code below:

```suggestion
        respawn_max_retries: int = -1,
```

launch/launch/actions/execute_local.py
```diff
@@ -182,6 +183,8 @@ def __init__(
         :param: respawn if 'True', relaunch the process that abnormally died.
             Either a boolean or a Substitution to be resolved at runtime. Defaults to 'False'.
         :param: respawn_delay a delay time to relaunch the died process if respawn is 'True'.
+        :param: respawn_max_retries number of times to respawn the process.
+            Defaults to infinite times.
```
clalancette, line 187:
```suggestion
            A negative value will respawn an infinite number of times (this is the default behavior).
```

launch/launch/actions/execute_local.py
```diff
@@ -207,6 +210,21 @@ def __init__(
...
+            if respawn_max_retries >= 0:
+                self.__respawn_max_retries = respawn_max_retries
+            else:
+                self.__logger = launch.logging.get_logger(
+                                    self.__process_description.final_name)
+                self.__logger.warning(
+                    "respawn_max_retries should be a positive integer.\
+                    Setting respawn_max_retries to infinite as default."
+                    )
+                self.__respawn_max_retries = -1
```
clalancette, line 226:
```suggestion
        self.__respawn_max_retries = respawn_max_retries
        self.__respawn_retries = 0
```

launch/launch/actions/execute_local.py
```diff
@@ -598,7 +616,12 @@ async def __execute_process(self, context: LaunchContext) -> None:
         if not context.is_shutdown\
                 and self.__shutdown_future is not None\
                 and not self.__shutdown_future.done()\
-                and self.__respawn:
+                and self.__respawn and abs(self.__respawn_max_retries) > 0:
+            # if self.__respawn_max_retries == None, we respawn the process infinite times
+            # by using abs(self.__respawn_max_retries) we can track the
+            # total amount of respawns using negative values.
+            # discount 1 to self.__respawn_max_retries
+            self.__respawn_max_retries -= 1
```
clalancette, line 624:
```suggestion
                and self.__respawn and (self.__respawn_max_retries < 0 or self.__respawn_retries < self.__respawn_max_retries):
            self.__respawn_retries += 1
```

clalancette (changes requested):
I've left some thoughts on how to simplify this implementation and make it easier to understand.

Santti4go:
> I've left some thoughts on how to simplify this implementation and make it easier to understand.

Thanks for the review @clalancette  comments already addressed!

launch/launch/actions/execute_local.py
Santti4go: Replicated the same suggestion on `execute_process.py`

adityapande-1995 (approved):
Looks good ! A minor point though, could you add to the docstring a clause that says that the respawn_max_retires parameter will be ignored if respawn is set to False ? Should be obvious, but just to be sure.

adityapande-1995 (changes requested):
Actually, could you also add the parsing logic over here : https://github.com/ros2/launch/blob/3bc47a5a6c99490a3a0ba5cbdf1342eb77f0ec39/launch/launch/actions/execute_process.py#L342-L355 ?

Currently, this can only be used via the python launch files, ideally we also want this available via XML and YAMlL.

Santti4go:
> Looks good ! A minor point though, could you add to the docstring a clause that says that the respawn_max_retires parameter will be ignored if respawn is set to False ? Should be obvious, but just to be clear.

Nice catch @adityapande-1995 !
I just uploaded that and repeated the same test described in the PR description but using the [XML launch file](https://github.com/ros2/demos/blob/rolling/demo_nodes_cpp/launch/topics/talker_listener_launch.xml) version.
Thanks for your review!

clalancette (changes requested):
Sorry I missed this on the last review, but we should also have a test for this functionality in https://github.com/ros2/launch/blob/rolling/launch/test/launch/test_execute_local.py and https://github.com/ros2/launch/blob/rolling/launch/test/launch/test_execute_process.py .  Once that is in place, I'll be happy with this.

Santti4go:
> Sorry I missed this on the last review, but we should also have a test for this functionality in https://github.com/ros2/launch/blob/rolling/launch/test/launch/test_execute_local.py and https://github.com/ros2/launch/blob/rolling/launch/test/launch/test_execute_process.py . Once that is in place, I'll be happy with this.

Sure thing, I'll work on that as soon as possible and let you know once it's done. @clalancette

launch/test/launch/test_execute_local.py
```diff
@@ -126,6 +126,38 @@ def generate_launch_description():
...
 
 
+def test_execute_process_with_respawn_max_retries():
+    """Test launching a process with respaw_max_retries attribute."""
+    def on_exit_callback(event, context):
+        on_exit_callback.called_count = on_exit_callback.called_count + 1
+    on_exit_callback.called_count = 0
+
+    respawn_max_retries = 2   # we want the process to respawn this amount of times
+    shutdown_time = 10.0   # leave enough time to actually test the respawn_max_retries
```
clalancette, line 136: So the problem with these tests is that they will *always* take 10 seconds to expire, even when we were successful after a very short amount of time.  That's not great for the buildfarm.

Can we instead do something smarter in the `on_exit_callback`, and have it quit early if we get the expected number of calls before the timeout?

clalancette (changes requested):

launch/test/launch/test_execute_local.py
Santti4go: Sure thing, I didn't do it that way at first because I didn't want to disrupt with the other tests.
Please take a look at the new approach and let me know what you think. @clalancette 

Basically what I did was trigger a Timer action (which would trigger a Shutdown signal) on the `on_exit_callback` when the `expected_called_count` is reached. I believe this is better than just trigger a Shutdown action because by waiting for a few seconds before shutting down we are really testing whether or not the `respawn_max_retries` quantity is being respected, i.e. the process doesn't keep respawning itself when it shouldn't.
I decided to leave the original 10secs timer as a safeguard in case something goes wrong one day and the `expected_called_count` is never reached.

Test results comparisson:
With this changes both the `test_execute_local`  and `test_execute_process` tooked 2secs, 5x less than before (see previous images [here](https://github.com/ros2/launch/pull/696#issuecomment-1691849061)).

image

image

clalancette (approved):
This looks good to me.  Thank you for iterating!  I'll run full CI on it next.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19483/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14003/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20200/)

Santti4go:
> This looks good to me. Thank you for iterating! I'll run full CI on it next.

It was a pleasure! Thank you @clalancette  and @adityapande-1995 for you time.
CI is looking good!

Issue #699, opened 2023-04-08
`ros2 launch` bug for `$OVERRIDE_LAUNCH_PROCESS_OUTPUT`


cfveeden:
## Bug report

**Required Info:**

- Operating System:
  - 20.04
- Installation type:
  - Source
- Version or commit hash:
  - Galactic and Humble
- DDS implementation:
  - rmw_cyclonedds_cpp
- Client library:
  - N/A

#### Steps to reproduce issue
```
export ROS_LOG_DIR=~/.ros2/reproduce/
export OVERRIDE_LAUNCH_PROCESS_OUTPUT=screen
ros2 launch your_package your_launch.xml # launch anything that logs
```

#### Expected behavior
The launch specific logging messages should only be output to the screen and should not be written to `launch.log`.

#### Actual behavior
Irrespective of `$OVERRIDE_LAUNCH_PROCESS_OUTPUT`'s value, launch specific logs got to `$ROS_LOG_DIR/*/launch.log`.

#### Additional information

Additionally, it would be good if users could disable launch specific logging entirely since this option is not provided at the moment.

PR #700, opened 2023-04-08
Fix type error and add 'none' option for `$OVERRIDE_LAUNCH_PROCESS_OUTPUT`

cfveeden:
Fixes type error that prevented `$OVERRIDE_LAUNCH_PROCESS_OUTPUT` from having any effect. Also add 'none' as an option.

Issue: #699

### DIFF
--- a/launch/launch/logging/__init__.py
+++ b/launch/launch/logging/__init__.py
@@ -26,6 +26,7 @@ import sys
 
 from typing import Any
 from typing import List
+from typing import Text
 
 from . import handlers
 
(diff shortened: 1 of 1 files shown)


launch/launch/logging/__init__.py
```diff
@@ -340,7 +341,7 @@ def _normalize_output_configuration(config):
     normalized_config = {
         'both': set(), 'stdout': set(), 'stderr': set()
     }
-    if isinstance(config, str):
+    if isinstance(config, Text):
```
clalancette, line 344: Can you please explain why you made this change?  Text is a `typing` type, not a regular Python type, so I'm not sure that this is necessary.

cfveeden: In trying to answer your question, I tried again to reproduce the issue and found that I couldn't. I suspect that I set `$OVERRIDE_LAUNCH_PROCESS_OUTPUT` but did not export it when I first tested. This PR can be closed since the env var actually does take effect.

clalancette: All right, thank you for trying again and the contribution.  I'll close this.

Issue #638, opened 2022-08-02
Empty string in ExecuteProcess command should be ignored


jacobperron:
#### Feature description

When passing a command to `ExecuteProcess`, it would be nice if empty strings were filtered out and not passed to the executable.
For example,

```py
# Executes: "ls foo '' bar"
ExecuteProcess(cmd=['ls', 'foo', '', 'bar'])
```

It would be better, IMO, if the empty string was not passed to the process, and we got the output of `ls foo bar` instead of `ls foo '' bar`.

This came up in the following scenario where we want to construct the command list at runtime via substitutions (like `LaunchConfiguration`):

```py
import launch
from launch.actions import DeclareLaunchArgument
from launch.actions import ExecuteProcess
from launch.substitutions import LaunchConfiguration


def generate_launch_description():
    command = ['./count_args.sh', 'foo', LaunchConfiguration('bar')]
    return launch.LaunchDescription([
        DeclareLaunchArgument('bar', default_value=''),
        ExecuteProcess(cmd=command, output='screen'),
    ])
```

In the above example, `LaunchConfiguration` may result in an empty string, in which case we want to ignore it.

#### Implementation considerations

Since it is plausible that there are use-cases for passing empty strings, we should make sure this is still possible.
We could achieve this by allowing users to pass a pair of nested quotes, e.g.

```py
ExecuteProcess(cmd=['echo', 'foo', "''", 'bar'])
```

jacobperron:
Tagging issues related to `ExecuteProcess` in case we change the syntax in the future: https://github.com/ros2/launch/issues/394 and https://github.com/ros2/launch/issues/263

PR #711, opened 2023-06-15
apply shlex.util to substitutions in cmd given to Executable

wjwwood:
We already do this for the prefix filter, and it should not break any existing uses, because of how `shlex.split` works.

It did prevent passing multiple command line arguments as a single launch configuration, because they would be grouped into a single `argv` entry in the main function of the program it was passed to when `shell=False` is used with subprocess.


wjwwood:
Looks like this breaks some tests, so I need to see why this might not be a safe change to make.

wjwwood:
I don't think this is right place/approach for this fix, and I think these issues are related:

- https://github.com/ros2/launch/issues/394
- https://github.com/ros2/launch/issues/638
- https://github.com/ros2/launch/issues/263
- https://github.com/ros2/launch/pull/454/files#r475755859

adityapande-1995 (commented):
I might be generalizing the problem here, but can we add an option to specify the delimiter here, instead of just a whitespace (which can be the default) ?

``ExecuteProcess`` for instance can be used any executable, some of which can accept arguments like ``some_exec flag1:=flag1Value`` instead of ``some_exec --flag1 flag1Value``, which would also require separation of arguments.

Either way a test for the ``:=`` would be nice.

wjwwood:
But in the case of `some_exec flag1:=flag1Value` the argument parser of the application would need to handle the splitting. I don't actually think we want launch splitting `flag1` and `flag1Value` apart.

The pull request only support whitespace delimiters because that's what shell's support.

> Either way a test for the := would be nice.

What would the test look like? Just something with `:=` in it?

methylDragon (commented):
Is there a chance this will break cases where `$(exec cmd)` type args are passed in?

I tested this with shlex.split() and it seems like it splits it up instead of pre-expanding the command.

```
shlex.split("wow $(echo ls)")
# Out: ['wow', '$(echo', 'ls)']

# Granted a user could always... (and arguably should)
shlex.split('wow "$(echo ls)"')
# Out: ['wow', '$(echo ls)']
```

Similarly, we might want to set `comments=True` on shlex.
Otherwise comments might get interpreted as args (but I'm not too sure about this.)

```
>>> shlex.split('wow ab:=1\n  "some string"  WORKSPACE=a # aa a', comments=True)
['wow', 'ab:=1', 'some string', 'WORKSPACE=a']
>>> shlex.split('wow ab:=1\n  "some string"  WORKSPACE=a # aa a', comments=False)
['wow', 'ab:=1', 'some string', 'WORKSPACE=a', '#', 'aa', 'a']
```

methylDragon (approved):
Other than my comments this looks alright to me.

That's a nifty trick with the tests, using the exit code like that 😬

adityapande-1995:
> But in the case of `some_exec flag1:=flag1Value` the argument parser of the application would need to handle the splitting. I don't actually think we want launch splitting `flag1` and `flag1Value` apart.
> 
> The pull request only support whitespace delimiters because that's what shell's support.
> 
> > Either way a test for the := would be nice.
> 
> What would the test look like? Just something with `:=` in it?

Ah right got it nevermind, shell doesn't need to worry about it. Looks good to me !

adityapande-1995 (approved):
Looks good with green CI !

wjwwood:
> I tested this with shlex.split() and it seems like it splits it up instead of pre-expanding the command.

My understanding is that any of our substitutions should be processed before the changes in this pr get to it, or if you want to use bash-like things like `$(cmd args ...)` then you need to use `shell=True` and then this change has no effect.

And yes, there are cases where the user will just need to quote things to avoid them being split.

Also, I'll mention that the argument parsing that happens in xml/yaml is still very fragile, and I would love to improve the robustness of that, for example something like the xml `<executable cmd="python3 -c &quot;print('$(var some_launch_config)')&quot;" ... />` will break the parser because it tries to `shlex.split()` the string `python3 -c "print('` and then crashes because of unmatched quotes. But this is a pre-existing issue that this pr neither made better or worse. However, I didn't have time to fully track down a solution to that. It's a tricky problem for sure.

> Similarly, we might want to set comments=True on shlex.

I actually considered that, but I thought that might interfere with arguments that contain `#`, but then they could just be quoted and/or escaped, so maybe it would be good to do that, but on the other hand, why include comments in the command like that? I mean I guess it might be from user input or something, but there really shouldn't be comments in the cmd and if there are, then for them to be process, you should use `shell=True`. I don't think it's the responsibility/scope of the `split_arguments` option/idea to handle everything a shell might do.

So I'm kind of :-1: on turning on `comments=True` for `shlex.split()`, but I think I could be convinced otherwise.

wjwwood:
Thanks for the reviews so far!

CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19549/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14063/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20266/)

wjwwood:
Looks like the `shlex.split()` may be breaking something on Windows with the `\` path separators.

wjwwood:
I think [d469612](https://github.com/ros2/launch/pull/711/commits/d469612e36f81a2c9a98e5e270d9f0271a9d03fc) should fix it (and I think the right way, based on the docs and some discussions I found online).

New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19561/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14074/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20285/)

launch/launch/actions/execute_local.py
```diff
@@ -73,6 +74,8 @@
 from ..utilities.type_utils import normalize_typed_substitution
 from ..utilities.type_utils import perform_typed_substitution
 
+g_is_windows = 'win' in platform.system().lower()
```
wjwwood, line 77: I'll also comment that this is a good way to detect Windows proper vs Cygwin on Windows, versus other ways of detecting windows, like `os.name`.

methylDragon:
> > I tested this with shlex.split() and it seems like it splits it up instead of pre-expanding the command.
> 
> My understanding is that any of our substitutions should be processed before the changes in this pr get to it, or if you want to use bash-like things like `$(cmd args ...)` then you need to use `shell=True` and then this change has no effect.
> 
> And yes, there are cases where the user will just need to quote things to avoid them being split.
> 
> Also, I'll mention that the argument parsing that happens in xml/yaml is still very fragile, and I would love to improve the robustness of that, for example something like the xml `<executable cmd="python3 -c &quot;print('$(var some_launch_config)')&quot;" ... />` will break the parser because it tries to `shlex.split()` the string `python3 -c "print('` and then crashes because of unmatched quotes. But this is a pre-existing issue that this pr neither made better or worse. However, I didn't have time to fully track down a solution to that. It's a tricky problem for sure.
> 
> > Similarly, we might want to set comments=True on shlex.
> 
> I actually considered that, but I thought that might interfere with arguments that contain `#`, but then they could just be quoted and/or escaped, so maybe it would be good to do that, but on the other hand, why include comments in the command like that? I mean I guess it might be from user input or something, but there really shouldn't be comments in the cmd and if there are, then for them to be process, you should use `shell=True`. I don't think it's the responsibility/scope of the `split_arguments` option/idea to handle everything a shell might do.
> 
> So I'm kind of 👎 on turning on `comments=True` for `shlex.split()`, but I think I could be convinced otherwise.

No issues on both of those!

I was thinking comments in the command that's passed could be used as comments. But all contexts in which you'd conceivably use this would also have the ability to have comments, so there's no need for that.

Looking at the windows changes

launch/launch/actions/execute_local.py
```diff
@@ -714,6 +750,18 @@ def execute(self, context: LaunchContext) -> Optional[List[LaunchDescriptionEnti
...
+            # Update the cmd to respect split_arguments option.
+            if evaluate_condition_expression(context, self.__split_arguments):
+                if self.__shell:
+                    self.__logger.debug("Ignoring 'split_arguments=True' because 'shell=True'.")
+                else:
+                    self.__logger.debug("Splitting arguments because 'split_arguments=True'.")
+                    expanded_cmd = []
+                    assert self.__process_event_args is not None
+                    for token in self.__process_event_args['cmd']:
+                        expanded_cmd.extend(shlex.split(token, posix=g_is_windows))
```
methylDragon, line 762: One thing I'm confused by is...

It seems that the posix argument is default `True` according to the [`shlex` docs](https://docs.python.org/3/library/shlex.html#shlex.split)

I'm not sure why specifying this changes the behavior for windows? :o
If I read this correctly, it'll only change the behavior for non-windows platforms (setting posix to false for those platforms.)

methylDragon (approved):
Just a question, but otherwise I'm approving this conditioned on green CI

launch/launch/actions/execute_local.py
wjwwood: Well, I didn't test on Windows, so it's probably wrong? ... Let me add a test for the opposite for Posix. My guess is that it will still fail on Windows and doesn't matter what it is on Linux.

It's a good catch, sorry I'll have another look.

methylDragon: Speaking of which, it might also be good to add a testcase with a backslash in your tests :thinking:

methylDragon: You might be able to get by with just negating the posix argument:
```
>>> shlex.split("a\\w")
['aw']
>>> shlex.split("a\\w", posix=False)
['a\\w']
>>> shlex.split("a\\w", posix=False)
['a\\w']
```

methylDragon:
```suggestion
                        expanded_cmd.extend(shlex.split(token, posix=(not g_is_windows)))
```

adityapande-1995: +1

wjwwood: Ok, [0c5a017](https://github.com/ros2/launch/pull/711/commits/0c5a017e817122b01ea1bcd82d57c5e8ba94feee) should address this.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19570/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14082/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20294/)

methylDragon (approved):

methylDragon:
> CI:
> 
> * Linux [Build Status](http://ci.ros2.org/job/ci_linux/19570/)
> * Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14082/)
> * Windows [Build Status](http://ci.ros2.org/job/ci_windows/20294/)

Return code `None` 😱 😱 😱

wjwwood:
Yeah, there's still an issue on Windows. I'm in the process of spinning up a workspace on my Windows VM to debug it.

PR #712, opened 2023-06-18
Fix #666 handle SIGTERM

### LINKED ISSUE #666: Incorrect signal handling
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04 & Docker Hub ROS2 Humble images
- Installation type:
  - binaries
- Version or commit hash:
  - Humble
- DDS implementation:
  - Fast-RTPS & Cyclone
  - rclcpp & rclpy

#### Steps to reproduce issue

Start a ROS2 Launch process which spawns at least one node, such as

```
ros2 launch robot_state_publisher rsp-launch-urdf-inline.xml --noninteractive
```

Send a SIGTERM to the ROS2 launch process with

```
kill -TERM $ros2_launch_PID
```

**or**

Use the official ROS2 OCI compliant containers from Docker Hub and run with Docker or Podman

```
podman run --rm --name ros_signal_test ros:humble ros2 launch robot_state_publisher rsp-launch-urdf-inline.xml
```

And then stop the container with

```
podman stop ros_signal_test
```

#### Expected behavior

ROS2 Launch should catch the SIGTERM and propagate that signal to each of its child nodes.

#### Actual behavior

SIGTERM is ignored in the first instance, and in the second containerized example SIGTERM is still ignored but all processes are SIGKILL'd after the `podman stop` timeout.

#### Additional information

The [ROS2 Launch design article](https://design.ros2.org/articles/roslaunch.html) says

_"if someone attempts to SIGTERM the launch system, they probably did so out of impatience after sending SIGINT to the launch system, and therefore the launch system should attempt to exit quickly."_

This is not an accurate assumption.

The [Linux signal man page](https://man7.org/linux/man-pages/man7/signal.7.html) defines, according to the POSIX.1-1990 standard, "SIGINT" as an "Interrupt from keyboard" and "SIGTERM" as a "Termination signal".

The [GNU manual](https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html) entry on termination signals goes further, defining "SIGINT" as being "sent when the user types the INTR character (normally C-c)" and "SIGTERM" as "a generic signal used to cause program termination".

Neither of these sources make reference to any hierarchy or precedence of these two signals.

Users of the ROS2 Launch system are likely to first run into this issue when try to automate the running of their ROS2 nodes on embedded systems, in systemd services, podman/docker containers, or both!

By default [systemd services send SIGTERMs](https://www.freedesktop.org/software/systemd/man/systemd.kill.html) to their processes on shutdown, followed by SIGKILLs after a timeout.

Similarly, [Podman](https://docs.podman.io/en/latest/markdown/podman-stop.1.html)/[Docker](https://docs.docker.com/engine/reference/commandline/stop/) containers get sent a SIGTERM to PID1 when running, for example, `podman stop ros_signal_test`. 

The issue with containers cropped up here https://github.com/ros2/launch/issues/473, but it was only resolved for SIGINTs.
A the issues of SIGTERMs not being handled was mentioned here https://github.com/ros2/launch/issues/545#issuecomment-1191890187, but I've not been able to find any progress on it.

My current workaround for clean termination of an embedded ROS2 Launch system is to set `KillMode=SIGINT` for systemd services, or run Podman/Docker containers with the `--stop-signal=SIGINT` flag or set `STOPSIGNAL SIGINT` in the Dockerfile.
This only works when ROS2 Launch either detects it's running in non-interactive mode or that is set explicitly with `--noninteractive` https://github.com/ros2/launch/pull/475.

However, it's oxymoronic to have to both be in "non-interactive" mode and to send a "keyboard interrupt" signal (SIGINT) for clean termination!

----
#### Feature description
Have the ROS2 Launch system catch SIGTERMs and forward the same signal to their child nodes.

Windows??

ciandonovan:
Fixes #666

### SIGTERM

Instead of installing signal handlers with the OS directly, ROS2 Launch uses Python's `signal.set_wakeup_fd` to queue and forward signals to handlers registered from the async event loop.

Although one of these handlers has been set for SIGTERM, it never gets called since when that signal gets delivered, the kernel terminates the process immediately. The process never changes the default disposition of that signal. Delivery of SIGINT however, doesn't cause the kernel to terminate the process, and instead that signal get dutifully forwarded to the appropriate handler. This is despite there being no difference in the ROS2 Launch code in setting them both up.

After running an strace on a one-line Python program and seeing that a handler was still automatically installed for SIGINT, I came across this StackOverflow post. Depending on whether a Python script is run in some combination of foreground process and in an interactive shell, a handler will be installed for SIGINT for use in the `KeyboardInterrupt` exception. It just so happens that it also gets forwarded by `signal.set_wakeup_fd`, and since the handler prevents the program from being automatically terminated, it can actually do so.

This is why the handler for SIGINT by chance works, but the one for SIGTERM does not.

The fix for this was to simply install the default interrupt signal handler on any signal the async event loop requested handled, so that they would no longer cause the program to terminate and instead be forwarded to `signal.set_wakeup_fd`.

### SIGQUIT

SIGQUIT signals are often generated by the user hitting 'CTRL+\' in an interactive shell. However, when the terminal driver receives those characters, it doesn't just send a SIGQUIT to the running process, but to the entire foreground process group.

By default, the ROS2 Client Libraries don't register a handler for SIGQUIT, as they do for SIGINT and SIGTERM, so the ROS2 node will be terminated immediately and its core dumpted on receipt. So there is little reason for ROS2 Launch to catch this signal or the purposes of forwarding it on to its children.

Also, not catching SIGQUIT means that if there is a problem with the Python signal handling, then the user still has available a shortcut on their keyboards to unilaterally kill the process.

The signal handler for SIGQUIT was removed.

### Future work

This Pull Request allows ROS2 Launch to handle SIGTERMs and initiate shutdown, but sends SIGINTs instead of SIGTERMs to the child processes. This will usually be fine since the ROS2 Client Libraries install handlers for both SIGINT and SIGTERM by default, and is no worse than the current situation, but it would be more flexible to allow users to send whatever signals they want to their nodes.

To do this would require removing the `due_to_sigint` and `send_sigint` arguments of many functions and replacing them with an argument passing the name of the signal to be forwarded. Care would need to be taken of other signals besides SIGINT such as SIGQUIT and SIGTSTP that also tend to get sent to the entire process group to avoid duplicates. Currently if the child node doesn't shutdown within 5 seconds an "escalated" SIGTERM is sent - but if SIGTERM was the initial signal then should it progress straight to SIGKILL? Or wait 10 seconds?

If others feel this is worth developing let me know and I'll work on another PR, but I feel this is in good shape now and will already be of benefit to others upstream.

@wjwwood - mentioned in the relevant ToDo for this feature

### DIFF
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -185,15 +185,16 @@ class LaunchService:
 
             self.__this_task = this_task
             # Setup custom signal handlers for SIGINT & SIGTERM
-            signals_received: dict = {}
+            signals_received = set()
 
             def _on_signal(signum):
                 nonlocal signals_received
-                base_msg = 'caught ' + signal.Signals(signum).name
-                if signum not in signals_received or signals_received[signum] is not True:
-                    signals_received[signum] = True
+                signal_name = signal.Signals(signum).name
+                base_msg = 'caught ' + signal_name
+                if signal_name not in signals_received:
+                    signals_received.add(signal_name)
+                    due_to_sigint = True if signal_name == 'SIGINT' else False
                     self.__logger.warning(base_msg)
-                    due_to_sigint = True if signal.Signals(signum).name == 'SIGINT' else False
                     ret = self._shutdown(
                         reason=base_msg, due_to_sigint=due_to_sigint, force_sync=True
                     )
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -185,7 +185,7 @@ class LaunchService:
                 this_task = asyncio.Task.current_task(this_loop)
 
             self.__this_task = this_task
-            # Setup custom signal handlers for SIGINT, SIGTERM and maybe SIGQUIT.
+            # Setup custom signal handlers for SIGINT & SIGTERM
             sigint_received = False
 
             def _on_sigint(signum):
@@ -214,8 +214,6 @@ class LaunchService:
                 # Setup signal handlers
                 manager.handle(signal.SIGINT, _on_sigint)
                 manager.handle(signal.SIGTERM, _on_sigterm)
-                if platform.system() != 'Windows':
-                    manager.handle(signal.SIGQUIT, _on_sigterm)
                 # Yield asyncio loop and current task.
                 yield this_loop, this_task
         finally:
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -186,34 +186,25 @@ class LaunchService:
 
             self.__this_task = this_task
             # Setup custom signal handlers for SIGINT & SIGTERM
-            sigint_received = False
+            signals_received: dict = {}
 
-            def _on_sigint(signum):
-                nonlocal sigint_received
-                base_msg = 'user interrupted with ctrl-c (SIGINT)'
-                if not sigint_received:
+            def _on_signal(signum):
+                nonlocal signals_received
+                base_msg = 'caught ' + signal.Signals(signum).name
+                if signum not in signals_received or signals_received[signum] is not True:
+                    signals_received[signum] = True
                     self.__logger.warning(base_msg)
                     ret = self._shutdown(
-                        reason='ctrl-c (SIGINT)', due_to_sigint=True, force_sync=True
+                        reason=base_msg, due_to_sigint=True, force_sync=True
                     )
                     assert ret is None, ret
-                    sigint_received = True
                 else:
                     self.__logger.warning('{} again, ignoring...'.format(base_msg))
 
-            def _on_sigterm(signum):
-                signame = signal.Signals(signum).name
-                self.__logger.error(
-                    'user interrupted with ctrl-\\ ({}), terminating...'.format(signame))
-                # TODO(wjwwood): try to terminate running subprocesses before exiting.
-                self.__logger.error('using {} can result in orphaned processes'.format(signame))
-                self.__logger.error('make sure no processes launched are still running')
-                this_loop.call_soon(this_task.cancel)
-
             with AsyncSafeSignalManager(this_loop) as manager:
                 # Setup signal handlers
-                manager.handle(signal.SIGINT, _on_sigint)
-                manager.handle(signal.SIGTERM, _on_sigterm)
+                manager.handle(signal.SIGINT, _on_signal)
+                manager.handle(signal.SIGTERM, _on_signal)
                 # Yield asyncio loop and current task.
                 yield this_loop, this_task
         finally:
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -193,8 +193,9 @@ class LaunchService:
                 if signum not in signals_received or signals_received[signum] is not True:
                     signals_received[signum] = True
                     self.__logger.warning(base_msg)
+                    due_to_sigint = True if signal.Signals(signum).name == 'SIGINT' else False
                     ret = self._shutdown(
-                        reason=base_msg, due_to_sigint=True, force_sync=True
+                        reason=base_msg, due_to_sigint=due_to_sigint, force_sync=True
                     )
                     assert ret is None, ret
                 else:
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -18,7 +18,6 @@ import asyncio
 import collections.abc
 import contextlib
 import logging
-import platform
 import signal
 import threading
 import traceback
--- a/launch/launch/utilities/signal_management.py
+++ b/launch/launch/utilities/signal_management.py
@@ -235,6 +235,7 @@ class AsyncSafeSignalManager:
         :return: previous handler if any, otherwise None
         """
         signum = signal.Signals(signum)
+        signal.signal(signum, signal.default_int_handler)
         if handler is not None:
             if not callable(handler):
                 raise ValueError('signal handler must be a callable')
--- a/launch/test/launch/utilities/test_signal_management.py
+++ b/launch/test/launch/utilities/test_signal_management.py
@@ -25,19 +25,19 @@ import osrf_pycommon.process_utils
 
 
 def cap_signals(*signals):
-    def _noop(*args):
-        pass
-
     def _decorator(func):
         @functools.wraps(func)
         def _wrapper(*args, **kwargs):
             handlers = {}
             try:
                 for s in signals:
-                    handlers[s] = signal.signal(s, _noop)
+                    handlers[s] = signal.signal(s, signal.default_int_handler)
                 return func(*args, **kwargs)
+            except KeyboardInterrupt:
+                pass
             finally:
-                assert all(signal.signal(s, h) is _noop for s, h in handlers.items())
+                assert all(signal.signal(s, h) is signal.default_int_handler
+                           for s, h in handlers.items())
         return _wrapper
 
     return _decorator


ciandonovan:
I had to modify `test_signal_management.py` to account for the fact I'm using `signal.default_int_handler` by catching and ignoring `KeyboardInterrupt` like in the main code.

Maybe some sort of global scope `noop` function might be more appropriate, needs to be global otherwise the `assert` won't pass with separate `noop` functions in the main and test code.

launch/launch/launch_service.py
```diff
@@ -185,37 +184,26 @@ def _prepare_run_loop(self):
...
+                base_msg = 'caught ' + signal.Signals(signum).name
+                if signum not in signals_received or signals_received[signum] is not True:
                     self.__logger.warning(base_msg)
                     ret = self._shutdown(
-                        reason='ctrl-c (SIGINT)', due_to_sigint=True, force_sync=True
+                        reason=base_msg, due_to_sigint=True, force_sync=True
                     )
                     assert ret is None, ret
-                    sigint_received = True
+                    signals_received[signum] = True
```
fujitatomoya, line 199: i believe this store should be done right after if statement.

launch/launch/launch_service.py
```diff
@@ -185,37 +184,26 @@ def _prepare_run_loop(self):
...
-                base_msg = 'user interrupted with ctrl-c (SIGINT)'
-                if not sigint_received:
+            def _on_signal(signum):
+                nonlocal signals_received
+                base_msg = 'caught ' + signal.Signals(signum).name
+                if signum not in signals_received or signals_received[signum] is not True:
                     self.__logger.warning(base_msg)
                     ret = self._shutdown(
-                        reason='ctrl-c (SIGINT)', due_to_sigint=True, force_sync=True
+                        reason=base_msg, due_to_sigint=True, force_sync=True
```
fujitatomoya, line 196: the name should be changed into `due_to_signal`? cz it is not only with sigint anymore.

```suggestion
                        reason=base_msg, due_to_signal=True, force_sync=True
```

ciandonovan:
@fujitatomoya has been a long time since I worked on this, some other regressions were introduced during the rebase also. Agree with your suggestions, but might rework the patches more in general over the next week, take a fresh look at it.
Glad to see interest in getting this resolved, happy to work with you if you have further suggestions or ideas on approach. maybe aligning with that's being done in https://github.com/ros2/ros2cli/issues/895?

fujitatomoya:
@ciandonovan i believe your suggestion posted on here makes sense. `launch` and `run` process does not really know how to handle the signals, so that they should forward the signals to the application logic to shutdown or finalize the application. application by here could be, storing some data using memory cache, accessing physical hardware device or sensors. in these cases, application logics need to receive the signals to shutdown or finalize it properly, otherwise it could be left unexpected state or error result.

ciandonovan:
@fujitatomoya @wjwwood @clalancette is there a case for catching and handling any signals other than SIGINT and SIGTERM?

Currently only SIGINT is handled as no signal handler on the process level is ever installed for anything else, despite the current code assuming so.

My pull request as it stands catches both SIGINT and SIGTERM, and forwards SIGINT to the child processes in both cases. I wrote it to handle any signals installed to the AsyncSafeSignalManager.handle(), but this flexibility might not be required? Also, it doesn't forward any signal if launch is in "noninteractive" mode, which shouldn't be the case for SIGTERM at least as it has different semantics to SIGINT (which is commonly sent to the entire foreground process group on Ctrl-C from the terminal).

If it is the case that only those two signals are of interest, I'll adjust to handle both accordingly - both forwarded when nonintereactive, and only SIGTERM forwarded when interactive. The ROS 2 Client Libraries already handle both signals gracefully so shouldn't introduce any unexpected issues by forwarding SIGTERM.

launch/launch/launch_service.py
ciandonovan: It should be more generic yes, but currently the semantics of SIGINT are different to other signals, as when it's combined with the `noninteractive` flag the signal is either forwarded or not. Will require further changes in different files that rely on this variable, looking into that now.

fujitatomoya:
@ciandonovan i am not sure what signals `launch` process should deal, because `launch` does and provides much more than `ros2 run` that is just a wrapper to execute the command from the arguments and parameters. that is the main reason that i just forward the signals to child process how they want to deal with https://github.com/ros2/ros2cli/issues/895. in general sigint and sigterm need to be handled, but there could be specific signal that `launch` should handle... that is something i would like to know too.

ciandonovan:
```
    send_sigint? True : False
    +----------------+--------+---------+
    |                | SIGINT | SIGTERM |
    +----------------+--------+---------+
    | interactive    | False  | True    |
    | noninteractive | True   | True    |
    +----------------+--------+---------+
``` 
Launch either detects, through checking if there's a controlling terminal, or the value of the flag `--noninteractive`, to determine if launch was launched interactively or not.

If it was launched interactively, then the receipt of SIGINT is implicitly assumed to have come from the user pressing Ctrl-C
and their terminal sending a SIGINT to the entire foreground process group. In such case, it is redundant to also send SIGINT from launch to the child ROS nodes, however it should be noted that even if another signal was sent there would be no ill-effects.

In all other cases, assume the child nodes were not themselves signalled, so manually send SIGINT, and later escalate to SIGTERM after timeout, to the nodes.

ciandonovan:
@fujitatomoya from reading the Linux signal [man page](https://www.man7.org/linux/man-pages/man7/signal.7.html) it's hard to make a case for handling any signals other than SIGINT and SIGTERM.

The signals SIGABRT, SIGALRM, SIGCHLD, SIGIO, SIGLOST, SIGPIPE, SIGPOLL, SIGURG, SIGVTALRM,  are raised in response to actions taken by the program itself by design.
The signals SIGEMT, SIGPROF, SIGTRAP are only relevant for in-built debugging and profiling (not relevant here). 
The signals SIGBUS, SIGFPE, SIGILL, SIGSEGV, SIGSYS are unlikely to be raised in a Python program.
The signals SIGCONT, SIGSTOP, SIGTSTP, SIGTTIN, SIGTTOU shouldn't be handled as is their actions are normal expected behavior. 
The other signals are either unused, synonyms, irrelevant unhandleable, or would be delivered to all processes anyway.

I had considered SIGHUP, but then again if a user explicitly wants their remote processes to continue after a terminal disconnect, then that's already possible with the well-known `nohup`, so doesn't make sense to alter that default behavior.

Further, only SIGINT and SIGTERM are currently handled by the ROS 2 Client Libraries used by the nodes, so forwarding arbitrary other signals to them would just cause them to immediately terminate anyway, which is not desired.

I think there could be a bigger conversation around changing some of the existing semantics of Launch, in particular I don't think the current behavior of sending a SIGINT, then a SIGTERM after timeout makes much sense. The POSIX standard doesn't specify any order of precedence between them, and rcl handles them both identically. There is no escalation in practice, nor do the signals' semantics indicate so either. There's also code for handling SIGKILL which just isn't possible as that's an uncatchable signal. However, that should probably be addressed in a separate pull request.

As it stands, this pull request is essentially a bugfix. Its behavior is now as it seems it was designed for from reading the docs, the code's comments, and its structure. SIGTERM was never actually handled before at all, as no handler was installed, and now it works quite nicely both being run as a systemd service and in a Podman/Docker container. Also works nice as a Podman container running inside a systemd service ;)

launch/launch/launch_service.py
ciandonovan: Decided to keep the original `due_to_sigint` as for now we're only handling two signals, SIGINT & SIGTERM, so `due_to_sigint=False` unambiguously implies SIGTERM.

launch/launch/utilities/signal_management.py
```diff
@@ -235,6 +235,7 @@ def handle(
         :return: previous handler if any, otherwise None
         """
         signum = signal.Signals(signum)
+        signal.signal(signum, signal.default_int_handler)
```
hidmic, line 238: @ciandonovan thanks for catching this! I dropped the ball here, my bad.

That said, I don't think this is the complete fix. We should only override default signal handlers (chaining non-default ones), and we should restore whatever was in place on `AsyncSafeSignalManager` exit.

launch/test/launch/utilities/test_signal_management.py
```diff
@@ -25,19 +25,19 @@
...
         @functools.wraps(func)
         def _wrapper(*args, **kwargs):
             handlers = {}
             try:
                 for s in signals:
-                    handlers[s] = signal.signal(s, _noop)
+                    handlers[s] = signal.signal(s, signal.default_int_handler)
                 return func(*args, **kwargs)
+            except KeyboardInterrupt:
+                pass
```
hidmic, line 37: @ciandonovan hmm, this will skip the entire test with a green check. The `asyncio` loop won't deliver any handlers if it is interrupted.

launch/launch/launch_service.py
```diff
@@ -185,37 +184,28 @@ def _prepare_run_loop(self):
...
+                nonlocal signals_received
+                signal_name = signal.Signals(signum).name
+                base_msg = 'caught ' + signal_name
+                if signal_name not in signals_received:
+                    signals_received.add(signal_name)
+                    due_to_sigint = True if signal_name == 'SIGINT' else False
                     self.__logger.warning(base_msg)
                     ret = self._shutdown(
-                        reason='ctrl-c (SIGINT)', due_to_sigint=True, force_sync=True
+                        reason=base_msg, due_to_sigint=due_to_sigint, force_sync=True
```
hidmic, line 199: @ciandonovan meta: sending a shutown event is not the same as pulling the brakes on the event loop, but I can see how the previous implementation can cause problems when managing `launch` itself. Most process managers use SIGTERM. I guess you could argue that SIGKILL escalation is appropriate if a shutdown isn't enough.

hidmic (commented):
@ciandonovan thanks for cleaning after me :) 

About signal escalation vs. signal forwarding. `launch` was designed to work with non-ROS aware executables too. SIGINT and SIGTERM handling isn't equivalent in general. It does make sense, however, to bypass some of the signal escalation stages based on the signal that triggered it. Essentially skip the SIGINT stage on SIGTERM induced shutdowns, as you suggest somewhere above. HTH.

launch/launch/utilities/signal_management.py
ciandonovan: In this case I'm not overriding any default handlers, as SIGINT is already handled by `signal.default_int_handler`, and SIGTERM isn't handled at all, so I'm just setting it to be handled by the same function.

Ideally both would be handled by some global _noop function that just passes, letting `signal.set_wakeup_fd` pick it up and do its job. Not sure the best way to implement that though. There was a _noop function in `test_signal_management.py` previously, but the assert fails if a signal handler was installed for SIGTERM as it's pointed to a different handler address.

launch/test/launch/utilities/test_signal_management.py
ciandonovan: Could you clarify why, or what it's interrupting? My understanding is that the `assert` in the `finally` section will always get run no matter what exceptions occur so long as they're handled.

launch/launch/utilities/signal_management.py
hidmic: > In this case I'm not overriding any default handlers, as SIGINT is already handled by `signal.default_int_handler`, and SIGTERM isn't handled at all, so I'm just setting it to be handled by the same function.

Right, but we don't really know which signal number we will be required to handle, and even we did (or assumed so or constrained to be such and such), there is also the possibility that some other library, used by `launch` or using `launch`, may have set their own as well.

launch/test/launch/utilities/test_signal_management.py
hidmic: Right, those `assert` will be executed (and likely succeed), but all the `assert` in the `test_async_safe_signal_manager` body after the first signal is raised won't because a `KeyboardInterrupt` exception will be raised during `asyncio.run_until_complete`.

launch/launch/utilities/signal_management.py
wjwwood: So what should we do here? Is it ok to just override any existing signal handler with when this is called? How does this interact with `signal.set_wakeup_fd`? What if someone sets their own signal handler after initializing launch, will our handling still function?

wjwwood:
We've been discussing this pr and the ros2run signal handling pr (https://github.com/ros2/ros2cli/pull/899) in our Jazzy release channels and I think the plan is to make this a bugfix shortly after the release, and after it has some time to soak in Rolling.

So we'll try to get these merged, and then backport after the release as a bugfix.

launch/launch/utilities/signal_management.py
ciandonovan: The only existing signal handler at this point is the one the Python binary sets itself for SIGINT to generate the `KeyboardInterrupt` exception, which I don't change as I point all handlers towards that.
The only other possibility is that someone could have created a handler within their launchfile? Not sure if that's even possible though, and would be terrible practice anyway.
Signal handlers they install in their nodes are of course separate, as signals are a per-process property, and launch is but the parent of the node children.
In terms of `signal.set_wakeup_fd`, nothing extra needs to be done. Installing the signal handlers here simply prevents the kernel from immediately terminating the process as is the common default disposition, and allows Python to then forward the signal through the fd.

martincerven:
I've been running `ros2 launch` from compose and only with `stop_signal: SIGINT` I got proper shutdown.

jlack1987:
Any updates here or plans to move this forward? Just found my way here after realizing docker uses a default SIGTERM signal and it was causing my processes to not shutdown gracefully

PR #717, opened 2023-07-12
to open expected outpout file with an encoding parameter

iuhilnehc-ynos:
I need to add a test in https://github.com/ros2/rcutils/pull/426, but it does not work if a regex file contains some escape character. (e.g. "\b")


fujitatomoya (approved):

fujitatomoya (approved):
lgtm

fujitatomoya:
@adityapande-1995 could we have another review?

launch_testing/launch_testing/tools/output.py
```diff
@@ -64,7 +65,7 @@ def _filter(output):
     return _filter
 
 
-def expected_output_from_file(path):
+def expected_output_from_file(path, encoding: Optional[str] = None):
```
clalancette, line 68: While we are adding in new type annotations, can we add one to path as well?

```suggestion
def expected_output_from_file(path: str, encoding: Optional[str] = None):
```

clalancette (changes requested):

launch_testing/launch_testing/tools/output.py
iuhilnehc-ynos: Yes. Updated in https://github.com/ros2/launch/pull/717/commits/e445d13fa35765fbcff33b0271991744791ea61e

clalancette (approved):
Looks good to me with green CI.

iuhilnehc-ynos:
CI with [repos](https://gist.githubusercontent.com/iuhilnehc-ynos/a6544c0ac62eba5ef9599facc17dff4b/raw/9894a172176fe6dacf2d6d5c7a70d091a66c5725/ros2.repos):
build/test_args: `--packages-above-and-dependencies launch_testing`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19390/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13920/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20112/)
*

PR #719, opened 2023-07-12
Update the launch code for newer flake8 and mypy.

clalancette:
Newer flake8 complains if you use parentheses around assert statements, so fix that.

Newer mypy had a number of minor complaints about our types, so improve the type checking to account for that.

With both of these in place, the code is now clean under the versions of flake8 and mypy in Fedora 38 (5.0.3 and 1.4.0, respectively).


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19146/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13703/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19877/)

PR #723, opened 2023-07-24
Fix remaining occurrences of "There is no current event loop"

clalancette:
Depends on https://github.com/ros2/launch_ros/pull/372

To do that, we do 3 things:

1.  In places where we would have called `asyncio.get_event_loop()`, instead call `osrf_pycommon.process_utils.get_loop()`.  Since the latter is already suppressing the warning, the warning will be quieted there.
2.  Add some additional warning filters (in the tests) to catch the "There is no current event loop" warning.  These will go away when we update to Python >= 3.11
3.  Remove the `create_future_impl` utility.  That was only needed for compatibility with Python < 3.6, which we have no instances of anymore.  This particular change also requires a change to `launch_ros`.

With these changes in place, there are no more warnings coming from the `launch` tests.


launch/launch/utilities/__init__.py
```diff
@@ -27,7 +26,6 @@
     'is_a',
     'is_a_subclass',
     'isclassinstance',
-    'create_future',
```
clalancette, line 30: I should point out that the removal of this implementation is technically removing something from the public launch API, and thus in theory we should deprecate it first.  However, I looked through all of the released packages on Rolling, and the only one that used this was `launch_ros`.  Therefore, I think it is safe to say that this is mostly an implementation detail, and we can safely remove it.

mjcarroll (approved):

PR #728, opened 2023-09-04
Capture the environment variables in TimerAction

sjdalessandro:
This is a proposed solution for ros2/launch_ros#376

I've created a new action, `ReplaceEnvironmentVariables`, which keeps a copy of the environment variables present in the system when a `TimerAction` is executed. The system environment variables are then temporarily replaced by that copy before the scheduled actions are performed.

Feedback is welcome.

### DIFF
--- a/launch/launch/actions/replace_environment_variables.py
+++ b/launch/launch/actions/replace_environment_variables.py
@@ -1,0 +1,41 @@
+# Copyright 2022 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Module for the ReplaceEnvironmentVariables action."""
+
+from typing import Mapping
+from typing import Text
+
+from ..action import Action
+from ..launch_context import LaunchContext
+
+
+class ReplaceEnvironmentVariables(Action):
+    """
+    Action that replaces the environment variables in the current context.
+
+    The previous state can be saved by pushing the stack with the
+    :py:class:`launch.actions.PushEnvironment` action.
+    And can be restored by popping the stack with the
+    :py:class:`launch.actions.PopEnvironment` action.
+    """
+
+    def __init__(self, environment: Mapping[Text, Text], **kwargs) -> None:
+        """Create a ReplaceEnvironmentVariables action."""
+        super().__init__(**kwargs)
+        self.__environment = environment
+
+    def execute(self, context: LaunchContext):
+        """Execute the action."""
+        context._replace_environment(self.__environment)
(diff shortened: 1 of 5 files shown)


launch/launch/actions/replace_environment_variables.py
```diff
@@ -0,0 +1,41 @@
...
+    :py:class:`launch.actions.PushEnvironment` action.
+    And can be restored by popping the stack with the
+    :py:class:`launch.actions.PopEnvironment` action.
+    """
+
+    def __init__(self, environment: Mapping[Text, Text], **kwargs) -> None:
+        """Create a ReplaceEnvironmentVariables action."""
+        super().__init__(**kwargs)
+        self.__environment = environment
+
```
adityapande-1995, line 38: Lets add the @classmethod parse () method here, so that this works in xml/yaml style parsing as well. Also a test for that case. For reference : https://github.com/ros2/launch/blob/dd6ab0edf19c2f35b8e71aa8151df50b8a2feff9/launch/launch/actions/reset_environment.py#L39

launch/test/launch/actions/test_replace_environment_variables.py
```diff
@@ -0,0 +1,80 @@
+# Copyright 2022 Open Source Robotics Foundation, Inc.
```
adityapande-1995, line 1: Nit : 2023

launch/test/launch/actions/test_replace_environment_variables.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # replaces non empty state
+    context = LaunchContext()
+    context.environment.clear()
+    assert len(context.environment) == 0
+    context.environment['quux'] = 'quuux'
+    assert len(context.environment) == 1
+    assert 'quux' in context.environment
+    assert context.environment['quux'] == 'quuux'
+    ReplaceEnvironmentVariables([('foo', 'bar'), ('spam', 'eggs')]).visit(context)
```
adityapande-1995, line 58: Lets also add a test/expected behavior where we use this action on 'quux', which already exists in the environment.

adityapande-1995 (commented):
Thanks for adding this ! Could you add a test for the changed behavior of the Timer action ? The current one is for the new action only.

launch/launch/actions/replace_environment_variables.py
sjdalessandro: I think that `ReplaceEnvironmentVariables` should not be exposed to xml and yaml frontends. It's similar to `PopEnvironment` and `PushEnvironment` in that it should only be used by Launch internally and not by its users. But maybe it's just me who can't find a reason to expose it, so I made the suggested changes.

Please let me know if anything else is needed.

adityapande-1995 (approved):

launch/launch/actions/replace_environment_variables.py
adityapande-1995: Yeah, now that I think that makes sense.

ahcorde:
I was able to reproduce the issue, I resolved the conflicts @emersonknapp or @fujitatomoya do you mind to take a look ?

ahcorde (approved):

ahcorde:
Pulls: ros2/launch#728
Gist: https://gist.githubusercontent.com/ahcorde/3c353fbe051a310ff9e4b40bd95fd636/raw/b588957c4b0b3718ccc807506898838c1ecdbae1/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17638
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/26341/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/19989/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6040/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26081/)

PR #734, opened 2023-09-12
Add conditional substitution

### LINKED ISSUE #727: Parse arguments into their relevant type
#### Feature description
Substitutions are meant to provide values for a dynamic launch configuration and thus they evaluate to strings. Substitutions though are also used to execute code (see PythonExpression). When that's performed, the result of a substitution must be wrapped in quotes as it needs to be defined as a string, and then it is compared to other strings. For example

```python
PythonExpression(["'value1' if '", LaunchConfiguration("name"), "' == 'true' else 'value2'"])
```

That's a primitive and cumbersome way of working with arguments. It would be useful if it was possible for the launch configurations to be parsed into the relevant type of their argument.

We could hopefully have specialized launch configurations, `{Bool,Int,Float,etc}LaunchConfiguration`, that parse their value before returning it. Then, an expression like the above would become

```python
PythonExpression(["'value1' if ", BoolLaunchConfiguration("name"), " else 'value2'"])
```

Do you think that's reasonable and would you be interested in a PR?

nlamprian:
Closes #727

### DIFF
--- a/launch/doc/source/architecture.rst
+++ b/launch/doc/source/architecture.rst
@@ -131,6 +131,10 @@ There are many possible variations of a substitution, but here are some of the c
 
   - This substitution gets a launch configuration value, as a string, by name.
 
+- :class:`launch.substitutions.IfElseSubstitution`
+
+  - This substitution takes a substitution, and if it evaluates to true, then the result is the if_value, else the result is the else_value.
+
 - :class:`launch.substitutions.LaunchDescriptionArgument`
 
   - This substitution gets the value of a launch description argument, as a string, by name.
--- a/launch/launch/substitutions/__init__.py
+++ b/launch/launch/substitutions/__init__.py
@@ -25,6 +25,7 @@ from .environment_variable import EnvironmentVariable
 from .equals_substitution import EqualsSubstitution
 from .file_content import FileContent
 from .find_executable import FindExecutable
+from .if_else_substitution import IfElseSubstitution
 from .launch_configuration import LaunchConfiguration
 from .launch_log_dir import LaunchLogDir
 from .local_substitution import LocalSubstitution
@@ -46,6 +47,7 @@ __all__ = [
     'EnvironmentVariable',
     'FileContent',
     'FindExecutable',
+    'IfElseSubstitution',
     'LaunchConfiguration',
     'LaunchLogDir',
     'LocalSubstitution',
--- a/launch/launch/substitutions/if_else_substitution.py
+++ b/launch/launch/substitutions/if_else_substitution.py
@@ -18,13 +18,13 @@ from typing import List
 from typing import Sequence
 from typing import Text
 
+from .substitution_failure import SubstitutionFailure
 from ..frontend import expose_substitution
 from ..launch_context import LaunchContext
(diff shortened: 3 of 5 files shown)


clalancette (approved):
This looks fantastic to me, thanks for a nice clean pull request.  I'll run CI on it next.

clalancette (changes requested):
@nlamprian It looks like flake8 is unhappy here, mostly about double quotes but also about some import order stuff.  I don't know why the Rpr job didn't pick up on it, but if you setup your workspace locally it should show it to you.  That will need to be fixed before we merge this, thanks.

nlamprian:
Fixed, allegedly.
I did run the tests successfully last time. I am not able to reproduce the errors.
I am working on a fresh ros:rolling-ros-core container. I haven't identified what's different from the logs.

clalancette:
> Fixed, allegedly. I did run the tests successfully last time. I am not able to reproduce the errors. I am working on a fresh ros:rolling-ros-core container. I haven't identified what's different from the logs.

Out of curiosity, what does `pip3 list | grep flake8` in the container show you?

clalancette:
Locally I'm still seeing:

```
_________________________________ test_flake8 __________________________________
test/test_flake8.py:23: in test_flake8
    assert rc == 0, \
E   AssertionError: Found 1 code style errors / warnings:
E     ./launch/substitutions/if_else_substitution.py:27:1: I100 Import statements are in the wrong order. 'from .substitution_failure import SubstitutionFailure' should be before 'from ..utilities import perform_substitutions'
E   assert 1 == 0
```

nlamprian:
```
root@nlamprian:~/ros_ws# pip3 list | grep flake8
ament-flake8                         0.15.2
flake8                               4.0.1
```

The last error is hopefully fixed (the message can be misinterpreted).

clalancette:
> ```
> root@nlamprian:~/ros_ws# pip3 list | grep flake8
> ament-flake8                         0.15.2
> flake8                               4.0.1
> ```

OK, yeah.  You are missing a bunch of plugins, and flake8 is a bit annoying in that you can't specify which ones are required; it only runs the plugins you have installed.  My list looks like this:

```
ament-flake8                         0.15.2
flake8                               4.0.1
flake8-blind-except                  0.2.0
flake8-builtins                      1.5.3
flake8-class-newline                 1.6.0
flake8-comprehensions                3.8.0
flake8-deprecated                    1.3
flake8-docstrings                    1.6.0
flake8-import-order                  0.18.1
flake8-quotes                        3.3.1
```

We have a long-term plan to fix this up by making those plugins hard requirements, but we aren't there today.

Anyway, this looks good to me now (verified locally), so I'll run CI on this again.  Thanks for iterating.

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19514/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14032/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20233/)

federicociresola:
is it possible to include this feature also in the Humble branch?

Issue #736, opened 2023-10-10
How to add SysLogHandler


firesurfer:
Hi,
I would like to write all log messages which are written to `launch.log` per default additionally to the system log (and later on to a remote logging server).

What I tried so far is to add a SysLogHandler in my launch.py file to every registered logger:

```
import launch.logging
import logging.handlers
import logging

def generate_launch_description():

    for logger in launch.logging.LaunchLogger.all_loggers:
        handler = logging.handlers.SysLogHandler(address = '/dev/log')
        logger.addHandler(handler)
        logger.critical("test")
```

The "test" message is logged correctly but the actual output of all other loggers is not shown in the syslog. I guess the reason for this that the loggers are created when the LaunchDescription is executed.

So my question is: How can a add an additional handler to every logger created by ros2 launch in order to write all log messages which are written to `launch.log` also to the syslog.

PR #737, opened 2023-10-11
[Draft] Added syslog handler for launch log

firesurfer:
This PR adds initial support for writing to the syslog. 
It goes hand in hand with: https://github.com/ros2/rcl_logging/pull/105
In combination it allows to write the whole log a ros2 application creates to the syslog. Both PRs use 'local1' as facility which allows to redirect the whole log into a seperate log file or even a remote logging server without the need for additional configuration inside the ros2 logging system. (Everything can be configured in rsyslog.conf)

And also answers my question from #736 

## Why do we need it

In enterprise environments it can be desirable to store the whole application log on a remote logging server. Utilizing rsyslog for providing this functionality is an easy and established way.

Just grabbing `/rosout` won't suffice as we cannot grab all parts of the log. 

## What is missing so far:

1. Allow configuring the SysLogHandler (address, facility)
2. Allow to disable/enable the SysLogHandler
3. Implement custom formatter / do not reuse the file_formatter

## How to test it:

I redirected the 'local1' facility to `/var/log/local1.log`. This allows testing without spamming into your normal syslog.

```
sudo touch /var/log/local1.log
sudo chown syslog /var/log/local1.log
```

Edit ` /etc/rsyslog.d/50-default.conf`
And add the line:

```
local1.*                        -/var/log/local1.log
```
Restart rsyslog

```
sudo systemctl restart rsyslog.service 
```

Afterwards you should see the most basic output written into the file. 
Eg:

```
Oct 11 08:33:01 my_pc 1697005981.4697814 [INFO] [launch]: All log files can be found below /home/my_user/.ros/log/2023-10-11-08-33-01-469156-my_pc-2027270

```

## Some additional notes

The current implementation grabs on purpose only the parts of the log which can't be grabbed via rcl_logging  (https://github.com/ros2/rcl_logging/pull/105). Otherwise some log messages would occur twice in the syslog.

### DIFF
--- a/launch/launch/logging/__init__.py
+++ b/launch/launch/logging/__init__.py
@@ -110,6 +110,7 @@ class LaunchConfig:
         self._log_dir = None
         self.file_handlers = {}
         self.screen_handler = None
+        self.syslog_handler = None
         self.screen_formatter = None
         self.file_formatter = None
         self._log_handler_factory = None
@@ -250,6 +251,17 @@ class LaunchConfig:
             self.screen_handler.setFormatter(self.screen_formatter)
         return self.screen_handler
 
+    def get_syslog_handler(self):
+        """
+        Get the one and only syslog handler
+        """
+        if self.syslog_handler is None:
+            self.syslog_handler = logging.handlers.SysLogHandler(address='/dev/log', facility='local1')
+
+            if self.file_formatter:
+                self.syslog_handler.setFormatter(self.file_formatter)
+        return self.syslog_handler
+
     def set_log_format(self, log_format, *, log_style=None):
         """
         Set up launch log file format.
@@ -275,6 +287,8 @@ class LaunchConfig:
             )
             for handler in self.file_handlers.values():
                 handler.setFormatter(self.file_formatter)
+            if self.syslog_handler:
+                self.syslog_handler.setFormatter(self.file_formatter)
         else:
             self.file_formatter = None
 
@@ -328,6 +342,10 @@ def get_logger(name=None) -> logging.Logger:
     launch_log_file_handler = launch_config.get_log_file_handler()
     if launch_log_file_handler not in logger.handlers:
         logger.addHandler(launch_log_file_handler)
+    syslog_handler = launch_config.get_syslog_handler()
+    if syslog_handler not in logger.handlers:
+        logger.addHandler(syslog_handler)
+
     return logger
 
 


firesurfer:
Just a quick push, as I would really like to progress on these patches :)

sloretz:
@clalancette assigned you to provide thoughts on Windows 🧇

firesurfer:
@clalancette  As I already wrote in https://github.com/ros2/rcl_logging/pull/105 we could use the  Event Log on Windows instead: https://docs.python.org/3/library/logging.handlers.html#nteventloghandler

Additionally if desired by the uses one could support writing to a remote Syslog server (The python logging handler implementation supports this in contrast to spdlog)

```
logging.handlers.SysLogHandler(address=('localhost', SYSLOG_UDP_PORT), facility=LOG_USER, socktype=socket.SOCK_DGRAM)
```

130s:
> without the need for additional configuration inside the ros2 logging system. (Everything can be configured in rsyslog.conf)

I found this interesting. Would you elaborate the decision process?

Personally I prefer configuration of ROS itself (i.e. the destination of log that ROS generates, in this context) to be done in ROS level. But I want to hear other thoughts.

130s:
I guess I couldn't make the intent of my question clear enough. Let me retry. I did NOT ask why you want ROS to send logs to `syslog` (I do like aggregating all logs, not just the ones ROS generates, at one location e.g. `syslog` or any central place). 

Instead, what I asked is why you chose `rsyslog.conf` to be the place to configure ROS, so that rcl_logger sends log to `syslog`.

> Everything can be configured in rsyslog.conf

AFAIK `rsyslog.conf` is outside of ROS. Because I prefer configuring ROS to be done in ROS level, I want a way in ROS (e.g. `launch`, `node`) to tell ROS "hey, send logs to syslog".

I'm asking this as a non-blocking question for now.

firesurfer:
Sorry that I understood your question wrong. 

1. It is the default common logging facility of most linux systems 
2. It is way more flexible than what the ROS logging can do
3. Separate application and logging configuration on the system 
4. Additionally it is a way to obtain _all_ logs from multiple systems running a ROS application. At the moment just capturing `/ros_out` is not enough. 

> AFAIK rsyslog.conf is outside of ROS. Because I prefer configuring ROS to be done in ROS level, I want a way in ROS (e.g. launch, node) to tell ROS "hey, send logs to syslog".

Well thats what this patch is doing. You tell ROS send to syslog. 
What is syslog: It is in most cases rsyslogd which is configured via `rsyslog.conf`.

130s:
>  What is syslog: It is in most cases rsyslogd which is configured via `rsyslog.conf`.

On OS-level, sure. What I was saying is that I'd add an option somewhere in ROS, e.g. `launch` or else, so that I can configure ROS to send log to rsyslogd, without directly touching `rsyslog.conf`. Intent of that is to avoid customization on "`sudo`-protected" area as much as possible for the ease of configuration management.

But, that's just my preference, which should be opened separately as a feature enhancement discussion. I saw your https://github.com/ros2/rcl_logging/pull/105#issuecomment-2345337939 and now I understand all these syslog improvement is on hold anyways.

PR #752, opened 2024-01-12
Add exception type to error output

dyackzan:
Knowing the type of exception that occurred provides more context and makes launch exceptions easier to track down.


clalancette:
Our usual workflow is to get it onto the `rolling` branch first, and then use the mergify bot to backport it.  So I'm going to close this one for now.

PR #626, opened 2022-07-12
Improve launch file parsing error messages

timonegk:
This pull request improves the parsing error messages for launch files in two ways:
1. Parser errors clearly show that values are substituted and not part of the error message itself. I was recently confused by a message `Attribute value of type <class 'str'> not found in Entity let`, not realizing that the attribute "value" was missing and not some kind of "attribute value".
2. On an InvalidLaunchFileError, all parsing errors are logged.

PR #753, opened 2024-01-12
Add exception type to error output

dyackzan:
Knowing the type of exception that occurred provides more context and makes launch exceptions easier to track down.

### DIFF
--- a/launch/test/launch/test_invalid_launch_file_error.py
+++ b/launch/test/launch/test_invalid_launch_file_error.py
@@ -1,4 +1,4 @@
-# Copyright 2023 Open Source Robotics Foundation, Inc.
+# Copyright 2024 Open Source Robotics Foundation, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/launch/test/launch/test_invalid_launch_file_error.py
+++ b/launch/test/launch/test_invalid_launch_file_error.py
@@ -1,0 +1,33 @@
+# Copyright 2023 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


tylerjw (approved):
This looks like a nice change. Do you have a test for this?

dyackzan:
@tylerjw I just tested locally, but I'll work on getting a test added to this PR

launch/test/launch/test_invalid_launch_file_error.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2023 Open Source Robotics Foundation, Inc.
```
tylerjw, line 1:
```suggestion
# Copyright 2024 Open Source Robotics Foundation, Inc.
```

tylerjw (approved):
Thank you for adding the test. Will it be ok to backport this to Humble and Iron?

dyackzan:
@tylerjw I don't think a backport will work here...It looks like the functionality to report multiple exceptions was added into rolling and not backported to humble and iron ([from this PR](https://github.com/ros2/launch/pull/626)). So this change won't be 1-to-1 and the test I added for the multiple exception case wouldn't pass.
What's the recommended course of action here? Attempt to backport those changes first and then these? Or should I just open separate PRs

clalancette:
> What's the recommended course of action here? Attempt to backport those changes first and then these? Or should I just open separate PRs

Generally, we get the fixes into `rolling` first, and then open backports.  So even though this is going to have to be a "manual" backport (rather than using the bot), we should get this landed first.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20121/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14562/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20835/)

dyackzan:
> CI:
> 
> * Linux [Build Status](http://ci.ros2.org/job/ci_linux/20122/)
> * Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14562/)
> * Windows [Build Status](http://ci.ros2.org/job/ci_windows/20835/)

@clalancette in the last section of the [logs](https://ci.ros2.org/job/ci_linux/20122/consoleFull) it looks like the `test_xmllint.py` script failed. I see
```
14:07:00 test/test_xmllint.py .                                                   [ 38%]
14:07:03 FATAL: command execution failed
```
but I don't think this would be caused by this PR since I didn't commit any xml changes. Let me know what you think

clalancette:
> but I don't think this would be caused by this PR since I didn't commit any xml changes. Let me know what you think

Yeah, that was just an infrastructure problem.  I've kicked off the job again.

PR #755, opened 2024-01-25
Rework task exceptions loop.

clalancette:
The original reason to do this was because mypy was having a difficult time ensuring that the final raised exception was valid.  This fixes that problem.

But this also should be much faster.  Before we were iterating over the list 3 times; once to the collect the exceptions, once to filter out None, and once to them print them out. Instead, this just iterates once, dispatching everything properly.

### DIFF
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -327,14 +327,20 @@ class LaunchService:
                         return_when=asyncio.FIRST_COMPLETED
                     )
                     # Propagate exception from completed tasks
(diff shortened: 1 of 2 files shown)


adityapande-1995 (approved):

adityapande-1995:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20178/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14602/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20887/)

PR #758, opened 2024-02-05
Ignore ModuleNotFoundError in entry point discovery

mathias-luedtke:
This prevents pytest (or colcon test) from failing if a module imports a missing dependency.


audrow:
Hi @mathias-luedtke, could you give us a motivating example and add a test for this?

PR #759, opened 2024-02-13
Some small cleanups to launch

clalancette:
1.  Cleanup some type annotations.  In particular, use the inline style of type annotations, rather than the comments, as this is what we do everywhere else.  It will also make newer versions of mypy happier.
2.  Change the pytest markers tryfirst/trylast to using hookimpl, since the former are deprecated and throw warnings on newer versions of pytest.

### DIFF
--- a/launch/launch/launch_context.py
+++ b/launch/launch/launch_context.py
@@ -52,24 +52,24 @@ class LaunchContext:
         self.__argv = argv if argv is not None else []
         self.__noninteractive = noninteractive
 
-        self._event_queue = asyncio.Queue()  # type: asyncio.Queue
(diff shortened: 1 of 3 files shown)


mjcarroll (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20285/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14697/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21016/)

fujitatomoya (approved):
lgtm with green CI

Issue #762, opened 2024-03-04
Unexpected YAML parsing with get_attr and List[Entity] type


baconbrot:
## Bug report
The `get_attr` method from the [YAML entity](https://github.com/ros2/launch/blob/rolling/launch_yaml/launch_yaml/entity.py#L98) looks wrong to me when using `data_type=List[Entity]`.

The names of the entities returned by [following section](https://github.com/ros2/launch/blob/rolling/launch_yaml/launch_yaml/entity.py#L125)
```
        if check_is_list_entity(data_type):
            if isinstance(data, list) and isinstance(data[0], dict):
                return [Entity(child, name) for child in data]
            raise TypeError(
                "Attribute '{}' of Entity '{}' expected to be a list of dictionaries.".format(
                    name, self.type_name
                )
            )
```
are all set to the name of the parent attribute.
### Example
Changing the timer event parse method to accept an `actions` attribute of `data_type=List[Entity]` then the above code using following launch.yaml
```
launch:
  - timer:
      period: 3.0
      actions:
        - log:
            message: 'Timer msg 1!'
        - log:
            message: 'Timer msg 2!'
```
creates for each log entity an entity with `type_name=actions`  instead of  `type_name=log`.

The `children` property inside this entity class implements the correct (expected) logic.

Is this logic intended or a bug?

baconbrot:
Actually I am wrong! 
This is not a bug but intended behavior. 
An example use case is the [parsing of node params](https://github.com/ros2/launch_ros/blob/rolling/launch_ros/launch_ros/actions/node.py#L257).

PR #763, opened 2024-03-04
Fix ros2/launch#762

baconbrot:
Fixes ros2/launch#762

This fix is just a copy/paste of the working [children property getter](https://github.com/ros2/launch/blob/rolling/launch_yaml/launch_yaml/entity.py#L76).


baconbrot:
Closed because this is not a bug, see ros2/launch#762

Issue #741, opened 2023-11-04
Substitutions for string manipulations and xacro


nlamprian:
#### Feature description
Coming back with another request for a substitution... I tried using a path join substitution with a filename that depends on a launch configuration. The path join substitution does not allow one to concatenate parts of a filename. It ended up using a python expression.

```python
xacro_filename = PythonExpression(["'", LaunchConfiguration("model"), "' + '.xacro'"])
xacro_filepath = PathJoinSubstitution(["robot_description", "urdf", xacro_filename])
```

#### Implementation considerations
I would have hoped that text substitution could take multiple texts so that I could do

```python
TextSubstitution(LaunchConfiguration("model"), ".xacro")
```

But this is not compatible, given how text substitution was put together.

So, how about having a substitution that concatenates strings and then also having the separator parameterized?

In the simple case, one would do

```python
StringSubstitution([LaunchConfiguration("model"), ".xacro"])  # -> "model_value.xacro"
```

And then, this could be used in other practical ways, for example, to create a xacro substitution

```python
class XacroSubstitution(Command):
    def __init__(
        self, 
        filepath: SomeSubstitutionsType, 
        arg_str: SomeSubstitutionsType = "",  # xacro arguments that are passed as a launch argument
        arg_dict: Dict = None,  # xacro arguments that are defined in the launch file
    ):
        command = [FindExecutable(name="xacro"), " ", filepath, " "]
        arg_str, arg_dict = normalize_to_list_of_substitutions(arg_str), arg_dict or {}
        arg_list = [StringSubstitution([key, value], sep=":=") for key, value in arg_dict.items()]
        command.extend(StringSubstitution(arg_list + arg_str, sep=" "))
        super().__init__(command)
```

matosinho:
I did the same with Python expression for ".sdf". That was the exact thing I tried, to use a list with TextSubstitution, but then I discovered that it was not possible. Nice suggestion!

erichlf:
I attempted to setup a launch file for gazebo and I was hoping to set the environment variable needed for the model and world by using this same method. Of course, I can't because I can't convert the `PathJoinSubstitution` to a string.

christophebedard:
[...]

PR #768, opened 2024-03-19
Add TextJoinSubstitution

bektaskemal:
Addresses #741.
Don't use PythonExpression to join strings and be able to set seperators.


christophebedard:
Thanks for the PR.

Since this is specifically intended for `PathJoinSubstitution`, I don't think this is the right solution. Substitutions are already meant to be concatenated. I think the solution is to improve `PathJoinSubstitution`. Please see #838.

PR #770, opened 2024-03-21
launch_xml: fix xml syntax in README

scpeters:
I noticed some incorrect XML syntax in the `launch_xml` README, and this fixes it.


ahcorde (approved):

ahcorde:
This change is only related with documentation. I will merge it

Issue #779, opened 2024-05-16
Change ISSUE_TEMPLATE.md


DLu:
1. This repo uses an old version of [issue templates](https://github.com/ros2/launch/blob/rolling/.github/ISSUE_TEMPLATE.md) that Github warns against. 
2. The issue template directs people to ROS answers: https://answers.ros.org

peterborkuti:
Hi @audrow ,
I thought, I could help. See my PR.

PR #782, opened 2024-05-24
Add github templates for bug reports and features. Also links for important ROS related sites.

peterborkuti:
Hi,
I checked #779 and I thought, I could help.
This is what you were thinking about?

You can try the configurations here:
https://github.com/peterborkuti/launch/issues

Here is how it looks like:
image

image

image

I was thinking about creating drop downs with OS and DDS versions.

Hope you like it
Péter


DLu:
:+1:

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,62 @@
...
+    attributes:
+      label: "Version or commit hash:"
+    validations:
+      required: true
+  - type: input      
+    id: dds
+    attributes:
+      label: "DDS implementation:"
+    validations:
+      required: true
```
fujitatomoya, line 26: It does not have to be pull down, but can we add possible choices like, `rmw_fastrtps`, `rmw_connextdds` and `rmw_cyclonedds`? (there are more rmw implementations.) it would be easier for user to understand what needs to be filled in here especially beginners?

the same comment goes to `clientlib`.

fujitatomoya (commented):
IMO this is good enhancement! thanks for the PR.

.github/ISSUE_TEMPLATE/bug_report.yml
peterborkuti: Sure, I was thinking about it.
The possibilities:
1) Placeholder text
2) Some explanation before the label. Examples and how the user can check it, like: "ros2 doctor --report and check the line: middleware name".
3) Dropdown with the option "Other". In this case, we need another field for input for "Other".

In all the cases, I may keep the explanation.
You can choose, what you prefer. Or you can suggest another method.
Péter

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,69 @@
...
+  - type: input
+    id: version
+    attributes:
+      label: "Version or commit hash:"
+    validations:
+      required: true
+  - type: input      
+    id: dds
+    attributes:
+      label: "DDS implementation:"
```
fujitatomoya, line 24: DDS implementation is the one of the RMW implementation, this should be `RMW` (ROS Middle Ware).

```suggestion
      label: "RMW implementation:"
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,69 @@
...
+      description: |
+        **Examples:** *rmw_fastrtps_cpp*, *rmw_connextdds*, *rmw_cyclonedds_cpp*, ...
+        You can check it with the command: `ros2 doctor --report`
+        Find the line starting with `middleware name` in the report.
```
fujitatomoya, line 28: This looks good to me.

BTW, if we introduce `ros2 doctor --report` (which i think good idea), can we have dedicated field like `Output from ros2 doctor --report`?

fujitatomoya (changes requested):

.github/ISSUE_TEMPLATE/bug_report.yml
peterborkuti: Hi, done.
I also add some more descriptions.
My only concern is that ros2 doctor --report contains nearly every info (OS version, ros version, rmw) so in this case we are asking these info two times from the user.

fujitatomoya: that is true, lets see more feedback from the community.

peterborkuti: In the meantime, I formatted the ros2 doctor report and I changed the default value to be a placeholder to force the user  fill in the reproduction steps.

fujitatomoya (approved):

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,85 @@
...
+  - type: markdown
+    attributes:
+      value: "**Required Info:**"
+  - type: input
+    id: os
+    attributes:
+      label: "Operating System:"
+      description: |
+        Please try to be specific.
+        For linux please use the command `uname -a` from a terminal and copy paste its output here.
```
audrow, line 15:
```suggestion
        For Linux, please use the command `uname -a` from a terminal and copy paste its output here.
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,85 @@
...
+        For windows, open a terminal (Win key + R and type `cmd`), type the command `ver` and press enter. Then copy paste the output here.
```
audrow, line 16:
```suggestion
        For Windows, open a terminal (Win key + R and type `cmd`), type the command `ver` and press enter. Then copy paste the output here.
```

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,11 @@
...
+contact_links:
+  - name: Robotics Stack Exchange
+    url: https://robotics.stackexchange.com/
+    about: Please ask and answer questions here.
+  - name: Documentation for Active ROS Distributions
+    url: https://docs.ros.org/
+    about: Please check our documentation here.
+  - name: ROS Discourse
+    url: https://discourse.ros.org/
+    about: Talk about ROS and ROS related things.
```
audrow, line 11:
```suggestion
    about: Talk about ROS and ROS-related things.
```

audrow (commented):
This is great!

I made a few small suggestions.

It's also probably best to remove the existing issue template correct?

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,85 @@
...
+        Client libraries are the APIs that allow users to implement their ROS 2 code.
+    validations:
+      required: false
+  - type: textarea
+    id: doctor
+    attributes:
+      label: "'ros2 doctor --report' output"
+      description: |
+        It can help us knowing the details of your ROS environment.
+        Please use the command `ros2 doctor --report` and copy paste it's output here.
```
christophebedard, line 52:
```suggestion
        Please use the command `ros2 doctor --report` and copy paste its output here.
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,85 @@
...
+      render: Formatted
+    validations:
+      required: false
+  - type: textarea
+    id: repro
+    attributes:
+      label: "Steps to reproduce issue"
+      description: "How do you trigger this bug? Please walk us through it step by step."
```
christophebedard, line 60: Since this part is very important, I think we should be more specific here:

```suggestion
      description: |
        How do you trigger this bug? Please walk us through it step by step.
        Include all the commands you ran in the exact order you ran them so that anyone can reproduce the bug.
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,85 @@
+name: Bug Report
+description: File a bug report.
+title: "[Bug]: "
```
christophebedard, line 3: I just wanted to point out that this will result in "[Bug]: " being pre-filled in the issue title field. However, we do not really use "[Bug]:" or "[Feature]:" prefixes in issue titles in the ROS 2 repos.

@audrow thoughts?

audrow: Good point. I suppose it should be removed. We generally seem to prefer labels.

peterborkuti: Thanks, I removed both.

peterborkuti (commented):
Thank you for your suggestions.

audrow:
@peterborkuti, should the existing template be removed in this PR?

peterborkuti:
@audrow
I hope, its done. I removed the old template and fixed the DCO issues.

peterborkuti:
Hi,
Is there any issue with this version?

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,19 @@
+name: Feature request
+description: File a feature request.
+labels: ["feature"]
```
fujitatomoya, line 3: I would use the default labels provided by github, so that we cannot miss tagging.

```suggestion
labels: ["enhancement"]
```

fujitatomoya:
The current plan is to go with org template https://github.com/ros2/.github/pull/3, so i will go ahead to close this for now. if the repository specific templates are required, of course we can do that for each repository templete.

peterborkuti:
Dear @fujitatomoya 
As I was working on this for many months and I did 10+ times what you asked, I would be very happy if you would ask me to do it for the ros2 repo. Of course, you can do whatever yo want. But it is a bit rude, I may say.
Now I think, that it was a good thing, that I tested the ROS2 dev community with this pull request.

fujitatomoya:
@peterborkuti 

My apology for this decision change after all these iteration, but this sometimes happens after discussion with developers. I really appreciate you for being so responsive to apply the changes, thanks for that. i did not mean to be rude. as you pointed out, i should have been much louder on this change with you to share the progress. (i missed you on https://github.com/ros2/.github/pull/3, so added as co-author, sorry about that too.)

besides, if this is still better for this repository specific templates, feel free to reopen.

PR #790, opened 2024-07-18
Allows substitutions with Python Standard Path Objects with `PathSubstitutions`

InvincibleRMC:
Creates `PathSubstitutions` and update some types. For more information on Path check out the documentation [here](https://docs.python.org/3/library/pathlib.html)

PR #806, opened 2024-11-02
Cast pathlib.Path to str inside normalize_to_list_of_substitutions.

### LINKED ISSUE #805: [Enhancement] `normalize_to_list_of_substitutions` should cast `pathlib.Path` to `str`
#### Feature description

Currently, for `IncludeLaunchDescription`, I have to do this:

```python
pkg_uwu_sim = get_package_share_path("uwu_sim")
xacro = IncludeLaunchDescription(
   str(pkg_uwu_sim / "launch" / "xacro.launch.py"),
   launch_arguments={"robot": ROBOT_TYPE}.items(),
)
```

It would be nicer if I could just pass in `pathlib.Path` without casting to string first:

```python
xacro = IncludeLaunchDescription(
   pkg_uwu_sim / "launch" / "xacro.launch.py",
   launch_arguments={"robot": ROBOT_TYPE}.items(),
)
```

However, doing the above currently results in `TypeError: 'PosixPath' object is not iterable`. Most likely related to #345, but I'll send in a PR afterwards to fix this specific case in a way that might fix many others.

In short, always treat `pathlib.Path` as a `str` for `TextSubstitution`. Tracing the source code from `IncludeLaunchDescription`, these are the steps taken on the `launch_description_source` argument:

1. In `IncludeLaunchDescription`, if `launch_description_source` isn't an instance of `LaunchDescriptionSource`, wrap it in `AnyLaunchDescriptionSource`.
2. In `AnyLaunchDescriptionSource`, pass the `launch_file_path` to parent's `__init__()`.
3. In `LaunchDescriptionSource`, set `self.__location` to `normalize_to_list_of_substitutions(location)`.

It is in `normalize_to_list_of_substitutions` when the error occurs, as since `pathlib.Path` isn't a `str` nor `Substitution`, it attempts to cast it to an `Iterable` to normalize each item within it, which isn't valid.

My proposal and PR later:

```python
def normalize_to_list_of_substitutions(subs: SomeSubstitutionsType) -> List[Substitution]:
    """Return a list of Substitutions given a variety of starting inputs."""
    # Avoid recursive import
    from ..substitutions import TextSubstitution

    def normalize(x):
        if isinstance(x, Substitution):
            return x
        if isinstance(x, str):
            return TextSubstitution(text=x)
        raise TypeError(
            "Failed to normalize given item of type '{}', when only "

    ### >>> Insert
    if isinstance(subs, Path):
        return [TextSubstitution(text=str(subs.resolve()))]
    ### <<< End
    if isinstance(subs, str):
        return [TextSubstitution(text=subs)]
    if is_a_subclass(subs, Substitution):
        return [cast(Substitution, subs)]
```

Since `normalize_to_list_of_substitutions` is used in many places, simply casting `pathlib.Path` to `str` may fix a lot of things, where the previous behaviour was to attempt to cast it to `Iterable` and crash with `TypeError` as a result.

Interpause:
This resolves #805. More details can be found there.

After backporting the commit to `humble`, I have tested it to at least work on my launch files, allowing me to remove the `str()` cast.

Since this is my first time, I am not exactly sure how to run tests, but I think this means its good?

```
============================================================================================= test session starts =============================================================================================
platform linux -- Python 3.10.12, pytest-6.2.5, py-1.10.0, pluggy-0.13.0
rootdir: /workspaces/project-name/thirdparty/launch/launch_pytest, configfile: pytest.ini
plugins: launch-pytest-3.6.1, launch-testing-3.6.1, ament-copyright-0.12.11, ament-xmllint-0.12.11, launch-testing-ros-0.19.7, ament-lint-0.12.11, ament-flake8-0.12.11, ament-mypy-0.12.11, ament-pep257-0.12.11, colcon-core-0.18.1, cov-3.0.0
collected 18 items                                                                                                                                                                                            

test/test_copyright.py .                                                                                                                                                                                [  5%]
test/test_flake8.py .                                                                                                                                                                                   [ 11%]
test/test_pep257.py .                                                                                                                                                                                   [ 16%]
test/launch_pytest/test_plugin.py .............                                                                                                                                                         [ 88%]
test/launch_pytest/tools/test_process.py ..                                                                                                                                                             [100%]

============================================================================================= 18 passed in 17.48s =============================================================================================
```

EDIT:
Also would it be possible to backport this to humble? Thanks

### DIFF
--- a/launch/launch/utilities/normalize_to_list_of_substitutions_impl.py
+++ b/launch/launch/utilities/normalize_to_list_of_substitutions_impl.py
@@ -17,6 +17,7 @@
 from typing import cast
 from typing import Iterable
 from typing import List
+from pathlib import Path
 
 from .class_tools_impl import is_a_subclass
 from ..some_substitutions_type import SomeSubstitutionsType
@@ -37,6 +38,8 @@ def normalize_to_list_of_substitutions(subs: SomeSubstitutionsType) -> List[Subs
             "Failed to normalize given item of type '{}', when only "
             "'str' or 'launch.Substitution' were expected.".format(type(x)))
 
+    if isinstance(subs, Path):
+        return [TextSubstitution(text=str(subs.resolve()))]
     if isinstance(subs, str):
         return [TextSubstitution(text=subs)]
     if is_a_subclass(subs, Substitution):


Interpause:
Just saw #790. I think that one seems better as its more thorough & explicit, though that means more review work needed. Mine is closer to a quick workaround.

PR #179, opened 2019-02-07
Support required nodes

kyrofa:
Currently, the only way to specify that a given node is required for a given `LaunchDescription` is to register an `OnProcessExit` handler for the exit event of the required node that then emits a `Shutdown` event. Instead of requiring this boilerplate for a common use-case, This PR resolves #174 by adding an `on_exit` parameter to `ExecuteProcess` that can take actions, and adding a new action called `Shutdown` that immediately causes the launched system to shut down. This allows for the concept of a required node to be expressed simply with:

    launch_ros.actions.Node(
        # ...
        on_exit=Shutdown()
    )

PR #816, opened 2024-12-20
[Humble] Let XML executables/nodes be "required" (like in ROS 1) (#751)

monicaperezserrano:
Cherry pick of the commit discussed here: https://github.com/ros2/launch/pull/179

Supports "on_exit" for XML launch files in ROS 2 humble. 
I tested it locally and it works as expected: 
- Launch file with 2 nodes (both on_exit=shutdown)
- When I kill one node, the process exits with Exit Code = 0.

* Let XML nodes be "required"

Essentially on_exit="shutdown" is equivalent to ROS 1 required="true".

This feature is implemented using the python launchfile on_exit mechanism.

Right now "shutdown" is the only action accepted by on_exit, but theoretically more "on_exit" actions could be added later.

Example:
<executable cmd="ls" on_exit="shutdown"/>

* Added tests for yaml

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -23,7 +23,7 @@ from typing import Optional
 from typing import Text
 
 from .execute_local import ExecuteLocal
-
+from .shutdown_action import Shutdown
 from ..descriptions import Executable
 from ..frontend import Entity
 from ..frontend import expose_action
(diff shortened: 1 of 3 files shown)


monicaperezserrano:
Not sure why the DCO check is failing, I just cherry-picked the commit `986ea83`.
However, it says the email of the author is not the expected one, any ideas on how to solve this?

christophebedard:
Thanks for the PR. This ended up being backported to Humble in #863, so I will close this PR.

> Not sure why the DCO check is failing, I just cherry-picked the commit `986ea83`. However, it says the email of the author is not the expected one, any ideas on how to solve this?

It is usually because the user configured GitHub to use a specific email address when their commit is merged by GitHub (e.g., when their PR is merged), and that email is different from the email in the `Signed-off-by:` trailer line. We can just manually override the DCO check when we see this.

PR #820, opened 2025-01-23
Stop loading extensions during launch_{xml,yaml} tests.

clalancette:
The reason for this requires some explanation.  In current Windows builds, a "soft" error is generated every time these tests are run.  That error suggests that launch_xml or launch_yaml failed to load in the launch_ros component. But launch_xml and launch_yaml are in a repository that launch_ros depends on; how are they trying to load in launch_ros?

The answer has to do with the extension mechanism.  Because launch has the ability to load in extensions, it will attempt to do so anytime the Parser class is initialized.  One of the available extensions is the launch_ros parser, so it will attempt to load that.  And on Windows, that will fail to find rclpy.  This makes some sense; launch_xml doesn't depend on rclpy, so it doesn't necessarily have the paths setup properly to find it.  However, I don't understand why this only fails on Windows and not Linux.

Regardless, the solution presented here disables the loading of extensions in both launch_xml and launch_yaml.  This should ensure that it only loads in the things from this repository, avoiding the problem completely.

### DIFF
--- a/launch_xml/test/launch_xml/parser_no_extensions.py
+++ b/launch_xml/test/launch_xml/parser_no_extensions.py
@@ -1,0 +1,31 @@
+# Copyright 2025 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import TextIO
+from typing import Tuple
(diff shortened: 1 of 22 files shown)


clalancette:
Pulls: https://github.com/ros2/launch/pull/820
Gist: https://gist.githubusercontent.com/clalancette/a158460ee87a81d7c283f8acc24e2bf0/raw/9722b96a8a1c1761ba992159192ba052f270343f/ros2.repos
BUILD args:  --packages-above-and-dependencies launch_xml launch_yaml
TEST args:  --packages-above launch_xml launch_yaml
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15085
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22408/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16666/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1941/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23200/)

christophebedard (approved):
Seems reasonable to me.

Issue #799, opened 2024-09-28
Location of included launch files cannot be determined


dseredyn:
## Bug report

- Operating System:
  - Ubuntu 22.04
- Installation type:
  binary
- Version or commit hash:
  humble, from distro
- DDS implementation:
  - 
- Client library (if applicable):
  - 

#### Steps to reproduce issue
I'm trying to make a tool for examination of launch file structure and to visualize a tree of included launch files (python).

The following code, where ent is IncludeLaunchDescription object, should output a path to the included launch file, but instead it returns a string that refers to Python object pointer, e.g. "<launch.substitutions.path_join_substitution.PathJoinSubstitution object at 0x74b241481780>"

if isinstance(ent, IncludeLaunchDescription):
  print(isinstance(ent.launch_description_source.location, str))
  print(ent.launch_description_source.location)

ACTUAL output:
True
<launch.substitutions.path_join_substitution.PathJoinSubstitution object at 0x74b241481780>

EXPECTED output:
True
/home/user/ws/share/...

This should be fixed to enable error reporting for individual launch files included from other launch files. Now it is impossible to automatically determine which launch file has defined a problematic action.

I'm using PAL robotics scripts to build launch files. The function that includes a launch file is:

def include_launch_py_description(
        pkg_name: SomeSubstitutionsType,
        paths: List[SomeSubstitutionsType],
        **kwargs) -> Action:
    """
    Return IncludeLaunchDescription for the file inside pkg at paths.

    Example:
    -------
        include_launch_py_description('my_pkg', ['launch', 'my_file.launch.py'])
        returns file IncludeLaunchDescription from PATH_TO_MY_PKG_SHARE/launch/my_file.launch.py

    """
    pkg_dir = FindPackageShare(pkg_name)
    full_path = PathJoinSubstitution([pkg_dir] + paths)

    return IncludeLaunchDescription(
        PythonLaunchDescriptionSource(
            full_path),
        **kwargs)

fujitatomoya:
@wjwwood can you take a look about this?

christophebedard:
@dseredyn sorry we haven't followed up here. What's the latest status of your launch examination tool?

tl;dr substitutions need to be performed by calling `LaunchDescriptionSource.get_launch_description()` before using `LaunchDescriptionSource.location`

----

> Also, it looks like IncludeLaunchDescription convert Substitutions into a string <launch.substitutions.path_join_substitution.PathJoinSubstitution object at ...>:

[...]

PR #822, opened 2025-01-28
[Jazzy] Improve error reporting by adding file locations to exceptions

rolker:
In order to pinpoint which launch file caused an exception, this patch intercepts exceptions and adds notes that help determine the source of the error.

This might help with #799 as well.


christophebedard:
Would you be able to target the `rolling` branch instead? We usually do PRs against `rolling` and then backport to other distros as needed (and if possible).

rolker:
Opened a PR against rolling with the same commits. Should I leave this open to backport to Jazzy or is there another mechanism to do that?

christophebedard:
We can close this for now. We have a tool that backports PRs.

PR #824, opened 2025-02-01
Add test coverage for substitution edgecases involving E notation

danielcranston:
I extended some tests while doing some investigations in https://github.com/ros2/launch_ros/issues/384#issuecomment-2628904348. I figured it worthwhile to make a PR in case you find them useful.

I might be missing other tests that could get a similar treatment, let me know if that's the case!

### DIFF
--- a/launch/test/launch/substitutions/test_equals_substitution.py
+++ b/launch/test/launch/substitutions/test_equals_substitution.py
@@ -94,6 +94,14 @@ def test_equals_substitution():
     _permute_assertion('inf', '-inf', lc, 'false')
     _permute_assertion('-inf', '-inf', lc, 'true')
 
+    # Numerics (scientific E notation)
+    _permute_assertion('1234e1', '12340', lc, 'true')
+    _permute_assertion('1234E2', '123400', lc, 'true')
+    _permute_assertion('1234E2', '1234E2', lc, 'true')
+    _permute_assertion("'1234E2'", '123400', lc, 'false')
+    _permute_assertion("'1234E2'", '1234E2', lc, 'false')
+    _permute_assertion("'1234E2'", "'1234E2'", lc, 'true')
+
     # Strings
     _permute_assertion('wow', 'wow', lc, 'true')
     _permute_assertion('wow', True, lc, 'false')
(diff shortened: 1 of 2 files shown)


christophebedard (approved):
Looks good to me.

This is completely separate from https://github.com/ros2/launch_ros/issues/384 and https://github.com/ros2/launch_ros/pull/436, right? As in, the behaviour this is testing is correct?

christophebedard:
Pulls: ros2/launch#824
Gist: https://gist.githubusercontent.com/christophebedard/3a387055330bc86bf34c1c11a029c230/raw/3eefe972d23c9cd96fbc4243f027e4f8ba4e4d3d/ros2.repos
BUILD args:  --packages-up-to launch
TEST args:  --packages-select launch
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15115
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22445/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16697/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1973/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23234/)

danielcranston:
Yes I would say it is separate. 

> As in, the behaviour this is testing is correct?

If "correct" means "consistent with PYYAML 1.1", then yes I believe it's **largely** correct. I'm not an expert in the nuances of YAML specs, but from the discussion in https://github.com/ros2/launch_ros/issues/384#issuecomment-2634074244, I'm a bit on the fence on the `_permute_assertion` checks though.

If `_permute_assertion` intends to reflect YAML 1.1 / PYYAML, then there's a small inconsistency:

```python
    _permute_assertion('1234e1', '12340', lc, 'true')  # Arguably wrong. PYYAML treats LHS as a string, so shouldn't equal RHS
    _permute_assertion('1234E2', '123400', lc, 'true')  # Arguably wrong. PYYAML treats LHS as a string, so shouldn't equal RHS
    _permute_assertion('1234E2', '1234E2', lc, 'true')  # OK, PYYAML treats the LHS as a string anyway
    _permute_assertion("'1234E2'", '123400', lc, 'false')  # OK, LHS is coerced to string
    _permute_assertion("'1234E2'", '1234E2', lc, 'false')  # OK, LHS is coerced to string
    _permute_assertion("'1234E2'", "'1234E2'", lc, 'true')  # OK I guess.
```

Though I have a feeling that, for original authors, the measure of "correctness" isn't YAML 1.1 but something else (maybe some heuristics to reflect how ROS 1 `roslaunch` behaved?). I should also mention that it's not yet clear to me what piece of code makes `'1234e1'` become equal to `'12340'`.

It's also interesting that `'1234E1' == '12340'` and `'1234E1' == '1234E1'` are simultaneously true.

fujitatomoya (approved):
lgtm

PR #831, opened 2025-02-24
Updated `launch` typings

InvincibleRMC:
Update typings to be more accurate and cover more functions.


christophebedard:
Sorry, my PR created [more work](https://build.ros2.org/job/Rpr__launch__ubuntu_noble_amd64/117/consoleFull#console-section-59) for you :sweat_smile: I can take care of it if you want.

InvincibleRMC:
> Sorry, my PR created [more work](https://build.ros2.org/job/Rpr__launch__ubuntu_noble_amd64/117/consoleFull#console-section-59) for you 😅 I can take care of it if you want.

It's fine lol. All fixed up.

InvincibleRMC:
The most recent set of commits will help resolve some of https://github.com/ros2/ros2_tracing/issues/160. Since it makes the base class use `Sequence` which is `covariant` unlike `List` for functions like `execute`.

christophebedard:
That sounds great! Sorry we didn't get to it before the Kilted freeze. I will try to review it next week.

InvincibleRMC:
> That sounds great! Sorry we didn't get to it before the Kilted freeze. I will try to review it next week.

That's totally alright I understand all of the maintainers are very busy!

launch/launch/actions/conftest.py
```diff
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
```
christophebedard:
```suggestion
```

launch/launch/action.py
```diff
@@ -14,20 +14,33 @@
...
+from typing import Union
+
+from typing_extensions import Self
 
 from .condition import Condition
 from .launch_context import LaunchContext
 from .launch_description_entity import LaunchDescriptionEntity
 
-if False:
+
```
christophebedard:
```suggestion
```

launch/launch/action.py
```diff
@@ -14,20 +14,33 @@
...
+if TYPE_CHECKING:
     from .frontend import Entity  # noqa: F401
     from .frontend import Parser  # noqa: F401
+    from .conditions import IfCondition
+    from .conditions import UnlessCondition
+
+
+class ActionParsedDict(TypedDict, total=False):
+    condition: Union['IfCondition', 'UnlessCondition']
```
christophebedard: this makes sense, but the added burden is not very trivial. Looks like it doesn't affect child classes too much, although they need to extend the `TypedDict`. I think runtime behaviour won't be affected if child classes don't do any of this, correct? Will it make mypy even more unhappy if they don't make similar changes?

launch/launch/actions/append_environment_variable.py
```diff
@@ -72,18 +86,22 @@ def parse(
...
+    ) -> Tuple[Type[Self], AppendEnvironmentVariableParsedDict]:
         """Parse an 'append_env' entity."""
         _, kwargs = super().parse(entity, parser)
-        kwargs['name'] = parser.parse_substitution(entity.get_attr('name'))
-        kwargs['value'] = parser.parse_substitution(entity.get_attr('value'))
+
+        new_kwargs = AppendEnvironmentVariableParsedDict(
+            name=parser.parse_substitution(entity.get_attr('name')),
+            value=parser.parse_substitution(entity.get_attr('value')),
+            **kwargs)
```
christophebedard: I think I'd prefer to keep it looking like it did before, i.e., something like:

```suggestion
        new_kwargs = AppendEnvironmentVariableParsedDict(**kwargs)
        new_kwargs['name'] = parser.parse_substitution(entity.get_attr('name'))
        new_kwargs['value'] = parser.parse_substitution(entity.get_attr('value'))
```

What do you think? This way all kwargs are set with `new_kwargs['_____']`

Same for all other actions

launch/launch/actions/for_loop.py
```diff
@@ -22,12 +22,19 @@
 from typing import Optional
 from typing import Protocol
 from typing import Text
+from typing import Tuple
+from typing import Type
+
+from typing_extensions import NotRequired
+from typing_extensions import Self
+
```
christophebedard:
```suggestion
```

launch/launch/actions/include_launch_description.py
```diff
@@ -174,7 +187,9 @@ def execute(self, context: LaunchContext) -> List[LaunchDescriptionEntity]:
             argument_names = my_argument_names
             if ild_actions is not None:
                 for ild_action in ild_actions:
-                    argument_names.extend(ild_action._try_get_arguments_names_without_context())
+                    names = ild_action._try_get_arguments_names_without_context()
+                    if names:
+                        argument_names.extend(names)
```
christophebedard, line 186: `or []` would be simpler, but this is fine

launch/launch/actions/log_info.py
```diff
@@ -14,11 +14,17 @@
 
 """Module for the LogInfo action."""
 
+from typing import Any
 from typing import List
+from typing import Tuple
+from typing import Type
+
```
christophebedard:
```suggestion
```

launch/launch/actions/push_environment.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 """Module for the PushEnvironment action."""
+from typing import Any
```
christophebedard, line 17:
```suggestion

from typing import Any
```

launch/launch/actions/push_launch_configurations.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 """Module for the PushLaunchConfigurations action."""
+from typing import Any
```
christophebedard, line 17:
```suggestion

from typing import Any
```

launch/launch/actions/set_launch_configuration.py
```diff
@@ -13,10 +13,15 @@
 # limitations under the License.
 
 """Module for the SetLaunchConfiguration action."""
-
+from typing import Any
```
christophebedard, line 17:
```suggestion

from typing import Any
```

launch/launch/actions/timer_action.py
```diff
@@ -92,10 +102,10 @@ def __init__(
         self.__period = type_utils.normalize_typed_substitution(period, float)
         self.__actions = actions
         self.__context_locals: Dict[Text, Any] = {}
+        self._completed_future: Optional[asyncio.Future[None]] = None
         self.__context_launch_configuration: Dict[Any, Any] = {}
-        self._completed_future: Optional[asyncio.Future] = None
         self.__canceled = False
-        self._canceled_future: Optional[asyncio.Future] = None
+        self._canceled_future: Optional[asyncio.Future[bool]] = None
```
christophebedard: what's the reason for moving `self._completed_future`?

launch/launch/actions/unregister_event_handler.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 """Module for the UnregisterEventHandler action."""
+from typing import Any
```
christophebedard, line 17:
```suggestion

from typing import Any
```

launch/launch/actions/unset_environment_variable.py
```diff
@@ -13,10 +13,15 @@
 # limitations under the License.
 
 """Module for the UnsetEnvironmentVariable action."""
-
+from typing import Any
```
christophebedard, line 17:
```suggestion

from typing import Any
```

launch/launch/event_handlers/on_shutdown.py
```diff
@@ -73,6 +74,6 @@ def handler_description(self) -> Text:
         return '{}'.format(self.__on_shutdown)
 
     @property
-    def matcher_description(self):
+    def matcher_description(self) -> str:
```
christophebedard: why use `str` here but `Text` in `BaseEventHandler`?

launch/launch/events/process/process_started.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 """Module for ProcessStarted event."""
+from typing import Any
```
christophebedard, line 17:
```suggestion

from typing import Any
```

launch/launch/events/process/process_stderr.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 """Module for ProcessStderr event."""
+from typing import Any
```
christophebedard, line 17:
```suggestion

from typing import Any
```

launch/launch/events/process/process_stdin.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 """Module for ProcessStdin event."""
+from typing import Any
```
christophebedard, line 17:
```suggestion

from typing import Any
```

launch/launch/events/process/process_stdout.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 """Module for ProcessStdout event."""
+from typing import Any
```
christophebedard, line 17:
```suggestion

from typing import Any
```

launch/launch/invalid_launch_file_error.py
```diff
@@ -13,12 +13,14 @@
 # limitations under the License.
 
 """Implementation of `InvalidLaunchFileError` class."""
+from typing import List
```
christophebedard, line 17:
```suggestion

from typing import List
```

launch/launch/utilities/class_tools_impl.py
```diff
@@ -16,13 +16,17 @@
 
 import inspect
 
+from typing import Any
+from typing import Type
 
-def isclassinstance(obj):
+
+# Can be redone properly with TypeIs once mypy is version 1.10+ is available on all platforms
```
christophebedard:
```suggestion
# Can be redone properly with TypeIs once mypy version 1.10+ is available on all platforms
```

launch/launch/utilities/signal_management.py
```diff
@@ -107,20 +110,25 @@ def __enter__(self) -> 'AsyncSafeSignalManager':
...
+            if self.__close_sockets:
+                self.__close_sockets()
             self.__rsock = None
             self.__wsock = None
             self.__close_sockets = None
 
-    def __add_signal_readers(self):
+    def __add_signal_readers(self) -> None:
+        if self.__rsock is None:
+            raise RuntimeError('AsnycSafeSignalManager has not been initliazed.')
```
christophebedard:
```suggestion
            raise RuntimeError('AsyncSafeSignalManager has not been initialized.')
```

launch/launch/utilities/signal_management.py
```diff
@@ -133,24 +141,29 @@ def __add_signal_readers(self):
                 self.__loop.call_soon_threadsafe,
                 self.__handle_signal)
 
-            def run_background_loop():
+            def run_background_loop() -> None:
                 asyncio.set_event_loop(self.__background_loop)
+                if self.__background_loop is None:
+                    raise RuntimeError('AsnycSafeSignalManager has not been initliazed.')
```
christophebedard:
```suggestion
                    raise RuntimeError('AsyncSafeSignalManager has not been initialized.')
```

launch/launch/utilities/signal_management.py
```diff
@@ -133,24 +141,29 @@ def __add_signal_readers(self):
...
             self.__background_loop = None
         else:
-            self.__loop.remove_reader(self.__rsock.fileno())
+            if self.__rsock:
+                self.__loop.remove_reader(self.__rsock.fileno())
 
-    def __install_signal_writers(self):
+    def __install_signal_writers(self) -> None:
+        if self.__wsock is None:
+            raise RuntimeError('AsnycSafeSignalManager has not been initliazed.')
```
christophebedard:
```suggestion
            raise RuntimeError('AsyncSafeSignalManager has not been initialized.')
```

launch/launch/utilities/type_utils.py
```diff
@@ -354,6 +354,7 @@ def get_typed_value(
 # only return True when the passed tpe is either a substitution or a mixed
 # list of strings and substitutions. Indeed, there is no way that I could find
 # using overloads to describe "anything else than the above two types".
+# Can be redone properly with TypeIs once mypy is version 1.10+ is available on all platforms
```
christophebedard:
```suggestion
# Can be redone properly with TypeIs once mypy version 1.10+ is available on all platforms
```

christophebedard (commented):
Some minor comments, otherwise looks good. Thanks for your patience :grin: 

I do wonder: if mypy is currently not complaining about the lack of type info (provided by this PR), how do we ensure that future PRs won't just get rid or mess up the type info?

InvincibleRMC:
> I do wonder: if mypy is currently not complaining about the lack of type info (provided by this PR), how do we ensure that future PRs won't just get rid or mess up the type info?

By default `mypy` only type checks functions that have type hints. This is to better support gradual introduction of typing to existing code bases. Something I was hoping to change soon is the default `ament_mypy` settings to force all functions to be typed and not allowing untyped imports. I'm not sure how many packages actually hit 100% typing coverage especially since by default it attempts to type check unit tests which are rarely given type hints. In the iterim we can update this specific `ament_mypy` invocation but, updating the default is probably the best way forward in the long run.

launch/launch/actions/append_environment_variable.py
InvincibleRMC: Sadly doing it that way is considered untype safe since there are no guarentees that all the keys get added.
https://mypy-play.net/?mypy=latest&python=3.12&gist=67fac86e3320b4dc9002cc4344bd254a

launch/launch/action.py
InvincibleRMC: Seeing that all of the `.parse` methods are handled in the `@expose_action` decorator  for other launch types having typing is probably not that important. We could replace all of them with `Dict[str, Any]` and it would be fine imo even if less precise.

launch/launch/actions/append_environment_variable.py
christophebedard: unlike `ActionParsedDict`, your example doesn't set `total=False`; setting it resolves the issue

InvincibleRMC: That is true for keys that are optional like in `ActionParsedDict`. But marking keys that always exists as optional is not ideal. It might be to remove the `TypedDict` as mentioned above.

InvincibleRMC:
Since I was curious using the `--disallow-untyped-defs` flag currently exposes 390 errors in `launch`.

launch/launch/action.py
InvincibleRMC: @christophebedard Following up do agree using `Dict[str, Any]` would be better for the lower burden?

launch/launch/actions/append_environment_variable.py
christophebedard: We can always start with some simple `Dict[str, Any]` and reconsider later.

christophebedard:
I'm off until next Tuesday, so I'll do a proper re-review then.

> By default `mypy` only type checks functions that have type hints. This is to better support gradual introduction of typing to existing code bases.

Ah right, I think I asked the same question before :man_facepalming: sorry!

> Something I was hoping to change soon is the default `ament_mypy` settings to force all functions to be typed and not allowing untyped imports. I'm not sure how many packages actually hit 100% typing coverage especially since by default it attempts to type check unit tests which are rarely given type hints. In the iterim we can update this specific `ament_mypy` invocation but, updating the default is probably the best way forward in the long run.

I think making that the overall default (i.e., for _everyone_) in `ament_mypy` would be _quite_ disruptive, although opting out would be easy.

launch/launch/action.py
christophebedard: Yeah, like I mentioned in https://github.com/ros2/launch/pull/831#discussion_r2076560117, we can start with `Dict[str, Any]` and reconsider later.

InvincibleRMC:
> I think making that the overall default (i.e., for _everyone_) in `ament_mypy` would be _quite_ disruptive, although opting out would be easy.

That's a good point. Maybe having some stricter checks that includes `--disallow-untyped-defs` for the ros core library would be best and that way it wouldn't be disruptive for users.

christophebedard:
> > I think making that the overall default (i.e., for _everyone_) in `ament_mypy` would be _quite_ disruptive, although opting out would be easy.
> 
> That's a good point. Maybe having some stricter checks that includes `--disallow-untyped-defs` for the ros core library would be best and that way it wouldn't be disruptive for users.

We are slowly moving towards that, thanks to you!

launch/package.xml
```diff
@@ -25,6 +25,7 @@
   <exec_depend>python3-importlib-metadata</exec_depend>
   <exec_depend>python3-lark-parser</exec_depend>
   <exec_depend>python3-yaml</exec_depend>
+  <exec_depend>python3-typing-extensions</exec_depend>
```
christophebedard, line 28: One thing to note is that, for example, Debian has `python3-typing-extensions` version 4.4.0 (https://github.com/ros2/rclpy/pull/1458), unlike Ubuntu Noble, which has 4.10.0: https://packages.ubuntu.com/noble/python3-typing-extensions. It would be worth quickly checking that everything we import from `typing_extensions` here is available in <=4.4.0.

christophebedard (approved):
Thanks for iterating. We can consider using something more specific than `Dict[str, Any]` for `parse()` later.

Other than the quick validation for `python3-typing-extensions` <=4.4.0, this looks good to me!

launch/package.xml
InvincibleRMC: Double checking `python3-typing-extensions` 4.4.0 does have both the `Self` type as well as `TypeAlias`.

christophebedard: Thanks for checking.

christophebedard:
Pulls: ros2/launch#831
Gist: https://gist.githubusercontent.com/christophebedard/f577d8f45feb1401201a09971961eaba/raw/ef0bdd997a901c00a9827ebe776b0174b5df4734/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15988
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23630/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17733/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3094/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24349/)

christophebedard:
I retriggered the RHEL job.

launch/launch/launch_service.py
```diff
@@ -201,20 +203,21 @@ def _on_sigint(signum):
...
-                this_loop.call_soon(this_task.cancel)
+                if this_task:
+                    this_loop.call_soon(this_task.cancel)
 
             with AsyncSafeSignalManager(this_loop) as manager:
                 # Setup signal handlers
                 manager.handle(signal.SIGINT, _on_sigint)
                 manager.handle(signal.SIGTERM, _on_sigterm)
-                if platform.system() != 'Windows':
+                if sys.platform != 'win322':
```
christophebedard:
```suggestion
                if sys.platform != 'win32':
```

christophebedard: also, could you explain why this change is necessary?

InvincibleRMC:
> I retriggered the RHEL job.

You should probably re-trigger the Windows job as well.

launch/launch/launch_service.py
```diff
@@ -201,20 +203,21 @@ def _on_sigint(signum):
...
-                this_loop.call_soon(this_task.cancel)
+                if this_task:
+                    this_loop.call_soon(this_task.cancel)
 
             with AsyncSafeSignalManager(this_loop) as manager:
                 # Setup signal handlers
                 manager.handle(signal.SIGINT, _on_sigint)
                 manager.handle(signal.SIGTERM, _on_sigterm)
-                if platform.system() != 'Windows':
+                if sys.platform != 'win32':
```
christophebedard: Can you explain why this change is necessary? Why can't we keep using `platform.system()`?

InvincibleRMC: Type checkers don't currently understand that `platform.system()` narrows systems. Currently its only `sys.platform` as discussed [here](https://typing.python.org/en/latest/spec/directives.html#version-and-platform-checking). If you would prefer I could slap in a `type: ignore` instead of changing that line.

christophebedard: I'm slightly concerned because `platform.system()` and `sys.platform` aren't exactly the same and I want to avoid regressions. So yeah, I'd prefer keeping it as-is and just silencing mypy for the line below.

You could propose switching to `sys.platform` in a separate PR; then we could take a proper look at it.

christophebedard:
Pulls: ros2/launch#831
Gist: https://gist.githubusercontent.com/christophebedard/ba760d95a5fe8af14c4b3090bd6a0d6a/raw/ef0bdd997a901c00a9827ebe776b0174b5df4734/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16002
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23646/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17752/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3099/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24364/)

christophebedard:
Looks like mypy on RHEL isn't happy. Could you take a look? RHEL might have a different mypy version.

InvincibleRMC:
@christophebedard do you know if `__set_wakeup_fd` is a static or instance variable?

https://github.com/ros2/launch/blob/07ec3d42c5fc537db17d46487c2d0a60dd04a7c8/launch/launch/utilities/signal_management.py#L68

christophebedard:
In the context of `AsyncSafeSignalManager`, it's never reassigned, so just a static variable.

InvincibleRMC:
Thanks. Just wanted to double check. Should be ready for another re-run now.

christophebedard:
Pulls: ros2/launch#831
Gist: https://gist.githubusercontent.com/christophebedard/bec1426dae0c418251586a6bd316a7b7/raw/ef0bdd997a901c00a9827ebe776b0174b5df4734/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16046
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23717/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17811/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3155/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24421/)

christophebedard:
The test failures in the Linux jobs are known, but the failures in the Windows job are suspicious:

* https://ci.ros2.org/job/ci_windows/24421/testReport/
* https://ci.ros2.org/job/ci_windows/24410/testReport/

I don't see any other recent Windows jobs with these test failures. Can you take a look?

InvincibleRMC:
Similar `test_communication` package errors seem to exist in [nightly](https://ci.ros2.org/job/nightly_win-2025_release/6/testReport/junit/(root)/test_communication/test_publisher_subscriber__rclpy__rclcpp__rmw_connextdds_xunit_missing_result/). I'm not sure of the differences that don't cause these errors on `ci_windows`. Also I'm not sure what changes after [this](https://ci.ros2.org/job/ci_windows/24349/) run would cause errors.

christophebedard (approved):
Could be a (previous) change that made something flaky, but yeah, if it's also in the nightly job, then I think we're good.

Thanks, once again, for your patience :sweat_smile:

PR #836, opened 2025-03-15
Provide copy of launch configs to TimerAction's entities

### LINKED ISSUE #825: Bug report: error when combining/nesting GroupAction, IncludeLaunchDescription and TimerAction with launch arguments
I have a nested launch structure. In one launch file, I used `TimerAction` to launch a Node with a time delay. This launch file is included in another launch file using `IncludeLaunchDescription`. If this `IncludeLaunchDescription` is used directly, it works. If it is again nested inside a `GroupAction`, it gives me an error.

# Minimal example:

## `main_launch.py` has following content:
```python
from launch.actions import GroupAction
from launch_ros.actions import Node
from launch.actions import IncludeLaunchDescription



    nested_launch = IncludeLaunchDescription(
        PythonLaunchDescriptionSource("./nested_launch.py"),
        launch_arguments={'robot_namespace': "scout_mini"}.items()
    )

    my_group_action = GroupAction(
        actions=[nested_launch]
    )


    ld = LaunchDescription()
    ld.add_action(my_group_action)
    # ld.add_action(nested_launch)
    return ld
```

## `nested_launch.py` has following content:
```python
from launch_ros.actions import Node
from launch.actions import DeclareLaunchArgument
from launch.substitutions import LaunchConfiguration
from launch.actions import DeclareLaunchArgument, TimerAction



    robot_namespace = LaunchConfiguration('robot_namespace')
    
    declare_robot_namespace_cmd = DeclareLaunchArgument(
        name='robot_namespace',
        default_value='scout_mini'
    )

    example_timer_action = TimerAction(
        actions=[
            Node(
                package="demo_nodes_py",
                executable="listener",
                namespace=robot_namespace
            )
        ],
        period=2.0
    )


    ld = LaunchDescription()
    ld.add_action(declare_robot_namespace_cmd)
    ld.add_action(example_timer_action)
    return ld
```

## Error
Both launch files are in the same folder, I launch it with `ros2 launch ./main_launch.py` and get the following error:
```
[ERROR] [launch_ros.actions.node]: Error while expanding or validating node name or namespace for 'package=demo_nodes_py, executable=listener, name=None, namespace=<launch.substitutions.launch_configuration.LaunchConfiguration object at 0x7c5cff6e60b0>':
[ERROR] [launch]: Caught exception in launch (see debug for traceback): launch configuration 'robot_namespace' does not exist
```
(Similar behavior can be seen, if the launch files are properly organised in packages.)
Debugging in the upper level launch files was quite confusing, since putting something in a GroupAction shouldn't change its behavior much (from my understanding).

## How it works
It works, if the `main_launch.py` instead ends with
```python
    ld = LaunchDescription()
    # ld.add_action(my_group_action)
    ld.add_action(nested_launch)
    return ld
```
It also works, if alternatively you don't use `TimerAction` in `nested_launch.py` or if you have everything within one launch file. It seems to be a problem with this combination.

# System
Ubuntu 22.04.5 LTS
ROS2 humble

christophebedard:
Fixes #825

This PR modifies `TimerAction` so that its underlying actions have access to the state of the launch configurations at the time that the `TimerAction` itself executed. It also "scopes" the `TimerAction`'s actions so that their changes to the launch configurations do not leak.

----

Consider this minimal example:

```py
    import launch
    import launch.actions
        launch.actions.GroupAction(
            scoped=True,
            actions=[
                launch.actions.SetLaunchConfiguration('launch_arg', 'launch_arg_value'),
                launch.actions.TimerAction(
                    period=1.0,
                    actions=[
                        launch.actions.LogInfo(
                            msg=['launch_arg=', launch.substitutions.LaunchConfiguration('launch_arg')],
                        ),
                    ],
                ),
            ],
        ),
    ])
```

Launching this will result in an error because `launch_arg` does not exist:

```
[ERROR] [launch]: Caught exception in launch (see debug for traceback): launch configuration 'launch_arg' does not exist
```

This is because, after all of the `GroupAction`'s actions have executed, [its `scoped` feature restores the launch configurations to their original state (before the `GroupAction` executed).](https://github.com/ros2/launch/blob/151b024a7cfac06f0d56798cbe87ab180b9ea3ab/launch/launch/actions/group_action.py#L118-L136).

By the time the `TimerAction`'s actions execute, the launch configurations have been reset and `launch_arg` does not exist. As discussed in the issue, this doesn't really match our expectations. I believe that the `TimerAction`'s actions should have access to the state of the launch configurations at the time that the `TimerAction` itself executed. This is achieved by keeping a copy of the launch configs when the `TimerAction` executes and restoring it when the underlying actions execute, similar to how it already keeps a copy of the context locals. This makes the above example work:

```
[INFO] [launch.user]: launch_arg=launch_arg_value
```

This brings up another question: should the changes to the launch configs made by the `TimerAction`'s actions persist? In the above example, the `launch_arg` could leak out of the `GroupAction` (see test). I think it makes sense to _not_ enable time-delayed modifications to the launch configurations, which means that the `TimerAction` should scope its entities like `GroupAction` and reset the launch configs to the state they were in prior to them executing.

**This might potentially break existing launch files that rely on this timed-config behaviour.**

### DIFF
--- a/launch/launch/actions/timer_action.py
+++ b/launch/launch/actions/timer_action.py
@@ -30,6 +30,9 @@ import warnings
 import launch.logging
 
 from .opaque_function import OpaqueFunction
+from .pop_launch_configurations import PopLaunchConfigurations
+from .push_launch_configurations import PushLaunchConfigurations
+from .reset_launch_configurations import ResetLaunchConfigurations
 
 from ..action import Action
 from ..event_handler import EventHandler
@@ -54,6 +57,11 @@ class TimerAction(Action):
     Action that defers other entities until a period of time has passed, unless canceled.
 
     All timers are "one-shot", in that they only fire one time and never again.
+
+    Entities executed after the given period of time can access the launch configurations that
+    exist at the time that the timer action executed, but changes made by them will not persist.
+    This is similar to grouping the entities in a :class:`launch.actions.GroupAction` with
+    ``scoped=True``.
     """
 
     def __init__(
@@ -84,6 +92,7 @@ class TimerAction(Action):
         self.__period = type_utils.normalize_typed_substitution(period, float)
         self.__actions = actions
         self.__context_locals: Dict[Text, Any] = {}
+        self.__context_launch_configuration: Dict[Any, Any] = {}
         self._completed_future: Optional[asyncio.Future] = None
         self.__canceled = False
         self._canceled_future: Optional[asyncio.Future] = None
@@ -139,7 +148,14 @@ class TimerAction(Action):
     def handle(self, context: LaunchContext) -> Optional[SomeEntitiesType]:
         """Handle firing of timer."""
         context.extend_locals(self.__context_locals)
-        return self.__actions
+        # Reset the launch configurations to the state they were in when the timer action was
+        # executed, and make sure to push and pop them so that the changes don't persist and leak
+        return [
+            PushLaunchConfigurations(),
+            ResetLaunchConfigurations(self.__context_launch_configuration),
+            *self.__actions,
+            PopLaunchConfigurations(),
+        ]
 
     def cancel(self) -> None:
         """
@@ -189,8 +205,11 @@ class TimerAction(Action):
             ))
             setattr(context, '_TimerAction__event_handler_has_been_installed', True)
 
-        # Capture the current context locals so the yielded actions can make use of them too.
-        self.__context_locals = dict(context.get_locals_as_dict())  # Capture a copy
+        # Capture the current context locals and launch configuration so the yielded actions can
+        # make use of them too.
+        # Make sure to capture copies
+        self.__context_locals = dict(context.get_locals_as_dict())
+        self.__context_launch_configuration = context.launch_configurations.copy()
         context.asyncio_loop.create_task(self._wait_to_fire_event(context))
 
         # By default, the 'shutdown' event will cause timers to cancel so they don't hold up the
--- a/launch/test/launch/test_timer_action.py
+++ b/launch/test/launch/test_timer_action.py
@@ -19,6 +19,7 @@ import sys
 import launch
 import launch.actions
 import launch.event_handlers
+import launch.substitutions
 
 
 def test_multiple_launch_with_timers():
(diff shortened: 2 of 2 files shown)


christophebedard:
The original issue was about launch configurations, but it's kind of the same situation with environment variables.

mjcarroll:
> I think it makes sense to not enable time-delayed modifications to the launch configurations, which means that the TimerAction should scope its entities like GroupAction and reset the launch configs to the state they were in prior to them executing.

I agree with this.  Otherwise, it could potentially get race-y?

mjcarroll (approved):
LGTM with CI.

christophebedard:
Pulls: ros2/launch#836
Branch: christophebedard/timer-action-context-launch-config
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15698
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23164/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17373/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2631/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24032/)

christophebedard:
I missed this mypy error: https://ci.ros2.org/job/ci_linux-aarch64/17373/

```
19:58:47 launch/actions/timer_action.py: note: In member "handle" of class "TimerAction":
19:58:47 launch/actions/timer_action.py:155:39: error: Argument 1 to "ResetLaunchConfigurations" has incompatible type "dict[str, str]"; expected "dict[str | Substitution | Iterable[str | Substitution], str | Substitution | Iterable[str | Substitution]] | None"  [arg-type]
```

I'm re-triggering CI, but I think I'll just leave this for now, since it likely won't be done by midnight anyway. We can consider it for a Kilted bugfix later.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23168/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17392/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2635/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24045/)

christophebedard:
> We can consider it for a Kilted bugfix later.

This is a bug fix, so I'd like to merge it this week. It does change some behaviour, but that behaviour is very specific and users should not have relied on it.

@cottsay what do you think?

cottsay:
> what do you think?

Yeah, let's get it in now.

Issue #835, opened 2025-03-14
Document how multiple substitutions can be concatenated into a single string using a list


dvdmc:
This is useful to launch parameter files depending on a `launch` argument. It was common in ROS XML launch files:
```
<rosparam command="load" file="$(find my_pkg)/cfg/config_$(arg map).yml" />
```
or use execution dependant XACRO files #741 .

In the [launch architecture](https://docs.ros.org/en/jazzy/p/launch/architecture.html) website it's stated that TextSubstitutions are "... usually used to wrap literals in the launch description so they can be concatenated with other substitutions." But I couldn't find any reference on "how to concatenate substitutions".

It seems that a list concatenation worked for [some time](https://robotics.stackexchange.com/questions/96726/ros2-concatenate-string-to-launchargument) (seems more like a bug than a feature), but it seems to not work anymore since [Humble](https://robotics.stackexchange.com/questions/99879/ros2-launch-how-to-concatenate-launchconfiguration-with-string). Is really the only solution now to use a PythonExpression? Couldn't be the TextSubstitution be extended to create strings?

----

This comes down to missing documentation. This is probably the tutorial that should explicitly mention concatenating substitutions: https://docs.ros.org/en/jazzy/Tutorials/Intermediate/Launch/Using-Substitutions.html. Potentially here in the launch API docs too: https://docs.ros.org/en/jazzy/p/launch/architecture.html#substitutions.

christophebedard:
I can't really tell you why it changed, since I don't know why, but the correct way to concatenate a string and a substitution (which was also demonstrated in some of the Robotics StackExchange answers you linked to) now is to create a list.

Instead of this:

```py
node_namespace="aq_"+launch.substitutions.LaunchConfiguration("id"),
```

do this:

```py
node_namespace=["aq_", launch.substitutions.LaunchConfiguration("id")],
```

`"aq_"` will be concatenated with the result of the substitution. Note that, in this case, the `Node` action's `node_namespace` parameter was renamed to `namespace`.

Your ROS (1) XML example should work the same way in ROS 2 XML or YAML launch files. You can find examples here: https://docs.ros.org/en/jazzy/Tutorials/Intermediate/Launch/Using-Substitutions.html.

Since I think this answers your question, I'll close this, but feel free to keep discussing.

dvdmc:
[...]

PR #838, opened 2025-03-16
Allow concatenating each path component of PathJoinSubstitution

### LINKED ISSUE #741: Substitutions for string manipulations and xacro

christophebedard:
Closes #741

As mentioned in the above issue, users want to concatenate multiple substitutions into a single string, which, in this case, is a single path component. This is a very normal use of substitutions, but `PathJoinSubstitution` does not support it. It only supports joining individual substitutions into a path.

This PR changes `PathJoinSubstitution` to accept an `Iterable[SomeSubstitutionsType]` instead of being restricted to `Iterable[Union[Text, Substitution]]`. This enables the use-case presented in the issue:

```py
PathJoinSubstitution(['robot_description', 'urdf', [LaunchConfiguration('model'), '.xacro']])
```

If the `model` launch configuration was set to `my_model`, this would result in a path equal to:

```py
'robot_description/urdf/my_model.xacro'
```

As the existing test shows, this is backward compatible, since `Iterable[SomeSubstitutionsType]` is a superset of `Iterable[Union[Text, Substitution]]`.

I've added an example to the class documentation and expanded the test a bit.

### DIFF
--- a/launch/launch/substitutions/path_join_substitution.py
+++ b/launch/launch/substitutions/path_join_substitution.py
@@ -16,31 +16,70 @@
 
 import os
 from typing import Iterable
+from typing import List
 from typing import Text
-from typing import Union
 
 from ..launch_context import LaunchContext
+from ..some_substitutions_type import SomeSubstitutionsType
 from ..substitution import Substitution
+from ..utilities import perform_substitutions
 
 
 class PathJoinSubstitution(Substitution):
-    """Substitution that join paths, in a platform independent way."""
+    """
+    Substitution that join paths, in a platform independent way.
 
-    def __init__(self, substitutions: Iterable[Union[Text, Substitution]]) -> None:
-        """Create a PathJoinSubstitution."""
+    This takes in a list of path components as substitutions. The substitutions for each path
+    component are performed and concatenated, and then all path components are joined.
+
+    For example:
+
+    .. code-block:: python
+
+        PathJoinSubstitution([
+            EnvironmentVariable('SOME_DIR'),
+            'cfg',
+            ['config_', LaunchConfiguration('map'), '.yml']
+        ])
+
+    If the ``SOME_DIR`` environment variable was set to ``/home/user/dir`` and the ``map`` launch
+    configuration was set to ``my_map``, this would result in a path equal equivalent to (depending
+    on the platform):
+
+    .. code-block:: python
+
+        '/home/user/dir/cfg/config_my_map.yml'
+    """
+
+    def __init__(self, substitutions: Iterable[SomeSubstitutionsType]) -> None:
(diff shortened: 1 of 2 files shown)


christophebedard:
There's a request to backport this to Humble (and Jazzy while we're at it): https://github.com/ros2/launch/issues/835#issuecomment-2727688958

ahcorde (approved):

christophebedard:
Pulls: ros2/launch#838
Branch: christophebedard/path-join-substitution-list-of-somesubstitutionstype
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15694
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23162/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17369/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2627/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24028/)

christophebedard:
The test failures are unrelated (`ros2param` CLI flakes and Connext).

PR #843, opened 2025-03-18
Add StringJoinSubstitution substituion

baconbrot:
Add a substitution wrapper for python `str.join` method.
Supports concatenation of substitutions as implemented in ros2/launch#838

This PR would complement `PathJoinSubstitution` for general use cases like file names with separators (e.g. some.long.urdf.xacro), naming of entities (e.g. the_name_of_the_robot) or even namespaces (e.g. /env/group/robot/joint)

### Usecases
**Domains**
e.g. `https://docs.ros.org` with `subdomain` launch configuration equal to `docs`
```
StringJoinSubstitution(
    [
        ['https://', LaunchConfiguration('subdomain')],
        'ros',
        'org'
    ],
    delimiter='.'
)
```
**File names**
e.g. `robot.a.urdf.xacro` with `model` launch configuration equal to `a`
```
StringJoinSubstitution(
    [
        'robot'
        LaunchConfiguration('model'),
        'urdf',
        'xacro'
    ],
    delimiter='.'
)
```

### DIFF
--- a/launch/launch/substitutions/string_join_substitution.py
+++ b/launch/launch/substitutions/string_join_substitution.py
@@ -1,4 +1,4 @@
-# Copyright 2019 Open Source Robotics Foundation, Inc.
+# Copyright 2025 Open Source Robotics Foundation, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/launch/launch/substitutions/string_join_substitution.py
+++ b/launch/launch/substitutions/string_join_substitution.py
@@ -50,7 +50,7 @@ class StringJoinSubstitution(Substitution):
     def __init__(
         self,
         substitutions: Iterable[SomeSubstitutionsType],
-        delimiter: SomeSubstitutionsType = "",
+        delimiter: SomeSubstitutionsType = '',
     ) -> None:
         """
         Create a StringJoinSubstitution.
--- a/launch/launch/substitutions/string_join_substitution.py
+++ b/launch/launch/substitutions/string_join_substitution.py
@@ -72,7 +72,7 @@ class StringJoinSubstitution(Substitution):
         return self.__substitutions
 
     @property
-    def delimiter(self) -> Text:
+    def delimiter(self) -> List[Substitution]:
         """Getter for delimiter."""
         return self.__delimiter
 
--- a/launch/launch/substitutions/string_join_substitution.py
+++ b/launch/launch/substitutions/string_join_substitution.py
@@ -35,7 +35,7 @@ class StringJoinSubstitution(Substitution):
     .. code-block:: python
 
         StringJoinSubstitution(
-            ['https://', LaunchConfiguration('subdomain')], 'ros', 'org'],
+            [['https', '://'], LaunchConfiguration('subdomain')], 'ros', 'org'],
             delimiter='.'
         )
 
--- a/launch/launch/substitutions/string_join_substitution.py
+++ b/launch/launch/substitutions/string_join_substitution.py
@@ -1,0 +1,94 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Module for the StringJoinSubstitution substitution."""
+
+from typing import Iterable, List, Text
+
+from ..launch_context import LaunchContext
+from ..some_substitutions_type import SomeSubstitutionsType
+from ..substitution import Substitution
+from ..utilities import perform_substitutions
+
+
+class StringJoinSubstitution(Substitution):
+    """
+    Substitution that joins strings and/or other substitutions.
+
+    This takes in a list of string components as substitutions.
+    The substitutions for each string component are performed and concatenated,
+    and then all string components are joined with a specified delimiter as seperation.
+
+    For example:
+
+    .. code-block:: python
+
+        StringJoinSubstitution(
+            ['https://', LaunchConfiguration('subdomain')], 'ros', 'org'],
+            delimiter='.'
+        )
+
+    If the ``subdomain`` launch configuration was set to ``docs``
+    and the ``delimiter`` to ``.``, this would result in a string equal to
+
+    .. code-block:: python
+
+        'https://docs.ros.org'
+    """
+
+    def __init__(
+        self, substitutions: Iterable[SomeSubstitutionsType], delimiter: str = ''
+    ) -> None:
+        """
+        Create a StringJointSubstitution.
+
+        :param substitutions: the list of string component substitutions to join
+        :param delimiter: the text inbetween two consecutive components (default no text)
+        """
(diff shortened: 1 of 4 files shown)


launch/test/launch/substitutions/test_string_joint_substitution.py
```diff
@@ -0,0 +1,43 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
```
christophebedard:
```suggestion
# Copyright 2025 Open Source Robotics Foundation, Inc.
```

launch/launch/substitutions/string_join_substitution.py
```diff
@@ -0,0 +1,94 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
```
christophebedard:
```suggestion
# Copyright 2025 Open Source Robotics Foundation, Inc.
```

launch/launch/substitutions/string_join_substitution.py
```diff
@@ -0,0 +1,94 @@
...
+    If the ``subdomain`` launch configuration was set to ``docs``
+    and the ``delimiter`` to ``.``, this would result in a string equal to
+
+    .. code-block:: python
+
+        'https://docs.ros.org'
+    """
+
+    def __init__(
+        self, substitutions: Iterable[SomeSubstitutionsType], delimiter: str = ''
```
christophebedard: I think we should make `delimiter` substitutable, i.e., a `SomeSubstitutionsType`.

launch/launch/substitutions/string_join_substitution.py
```diff
@@ -0,0 +1,94 @@
...
+
+from typing import Iterable, List, Text
+
+from ..launch_context import LaunchContext
+from ..some_substitutions_type import SomeSubstitutionsType
+from ..substitution import Substitution
+from ..utilities import perform_substitutions
+
+
+class StringJoinSubstitution(Substitution):
```
christophebedard, line 25: Note: It should probably use `PathJoinSubstitution` _once_ #838 is merged.

christophebedard (commented):
This seems fine, ~~but could you give a real-world example?~~ Ah, sorry, I glossed over the last example in the PR description, although I kind of think it could simply be `['robot.', LaunchConfiguration('model'), '.urdf.xacro']`.

Also, do you think we should support frontends, i.e., by implementing `parse()`?

launch/launch/substitutions/string_join_substitution.py
baconbrot: Good suggestion!

launch/launch/substitutions/string_join_substitution.py
baconbrot: You mean like extending `PathJoinSubstitution`?

christophebedard: Oops, the other way: `PathJoinSubstitution` could extend this with `delimiter=os.pathsep`. However, looking into it, that wouldn't be the same, since `os.path.join()` does a bit more than just that, so nevermind!

baconbrot:
@christophebedard frontend support would definitly be a great addition for #838 because of cross-platform path handling.

For both cases, PathJoinSubstitution and StringJoinSubstitution, I come across following problem: 
Consider previous example for python launch
```
StringJoinSubstitution(
    [
        'robot'
        LaunchConfiguration('model'),
        'urdf',
        'xacro'
    ],
    delimiter='.'
)
```
I dont know how this would translate to yaml launch. Maybe to something like this
```
- arg:
  - name: "model"
- join-string:
      delimiter: "."
        children:
          - "robot"
          - "$(var model)"
          - "urdf"
          - "xacro"
```
but the join-string result is not accessible. Perhaps something like this
```
- arg:
  - name: "model"
- let:
    name: "some_string"
    value:
      join-string:
      delimiter: "."
      children:
        - "robot"
        - "$(var model)"
        - "urdf"
        - "xacro"
```
could work

christophebedard:
Substitutions in frontends look like `$(<name> <args>)`: `$(var ...)` is the `LaunchConfiguration` substitution for frontends, see https://github.com/ros2/launch/blob/7a79d6b3df71cd28ebc1ccc78a186f6be92072a5/launch/launch/substitutions/launch_configuration.py#L33-L34

So for `PathJoinSubstitution`, it would be `$(path-join ...)` and look like this:

```yaml
launch:
- let:
    name: 'model'
    value: 'a'
- log:
    message: '$(path-join robot $(var model) urdf xacro)'
```

This would log `robot/a/urdf/xacro`. Everything after `path-join` in `$()` is provided to the `*Substitution` class' `parse()` function as a sequence, see the `data` param here for example: https://github.com/ros2/launch/blob/7a79d6b3df71cd28ebc1ccc78a186f6be92072a5/launch/launch/substitutions/for_loop_var.py#L77 It gets a bit weird if there are spaces, but it works if you quote the item that has spaces: you can do `$(path-join robot $(var model) "ur df" xacro)` which results in `robot/a/ur df/xacro`. You can also nest substitutions, which is kind of the point of #838 (but for Python). In frontends, it looks like:

```yaml
launch:
- let:
    name: 'model'
    value: 'a'
- let:
    name: 'id'
    value: '2'
- log:
    message: '$(path-join "robot$(var id)" $(var model) "ur df" xacro)'
```

which outputs `robot2/a/ur df/xacro`. I think having the correct/platform-specific path separator might not actually matter in launch files, because the underlying Python code probably translates it into the correct path string, but it might still be worth supporting. I'll open a PR to add this to `PathJoinSubstitution` once #838 is merged.

For `StringJoinSubstitution`, the first element in `$(str-join ...)` would have to be the delimiter, then the rest would be the components to join. This can be added in a follow-up PR, though. Once `delimiter` supports substitutions, I think this PR is good to go.

launch/launch/substitutions/string_join_substitution.py
```diff
@@ -0,0 +1,98 @@
...
+    and then all string components are joined with a specified delimiter as seperation.
+
+    For example:
+
+    .. code-block:: python
+
+        StringJoinSubstitution(
+            ['https://', LaunchConfiguration('subdomain')], 'ros', 'org'],
+            delimiter='.'
+        )
```
christophebedard: I would augment this example (without changing the resulting string) to show that each component can be a list that gets concatenated, similar to some of your tests and similar to the example I added in #838 + https://github.com/ros2/ros2_documentation/pull/5157 (see #835).

Maybe something simple, like splitting `'https://'` in 2:

```suggestion
        StringJoinSubstitution(
            [['https', '://'], LaunchConfiguration('subdomain')], 'ros', 'org'],
            delimiter='.'
        )
```

christophebedard (approved):
Just a small suggestion, otherwise this looks good to me!

baconbrot:
@christophebedard Alright, thanks for your review! :D

christophebedard:
Pulls: ros2/launch#843
Gist: https://gist.githubusercontent.com/christophebedard/5aae799688178d79a6898bdefa08f849/raw/224dcbb05e9fa97b38d47aa82b744b2d642e36cf/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15605
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23051/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17269/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2531/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23922/)

christophebedard:
The test failures seem to be unrelated.

christophebedard:
Would you be willing to add support for frontends, i.e., by implementing `parse()`? Otherwise, that's fine, I can open an issue and someone can eventually get to it.

baconbrot:
Yes! I can do that

christophebedard:
Great!

PR #845, opened 2025-03-23
Document substitutions concatenation in architecture doc

christophebedard:
Relates to #835

From a user's perspective, this part should've mentioned _how_ substitutions are concatenated, see: https://github.com/ros2/launch/issues/835#issuecomment-2746363129. This is the architecture doc and not user documentation, but I think we can still provide a clue as to how substitutions are concatenated here.

### DIFF
--- a/launch/doc/source/architecture.rst
+++ b/launch/doc/source/architecture.rst
@@ -120,7 +120,8 @@ There are many possible variations of a substitution, but here are some of the c
 - :class:`launch.substitutions.Text`
 
   - This substitution simply returns the given string when evaluated.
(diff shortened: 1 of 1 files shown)


fujitatomoya (approved):
i also think that is fine to add `how` here.

christophebedard:
Running a ci_linux job just to be safe since there's no Rpr job: [Build Status](https://ci.ros2.org/job/ci_linux/22850/)

christophebedard:
I plan to backport this, since people don't really look at the API docs for Rolling

christophebedard:
@mergifyio backport jazzy humble

PR #858, opened 2025-04-13
Add Other Logging Implementations

### LINKED ISSUE #821: Add logging actions for severity levels other than 'INFO'
Currently there is no action analogous to `LogInfo` for other severity levels such as 'WARN', 'DEBUG' and 'ERROR'.
I was wondering whether this is a design choice or a feature that is yet to be implemented.

InvincibleRMC:
Added generic `Log` class and Specializations for `LogInfo`, `LogDebug`, `LogError`, `LogWarning`.

Closes #821


launch/launch/actions/log_info.py
```diff
@@ -1,63 +0,0 @@
...
-from ..frontend import expose_action
-from ..frontend import Parser  # noqa: F401
-from ..launch_context import LaunchContext
-from ..some_substitutions_type import SomeSubstitutionsType
-from ..substitution import Substitution
-from ..utilities import normalize_to_list_of_substitutions
-
-
-@expose_action('log')
-class LogInfo(Action):
```
christophebedard, line 32: The downside to removing this is that, if someone is importing `LogInfo` by doing:

```py
from launch.actions.log_info import LogInfo
```

instead of:

```py
from launch.actions import LogInfo
```

it will just break. It's probably just a small percentage of imports, but I think it would still be worth keeping this file and re-exporting `LogInfo`. You could also add a warning in this file so that it gets printed if the file is imported directly (I _think_ that should work). It's probably overkill, but it's not hard to do, so might as well.

launch/launch/actions/log.py
```diff
@@ -0,0 +1,127 @@
...
+        """Getter for self.__level."""
+        return self.__level
+
+    def execute(self, context: LaunchContext) -> None:
+        """Execute the action."""
+        level_sub = ''.join([context.perform_substitution(sub)
+                             for sub in self.level]).upper()
+
+        self.__logger.log(
+            logging.getLevelNamesMapping()[level_sub],
```
christophebedard: I would do the mapping before the call to `self.__logger.log()` and raise a nicer, more helpful error message instead of letting invalid `level` values lead to a simple `KeyError`. Something like:

```py
level_map = logging.getLevelNamesMapping()
if level_sub not in level_map:
    raise KeyError(f"Invalid log level '{level_sub}', expected: {level_map.keys()}")
```

I would also test that an error is raised/that the launch file execution correctly leads to an error if an invalid `level` is used. Doing that just for the Python tests is fine.

launch/launch/actions/log.py
```diff
@@ -0,0 +1,127 @@
...
+        kwargs['msg'] = parser.parse_substitution(entity.get_attr('message'))
+
+        # Check if still using old log action
+        level = entity.get_attr('level', optional=True)
+        if level is None:
+            warnings.warn(
+                'The action log now expects a log level.'
+                ' Either provide one or switch to using the log_info action',
+                stacklevel=2)
+            level = 'INFO'
```
christophebedard: Can you add a comment saying something like this here:

```py
# TODO: Make 'level' mandatory in Rolling after L-turtle release
```

with other similar comments everywhere where warnings/temporary code should be updated later. This way it's easy to find and update these in ~1 year :grin:

christophebedard (commented):
A few minor comments, otherwise this looks good

launch/launch/actions/log_info.py
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+
+"""Module for the LogInfo action."""
+
+import warnings
+
+from .log import LogInfo as LogInfo  # noqa: F401
+
+# TODO: Remove log_info.py for Release after L-turtle release
+warnings.warn('importing from log_info.py is deprecate; import LogInfo from log.py instead.')
```
christophebedard:
```suggestion
warnings.warn('importing from log_info.py is deprecated; import LogInfo from log.py instead.')
```

launch/test/launch/actions/test_log.py
```diff
@@ -81,3 +81,11 @@ def test_log_execute():
...
     launch_context = LaunchContext()
     assert log.visit(launch_context) is None
+
+
+def test_log_level_error():
+    """Checks for error message to be raised given invalid level."""
+    try:
+        Log(msg='foo', level='foo')
+    except KeyError as e:
+        assert 'Invalid log level ' in str(e)
```
christophebedard: This doesn't actually assert that an error is raised. You can use `with pytest.raises(KeyError, match=r'Invalid log level.*'):`: https://docs.pytest.org/en/stable/how-to/assert.html#matching-exception-messages

And an error wouldn't be raised in this case if you're just calling the constructor; you have to call `visit()` on it so that `execute()` is called.

christophebedard (approved):

christophebedard:
Pulls: ros2/launch#858
Branch: mergify/bp/kilted/pr-36
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15809
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23277/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17494/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2744/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24161/)

christophebedard:
The test failures are unrelated. Thanks for the PR!

christophebedard:
Pulls: ros2/launch#858
Gist: https://gist.githubusercontent.com/christophebedard/0eed6b9363abeb033c16bff0396696cf/raw/6f35f009047a5ffc1577a49c27074f1c4e1213e1/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15813
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23281/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17498/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2748/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24166/)

christophebedard:
So I just noticed that I didn't properly test this: the first CI run is testing branch `mergify/bp/kilted/pr-36` across all repos, which ends up testing https://github.com/ros2/ament_cmake_ros/pull/36 and not this PR. This is because I'm using the [`ros-ci-for-pr` script](https://github.com/ros-tooling/ros-github-scripts) and re-used a previous command without removing the [`--branch` option](https://github.com/ros-tooling/ros-github-scripts#with-a-branch-name). It commented on this PR but didn't test this PR. I only noticed now when triggering CI for another PR.

I think this PR _would_ be fine, but it uses [`logging.getLevelNamesMapping()`](https://docs.python.org/3/library/logging.html#logging.getLevelNamesMapping), which is only available in Python 3.11. That is greater than the minimum Python language requirement for Rolling/Kilted: https://www.ros.org/reps/rep-2000.html#kilted-kaiju-may-2025-november-2026. Therefore, based on this, I'll revert this PR right away.

christophebedard:
Reverted in #865. We'll see if CI passes above, but just based on the minimum Python requirement, we'll have to find an alternative to `logging.getLevelNamesMapping()`.

@InvincibleRMC would you be willing to open another PR that doesn't use `logging.getLevelNamesMapping()`?

InvincibleRMC:
> @InvincibleRMC would you be willing to open another PR that doesn't use `logging.getLevelNamesMapping()`?

Yeah definitely.

christophebedard:
So it does fail on RHEL: https://ci.ros2.org/job/ci_linux-rhel/2748/testReport/junit/launch.test.launch.actions/test_log/test_log_execute/. Windows will probably fail too, knowing that it has an older Python version, but I'll just cancel it.

PR #861, opened 2025-04-20
use custom log_file name as per the user setting

### LINKED ISSUE #380: Consider providing user control of log file base names
#### Feature description
I have reported an issue in the `ros2cli` repo recently (https://github.com/ros2/ros2cli/issues/856) which hopefully will gain traction.

The general idea in the above issue -- and please read that for more details -- is that the default log file names right now are prepended with the executable name (in the C++ case) or the Python executable name (e.g., `python3`) in the Python case.

The investigation is leading to maybe providing a user-configurable parameter for `ros2 run` that allows the user to provide their own base name for log files.

Per @fujitatomoya's suggestion, I wanted to open this up here as well to ensure that the `launch_ros` tools track the above issue and have a way to expose this through `ros2 launch`, and the launch APIs, as well.

izzaaaatanishq:
closes https://github.com/ros2/launch_ros/issues/380

### DIFF
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -56,7 +56,8 @@ class LaunchService:
         *,
         argv: Optional[Iterable[Text]] = None,
         noninteractive: bool = False,
-        debug: bool = False
+        debug: bool = False,
+        log_file_name: str = 'launch.log'
     ) -> None:
         """
         Create a LaunchService.
@@ -68,11 +69,17 @@ class LaunchService:
         """
         # Setup logging and debugging.
         launch.logging.launch_config.level = logging.DEBUG if debug else logging.INFO
+        self._log_file_name = log_file_name
+        # Ensure the log file name ends with `.log`
+        if not self._log_file_name.endswith('.log'):
+            self._log_file_name += '.log'
+        launch.logging.launch_config.log_file_name = self._log_file_name
+        # Setup logging
+        self._logger_name = self._log_file_name.removesuffix('.log')
+        self.__logger = launch.logging.get_logger(self._logger_name)
+
         self.__debug = debug
         self.__argv = argv if argv is not None else []
-
-        # Setup logging
-        self.__logger = launch.logging.get_logger('launch')
 
         # Setup context and register a built-in event handler for bootstrapping.
         self.__context = LaunchContext(argv=self.__argv, noninteractive=noninteractive)
--- a/launch/launch/logging/__init__.py
+++ b/launch/launch/logging/__init__.py
@@ -483,6 +483,8 @@ def get_output_loggers(process_name: str, output_config: Union[str, Dict[str, An
     :param process_name: the process-like action whose outputs want to be logged.
     :param output_config: configuration for the output loggers,
         see above for details.
+    :param main_log_file_name: the name of the main log file to be used.
+        Defaults to 'launch.log'.
     :returns: a tuple with the stdout and stderr output loggers.
     """
     output_config = _normalize_output_configuration(output_config)
--- a/launch/launch/logging/__init__.py
+++ b/launch/launch/logging/__init__.py
@@ -19,20 +19,11 @@ import datetime
 import locale
 import logging
 import logging.handlers
-
 import os
 import socket
 import sys
-
-from typing import Any
-from typing import Dict
-from typing import List
-from typing import Literal
-from typing import Optional
-from typing import Protocol
-from typing import Set
-from typing import Tuple
-from typing import Union
+from typing import (Any, Dict, List, Literal, Optional, Protocol, Set, Tuple,
+                    Union)
 
 from typing_extensions import TypeAlias
 
@@ -125,6 +116,7 @@ class LaunchConfig:
 
     def reset(self) -> None:
         self._log_dir = None
+        self._log_file_name = 'launch.log'
         self.file_handlers: Dict[str, logging.Handler] = {}
         self.screen_handler = None
         self.screen_formatter: Optional[logging.Formatter] = None
@@ -146,6 +138,20 @@ class LaunchConfig:
         :param new_level: the default log level used for all loggers.
         """
         logging.root.setLevel(new_level)
+
+    @property
+    def log_file_name(self) -> str:
+        """Get the current log file name."""
+        return self._log_file_name
+
+    @log_file_name.setter
+    def log_file_name(self, log_file_name: str):
+        """
+        Set the name of the log file.
+
+        :param log_file_name: the name of the log file where logger output should be written.
+        """
+        self._log_file_name = log_file_name
 
     @property
     def log_dir(self):
@@ -366,7 +372,7 @@ def get_logger(name: Optional[str] = None) -> logging.Logger:
     screen_handler = launch_config.get_screen_handler()
     if screen_handler not in logger.handlers:
         logger.addHandler(screen_handler)
-    launch_log_file_handler = launch_config.get_log_file_handler()
+    launch_log_file_handler = launch_config.get_log_file_handler(launch_config.log_file_name)
     if launch_log_file_handler not in logger.handlers:
         logger.addHandler(launch_log_file_handler)
     return logger
@@ -435,8 +441,8 @@ def _normalize_output_configuration(config: Union[str, Dict[str, Any]]) -> Dict[
     return normalized_config
 
 
-def get_output_loggers(process_name: str, output_config: Union[str, Dict[str, Any]]
-                       ) -> Tuple[logging.Logger, logging.Logger]:
+def get_output_loggers(process_name: str, output_config: Union[str, Dict[str, Any]],
+                       main_log_file_name='launch.log') -> Tuple[logging.Logger, logging.Logger]:
     """
     Get the stdout and stderr output loggers for the given process name.
 
@@ -496,7 +502,7 @@ def get_output_loggers(process_name: str, output_config: Union[str, Dict[str, An
         # If a 'log' output is configured for this source or for
         # 'both' sources, this logger should output to launch main log file.
         if 'log' in (output_config['both'] | output_config[source]):
-            launch_log_file_handler = launch_config.get_log_file_handler()
+            launch_log_file_handler = launch_config.get_log_file_handler(main_log_file_name)
             # Add launch main log file handler if necessary.
             if launch_log_file_handler not in logger.handlers:
                 launch_log_file_handler.setFormatterFor(
--- a/launch/test/launch/test_launch_service.py
+++ b/launch/test/launch/test_launch_service.py
@@ -26,7 +26,18 @@ import osrf_pycommon
 
 def test_launch_service_constructors():
     """Test the constructors for LaunchService class."""
-    LaunchService()
+    # When no name is provided for log file
+    ls = LaunchService()
+    assert ls._log_file_name == 'launch.log'
+    assert ls._logger_name == 'launch'
+    # WHen log file name is provided without .log
+    ls = LaunchService(log_file_name='custom_logger')
+    assert ls._log_file_name == 'custom_logger.log'
+    assert ls._logger_name == 'custom_logger'
+    # When log file name is provided with .log
+    ls = LaunchService(log_file_name='custom_logger.log')
+    assert ls._log_file_name == 'custom_logger.log'
+    assert ls._logger_name == 'custom_logger'
     LaunchService(debug=True)
     LaunchService(debug=False)
 
--- a/launch/test/launch/test_logging.py
+++ b/launch/test/launch/test_logging.py
@@ -67,7 +67,7 @@ def test_output_loggers_bad_configuration(log_dir):
         launch.logging.get_output_loggers('some-proc', {'stdout': {'garbage'}})
 
 
-@pytest.mark.parametrize('config,checks', [
+configs = [
     ('screen', {'stdout': {'screen'}, 'stderr': {'screen'}}),
     ('log', {'stdout': {'log'}, 'stderr': {'log', 'screen'}}),
     ('both', {'both': {'log', 'screen'}}),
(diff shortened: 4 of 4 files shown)


launch/launch/launch_service.py
```diff
@@ -67,6 +68,9 @@ def __init__(
         """
         # Setup logging and debugging.
         launch.logging.launch_config.level = logging.DEBUG if debug else logging.INFO
+        # Set the log file name.
+        if log_file_name is not None:
+            launch.logging.launch_config.file_name = log_file_name
```
fujitatomoya: how about applying this to logger name line 78 as well?

```suggestion
            # Ensure the log file name ends with `.log`
            if not log_file_name.endswith('.log'):
                log_file_name += '.log'
            launch.logging.launch_config.file_name = log_file_name
```

fujitatomoya (commented):
test would be ideal to make sure the specified log file name is generated with `ros2 launch --log-file-name` command.

launch/launch/launch_service.py
izzaaaatanishq: suggestion applied

izzaaaatanishq:
> test would be ideal to make sure the specified log file name is generated with `ros2 launch --log-file-name` command.

@fujitatomoya 
I am not clear what you mean here.

launch/launch/logging/__init__.py
```diff
@@ -307,6 +319,7 @@ def get_log_file_path(self, file_name='launch.log'):
         :param: file_name of the log file from which to get the absolute path.
         :return: the absolute path to the log file.
         """
+        file_name = self.file_name
```
christophebedard: I don't think this is right. This method should use the `file_name` arg if provided, otherwise it should use `self._file_name`.

Tanishq30052002: updated

Tanishq30052002:
> > test would be ideal to make sure the specified log file name is generated with `ros2 launch --log-file-name` command.
> 
> @fujitatomoya 
I am not clear what you mean here.

launch/launch/logging/__init__.py
izzaaaatanishq: @christophebedard I have updated the functionality of the function

christophebedard: I think you need to do to `get_log_file_path()` what you did to `get_log_file_handler()` so that it uses the provided `file_name` arg if there is one.

christophebedard:
Also, please take a look at the test results here: https://build.ros2.org/job/Rpr__launch__ubuntu_noble_amd64/164/testReport/. There are some flake8 linter issues and some test failures that seem to be related to your change. You can run the tests locally

launch/launch/logging/__init__.py
izzaaaatanishq: I don't think so, since in `get_log_file_handler` we are even checking if the file name is present in the file handlers. Also, why disturb the functionality of the `get_log_file_path`? Its purpose should just be to complete the path with the log file name and return.

izzaaaatanishq:
@christophebedard i have reolved all the issues and resolved the failing test. 

@fujitatomoya can you please review it once again?

launch/launch/logging/__init__.py
izzaaaatanishq: @christophebedard can we resolve this conversation now?

launch/launch/logging/__init__.py
```diff
@@ -309,7 +321,7 @@ def get_log_file_path(self, file_name='launch.log'):
         """
         return os.path.join(self.log_dir, file_name)
 
-    def get_log_file_handler(self, file_name='launch.log'):
+    def get_log_file_handler(self, file_name=None):
```
fujitatomoya: for me, this method API is confusing now. (and `get_log_file_path` method above.)
with this PR, file name now belongs to `LaunchConfig` class with setter and getter.
so i think that user application is expected to set the file name either constructor or setter, why it provides the file_name argument here? probably this is just a backward compatibility or temporary override the file name when this method is called? maybe at least we can update the docstring?

fujitatomoya (commented):
@Tanishq30052002 thanks for working on this, had a couple of minor comments.

fujitatomoya:
> > test would be ideal to make sure the specified log file name is generated with `ros2 launch --log-file-name` command.
> 
> @fujitatomoya I am not clear what you mean here.

i am just suggesting unit tests for CI/CD to make sure nothing is broken during development.
since this PR changes `LaunchService` and `LaunchConfig`, i think we can add test to ensure file_name is properly handled in those classes. and as user interface, that would be nice to add test to https://github.com/ros2/launch_ros/tree/rolling/ros2launch/test. but i do not see any actually test cases exist for using `ros2launch`, so this one is probably nice to have.

launch/launch/logging/__init__.py
Tanishq30052002: Because there are currently many calls to this function, with different file names, we can consider "own log" cases.
So I will check if nothing is provided then use the log file name provided by the user, else take the arg provided by function call.

izzaaaatanishq: @fujitatomoya i have updated the functionality of these function, and done changes in the effecting tests

izzaaaatanishq:
> am just suggesting unit tests for CI/CD to make sure nothing is broken during development.
> since this PR changes `LaunchService` and `LaunchConfig`, i think we can add test to ensure file_name is properly handled in those classes. and as user interface, that would be nice to add test to

ok, sounds good. I will create test for the same.
currently functionality is working and it is working as expected with failing any current tests

izzaaaatanishq:
> > am just suggesting unit tests for CI/CD to make sure nothing is broken during development.
> > since this PR changes `LaunchService` and `LaunchConfig`, i think we can add test to ensure file_name is properly handled in those classes. and as user interface, that would be nice to add test to
> 
> ok, sounds good. I will create test for the same. currently functionality is working and it is working as expected with failing any current tests

@fujitatomoya i have written tests for launch service and launch config. you may review it once and let me know if something more can be added.

launch/launch/launch_service.py
```diff
@@ -55,7 +55,8 @@ def __init__(
         *,
         argv: Optional[Iterable[Text]] = None,
         noninteractive: bool = False,
-        debug: bool = False
+        debug: bool = False,
+        log_file_name: Optional[Text] = None
```
fujitatomoya: This can be `launch` in default? the following checks and setting looks like exactly default assignment?

launch/launch/logging/__init__.py
```diff
@@ -130,6 +131,22 @@ def level(self, new_level):
         """
         logging.root.setLevel(new_level)
 
+    @property
+    def log_file_name(self) -> str:
+        """Get the current log file name."""
+        if self._log_file_name is None:
+            self._log_file_name = 'launch.log'
+        return self._log_file_name
```
fujitatomoya: i think this should return the `self._log_file_name` as it is without storing the internal default value. this generates the bug something like,

```python
if logging_config.log_file_name is None:
    logging_config.log_file_name('my_launch.log')
```

this is never be set.

launch/launch/logging/__init__.py
```diff
@@ -300,33 +317,30 @@ def set_log_format(self, log_format, *, log_style=None):
         else:
             self.file_formatter = None
 
-    def get_log_file_path(self, file_name='launch.log'):
+    def get_log_file_path(self):
```
fujitatomoya: https://github.com/ros2/launch/pull/861#discussion_r2060690548 was originally question. for backward compatibility, i think we do not have to break the API because of this. so maybe rewind this change as it was before?

fujitatomoya (commented):
@Tanishq30052002 we are getting closer, and thanks for being patient and iterating with me. I have only a couple of minor things, could you please have a look? (thanks for adding test though.)

launch/launch/launch_service.py
izzaaaatanishq: can you please elaborate a bit? I am not clear.

launch/launch/logging/__init__.py
izzaaaatanishq: But since now file name is a class variable so why do we need to pass it in the function.
I am not clear

launch/launch/logging/__init__.py
izzaaaatanishq: but we know that log file name will return the file name, and willl set some default value if no name is already set.

launch/launch/launch_service.py
izzaaaatanishq: @fujitatomoya would be glad if you could elaborate what you mean by last 3 comments.

fujitatomoya: this one, i am just saying that we can assign the default value for `log_file_name`. that is what it does the following process.

launch/launch/logging/__init__.py
fujitatomoya: in that case, default name should be set in the 1st place? because it is default. and then this getter property simply returns the value in `_log_file_name`. otherwise, above example does not work.

launch/launch/logging/__init__.py
fujitatomoya: to keep the backward compatibility. i do not think there are many applications rely on these functions, but there can be. and if we change this API like this, it will break their userspace. i mean we can do (actually we do) that if necessary such as new features, security issue we must fix. but this is to add the log file name option to `ros2 launch`, so i think maybe we just want to keep the backward compatibility, so that we do not necessarily break the user space.

launch/launch/launch_service.py
izzaaaatanishq: i have resolved this issue

launch/launch/logging/__init__.py
izzaaaatanishq: but if someone wants to change the name he is allowed to do so. also what will be the benefit of knowing file name is none?

launch/launch/logging/__init__.py
izzaaaatanishq: i think we shouldnt go with this idea, otherwise we might need to check if file name is provided use it else check for class variable. 

also since this is a module which is not user accessible. this is something specific to development of ros2 so we should not bother.

launch/launch/logging/__init__.py
fujitatomoya: I really do not understand why you want to go this way, can you explain the reason against the following points?

- `_log_file_name` is initialized lazily, i.e., only when the property is accessed for the first time.
- why getter property sets the internal value in the very 1st place? user does not expect this behavior.
- If any code tries to directly access `_log_file_name` before the `log_file_name` property is accessed, it would find `None` instead of the intended default ('launch.log').
- initialization of `_log_file_name` relies on this getter property, and if `_log_file_name` is not set correctly upfront, it might lead to unexpected behavior.

launch/launch/logging/__init__.py
fujitatomoya: > i think we shouldnt go with this idea, otherwise we might need to check if file name is provided use it else check for class variable.

i disagree. that i think it is not worth to break the user space for this. i really think we should keep the backward compatibility with this fix, and that is not even hard to do so. we can take the file_name argument and set it to `_log_file_name`, probably docstring could be added to explain that.

> also since this is a module which is not user accessible. this is something specific to development of ros2 so we should not bother.

that is incorrect.
this is accessible by user application, you already changes the test code because of this. test code is the user application, and there could be thousands application rely on this API. i say we should not bother the user application because of this change.

fujitatomoya (changes requested):
@Tanishq30052002 i am not convinced to take this current change. please review my comments and let's not break the user space.

izzaaaatanishq:
> @Tanishq30052002 i am not convinced to take this current change. please review my comments and let's not break the user space.

I got your point, I will try to resolve these issues

launch/launch/logging/__init__.py
izzaaaatanishq: so for this i will set the _log_file_name to 'launch.log' directly during initialization.
I hope it will serve the purpose

launch/launch/logging/__init__.py
izzaaaatanishq: @fujitatomoya i have modified as per backward compatibility. Have a look once.
But it might have one limitation. 
If we change log_file_name for default case we need to modify at 2 places now.

launch/launch/logging/__init__.py
```diff
@@ -300,33 +315,32 @@ def set_log_format(self, log_format, *, log_style=None):
         else:
             self.file_formatter = None
 
-    def get_log_file_path(self, file_name='launch.log'):
+    def get_log_file_path(self, log_file_name='launch.log'):
```
fujitatomoya: this still breaks the user application, we should keep `file_name` here.

```suggestion
    def get_log_file_path(self, file_name='launch.log'):
```

launch/launch/logging/__init__.py
```diff
@@ -300,33 +315,32 @@ def set_log_format(self, log_format, *, log_style=None):
...
 
-        :param: file_name of the log file from which to get the absolute path.
+        :param log_file_name: name of the file for which we want full path
         :return: the absolute path to the log file.
         """
-        return os.path.join(self.log_dir, file_name)
+        return os.path.join(self.log_dir, log_file_name)
 
-    def get_log_file_handler(self, file_name='launch.log'):
+    def get_log_file_handler(self):
```
fujitatomoya: same here. also corresponding code including test needs to roll back as original.

fujitatomoya (changes requested):

izzaaaatanishq:
@fujitatomoya i have reverted back changes and also squash multiple commits to keep the history clean.
Also there are some changes in test of logging which are neccessory for testing cutom file name logging.

launch/launch/logging/__init__.py
```diff
@@ -347,7 +362,7 @@ def get_logger(name=None) -> logging.Logger:
     screen_handler = launch_config.get_screen_handler()
     if screen_handler not in logger.handlers:
         logger.addHandler(screen_handler)
-    launch_log_file_handler = launch_config.get_log_file_handler()
+    launch_log_file_handler = launch_config.get_log_file_handler(launch_config.log_file_name)
```
fujitatomoya, line 375: i do not think this change is required.

instead, i think we need to set `self._log_file_name` on `get_log_file_handler` call to keep compatible behavior with docstring update?

izzaaaatanishq:
@fujitatomoya any changes to which needs to be done yet?

fujitatomoya:
@Tanishq30052002 thanks for the ping, i will take a look in this week.

launch/launch/logging/__init__.py
izzaaaatanishq: I dont think we should change this, bcuz if we do so then file name get updated to the log file name as per class member even though user provide the arg for file name.

izzaaaatanishq: @fujitatomoya ?

izzaaaatanishq: @fujitatomoya can we finalize changes? Pending from a long time.
If nothing is required from the logic side then i will resolve the merge conflict

fujitatomoya: @Tanishq30052002 sorry i missed that, i will take a look today or tomorrow for sure.

fujitatomoya:
@Mergifyio rebase

fujitatomoya:
@Tanishq30052002 can you rebase this? there is a conflict, mergifyio cannot do that.

izzaaaatanishq:
I will do it in the morning, currently equipped with other tasks.

launch/launch/logging/__init__.py
fujitatomoya: i think i was mixed up on this, it seems to be correct change.

izzaaaatanishq:
> I will do it in the morning, currently equipped with other tasks.

@fujitatomoya rebased

fujitatomoya (approved):

fujitatomoya:
Pulls: ros2/launch#861, ros2/launch_ros#461
Gist: https://gist.githubusercontent.com/fujitatomoya/99be9019fa1e6cc34a51fba991d9df57/raw/489985c08e7696fc50178cb8b0248c3e758490b5/ros2.repos
BUILD args:  --packages-above-and-dependencies launch ros2launch
TEST args:  --packages-above launch ros2launch
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16216
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23957/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18014/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3358/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24605/)

izzaaaatanishq:
@fujitatomoya can you check this once.

ahcorde (approved):

PR #862, opened 2025-04-23
Scope launch file dir/path locals to included launch file

### LINKED ISSUE #618: $(dirname) is affected by previous ($find-pkg-share ...)
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04...
- Installation type:
  - binaries
- Version or commit hash:
  - up to date
- DDS implementation:
  - irrelevant
  - irrelevant

#### Steps to reproduce issue

Create file `test.launch.yaml`.

```yaml
launch:
  - executable:
      cmd: echo $(dirname)
      name: obvious
      output: screen
  - include:
      file: $(find-pkg-share some_package)/file.launch.yaml
  - executable:
      cmd: echo $(dirname)
      name: surprise
      output: screen
```

The file `file.launch.yaml` in `some_package`:
```yaml
launch: []
```

Run `ros2 launch test.launch.yaml`.

#### Expected behavior

`echo $(dirname)` outputs the path to `test.launch.yaml` each time.

#### Actual behavior

`echo $(dirname)` outputs the path to `test.launch.yaml` the first time and the path to the package share of `some_package` the second time. It appears that "the context" is somehow not reset. Is this really intentional? I also could not find the docs for that substitution.

#### Additional information

```yaml
launch:
  - executable:
      cmd: echo $(dirname)
      name: obvious
      output: screen
  - executable:
      cmd: echo $(find-pkg-share some_package)
      name: testing_this
      output: screen
  - executable:
      cmd: echo $(dirname)
      name: surprise
      output: screen
```

In this setting `echo $(dirname)` outputs the path to `test.launch.yaml` each time.

It is very unexpected but also breaking when changed.

christophebedard:
Resolves #618

Some context locals are set by `IncludeLaunchDescription` with the path to the launch file (`current_launch_file_path`) and the path to the directory containing the launch file (`current_launch_file_directory`). They can then be used through substitutions: `ThisLaunchFile()` and `ThisLaunchFileDir()`, respectively. These only make sense in the included launch file itself, and these locals should not be set when not in a(n included) launch file, e.g., when in a launch script. That usually happens when calling the launch service manually, which is common in tests.

The issue reported in #618 is that `IncludeLaunchDescription` doesn't reset the context locals to the previous values. This means that the two launch file location locals point to the included launch file even _after_ the `IncludeLaunchDescription` action in the parent launch file. See the reproducer in the original issue.

The fix is to scope the two launch file location context locals to `IncludeLaunchDescription`. We also have to _remove_ the context locals after `IncludeLaunchDescription` if it was included by a launch script (see test). Since there is no easy way/API to _remove_ locals, I had to work around it a bit. I'm open to ideas to make that better.

Note that the solution _only_ scopes the two launch file location locals to `IncludeLaunchDescription`. Other locals are kept (see test), since included launch files are meant to act as if they were literally included in the parent launch file.

### DIFF
--- a/launch/launch/actions/include_launch_description.py
+++ b/launch/launch/actions/include_launch_description.py
@@ -28,7 +28,7 @@ from typing import Union
 
 import launch.logging
 
-
+from .opaque_function import OpaqueFunction
 from .set_launch_configuration import SetLaunchConfiguration
 from ..action import Action
 from ..frontend import Entity
@@ -212,13 +212,7 @@ class IncludeLaunchDescription(Action):
                                                             LaunchDescriptionEntity]]:
         """Execute the action."""
         launch_description = self.__launch_description_source.get_launch_description(context)
-        # If the location does not exist, then it's likely set to '<script>' or something.
-        context.extend_locals({
-            'current_launch_file_path': self._get_launch_file(),
-        })
-        context.extend_locals({
-            'current_launch_file_directory': self._get_launch_file_directory(),
-        })
+        self._set_launch_file_location_locals(context)
 
         # Do best effort checking to see if non-optional, non-default declared arguments
         # are being satisfied.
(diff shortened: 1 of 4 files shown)


ahcorde:
Pulls: ros2/launch#862
Gist: https://gist.githubusercontent.com/ahcorde/75fc455634320b6afd08a30d2d6b6da8/raw/76bd55d96a58f8678f35912fa2b64de3c5d374c3/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17642
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/26345/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/19993/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6044/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26085/)

fujitatomoya (approved):
this looks good to me, but i would like to have one more approval on this before merge.

ahcorde (approved):

christophebedard:
Thanks for the reviews and CI!

peci1:
How is it with backportability of this PR to Kilted/Jazzy?

PR #865, opened 2025-04-28
Revert "Add Other Logging Implementations"

christophebedard:
Reverts ros2/launch#858

See https://github.com/ros2/launch/pull/858#issuecomment-2836447242


christophebedard:
This is a simple revert and the Rpr job looks good, so merging

PR #866, opened 2025-04-28
Other Logging Implementations with `getLevelNamesMapping` fix

InvincibleRMC:
Redo of #858 

With fix.


christophebedard:
Can you rebase on `rolling`?

launch/launch/actions/log.py
```diff
@@ -0,0 +1,144 @@
...
+if sys.version_info >= (3, 11):
+    log_levels = logging.getLevelNamesMapping()
+else:
+    # TODO: Remove after Python 3.11+ is minimum support
+    log_levels = {'NOTSET': logging.NOTSET,
+                  'DEBUG': logging.DEBUG,
+                  'INFO': logging.INFO,
+                  'WARNING': logging.WARNING,
+                  'ERROR': logging.ERROR,
+                  'CRITICAL': logging.CRITICAL}
```
christophebedard: Very minor, but I'd prefer doing this directly in `Log.execute()` if `log_levels` is not used anywhere else.

You could also just check if `hasattr(logging, 'getLevelNamesMapping')` instead of checking `sys.version`.

InvincibleRMC: I typically prefer using sys.version_info since I find it more human readable at a glance when branches can be removed in the future. I also find it plays better with type checkers. However, if you prefer `hasattr` I will use it.

christophebedard: No, that's fine! Thanks for moving it.

christophebedard (approved):
Thanks for the PR (again) and for making the fix!

Now I'll trigger CI and triple-check :laughing:

christophebedard:
Pulls: ros2/launch#866
Gist: https://gist.githubusercontent.com/christophebedard/45312c1e520537aae1aebfc9f7e490b1/raw/6f35f009047a5ffc1577a49c27074f1c4e1213e1/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15820
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23288/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17505/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2779/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24173/)

christophebedard:
Alright, CI looks good. The test failures are unrelated.

emersonknapp:
@InvincibleRMC first off, great stuff! really happy to see this feature. noticed a small issue logged in https://github.com/ros2/launch/issues/882, that frontends using these print a `UserWarning` even though it actually works fine. not sure at first glance how to fix it - if you know easily then that would be helpful :)

InvincibleRMC:
> @InvincibleRMC first off, great stuff! really happy to see this feature. noticed a small issue logged in #882, that frontends using these print a `UserWarning` even though it actually works fine. not sure at first glance how to fix it - if you know easily then that would be helpful :)

Oops. Good catch. I introduced a fix in #883.

PR #891, opened 2025-06-27
Remove `LaunchDescriptionArgument` from documentation

### LINKED ISSUE #890: `LaunchDescriptionArgument` from documentation doesn't exist
### Problem

I was going through the launch documentation here: https://docs.ros.org/en/humble/p/launch/architecture.html#substitutions and noticed `LaunchDescriptionArgument` but didn't really the description for what function it served. I searched the code base, and there was no associated launch class at least as far back as foxy - though in foxy, there was a `DeclareLaunchDescriptionArgument` action, so those seemed to be connected in some way (NOTE: even though `DeclareLaunchDescriptionArgument` shows up in the `foxy` documentation, the code shows that the class was called `DeclareLaunchArgument`, meaning that documentation entry was also not correct).

### Proposed Change

Pretty basic, just remove that `LaunchDescriptionArgument` entry in the documentation. I can put up a PR for it, so long as the change makes sense and I'm not accidentally missing something. I assume that would go into rolling, then get backported to the relevant branches.

chen-harrison:
## Description

Remove nonexistent substitution class LaunchDescriptionArgument from documentation

Fixes #890

### DIFF
--- a/launch/doc/source/architecture.rst
+++ b/launch/doc/source/architecture.rst
@@ -136,10 +136,6 @@ There are many possible variations of a substitution, but here are some of the c
 
   - This substitution takes a substitution, and if it evaluates to true, then the result is the if_value, else the result is the else_value.
 
-- :class:`launch.substitutions.LaunchDescriptionArgument`
-
-  - This substitution gets the value of a launch description argument, as a string, by name.
-
 - :class:`launch.substitutions.LocalSubstitution`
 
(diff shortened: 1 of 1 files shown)


christophebedard (approved):
Since this only changes non-code documentation, I'll skip CI.

christophebedard:
@Mergifyio backport kilted jazzy humble

PR #901, opened 2025-07-17
Update is_idle function to also consider async tasks

### LINKED ISSUE #747: Launch doesn't stop running nodes when exception thrown
## Bug report

**Required Info:**

- Operating System:
Ubuntu 22.04
- Installation type:
Binary
- Version or commit hash:
Iron 2.0.2
- DDS implementation:
Default
N/A

#### Steps to reproduce issue

Create launch file `my_launch.py`

```py
import launch
import launch_ros.actions


def generate_launch_description() -> LaunchDescription:
    return launch.LaunchDescription(
        [
            # Real nodes - these get left running when launch file throws exception
            launch_ros.actions.Node(
                package="examples_rclcpp_minimal_publisher",
                executable="publisher_member_function",
            ),
            launch_ros.actions.Node(
                package="examples_rclcpp_minimal_subscriber",
                executable="subscriber_member_function",
            ),
            # Fake node - this will cause an exception to be thrown.
            launch_ros.actions.Node(
                package="examples_rclcpp_minimal_subscriber",
                executable="fake_node",
            ),
        ]
    )
```
Launch it
```
$ ros2 launch my_launch.py
```

#### Expected behavior
Launch catches exception and stops, including stopping any nodes that were started before the exception was caught, i.e., `ros2 node list` returns nothing afterwards.

#### Actual behavior
Launch catches exception and stops, but doesn't kill any of the already started nodes.
```
$ ros2 launch my_launch.py
[INFO] [launch]: All log files can be found below /home/jackie/.ros/log/2023-12-05-14-26-16-767994-stealth-126383
[ERROR] [launch]: Caught exception in launch (see debug for traceback): executable 'fake_node' not found on the libexec directory '/opt/ros/iron/lib/examples_rclcpp_minimal_subscriber' 
$ ros2 node list
/minimal_publisher
/minimal_subscriber
```

#### Additional information

Ideally of course the launch file is correctly written such that no exceptions are thrown in the first place. However, when in the process of writing a new launch file (especially one much larger than this example), it is hard to remember, unexpected and tedious to individually kill every node that was started before the exception was thrown. 

In my own troubleshooting I looked at what happens when the launch file is Ctrl+C'd vs. what happens when an exception is caught. Perhaps there's something in there (namely, some behavior difference when `LaunchService._shutdown()` is passed `due_to_sigint=True` and `force_sync=True`)?

https://github.com/ros2/launch/blob/f15691c341d32e53970fa18a9a22c9206c2edfbe/launch/launch/launch_service.py#L191-L202

https://github.com/ros2/launch/blob/f15691c341d32e53970fa18a9a22c9206c2edfbe/launch/launch/launch_service.py#L345-L355

MegaIng:
## Description

This PR changes the `LaunchService._is_idle` function to also consider running background tasks. This is to mitigate the error path described in [this comment](https://github.com/ros2/launch/issues/747#issuecomment-3071234903).

The core point is that there might be background tasks running that will generate events in the future, which isn't being taken into consideration, therefore the process exits despite there still being stuff to do.

Fixes #747 

Hopefully and probably not. All old tests still work, but people might be relying on interactions that I can't predict with my limited experience with this software.

### Additional Information

Another person I worked with had some issues of a test breaking in a flaky way - but it's not clear if it was caused by this change, the exact changes have changed since then and I couldn't reproduce it. Hopefully that doesn't actually relate to these changes.

### DIFF
--- a/launch/launch/launch_service.py
+++ b/launch/launch/launch_service.py
@@ -94,7 +94,7 @@ class LaunchService:
         # it being set to None by run() as it exits.
         self.__loop_from_run_thread_lock = threading.RLock()
         self.__loop_from_run_thread = None
-        self.__this_task: Optional[asyncio.Future[None]] = None
+        self.__this_task: Optional[asyncio.Task[None]] = None
 
         # Used to indicate when shutdown() has been called.
         self.__shutting_down = False
@@ -156,7 +156,14 @@ class LaunchService:
     def _is_idle(self) -> bool:
         number_of_entity_future_pairs = self._prune_and_count_entity_future_pairs()
         number_of_entity_future_pairs += self._prune_and_count_context_completion_futures()
-        return number_of_entity_future_pairs == 0 and self.__context._event_queue.empty()
+        if self.event_loop is not None and self.__this_task is not None:
+            tasks = asyncio.all_tasks(self.event_loop)
+            tasks.remove(self.__this_task)
+        else:
+            tasks = set()
+        return (number_of_entity_future_pairs == 0 and
+                self.__context._event_queue.empty() and
+                len(tasks) == 0)
 
     @contextlib.contextmanager
     def _prepare_run_loop(
--- a/launch/test/launch/test_execute_local.py
+++ b/launch/test/launch/test_execute_local.py
@@ -178,3 +178,48 @@ def test_execute_process_with_output_dictionary():
     ls = LaunchService()
     ls.include_launch_description(ld)
     assert 0 == ls.run()
+
+
+def test_execute_process_with_shutdown_on_error():
+    """Test proper shutdown of children after exception during launch."""
+    exited_processes = 0
+
+    def on_exit(event, context):
+        nonlocal exited_processes
+        print(event, context)
+        exited_processes += 1
+
+    executable_1 = ExecuteLocal(
+        process_description=Executable(
+            cmd=[sys.executable, '-c', 'while True: pass']
+        ),
+        output={'stdout': 'screen', 'stderr': 'screen'},
+        on_exit=on_exit,
+    )
+    executable_2 = ExecuteLocal(
+        process_description=Executable(
+            cmd=[sys.executable, '-c', 'while True: pass']
+        ),
+        output={'stdout': 'screen', 'stderr': 'screen'},
+        on_exit=on_exit,
+    )
+
+    # It's slightly tricky to coerce the standard implementation to fail in
+    # this way. However, launch_ros's Node class can fail similar to this and
+    # this case therefore needs to be handled correctly.
+    class ExecutableThatFails(ExecuteLocal):
+
+        def execute(self, context):
+            raise Exception('Execute Local failed')
+
+    executable_invalid = ExecutableThatFails(
+        process_description=Executable(
+            cmd=['fake_process_that_doesnt_exists']
+        ),
+        output={'stdout': 'screen', 'stderr': 'screen'},
+    )
+    ld = LaunchDescription([executable_1, executable_2, executable_invalid])
+    ls = LaunchService()
+    ls.include_launch_description(ld)
+    assert ls.run() == 1
+    assert exited_processes == 2


Flova:
Any update or feedback regarding this PR? @wjwwood

Flova:
@ahcorde is there a change we get this moving? Nearly everybody I know who uses ROS2 experienced this issue at some point.

SammyRamone:
It would be good to merge this, as this error is affecting many people and is quiet annoying.

launch/launch/launch_service.py
```diff
@@ -156,7 +156,14 @@ def _prune_and_count_context_completion_futures(self) -> int:
     def _is_idle(self) -> bool:
         number_of_entity_future_pairs = self._prune_and_count_entity_future_pairs()
         number_of_entity_future_pairs += self._prune_and_count_context_completion_futures()
-        return number_of_entity_future_pairs == 0 and self.__context._event_queue.empty()
+        if self.event_loop is not None and self.__this_task is not None:
+            tasks = asyncio.all_tasks(self.event_loop)
+            tasks.remove(self.__this_task)
```
wjwwood, line 161: What if `self.__this_task` has completed or otherwise not present in the tasks? Maybe we should use `.discard()` to avoid a potential `KeyError`? I don't think it's actually a problem if `self.__this_task` isn't in the set, but it could happen I think.

wjwwood, line 161: You're excluding `self.__this_task`, but what about the `process_one_event_task`? I feel like you could get into a situation where that task is present, waiting for an event that will never come, but it is not excluded from the list of all tasks, there for `is_idle()` is also false, and you end up in a dead-lock.

Would it make more sense to track non-internal tasks in the context instead? Rather than grabbing all tasks and trying to exclude various tasks that don't matter for the shutdown condition? Maybe using `context._completion_futures` to track this?

wjwwood (commented):
Sorry for the long delay in reviewing this. I needed some time to really look into it, since I deem it a pretty high risk change. I left some comments/concerns about it and would appreciate any feedback on them you guys could give, but I understand you may not have time or the issue may not be fresh for you any longer.

I opened https://github.com/ros2/launch/pull/996 with my suggestions (preserving your tests).

launch/launch/launch_service.py
MegaIng: In what situation would `self.__this_task` be non-`None` and not being executed? To reach this point we must be inside an async task, and if we don't know which task we ourselves are because this task wasn't correctly being kept track of, we can't implement this check correctly.

We can switch to `.discard` to avoid a `KeyError`, but I suspect that this would just hide a programming error somewhere else.

MegaIng: We are excluding the currently running tasks because we want to check if there is any *other* work that may be running.

The issue with ignoring stuff like "internal tasks" is that it may again reintroduce the error that leads to this issue - i.e. canceling stuff that may still produce events. (if there currently are no such internal tasks, they may be added in the future)

Maybe the `process_one_event` task should have a timeout for how long it waits for an event? It would then get rescheduled by the main loop as needed, but still giving the main loop a chance to check for being idle.

I also want to point out that this patch has been used in production, and this has not been an issue, although I am not currently deep enough in the logic to explain why it's not an issue.

wjwwood: That's fair, I suppose it's likely that if `__this_task` is not in the tasks list then we're in some kind of bug state, in which case the `KeyError` might be appropriate. I was also thinking about thread-safety here, but ideally that shouldn't be a problem in asyncio, i.e. there's a race between checking `self.__this_task` and using it in the `remove()`, but again ideally no one would be changing that under our feet.

The pattern I've seen online is using `asyncio.current_task()` and storing the result of that locally until you're done with it.

We actually use that conditionally here:

https://github.com/MegaIng/ros2-launch/blob/18662d5df0ec206b64c4bef153dbb997dbb82b90/launch/launch/launch_service.py#L197-L202

But that is used to provide access to the "LaunchService" task via the getter, whereas in the place where we're using it, we actually mean "the async task we're currently in" because that's the only one we actually care about ignoring when considering the task list. It just so happens those are the same thing most of the time (perhaps all of the time).

Also, online folks tend to filter the output of all tasks against their done state, e.g. `other_pending = [t for t in asyncio.all_tasks() if not t.done() and t !== current]`. This also avoids the `KeyError` and maybe is the better approach here.

I know the existing patch has the benefit of some battle testing in production, but I also want to merge code that conceptually covers everything as well. What do you think?

wjwwood: > Maybe the process_one_event task should have a timeout for how long it waits for an event? It would then get rescheduled by the main loop as needed, but still giving the main loop a chance to check for being idle.

We could do that, but I dislike the idea that we have to choose a number for that timeout that either a) is very short and therefore we end up busy-waiting and wasting CPU cycles, or b) is kind of long and means that shutdown may be delayed by up to that amount of time when it could have just exited immediately (this would present on the console as a hesitation when ctrl-c'ing launch for example). Instead it would be better to not consider it in the calculation, at least in my opinion.

As for why this isn't a problem in production, I don't know either. It could be that the sequencing is already that this idle check happens after a `process_one_event` but before the next one is queued. My concern is that while it may not affect you in your production use case, it could be simply that the way you're using launch doesn't excite this part of the system (as in control theory).

My thinking was just that `process_one_event` alone should definitely not block shutdown, if present, so it should be ignored when checking for is_idle.

wjwwood (commented):
Thanks for iterating. I left a few more responses, but if we don't agree to any changes, I'm inclined merge this one as-is and open follow ups.

Thanks again for the contribution!

launch/launch/launch_service.py
MegaIng: I agree that conceptually `self.__this_task` and current tasks are slightly different, and that changing that part is ok.

I am slightly worried about throwing away any Tasks that are marked as `done()` because AFAIK that doesn't mean that their return value has been used. If that return value would be used by the main loop, that could generate events, no?

MegaIng: > My thinking was just that process_one_event alone should definitely not block shutdown

Issue is that `await queue.get()` could have retrieved an item, emptying the queue and then given control to the other task. This would mean `process_one_event` still has one event to process but `is_idle` would be True if we were ignoring `process_one_event`.

We should probably be using the `task_done` machinery of `queue`? I.e. check if the queue is empty, and all events are processed, and then I think it's safe to ignore `process_one_event`.

MegaIng: And race conditions really shouldn't be an issue as long as we don't have any await points inbetween. That's the central promise of async compared to threaded.

PR #915, opened 2025-10-23
Add FindLaunchfile substitution

### LINKED ISSUE #849: IncludeLaunchDescription - request name without suffix/extension
### Description

Allow LaunchDescription inclusion to happen with an incomplete name, having auto-completion fill in the suffix/extension, throwing an error if it's ambiguous.

Examples:

Commandline
```
ros2 launch nav2_bringup bringup
```

Python launchfile
```
from pathlib import Path

from ament_index_python.packages import get_package_share_directory
from launch.actions import  IncludeLaunchDescription

...

  launch_dir = Path(get_package_share_directory('nav2_bringup') / 'launch' 
  ...
  IncludeLaunchDescription(launch_dir / 'bringup'))
```

XML
```
<include file="$(find-pkg-share nav2_bringup)/launch/bringup" />
```

YAML
```
- include:
    file: "$(find-pkg-share nav2_bringup)/launch/bringup"
```

All of these would resolve to actually including `bringup_launch.py` without having to say launch twice, or know it's implemented in python

### Motivation

Given that users can write launchfiles in several languages, we have a usability/discoverability problem.

Launchfile authors want to put a syntax-appropriate suffix on the file, so that code editors will highlight the syntax properly.

However from a user perspective, the action of including a launchfile is a semantic one. I am saying "please launch this semantic ros subgraph/system", and I _explicitly_ do not care what language the description of that subsystem was implemented in. Compare: I don't care what language client is publishing the topic I am subscribing to, I should not have to say `subscribe_to_python_publisher('topic_name_py')`.

The discussion of using a single common suffix for launchfiles is pretty split, at least within the core contributor base, so globally using `.launch` is probably never going to be the solution for this.

So, based on some discussions around this limitation, this way of approaching the problem came up.

This allows authors to
1. Properly suffix the file for syntax highlighting in editor
2. Change syntaxes later without breaking users who include the particular launch description

And allows users to
1. Not know or care which language was used to provide a description, focusing on the package's exposed semantic interface

Probably have to do:
- If file of given name exists, OK
- Else check `path.parent` for files with that prefix
- If there are multiple, throw error
- Else launch that file

Maybe we want to limit this to a specified list of "allowed suffixes" so not any suitable prefix would match?
```
.launch
.xml
.y[a]ml
.py
_launch.[xml|y[a]ml|py]
.launch.[xml|y[a]ml|py]
```

### Additional Information

When implemented, tutorials and demos should be updated to provide this as the default usage pattern.

`ros2launch` CLI should show semantic prefixes as the autocomplete options

emersonknapp:
## Description

Fixes #849

 `FindLaunchfile` substitution can find launch files in a directory by stem name only, if it matches valid suffixes.

This enables users to _not care_ what language an included launchfile is written in, or whether it uses `.launch` or `_launch` style.

So instead of

```
IncludeLaunchDescription(launch_dir / 'radical_cool_autonomy.launch.xml')
```

the user can do:

```
IncludeLaunchDescription(FindLaunchfile(path=launch_dir, name='radical_cool_autonomy'))
```

This is slightly more verbose, but incredibly more flexible.

Now if the package maintainer migrates their Python launchfile to YAML, for example - we as users don't have to change our launchfiles and we'll automatically include the new one.

This change enables new features in `launch_ros` that I'll push soon: `FindPackageLaunch(pkg, name)` and `IncludePackageLaunch(pkg, name)`. These will utilize `FindPackageShareDirectory` + `FindLaunchfile` in combination.

CAVEAT: if maintainers follow the "redirect migration strategy" that I recommend in my upcoming roscon talk (below), there will now be two launchfiles matching the searched stem, which will fail the substitution. But I'm not sure what the strategy should be to mitigate that.

```
# migrated_launch.py
from launch.actions import LogWarning, IncludeLaunchDescription
from launch.substitutions import PathSubstitution, ThisLaunchDir

    return LaunchDescription([
        LogWarning(msg='migrated_launch.py is deprecated: redirecting to migrated_launch.yaml'),
        IncludeLaunchDescription(PathSubstitution(ThisLaunchDir()), 'autonomy_launch.yaml'),
    ])
```

I can think of these possible strategies:
* user side: use a new stem when following this migration strategy
* launch side: allow an optional "soft language preference" option which would help choose from multiple matches, e.g. `FindLaunchfile(name='stem', path=launch_dir, prefer='yaml')`. If there's _only_ python, it will return that. If py and yaml, then the yaml one returned. If py and xml, then... error? This is probably too confusing of an idea. 

Probably the "you should only have one launchfile per stem" soft constraint is better.

New feature!

No

### Additional Information

N/A


launch/launch/substitutions/find_launchfile.py
```diff
@@ -0,0 +1,126 @@
...
+        self.__path = normalize_to_list_of_substitutions(path)
+
+    @classmethod
+    def parse(cls, data: Sequence[SomeSubstitutionsType]
+              ) -> Tuple[Type['FindLaunchfile'], Dict[str, Any]]:
+        """Parse `FindLaunchfile` substitution."""
+        if len(data) != 2:
+            raise AttributeError(
+                f'find-launchfile substitution expects 2 argument, {len(data)} given')
+        return cls, {'name': data[0], 'path': data[1]}
```
SuperJappie08, line 92: Wouldn't it make more sense to make the arguments `$(find-launchfile PATH FILE)` instead of FILE PATH?

launch/launch/substitutions/find_launchfile.py
```diff
@@ -0,0 +1,126 @@
...
+from .substitution_failure import SubstitutionFailure
+from ..frontend import expose_substitution
+from ..launch_context import LaunchContext
+from ..some_substitutions_type import SomeSubstitutionsType
+from ..substitution import Substitution
+
+
+def name_matches_launchfile(name: Text, file: Path) -> bool:
+    from ..frontend import Parser  # avoid circular import
+    valid_extensions = {'py', *Parser.get_available_extensions()}
```
SuperJappie08, line 36: Wouldn't py already be part of  `Parser.get_available_extensions()`
(I guess it wouldn't hurt to have it as an extra fallback here.)

emersonknapp: It is not - python is the "backend", Parser is for "frontends" which are provided by launch_xml and launch_yaml. There's explicit checking for .py in the loading process as well.

launch/launch/substitutions/find_launchfile.py
emersonknapp: Open for debate, my reasoning was "find name in dir" which matches "grep pattern in location". 

I'm sure a sensible argument could be made for the opposite as well. But I'll stick with this implementation as my suggestion.

SuperJappie08: Yeah, I guess it is a matter of opinion.

An argument for my alternative would be that it would match the `ros2launch` argument ordering (in the package case), which makes the most sense for the suggested `IncludePackageLaunch`.
Keeping the same order of `[package/folder, launchfile']` for this substitution would keep everything consistent.

launch/launch/substitutions/find_launchfile.py
```diff
@@ -0,0 +1,126 @@
...
+    valid_suffixes = {'', '.launch', '_launch'}
```
SuperJappie08, line 37: Wouldn't allowing a non-suffix lead to matching Python package files as well?

I could imagine a package containing `my_node.py` and `my_node.launch.py`, which might introduce issues in the `IncludePackageLaunch` case.

SuperJappie08:
@emersonknapp, have you considered what to do with subfolders?

e.g:
```
launch/
├─ subsystem/
│  ├─ sublaunch.launch.xml
├─ general_launch_file.launch.xml
```

Should it be able to find the `sublaunch.launch.xml` from the `launch` folder?

emersonknapp:
@SuperJappie08 I thought about it but didn't come to a conclusion. Maybe it should take a recursive boolean argument?

launch/launch/substitutions/find_launchfile.py
emersonknapp: That is sensible, yes. Internal consistency is a good argument 👍 I'll change it

SuperJappie08:
I think it really depends on the way this substitution is framed:
- Now, as `FindLaunchfile`, it hints that a search and (possibly optional) recursion would make sense.
- If it were phrased as 'get this launch file, with any extension,' recursion would make significantly less sense, as the specific launch file base name needs to be specified. (`GetLaunchfile` could be a clearer name for this function.)

If recursion were to be implemented, I can think of 3 options:
- A trailing optional boolean argument to enable it. However, this would be the only substitution with a flag, which feels off.
- A separate substitution, which could be called `FindLaunchfileRecursive`. Also probably not the nicest and more maintance.
- Allowing some globbing in the launch file name. So if you would want to select something from any folder, use `'**/my_file'`. It's a bit weird and harder to parse, but it would keep the user API relatively clean.

But after writing this, I feel the best option could just be to rename it to `GetLaunchfile` and not support recursion, as it will only do exact matches on the base name.
This also prevents edge cases, for when there are multiple folders containing launch files with the same base name.
This also has great user explanation, that you probably should know ~~exactly~~ what the included launch file does and not just pick a random one, since it could be a safety issue. 
(For example, I move similar launch files in subfolders, one of the folders gets renamed, and the preferred launch file now changes, changing the behavior of the system completely. It is a bit of a stretch, but this would probably be some user who would just do that.

PR #916, opened 2025-10-27
Frontend expose unexposed Actions - that can be

emersonknapp:
## Description

Working on a "completeness" pass for the launch frontends. These Actions were trivial to expose but hadn't been. That should cover everything that doesn't involve Events or "Opaque" processes.

New feature

no

### Additional Information

N/A

### DIFF
--- a/launch/launch/actions/pop_environment.py
+++ b/launch/launch/actions/pop_environment.py
@@ -15,11 +15,18 @@
 """Module for the PopEnvironment action."""
 
 from typing import Any
+from typing import Dict
+from typing import Tuple
+from typing import Type
 
 from ..action import Action
+from ..frontend import Entity
+from ..frontend import expose_action
+from ..frontend import Parser
 from ..launch_context import LaunchContext
 
 
+@expose_action('pop_env')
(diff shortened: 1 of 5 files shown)


christophebedard:
Would you mind adding some simple tests?

fujitatomoya (approved):
implementation looks good, test would be ideal.

ahcorde (commented):
@emersonknapp do you have plans to add some tests here ?

SuperJappie08:
@emersonknapp, Is it wise to expose `Pop` and `Push` actions?
As they are internally used to handle the scoping in Actions such as `GroupAction`, I assumed they were ment to be internal (back-end) actions.

My fear stems from the case where a user were to push and pop these incorrectly.
This could result in breaking these other 'scoping' actions cause trouble down the line.
An example would be pushing and not popping (while in a group).

Does exposing those, what I assumed to be internal, scope setting Actions to the user add to the usefulness in the front-end.
As the scoping can already be achieved by using the `GroupAction`.

EDIT: Phrasing

PR #930, opened 2025-11-22
change docs file structure to work with recent rosdoc2

ottojo:
## Description
This moves the sphinx docs sources one directory up, as expected by rosdoc2. Furthermore, the sphinx sourcedir is configured in rosdoc2.yaml, which is how to tell rosdoc2 to use the existing index.rst instead of wrapping it (https://github.com/ros-infrastructure/rosdoc2/blob/main/doc/index.rst.md).
I have added toctree entries for the autogenerated documentation elements by rosdoc2.

This should also fix https://github.com/ros2/ros2_documentation/issues/6042

Yes, the documentation page will change. It seems however (as apparent in the issue above), that the resulting link to architecture.html was changed inadvertently at some point. This reverts it back to the original link.

No.

### Additional info

Needs `@Mergifyio backport humble jazzy rolling`


launch/doc/.gitignore
```diff
@@ -1 +0,0 @@
-build
```
fujitatomoya, line 1: https://github.com/ros2/rmw_fastrtps/pull/855#discussion_r2545027166

probably we want to wait until the above discussion result?

fujitatomoya (commented):
@ottojo my understanding is that this PR replaces https://github.com/ros2/ros2_documentation/pull/6043, right? IMO this fix makes much better than https://github.com/ros2/ros2_documentation/pull/6043.

launch/doc/.gitignore
fujitatomoya: ah i see https://github.com/ros2/ros2_documentation/pull/6043#issuecomment-3566681206, thanks 👍

ottojo:
> @ottojo my understanding is that this PR replaces [ros2/ros2_documentation#6043](https://github.com/ros2/ros2_documentation/pull/6043), right? IMO this fix makes much better than [ros2/ros2_documentation#6043](https://github.com/ros2/ros2_documentation/pull/6043).

yes, this is intended to replace 6043

launch/doc/.gitignore
ottojo: i removed those files which look like a standalone sphinx-project (without using rosdoc2)

launch/doc/index.rst
```diff
@@ -11,7 +11,10 @@ Welcome to launch's documentation!
    :caption: Contents:
 
    architecture
-   modules
+   Links <__links>
+   Python API <modules>
+   Standard Documents <__standards>
+   ROS Package Dependencies <__ros_package_dependencies>
```
ahcorde, line 17: I can see this error

```bash
WARNING: toctree contains reference to nonexisting document '__ros_package_dependencies' [toc.not_readable]
```

launch/doc/index.rst
ahcorde, line 29:
```suggestion
   ROS Package Dependencies <__ros_package_dependencies>

Indices and tables
==================
```

ahcorde (changes requested):

launch/doc/index.rst
ottojo: I can't reproduce this, is this on the buildfarm or locally for you? This probably happens with a (perhaps just slightly) out of date rosdoc2, as this only got merged recently: https://github.com/ros-infrastructure/rosdoc2/pull/205

ahcorde: I'm running version `rosdoc2 0.2.2.dev0`, I used a `virtual env` to install `rosdoc2` with the latest changes but I can't see this directory.

I'm using this cmd:

```bash
rosdoc2 build   --package-path ./src/ros2/launch/launch
```

fujitatomoya:
@ottojo can you address the comments?

launch/doc/index.rst
ottojo: i could reproduce the same warning with rosdoc2 installed via the `python3-rosdoc2` apt package on kilted (after making this change in the rosdoc2 yaml, which is not required in a more recent version).
```diff
diff --git a/launch/rosdoc2.yaml b/launch/rosdoc2.yaml
index d2f5855..c243fe9 100644
--- a/launch/rosdoc2.yaml
+++ b/launch/rosdoc2.yaml
@@ -3,4 +3,4 @@ version: 1
 ---
 settings: {}
 builders:
-  - sphinx: { sphinx_sourcedir: "doc" }
+  - sphinx: { name: "launch", output_dir: "", sphinx_sourcedir: "doc" }
```

 I'm not sure how to proceed here, to me this looks to me like a rosdoc2 version issue, unfortunately the version has not been bumped in 2 years despite significant changes (https://github.com/ros-infrastructure/rosdoc2/blob/main/rosdoc2/__init__.py#L15)

ottojo:
@fujitatomoya I have applied the suggestions, and added a redirect since the link change in the ros2-docs got merged in the meantime, so both versions stay working.

I'm not sure how to proceed with the warning seen by @ahcorde , maybe someone else could verify if this is or isn't an issue, or maybe it's possible somehow to run the buildfarm doc job for this PR to check if that works?

Issue #939, opened 2026-01-29
ROS 2 YAML launch: "choices" is not supported in arg


rafal-gorecki:
### Description:
Using the "choices" key in an arg inside a ROS 2 YAML launch file causes a runtime error, but this limitation is not clearly documented.

### Steps to reproduce:
1. Create YAML launch file:

```yaml
launch:
  - arg:
      name: rviz
      default: "true"
      description: "Launch RViz2"
      choices: ["true", "false"]
```

2. Run:
```bash
ros2 launch <package> <file>.yaml
```

### Observed behavior:
Launch fails with:
```bash
ValueError: Unexpected key(s) found in `arg`: {'choices'}
```

### Expected behavior:
- Support "choices" in YAML launch files

### Environment:
- ROS 2: Jazzy

christophebedard:
This is totally not well documented (or even tested/used anywhere in the ROS 2 core), but looking at [the implementation](https://github.com/ros2/launch/blob/rolling/launch/launch/actions/declare_launch_argument.py#L180-L184), the right syntax is:

```yaml
launch:
  - arg:
      name: rviz
      default: "true"
      description: "Launch RViz2"
      choice:
        - value: "true"
        - value: "false"
```

and the XML syntax is:

```xml
<launch>
    <arg name="rviz" default="true" description="Launch RViz2">
        <choice value="true"/>
        <choice value="false"/>
    </arg>
</launch>
```

We should:

1. Include a full YAML example and improve the XML example in the underlying [`DeclareLaunchArgument` Python class](https://docs.ros.org/en/rolling/p/launch/launch.actions.html#launch.actions.DeclareLaunchArgument) docstring: https://github.com/ros2/launch/blob/57c1aeec7b18be4e0e93c91a4b069d4437f5b839/launch/launch/actions/declare_launch_argument.py#L99-L107
2. Add a YAML test for it in `launch_yaml` similar to this XML test: https://github.com/ros2/launch/blob/57c1aeec7b18be4e0e93c91a4b069d4437f5b839/launch_xml/test/launch_xml/test_arg.py#L41-L58
3. Add examples somewhere in the ROS 2 docs

Any help would be appreciated!

mjcarroll:
Oh, @christophebedard and I looked at this at the same time...

Well I added a test that demonstrates in #941

christophebedard:
ahh you had assigned it to yourself, sorry!

rafal-gorecki:
[...]

PR #941, opened 2026-01-30
launch_yaml: add test using choice in launch args

mjcarroll:
Ref: #939 

Built-with: Gemini Flash 3


mjcarroll:
@christophebedard hold off on reviewing this one, I didn't see the similar XML test, I'm going to rework this one to look a little closer to that one.

PR #945, opened 2026-02-18
add functionality to enable color logging for ros2launch messages

### LINKED ISSUE #942: Add colorlog support
### Description

It would be very nice if support for [colorlog](https://pypi.org/project/colorlog/) could be added, for colorizing log outputs for launch files.

### Motivation

Makes logs look nicer & easier to skim over

Should be relatively easy to implement, just check if `RCUTILS_COLORIZED_OUTPUT` is set to `1` (maybe add another variable?) or the process is detected to be running in a tty, and then replace the `Formatter` with a `ColoredFormatter`:

https://github.com/ros2/launch/blob/57c1aeec7b18be4e0e93c91a4b069d4437f5b839/launch/launch/logging/__init__.py#L320-L322

### Additional Information

_No response_

andreas-loeffler:
## Description

This PR adds color logging support to the launch system. Currently the logs of individual nodes run from the launch file can be colorized using e.g. *emulate_tty*, however the messages generated by ros2launch itself (e.g., "process started", "process has died", and log tags [INFO] or [ERROR]) remained unaffected/non-colored.

Corresponding tests to the colorlog functionality have been added too. 

(a) ros2 launch using the newly added colorlogging 
image

(b) old behavior in which *RCUTILS_COLORIZED_OUTPUT* or *emulaty_tty* did not affect the color logging ability of ros2launch messages
image

Fixes #942 

Yes. Users will now see colorized log output for ros2launch events when running in a tty or when *RCUTILS_COLORIZED_OUTPUT=1*

No

### Additional Information
- The *contains_ansicode* helper function used for the added tests is derived from the ansi regex check posted here: https://stackoverflow.com/questions/68556358/how-to-check-for-ansi-character-in-python

- Since the PR adds a new python3 dependency (python3-colorlog) the corresponding rosdep key was already added. See https://github.com/ros/rosdistro/pull/49860

### DIFF
--- a/launch/launch/logging/__init__.py
+++ b/launch/launch/logging/__init__.py
@@ -25,6 +25,8 @@ import sys
 from typing import (Any, Dict, List, Literal, Optional, Protocol, Set, Tuple,
                     Union)
 
+import colorlog
+
 from typing_extensions import TypeAlias
 
 from . import handlers
@@ -125,6 +127,27 @@ class LaunchConfig:
         logging.root.setLevel(logging.INFO)
         self.set_screen_format('default')
         self.set_log_format('default')
+
+    def _logging_colors_enabled(self) -> bool:
+        """
+        Determine if colored output should be used for console logging.
+
+        Colors are enabled when:
+        - RCUTILS_COLORIZED_OUTPUT=1 is explicitly set, OR
+        - RCUTILS_COLORIZED_OUTPUT is unset AND stdout is a TTY
+
+        Colors are disabled when:
+        - RCUTILS_COLORIZED_OUTPUT=0 is explicitly set
+
+        :return: True if colors should be used, False otherwise
+        """
+        colorized_output = os.environ.get('RCUTILS_COLORIZED_OUTPUT')
+        if colorized_output == '0':
+            return False
+        elif colorized_output == '1':
+            return True
+        else:
+            return sys.stdout.isatty()
 
     @property
     def level(self) -> int:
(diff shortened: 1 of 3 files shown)


andreas-loeffler:
Hi @asymingt :) just wanted to check in on this PR. Is there anything else you need from my side to move this fwd, or any feedback I can address?

Wojcik98:
Are there any updates on this PR? Colored logs would be very useful.

launch/launch/logging/__init__.py
```diff
@@ -126,6 +128,27 @@ def reset(self) -> None:
...
+
+        :return: True if colors should be used, False otherwise
+        """
+        colorized_output = os.environ.get('RCUTILS_COLORIZED_OUTPUT')
+        if colorized_output == '0':
+            return False
+        elif colorized_output == '1':
+            return True
+        else:
+            return sys.stdout.isatty()
```
asymingt, line 150: If the `RCUTILS_COLORIZED_OUTPUT` environment variable is not set it looks like the terminal will have colors added by default. My only fear is that some ROS user might be relying on log parsing that might not be expecting the escape sequences used to add the color. Should we perhaps leave the default behavior as no colors, and have the user opt-in to the new behavior?

What do you think @mjcarroll?

launch/test/launch/test_logging.py
```diff
@@ -97,6 +115,111 @@ def test_output_loggers_bad_configuration(log_dir):
 ]
 
 
+def test_colorized_output_with_tty(capsys, monkeypatch, mock_clean_env):
+    """
+    Test color output with TTY.
+
+    Colors should be enabled when stdout is set to a TTY and
```
asymingt, line 122: This is the exact behavior I'm proposing we consider inverting.

launch/launch/logging/__init__.py
fujitatomoya: agree with this.

besides, scheme won't match rcutils.
`ColoredFormatter` is constructed without an explicit log_colors=, so it uses colorlog's defaults, notably INFO=green, DEBUG=cyan, those are not consistent with rcutils convention.
so launch's own messages will be colored differently from the node output they're interleaved with.
I'd ask for an explicit log_colors mapping that mirrors rcutils for visual consistency.

PR #953, opened 2026-03-19
Support `scoped=True` for `IncludeLaunchDescription` action

paulsohn:
## Description

Currently `IncludeLaunchDescription` (equivalently, `<include>` in xml) does not have any scoping option, so it leaks the launch configurations and set environment variables to the caller.

This variable leak is actually a feature that some ROS projects (e.g. Autoware) deliberately exploit to reuse common preset (i.e. the set of variables) -- so the backward compatibility should be kept, while we also need the ability to lint whether this leakage is intentional.

The behavior of `<include scoped="true" (...) />` is equivalent to already possible `<group scoped="true"><include (...) /></group>`.

The implementation is mostly a duplication of `group_action.py`. The only difference is that I deferred the support of `forwarding` for now (it acts like `forwarding=True`.)

~~Fixes `#801`~~ after a discussion I found that #801 also exposes other issues, but nevertheless this PR is highly relevant for what we are missing.
Compared to its own importance, we've been spending too much time on orthogonal issue about how to make `forwarding=False` / filter what to forward to the includee.

The change is backward compatible, as `IncludeLaunchDescription(scoped=False)` is the default behavior. On the other hand the user can specify `IncludeLaunchDescription(scoped=True)` to make scoped include.

The implementation part is done by myself.
Claude Code (Opus 4.6) helped me in generating test cases. A thorough human review involved.

### DIFF
--- a/launch/launch/actions/include_launch_description.py
+++ b/launch/launch/actions/include_launch_description.py
@@ -29,6 +29,10 @@ from typing import Union
 import launch.logging
 
 from .opaque_function import OpaqueFunction
+from .pop_environment import PopEnvironment
+from .pop_launch_configurations import PopLaunchConfigurations
+from .push_environment import PushEnvironment
+from .push_launch_configurations import PushLaunchConfigurations
 from .set_launch_configuration import SetLaunchConfiguration
 from ..action import Action
 from ..frontend import Entity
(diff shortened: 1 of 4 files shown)


SuperJappie08:
I'm unsure if this completely addresses #801, as this forwards the external scope.

This unintentional forwarding could also affect the inner launchfile in unintended ways if it is not accounted for.

There are situations where this unintended passing is intended; however, this is not the case described in #801.

wjwwood:
Since this pull request lead me to it, cross-reference to my post on (one of) the originating issues: https://github.com/ros2/launch/issues/801#issuecomment-5335696941

PR #971, opened 2026-04-28
Fix #931:add 'actions' key support for YAML,keeping the existing 'children' key as a fallback

### LINKED ISSUE #931: YAML: `group` should also accept a key `actions:` like the Python API
### Description

Especially given the lack of documentation, `children:` in the YAML frontend is an opaque implementation detail. The Python API takes `Group(actions=...)` so I believe so should YAML.

### Motivation

More intuitive usage for YAML

The `children:` special key is an implementation detail that is not present in XML because of its structure.

```xml
<group if="condition">
  <child />
</group>
```

In YAML the various lists of items that would go into the XML tag body are instead implemented as keys. The catch-all key is `children:`

```yaml
group:
  if: condition
  children:
    - child:
```

In some other cases the children may be heterogenous attribute lists instead of Actions, so those are always implemented with an explicit name, 

```xml
<node>
  <param name="x" value="y" />
</node>
```

```yaml
- node:
    param:
      - name: x
        value: y
```

I think it's fine to continue to have `children:` as the fallback, especially to not break existing usage, but adding an additional option for `actions:` to match the Python API makes a lot of sense. This would apply to any other Actions that take child actions (as noted, attributes must always be explicitly named so that's not a problem), if there are any

### Additional Information

_No response_

SanjayS66:
## Description
This PR adds support for YAML to accept 'actions' key to match the Python API, while continuing to have 'children' as a fallback to not break the existing usage.The ValueError message in the parser has also been updated to reflect both 'actions' and 'children' as valid keys

Fixes #931

yes. User can now use 'actions' to define nested entities.Existing files using 'children' will continue to work as that children is added as fallback.Error message also has been updated to reflect support for both actions and children keys.

No

### Additional Information

### DIFF
--- a/launch_yaml/launch_yaml/entity.py
+++ b/launch_yaml/launch_yaml/entity.py
@@ -63,13 +63,17 @@ class Entity(BaseEntity):
                 f'\n---\n{self.__element}\n---'
             )
         if isinstance(self.__element, dict):
-            if 'children' not in self.__element:
+            if 'actions' in self.__element:
+                self.__read_keys.add('actions')
+                children = self.__element['actions']
+            elif 'children' in self.__element:
+                self.__read_keys.add('children')
+                children = self.__element['children']
+            else :
                 raise ValueError(
                     f'Expected entity `{self.__type_name}` to have children entities.'
-                    f'That can be a list of subentities or a dictionary with a `children` '
-                    'list element')
-            self.__read_keys.add('children')
-            children = self.__element['children']
+                    f'That can be a list of subentities or a dictionary with an `actions` '
+                    f'or `children` list element')
         else:
             children = self.__element
         entities = []


emersonknapp (commented):
I'm concerned that this is global, but `children` may not always == `actions`. It is equivalent for `<group>` but are there other cases where XML frontend uses `children` but the Python API doesn't call them `actions`?

emersonknapp (approved):
Okay I did an investigation on this and found that children is used in
- `GroupAction`: `kwargs['actions'] = children`
- (launch_ros) `TimerAction`: `kwargs['actions'] = children`
- `ForEach`: Python API is `function=`, children parsed into a closure
- `ForLoop`: same

In all these contexts, the relevant children are indeed Actions. I believe that it would be semantically nonsensical for an action to provide an interface that exposes `actions` as a YAML input and also calls `entity.children`. In any case, the `assert_entity_completely_parsed` would catch that one of the two was never read.

This seems good to me, thanks for the contribution.

emersonknapp:
Pulls: ros2/launch#971
Gist: https://gist.githubusercontent.com/emersonknapp/666309cbd8f330c8c42d41f80688f0c4/raw/82dfd66aafdaa47c3af6dbbb76aa6e7b5f85bbbd/ros2.repos
BUILD args:  --packages-above-and-dependencies launch_yaml
TEST args:  --packages-above launch_yaml
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/19356
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29167/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21960/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9052/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28123/)

PR #985, opened 2026-07-11
Support substitutions in ExecuteProcess timeouts

Old-Ding:
## Summary

- allow XML and YAML `sigterm_timeout` and `sigkill_timeout` attributes to contain launch substitutions
- preserve parse-time non-negative validation for numeric values and reject invalid literal strings
- cover both frontends with regression tests

## Testing

- `python -m pytest -q launch/test/launch` (279 passed)
- `python -m pytest -q launch_xml/test/launch_xml` (31 passed)
- `python -m pytest -q launch_yaml/test/launch_yaml` (19 passed)
- package `ament_flake8` tests for `launch`, `launch_xml`, and `launch_yaml`
- `python -m mypy --ignore-missing-imports launch/actions/execute_process.py`
- `python -m compileall -q launch/launch launch_xml/launch_xml launch_yaml/launch_yaml`
- `git diff --check`

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -377,25 +377,28 @@ class ExecuteProcess(ExecuteLocal):
                     )
                 kwargs['respawn_delay'] = respawn_delay
 
-        if 'sigkill_timeout' not in ignore:
-            sigkill_timeout = entity.get_attr('sigkill_timeout', data_type=float, optional=True)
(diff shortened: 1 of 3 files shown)


sylvesterkaczmarek (commented):
Could substituted timeouts be validated after evaluation too? A literal -1 is rejected, but $(var timeout) resolving to -1 bypasses this check and reaches TimerAction, effectively causing immediate escalation.

PR #990, opened 2026-07-12
Parse YAML respawn delay strings

Old-Ding:
## Description

The YAML frontend accepts strings for attributes declared as `float`, but `ExecuteProcess.parse()` compared a string-valued `respawn_delay` with `0.0` before converting it. A launch file using `respawn_delay: '2.0'` therefore failed during parsing with `TypeError: '<' not supported between instances of 'str' and 'float'`.

Convert string-valued respawn delays at the existing frontend parsing boundary. Valid values reach `ExecuteLocal` as floats, while invalid strings fail immediately with a descriptive `ValueError`. Native float behavior and the runtime respawn state machine are unchanged.

Yes. Quoted numeric `respawn_delay` values in YAML launch files now work consistently instead of failing with a comparison `TypeError`. Invalid strings produce a parse-time error.

Yes. OpenAI Codex assisted with tracing the frontend data flow, preparing the focused fix and tests, and reviewing the result. I reviewed the diff and verification output.

### Additional Information

Local verification:
- reproduced the Rolling failure with `respawn_delay: '2.0'`
- all 279 `launch` functional tests passed
- all 18 `launch_yaml` functional tests passed
- both package flake8 tests passed
- targeted `ament_mypy` passed with zero errors
- `ament_flake8`, `ament_pep257`, and `ament_copyright` passed for both changed files
- `compileall` and `git diff --check` passed

### DIFF
--- a/launch/launch/actions/execute_process.py
+++ b/launch/launch/actions/execute_process.py
@@ -370,6 +370,15 @@ class ExecuteProcess(ExecuteLocal):
         if 'respawn_delay' not in ignore:
             respawn_delay = entity.get_attr('respawn_delay', data_type=float, optional=True)
             if respawn_delay is not None:
+                if isinstance(respawn_delay, str):
+                    try:
+                        respawn_delay = float(respawn_delay)
+                    except ValueError:
+                        raise ValueError(
+                            'Attribute respawn_delay of Entity node expected to be '
+                            'a non-negative floating-point value but got `{}`'.format(
+                                respawn_delay)
+                        ) from None
                 if respawn_delay < 0.0:
                     raise ValueError(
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
One edge case: float('nan') passes the < 0.0 check. Later nan > 0.0 is also false, so 'nan' effectively becomes an immediate respawn. Could non-finite values be rejected explicitly?

sylvesterkaczmarek (commented):
One edge case: float('nan') passes the < 0.0 check. Later nan > 0.0 is also false, so 'nan' effectively becomes an immediate respawn. Could non-finite values be rejected explicitly?

PR #1001, opened 2026-09-04
Sandbox environment variables in launch_xml/launch_yaml tests

mjcarroll:
`LaunchContext.environment` is `os.environ` itself, so `test_rep_env.py` — which executes `ReplaceEnvironmentVariables` against a real context and asserts the environment ends up empty — clears `PATH` for every test that runs after it in the same pytest process. That was harmless while pytest 7 collected `test/test_*.py` first, but pytest 9 collects `test/<pkg>/` first, so `test_rep_env.py` now runs ahead of `test_xmllint.py` and `shutil.which('xmllint')` returns `None`.

This adds an autouse fixture at the `test/` level in both packages that restores `os.environ` after each test, which fixes the failure without making it depend on collection order again.

### DIFF
--- a/launch_xml/test/conftest.py
+++ b/launch_xml/test/conftest.py
@@ -1,0 +1,43 @@
+# Copyright 2026 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Fixtures shared by all tests in this package."""
+
+import os
+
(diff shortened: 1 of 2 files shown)


fujitatomoya (approved):
lgtm

christophebedard (approved):

christophebedard:
Pulls: ros2/launch#1001
Gist: https://gist.githubusercontent.com/christophebedard/5a3231cdf060b61cfb1365e41646a6cf/raw/536123e00b58a5f3cfe7ad4d2995ebdc68d47d7a/ros2.repos
BUILD args:  --packages-above-and-dependencies launch_xml launch_yaml
TEST args:  --packages-above launch_xml launch_yaml
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/20344
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30443/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/23137/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/10242/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/29314/)

christophebedard:
Sounds like it would be a good idea to do this kinda everywhere?

ahcorde (approved):

PR #1005, opened 2026-09-14
Robust pytest_ignore_collect for multi-version Pytest compatibility (backport #974)

nim65s:
Hi,

This backports #974 on kilted, and replace #994 

I tested that it fix my downstream use cases with pytest 9.1.1

PR #972, opened 2026-05-04
Fix Pytest 8/9 compatibility in hooks

mjcarroll:
This PR fixes deprecated hook signatures in launch and launch_testing that cause failures in newer pytest versions (e.g. Pytest 9.0).

PR #1007, opened 2026-09-16
Fix Pytest 8/9 compatibility in hooks (backport #972)

nim65s:
#972

see #1005


christophebedard (approved):
Running CI with combined changes with #1005: https://github.com/ros2/launch/pull/1005#issuecomment-5706139893

PR #1006, opened 2026-09-16
 Robust pytest_ignore_collect for multi-version Pytest compatibility (backport #974)

nim65s:
backport #974 to Jazzy

same as #1005

PR #1008, opened 2026-09-16
Fix Pytest 8/9 compatibility and coroutine leaks in launch_pytest (backport #972)

nim65s:
backport #972 to Jazzy

same as #1007


christophebedard (approved):
Running CI with combined changes with #1006: https://github.com/ros2/launch/pull/1006#issuecomment-5718539927

PR #1010, opened 2026-09-20
Fix windows launch file calling python node

gftabor:
## Description

Ros2 run does a check on windows for python files. They need to be executed through the python interpreter rather than directly from the system execute process call. (If the python files were "installed" through cmake)

Here we copy the logic over to launch. This solves 

```
OSError: [WinError 193] %1 is not a valid Win32 application
```

Windows won't crash anymore.

The edit was suggested by chatGPT looking at the error traceback and the ros2 run source code.

### Additional Information

Even getting to this point requires setting an environment variable 
set PATHEXT=%PATHEXT%;.PY

Because there is a check for "is executable" that looks at a hard coded list of file extensions that windows knows how to execute.


gftabor:
The PR https://github.com/ros2/launch_ros/pull/577 replaces the need for the  environment variable.

Launch_ros PR lets the python file get past the initial check.
Launch PR lets the python file actually run

gftabor:
Ros2 run logic here 

https://github.com/ros2/ros2cli/blob/4eaefee77c3a66c2f724f56096539f009bb576a8/ros2run/ros2run/api/__init__.py#L54

